PR #2, opened 2022-10-05
Create deb_release.yml

yphus:
New deb release workflow

Manage RC and stable releases (with an option for dry mode)

Since workflow_dispatch only support a maximum of 10 user inputs, all config options are grouped into a single json parameter.

A typical json config looks like this:

```
{
    "mode": "testing",
    "dry_run": true,
    "checkbox-ng": true,
    "checkbox-support": true,
    "provider-base": true,
    "provider-resource": true,
    "provider-tpm2": true,
    "provider-sru": true,
    "provider-certification-server": true,
    "provider-certification-client": true,
    "provider-gpgpu": true,
    "provider-ipdt": true,
    "provider-phoronix": true
}
```
Successfully tested using the staging repo: 
https://github.com/canonical/checkbox-staging/actions/workflows/deb_release.yml

### DIFF
--- a/.github/workflows/deb_release.yml
+++ b/.github/workflows/deb_release.yml
@@ -1,0 +1,65 @@
+name: Debian packages release
+
+on:
+  workflow_dispatch:
+    inputs:
+      config:
+        description: JSON of options
+        type: string
+        default: '{}'
+        required: true
+
+jobs:
+  Release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt update -qq
+          sudo apt install -qq -y bumpversion devscripts python3-launchpadlib
+      - name: Checkout checkbox monorepo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check for new commits
+        run: |
+          echo '${{ inputs.config }}' | jq > config.json
+          tools/release/release_deb_monorepo.py check --config config.json
+          rm -f config.json
+      - name: Bump versions
+        run: |
+          git config --global user.email "robot@canonical.com"
+          git config --global user.name "Devices Certification Bot"
+          tools/release/release_deb_monorepo.py bump --part major
+      - name: Create changelog
+        run: |
+          tools/release/release_deb_monorepo.py changelog
+      - name: Archive changelog
+        uses: actions/upload-artifact@v3
+        with:
+          name: Changelog
+          path: changelog
+      - name: Push release tags
+        run: |
+          tools/release/release_deb_monorepo.py push
+      - name: Update the PPA recipe and kick-off the builds
+        env:
+          LP_CREDENTIALS: ${{ secrets.LP_CREDS }}
+          CHECKBOX_REPO: ${{ github.repository }}
+        run: |
+          tools/release/release_deb_monorepo.py build
+      - name: Open a new release for development
+        if: fromJson(inputs.config).mode == 'stable'
+        env:
+          DEBEMAIL: robot@canonical.com
+          DEBFULLNAME: Devices Certification Bot
+        run: |
+          tools/release/release_deb_monorepo.py open
+      - name: Create pull request
+        if: |
+          (fromJson(inputs.config).mode == 'stable' &&
+           fromJson(inputs.config).dry_run != true)
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr create -H release --title 'Open a new release for development' --body 'Created by Github action'
--- a/tools/release/README.md
+++ b/tools/release/README.md
@@ -1,0 +1,168 @@
+# Checkbox release process (Debian packages)
+
+## PPA/Repositories
+
+* [Stable](https://launchpad.net/~hardware-certification/+archive/ubuntu/public): The official release of Checkbox.
+* [Testing](https://code.launchpad.net/~checkbox-dev/+archive/ubuntu/testing): Release candidates of Checkbox before it becomes the official release.
+* [Development](https://code.launchpad.net/~checkbox-dev/+archive/ubuntu/ppa): Daily builds (that may contain experimental features).
+
+## Projects released as Debian packages
+
+* [checkbox-ng](https://github.com/canonical/checkbox/tree/main/checkbox-ng)
+* [checkbox-support](https://github.com/canonical/checkbox/tree/main/checkbox-support)
+* [providers/base](https://github.com/canonical/checkbox/tree/main/providers/base)
+* [providers/resource](https://github.com/canonical/checkbox/tree/main/providers/resource)
+* [providers/certification-client](https://github.com/canonical/checkbox/tree/main/providers/certification-client)
+* [providers/certification-server](https://github.com/canonical/checkbox/tree/main/providers/certification-server)
+* [providers/sru](https://github.com/canonical/checkbox/tree/main/providers/sru)
+* [providers/tpm2](https://github.com/canonical/checkbox/tree/main/providers/tpm2)
+* [providers/ipdt](https://github.com/canonical/checkbox/tree/main/providers/ipdt)
+* [providers/phoronix](https://github.com/canonical/checkbox/tree/main/providers/phoronix)
+* [providers/gpgpu](https://github.com/canonical/checkbox/tree/main/providers/gpgpu)
+
+## Release steps summary
+
+Steps | Release candidate(s) (RC) | Stable release | Dry mode
+:--- | :---: | :---: | :---:
+Parse each project and check for new commits | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
+Bump and tag versions | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
+Create release changelog (since the latest stable tag) | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
+Push release tags to Github | :heavy_check_mark: | :heavy_check_mark: | :x:[^1]
+Update the PPA recipes and kick-off the builds | :heavy_check_mark: | :heavy_check_mark: | :x:
+Open a new release for development | :x: | :heavy_check_mark: | :x:
+Create a pull request including the new [SemVer](https://semver.org/spec/v2.0.0.html) | :x: | :heavy_check_mark: | :x:
+
+## How to trigger the GitHub Actions release workflow
+
+All the release steps above are fully automated but initiating a release is a manual
+process requiring some user input.
+
+The same workflow support two types of release, **testing** and **stable**.
+Additionally, the release manager can perform a **dry run** to:
+* Identify which projects are going to be released 
+* Review the release changelog
+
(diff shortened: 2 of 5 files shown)


.github/workflows/deb_release.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: Changelog
+          path: changelog
+      - name: Push release tags
+        run: |
+          support/release/git/release_deb_monorepo.py push
+      - name: Update the PPA recipe and kick-off the builds
+        env:
+          LP_CREDS: credentials
```
mz2, line 56: I don't quite follow this one. a) I don't understand how it works (I thought `env` is setting prerequisites for the `run` so I didn't think `credentials` would exist) and b) I would have thought '${{ secrets.LP_CREDS }}' belongs here instead:

```yaml
env:
  LP_CREDS: "${{ secrets.LP_CREDS }}"
```

mz2: Also, if presently the script requires a file called `credentials`, it'd be better to change it to accept an env var instead so there is no temptation to write them out to disk.

yphus: it's a file actually https://help.launchpad.net/API/launchpadlib#Authenticated_access

.github/workflows/deb_release.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Archive the release configuration
+        uses: actions/upload-artifact@v3
+        with:
+          name: config.json
+          path: config.json
+      - name: Install dependencies
+        run: |
+          sudo apt update -qq
+          sudo apt install -qq -y bumpversion devscripts python3-launchpadlib
+          git clone https://git.launchpad.net/~checkbox-dev/checkbox/+git/support
```
kissiel, line 28: This is a part of Checkbox, so it feels like having this outside monorepo defeats the purpose of monorepo.

.github/workflows/deb_release.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Checkout checkbox monorepo
+        uses: actions/checkout@v3
+        with:
+          repository: canonical/checkbox
+          fetch-depth: 0
+          path: src/checkbox
+      - name: Check for new commits
+        run: |
+          support/release/git/release_deb_monorepo.py clone --config config.json
```
kissiel, line 37: the program used here was pushed to that repo without any review process, and it includes _many_ bugs, and copy-pasta legacy that's not needed

.github/workflows/deb_release.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Bump versions
+        run: |
+          git config --global user.email "sylvain.pineau@canonical.com"
```
kissiel, line 40: I guess it would be better to use ce-certification-qa's deets here.

.github/workflows/deb_release.yml
kissiel, line 56: this is a _path_ to credentials, not actual credentials, meaning we have multiple problems with it:
* automation doesn't consider this to be secret, so the safety precautions may not work (like redacting stuff on the fly)
* this is a relative path (as in ./credentials) that's super fragile
* the rm invocation later on is not guaranted, so this credential file should be made more ephemeral (like used from secrets directly)

.github/workflows/deb_release.yml
```diff
@@ -0,0 +1,73 @@
...
+          rm -f credentials
+      - name: Open a new release for development
+        if: fromJson(inputs.config).mode == 'stable'
+        run: |
+          support/release/git/release_deb_monorepo.py open
+      - name: Create pull request
+        if: fromJson(inputs.config).mode == 'stable'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DEBEMAIL: sylvain.pineau@canonical.com
```
kissiel, line 69: with ce-certification-qa "person" having ownership of many parts of the project, maybe we should make them own this too.

kissiel (changes requested):
See inline comments, including one security risk.

Also, most of what this action is supposed to do is hidden away behind the external tools. There's no info here what is actually going to happen when this action runs.
From what I can tell, this action when running on github will collect stuff from launchpad,
run tooling that will make changes to this repository to then do other LP related things.

Also, the main tool behind all of this, the "release_deb_monorepo.py" is a copy pasta with modifications that has not been reviewed nor approved by anyone, and using it in production feels bad.

.github/workflows/deb_release.yml
```diff
@@ -0,0 +1,73 @@
...
+          DEBFULLNAME: Sylvain Pineau
+        working-directory: src/checkbox
+        run: |
+          gh pr create -H release --title 'Open a new release for development' --body 'Created by Github action'
```
mz2, line 73: I can live with this as a temporary measure if bump-version really doesn't have a useful behaviour with `--no-commit`, but it's also a tiny bit weird to do this, so it will be a temporary measure for sure.

.github/workflows/deb_release.yml
mz2: The credentials do not in my understanding need to be passed in a file.

.github/workflows/deb_release.yml
mz2: We don't have to solve all legacy problems in this PR, but whatever edits were done to that script of course should be reviewed before we call this done, no question about that.

.github/workflows/deb_release.yml
yphus: I've update the legacy lp scripts to get credentials from the env var instead

.github/workflows/deb_release.yml
yphus: deb release scripts are now part of the monorepo (tools/release/)

.github/workflows/deb_release.yml
yphus: Fixed

yphus:
README file added in tools/release

tools/release/release_deb_monorepo.py
```diff
@@ -0,0 +1,422 @@
...
+                    #     cwd=project_path, check=True)
+            elif mode == 'testing':
+                if "rc" in last_tag:
+                    # bump to jump to rc+1
+                    run(['bumpversion', '--current-version', last_tag,
+                         '--tag', '--no-commit', 'N'],
+                        cwd=project_path, check=True)
+                else:
+                    # bump to jump to rc1
+                    run(['bumpversion', '--current-version',
```
mz2, line 222: What's the fixed 1.99.0 about?

tools/release/release_deb_monorepo.py
```diff
@@ -0,0 +1,422 @@
...
+                         '1.99.0',  # FIXME last_tag,
+                         '--tag', '--no-commit', self.args.part],
+                        cwd=project_path, check=True)
+            # Only keep the tags, discard .bumpversion file updates
+            run('git checkout .',
```
mz2, line 227: Hmm, should you also do a `git clean -dff`? (possible untracked files)

tools/release/release_deb_monorepo.py
```diff
@@ -0,0 +1,422 @@
...
+        logger.info("".center(80, '#'))
+        logger.info("# Parsing local repo...")
+        logger.info("".center(80, '#'))
+        for path, dirs, files in os.walk('.'):
+            if "debian" in dirs:
+                project_path = str(Path(*Path(path).parts))
+                project_root = '.'
+                project_name = str(project_path).replace('s/', '-')
+                if not config[project_name]:
+                    self.settings["projects"][project_name] = {
```
mz2, line 103: A subjective stylistic point made as a general grumble: to me the whole logic would come out easier to follow if there was actually no mutable `self.settings`, but there were free functions / class methods that took `settings` as an argument, and returned a copy of the modified `settings` as a return value, for the next party to then mutate.

tools/release/lp-request-import.py
```diff
@@ -0,0 +1,63 @@
...
+# Checkbox is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Script that forces a code import from Github in Launchpad.
```
mz2, line 21: I thought it would be nice to link to a page that explains what a "code import" is exactly (I was left wondering if it is a matter of a pull kind of mirroring step or something more elaborate), so I found https://help.launchpad.net/Code/Imports, but it looks like it was last updated 3AD 🙃 Anyhow, it might still be a handy link to include.

mz2 (changes requested):
Some requested README changes requested, and more subjective points on the release script.

tools/release/README.md
```diff
@@ -0,0 +1,49 @@
+# Checkbox Debian packages release tools
+
+Manage RC and stable releases (with an option for dry mode)
+
+The following projects are released as Debian packages:
```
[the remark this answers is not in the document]
yphus: Not all the components are released as debian packages. A top level README will be added via another PR

tools/release/README.md
```diff
@@ -0,0 +1,49 @@
...
+    "provider-sru": true,
+    "provider-certification-server": true,
+    "provider-certification-client": true,
+    "provider-gpgpu": true,
+    "provider-ipdt": true,
+    "provider-phoronix": true
+}
+```
+
+## PPA
```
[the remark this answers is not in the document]
yphus: snaps releases have their own workflows and will be covered in Maciej's PR

tools/release/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+* Release Candidates: https://code.launchpad.net/~checkbox-dev/+archive/ubuntu/testing
+* Stable releases: https://launchpad.net/~hardware-certification/+archive/ubuntu/public
+
+## TODO
+
+* Remove the forced major bump (hardcoded 1.99) and `--bump major`
```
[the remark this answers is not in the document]
yphus: GH issues?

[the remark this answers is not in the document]
mz2: Jira, please (I am investigating getting syncing in place for GitHub issues, at which point we can resolve the awkward Launchpad / Jira / GitHub triangle).

tools/release/release_deb_monorepo.py
```diff
@@ -221,7 +221,13 @@ def bump(self):
                 else:
                     # bump to jump to rc1
                     run(['bumpversion', '--current-version',
-                         '1.99.0',  # FIXME last_tag,
+                         # FIXME Using 1.99.0 here ensures all packages will
```
mz2, line 224: I can't say I follow your thought there fully, but sounds like we have a follow-up task to consistently version the packages so we don't have differing major (or indeed minor) versions across packages. Presumably this also conflicts with giving per package control when releasing too, so needs some thought.

pieqq:
+1 for me, since you addressed Maciek and Matias' points.

PR #3, opened 2022-10-06
Add checkbox snaps code

kissiel:
This PR import checkbox snaps repos without any modification for better tracking of the changes.


yphus:
checkbox-ng-packaging and checkbox-support-packaging are not needed. The debian/ folders for those two are already included (see checkbox-ng/debian and checkbox-support/debian)

PR #6, opened 2022-10-11
Add RTD config

pieqq:
After gathering everything into a monorepo, RTD complains about building the documentation that lives in checkbox-ng/docs. Using a .readthedocs.yaml config and specifying a requirements.txt file to fix the issue.


yphus (approved):
LGTM

PR #10, opened 2022-10-17
bin:audio_test.py: use pactl get-default-source and get-default-sink

medicalwei:
Get default source/sink directly instead of manually filtering through the list.

## Description

Describe your changes here:

- The original implementation of getting current default source/sink is by filtering through the list of sources/sinks and get the first item.  However, this wouldn't work if the default source/sink is not the first one.
- This fixes the issue by using the command `pactl get-default-(source|sink)` to get the identifier string of the device.
- Note that the other commands `pactl set-{source,sink}-volume` and `pactl set-{source,sink}-mute` can be used with identifier string instead of numeric id to control the device.

To test this issue:
1. Find a device with multiple audio outputs (laptop with HDMI audio out, for example), or multiple audio inputs.
2. Set the pactl default device sink (using `pactl set-default-sink`) to the second one of `pactl list sinks short | grep -v monitor | grep -v autonull` (or replace sink with source for inputs)
3. Run `audio_test.py` and it should use the device assigned above.

This test script didn't have an automated test before.  If a test script is required along with this PR, I can attempt to make one.

## Resolved issues

This PR is a migration of the Launchpad MP: https://code.launchpad.net/~checkbox-dev/plainbox-provider-checkbox/+git/plainbox-provider-checkbox/+merge/431179

Has no public LP bug / Jira issue opened.

## Documentation

- [ ] Automated tests are included for the changed functionality in this PR. If to be merged without tests, please elaborate why.
- [ ] Necessary documentation is provided for the changed functionality in this PR (tests are documentation, too).

## Tests

- [X] Steps to follow for reviewer to run & manually test are provided.
- [ ] A list of what tests were run and on what platform/configuration is provided.

### DIFF
--- a/providers/base/bin/audio_test.py
+++ b/providers/base/bin/audio_test.py
@@ -128,10 +128,12 @@ class PAVolumeController(object):
             return False
         if not self.identifier:
             return False
-        command = ['pactl',
-                   'set-%s-volume' % (self.pa_types[self.type]),
-                   str(self.identifier[0]),
-                   str(int(volume)) + "%"]
+        command = [
+            "pactl",
+            "set-%s-volume" % (self.pa_types[self.type]),
+            self.identifier,
+            str(int(volume)) + "%",
+        ]
         self._pactl_output(command)
         self._volume = volume
         return True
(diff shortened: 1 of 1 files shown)


medicalwei:
Hi, since another colleague discovered the same issue and forwards the ticket to me, could anyone review this changeset? Thanks.

medicalwei:
The following tests has been run:
* On a Ubuntu 22.04 VM
* On a Ubuntu 20.04 VM
* On an actual 22.04 test unit that has such problem (Dell 0C1B)
* On a Debian system that turns off audio profiles (so that only dummy input/output is available)

pieqq:
Thanks for this improvement, and thanks for testing on different platforms!

Looks good to me!

PR #25, opened 2022-10-24
collect mac to submission

stanley31huang:
## Description
adding a device job into the depends field of miscellanea/submission-resources job. So we could capture all hardware components while we submit a submission to C3.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A


pieqq:
Looks good to me.

Thanks!

PR #33, opened 2022-11-11
Fix typo in Bluetooth jobs

pieqq:
After fixing this, running

```
./manage.py validate
```

works. Before that, it would complain with:

```
error: ???: Cannot load job definitions from
'/<<PKGBUILDDIR>>/providers/base/units/bluetooth/jobs.pxu': Unexpected multi-line value (jobs.pxu, line 420)
```

## Resolved issues

Because of this typo, Debian packages cannot be built. See for instance:

https://launchpadlibrarian.net/633100725/buildlog_ubuntu-bionic-amd64.checkbox-provider-base_2.0.0~rc2~ppa~ubuntu18.04.1_BUILDING.txt.gz


ltmvic51 (approved):
+1, thank you for the quick fix.

PR #34, opened 2022-11-14
Metabox refresh

yphus:
## Description

Set of patches to get metabox working 

## Resolved issues

- Resume from LXD snapshots (metabox config are saved in the container but file operations now require a running container)
- Random issues with the expect command (on success the stdout buffer was totally consumed/reset)
- Tests requiring a desktop user session were failing since more env var are now required (in addition of just $DISPLAY)
- Failing scenarios (outdated test plan names, dummy timeouts)
- Noisy logging from ws4py

## Tests

tested using:

`$ metabox metabox/configs/testing-ppa-config.py --do-not-dispose`


yphus:
Landing this one to allow scenario development

PR #38, opened 2022-11-18
Remove pts from sru test plan

yphus:
## Description

Removal of the phoronix tests from the SRU test plan

## Resolved issues

PTS based tests are always failing and became useless in the scope of SRU testing


jocave (approved):
No issues from me

PR #42, opened 2022-11-23
Tox4all

yphus:
## Description

Collection of new github action workflows running tox on monorepo sub-projects aiming at replacing the previous PMR based solution running tests in LXC on supported Ubuntu releases.

## Resolved issues

Fixes [CHECKBOX-174](https://warthogs.atlassian.net/browse/CHECKBOX-174)

## Tests

Tested successfully in the staging repo:
-  https://github.com/canonical/checkbox-staging/actions/workflows/checkbox-ng.yaml
- https://github.com/canonical/checkbox-staging/actions/workflows/checkbox-support.yaml
- https://github.com/canonical/checkbox-staging/actions/workflows/provider-base.yaml
- https://github.com/canonical/checkbox-staging/actions/workflows/provider-resource.yaml

Tested with an unsupported python3.5 syntax (f-strings), expected failed test:
- https://github.com/canonical/checkbox-staging/actions/runs/3515959411


providers/certification-client/tox.ini
```diff
@@ -0,0 +1,68 @@
+[tox]
+envlist = py35,py36,py38,py310
+skip_missing_interpreters = true
+skipsdist=True
+
+[testenv]
+commands =
+    pip -q install -e ../../checkbox-ng
+    - {envbindir}/python3 ../resource/manage.py develop
```
pieqq, line 9: 2 things:

1. Maybe there should be a comment to explain why we run these ("Required because this provider depends on the resource and the base providers", or something like that). This should be the same for all other providers that depend on other providers.
2. In [Tox doc](https://tox.wiki/en/latest/index.html#system-overview), it says:

> starting a command with a single dash character means ignore exit code

Any particular reason it's used here? (and in the other tox.ini for projects depending on other projects)

providers/docker/tox.ini
```diff
@@ -0,0 +1,67 @@
...
+envlist = py35,py36,py38,py310
+skip_missing_interpreters = true
+skipsdist=True
+
+[testenv]
+commands =
+    pip -q install -e ../../checkbox-ng
+    - {envbindir}/python3 ../resource/manage.py develop
+    - {envbindir}/python3 ../base/manage.py develop
+    {envbindir}/python3 manage.py validate
```
pieqq, line 11: Is it normal the `manage.py test` command is not run for this provider?

pieqq (commented):
Great work!!! I can't imagine how long it took to gather all the info to get these written (including the painful process of reviewing what version of each package is available in each of the Ubuntu versions we still maintain!)

I left a few inline comments.

providers/certification-client/tox.ini
yphus: 1. Good idea, I'll add inline comments.

2. I think it's not needed in CI. I had to use this trick locally as the repo provider units were conflicting with previous tox invocations (which left developed providers in `/var/tmp`). Good catch

providers/docker/tox.ini
yphus: Actually `manage.py test` is not run for two providers, docker and tpm2.

- Docker: too many shellcheck errors are reported by manage.py test, unless the original maintainer (GarryW) helps us to fix them all, I'd recommend to just perform the usual provider validation on, this one.

- TPM2: problem with this one is the `bin/` dir which contains upstream tpm2-tools scripts copies. Again let's just run the provider validation.

providers/tpm2/tox.ini
```diff
@@ -4,9 +4,12 @@ skip_missing_interpreters = true
 skipsdist=True
 
 [testenv]
+whitelist_externals = rm
```
pieqq, line 7: According to [the documentation](https://tox.wiki/en/latest/changelog.html#v3-18-0-2020-07-23), this parameter has been deprecated:

> Add `allowlist_externals` alias to `whitelist_externals` (`whitelist_externals` is now deprecated). - by [@dajose](https://github.com/dajose) [#1491](https://github.com/tox-dev/tox/issues/1491)

Might be better to stick to the new alias to avoid problems later :)

pieqq (commented):
All good minus a little comment inline regarding `whitelist_externals` keyword.

providers/tpm2/tox.ini
yphus: thanks for the review, this is fixed.

pieqq (approved):
All good then! Thanks again!

PR #44, opened 2022-11-24
refresh the checkbox snap with the lastest commits from their launchpad counterparts

yphus:
## Description

A few commits were missing in the checkbox-snap series since the last snap release occurred after the monorepo setup

- Bump to 2.4.dev
- odm launchers


checkbox-snap/series_classic18/launchers/odm-certification-app
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+print_usage() {
+    echo "Usage:"
+    echo "  cdts.odm-certification IP"
```
pieqq, line 5: are we keeping the cdts snap?

pieqq (approved):
Looks OK to me, just one question inline regarding `cdts`.

PR #45, opened 2022-11-25
New workflows to get daily builds of checkbox and checkbox core snaps

yphus:
## Description

The workflow itself runs daily but the builds are actually triggered only if new commits were created for:

- the checkbox snap: only in checkbox-snap/
- the checkbox core snap: checkbox-ng/, checkbox-support/, providers/ and checkbox-core-snap/

## Tests

Tested using the staging repo:

- checkbox core snap: https://github.com/canonical/checkbox-staging/actions/runs/3546758036
- checkbox snap: https://github.com/canonical/checkbox-staging/actions/runs/3546748981

Note that the checkbox snap for 22.04 is uploaded to both **22.04/edge** and **latest/edge**

### DIFF
--- a/.github/workflows/checkbox-snap-daily-builds.yml
+++ b/.github/workflows/checkbox-snap-daily-builds.yml
@@ -1,0 +1,62 @@
+name: checkbox snap daily build
+
+on:
+  schedule:
+    - cron: '00 04 * * *'
+  workflow_dispatch:
+
+jobs:
+  snap:
+    strategy:
+      matrix:
+        type: [classic, uc]
+        releases: [16, 18, 20, 22]
+    runs-on: ubuntu-latest
+    env:
+      SERIES: series_${{ matrix.type }}${{ matrix.releases }}
+      SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT7_CREDS }}
+    name: Checkbox snap for series ${{ matrix.type }}${{ matrix.releases }}
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check for checkbox snap new commits
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/checkbox-snap-daily-builds.yml
```diff
@@ -0,0 +1,62 @@
...
+          mkdir -p ~/.local/share/snapcraft/provider/launchpad/
+          echo '${{ secrets.LP_CREDS }}' > ~/.local/share/snapcraft/provider/launchpad/credentials
+          git config --global user.email "robot@lists.canonical.com"
+          git config --global user.name "Certification bot"
+      - uses: snapcore/action-build@v1
+        id: snapcraft
+        with:
+          path: checkbox-snap/series_${{ matrix.type }}${{ matrix.releases }}
+          snapcraft-channel: 7.x/stable
+          snapcraft-args: remote-build --build-on amd64,arm64,armhf,i386 --launchpad-accept-public-upload
```
jocave, line 43: Just checking for my own interest... this build that is performed by `snapcraft remote-build ...` is it strictly "private"? I assume not, but not problem in this case?

yphus: It's not, all builds triggered via `snapcraft remote-build` are public. On success the ephemeral snap recipe on lp is cleanup up automatically but during the build or if it failed, it persists. See https://launchpad.net/~ce-certification-qa/+snaps

yphus: Ref: https://snapcraft.io/docs/remote-build

> Prerequisites
> Prospective snaps need to be open source, as the code will be publicly available, and you’ll need a [Launchpad account](https://login.launchpad.net/+new_account).

jocave (approved):
Looks really nice and preserves the current expectations of build frequency.

One small con that I think might be unavoidable when the the action is based on a cron schedule would be that you would get regular "Failed to build" outcomes and only by digging in to the workflow could you tell that it was "Checkbox for checkbox projects new commits" step that was the cause of the fail.

PR #252, opened 2022-11-28
Templates

beliaev-maksim:
add bug and ER templates to automatically apply labels


pieqq:
Just to clarify, this is so that any new bug is tagged with the label "bug"
where a FR is tagged with "enhancement"?

Does it have an impact on the launchpad bug import mechanism?

beliaev-maksim:
No influence on the import

This is only for creation of new bugs. With templates users will see two buttons, to open enhancement or to open a bug
Each will have a different set of fields

pieqq (approved):
+1. Thanks!

PR #267, opened 2022-12-09
info/secure-boot-check: Don't fail if the system does not support Sec…

dannf:
…ure Boot

If a system does not support secure-boot, `mokutil --sb-state` will report that and exit non-zero:

```
ubuntu@d05-4:~$ sudo mokutil --sb-state
This system doesn't support Secure Boot
ubuntu@d05-4:~$ echo $?
255
```

It seems to me that this is no reason to fail the test - the info was collected. Let's just ignore the error code from mokutil.

## Description

Describe your changes here:

- False positive test failure
- Ignore the error code when an error message itself is a satisfactory fulfillment of the info

## Resolved issues

N/A

## Documentation

- [x] Automated tests are included for the changed functionality in this PR. If to be merged without tests, please elaborate why.
I don't know that there's a reasonable way to add a test for this test.
- [x] Necessary documentation is provided for the changed functionality in this PR (tests are documentation, too).

## Tests

- [x] Steps to follow for reviewer to run & manually test are provided.
- [x] A list of what tests were run and on what platform/configuration is provided.

Ran it on a system that is known not to support Secure Boot.

### DIFF
--- a/providers/base/units/info/jobs.pxu
+++ b/providers/base/units/info/jobs.pxu
@@ -510,7 +510,7 @@ estimated_duration: 0.2
 requires:
  package.name == 'mokutil'
 command:
- mokutil --sb-state
+ mokutil --sb-state || true
 _summary: Check secure boot state
 _description:
  Output secure boot is enabled or disabled


bladernr (changes requested):
The his is a bit too simplistic. The goal of the test is to fail if secure boot is not enabled (where supported). I agree returning true for this error code is a good idea but we still need to fail if it is supported but not enabled.

dannf:
> The his is a bit too simplistic. The goal of the test is to fail if secure boot is not enabled (where supported). I agree returning true for this error code is a good idea but we still need to fail if it is supported but not enabled.

I thought that was the point of `miscellanea/secure_boot_mode`. This is `info/secure-boot-check`, which seems to just be about collecting output, not actually testing the feature. To emphasize that, note that `info/secure-boot-check` currently *passes* if the system supports secure boot and secure boot is disabled.

bladernr (approved):
Ack, I did, indeed, conflate the two similar jobs... works for me.

PR #271, opened 2022-12-14
metabox: Change container creation timeout

pieqq:
## Description

Change metabox container creation timeout from 2 minutes to 5 minutes. This helps initial creation scenario where a lot of data (images) has to be pulled from the network.

Thanks to this change, I can finally run

```
metabox configs/testing-ppa-config.py --log=TRACE --do-not-dispose
```

and see the command complete successfully!

(the code is still flimsy, and I will look into that later)


kissiel (approved):
+1

PR #272, opened 2022-12-15
Metabox fix container creation (and other various fixes)

pieqq:
## Description

Various fixes for metabox, mainly to prevent it from crashing early on when creating containers.

This PR also updates the calls to the `checkbox` snap, since both `checkbox-snappy` and `checkbox-snappy-classic` snaps have been removed in favor of the `checkbox` snap.

Finally, added reference to Jammy, added some logging info, and fixed a typo.

## Documentation

- There is currently zero documentation on Metabox, but this is something I hope to improve over the next cycle!

## Tests

1. Setup and activate venv: `python3 -m venv metabox && source metabox/bin/activate`
2. Deploy metabox in it in dev mode: `pip install -e .`
3. Run it with Testing PPA config: `metabox metabox/configs/testing-ppa-config.py --log=TRACE --do-not-dispose`
4. Run the same command again to make sure metabox reuses the same containers and there is no problem
5. Run it with Checkbox snap config: `metabox metabox/configs/checkbox-core-snap-classic-beta-config.py --log=TRACE --do-not-dispose`

All the required containers should be created. There should not be any complaint from LXC about not being able to create an instance because one with the same name already exists.

(The `urwid.testplan.UrwidTestPlanSelection` scenario will probably fail, this is expected, I will have to look into this next...)

### DIFF
--- a/metabox/configs/checkbox-core-snap-classic-beta-config.py
+++ b/metabox/configs/checkbox-core-snap-classic-beta-config.py
@@ -1,18 +1,18 @@
 configuration = {
     'local': {
-        'origin': 'classic_snap',
+        'origin': 'classic-snap',
         'checkbox_core_snap': {'risk': 'beta'},
         'checkbox_snap': {'risk': 'stable'},
         'releases': ['bionic', 'focal'],
     },
     'remote': {
-        'origin': 'classic_snap',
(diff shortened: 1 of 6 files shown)


kissiel (approved):
Goodies, +1

PR #279, opened 2022-12-21
Metabox workflow

yphus:
## Description

Describe your changes here:

- New Metabox workflow based on Tox (pytest)
- With pytest compatibility (`conftest.py`) only the failed scenarios will display the corresponding execution logs.

## Tests

Tested and deployed in the staging env: https://github.com/canonical/checkbox-staging/actions/runs/3747735301/jobs/6364279483

### DIFF
--- a/.github/workflows/metabox.yml
+++ b/.github/workflows/metabox.yml
@@ -1,0 +1,22 @@
+name: Metabox tests
+
+on:
+  workflow_dispatch:
+
+jobs:
+  Metabox:
+    defaults:
+      run:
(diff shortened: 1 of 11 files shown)


pieqq:
Very nice to have this integrated with GitHub Actions!

Tox can be used to run any command, not only Python tests. Wouldn't it be easier if it was setup to launch the `metabox` command directly, rather than having to maintain the extra pytest layer (`conftest.py`) you created just so that it can run a `pytest` command? (unless I'm missing something obvious!)

Issue #99, opened 2022-11-28
sleep_time_check.py failed to parse suspend/resume time


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1988521

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Confirmed	| 2022-09-02 07:51:02	| 8	| High	| False  	|

## Description
[Description]
sleep_time_check.py failed to parse 

    # find our times
    for line in lines:
        try:
            if "Average time to sleep" in line:
                sleep_time = float(line.split(':')[1].strip())
                sleep_times.append(sleep_time)
            elif "Average time to resume" in line:
                resume_time = float(line.split(':')[1].strip())
                resume_times.append(resume_time)
        except ValueError as e:
            print("ERROR: One or more times was not reported correctly:")
            print(e)
            return 1

I guess the "Average time to sleep/resume" likely missing or showing unexpected output.
In the failure case, the checkbox submission doesn't include "pm_test.reboot.3.log".
Thus, we won't know what's going on.

[Summary]
power-management/suspend-30-cycles-time-check-with-reboots test failed

[Steps to reproduce]
Run checkbox-cli stress test

[Checkbox job `com.canonical.certification::power-management/suspend-30-cycles-time-check-with-reboots` output]

stdout
------
ERROR: One or more times was not reported correctly:
could not convert string to float: 'N/A'

## Attachments
No attachments

Tags:
['ce-qa-concern', 'oem-priority', 'originate-from-1950728', 'originate-from-1950729', 'originate-from-1950737', 'originate-from-1970877', 'originate-from-1981804', 'somerville', 'stella']

beliaev-maksim:
This thread was migrated from launchpad.net
##### https://launchpad.net/~pieq wrote on 2022-10-28 02:26:32:
I suspect the logs generated by fwts_test.py spit out "N/A", and it makes checking the time impossible:

https://github.com/canonical/checkbox/blob/main/checkbox-support/checkbox_support/scripts/fwts_test.py#L128-L130

I will need to run this on an actual device and capture the logs to see what's the content of the file before being processed.

We see this error more and more, so I think this is important to get fixed.

##### https://launchpad.net/~pieq wrote on 2022-11-11 03:51:08:
Discussed with QA:

[...]

PR #300, opened 2023-01-11
Provide more information about suspend/resume times when running 30 cycles with reboots/poweroffs

pieqq:
## Description

As requested by QA, the output of `sleep_time_check.py` now includes more
info, namely:

- a warning if suspend/resume times are less than total number of runs
- the average times to suspend/resume for each cycle

Moreover, the script (and therefore the
`suspend-30-cycles-time-check-with-reboots` job) will fail if one or more
times are not reported correctly, which will allow QA team to
investigate.

Finally, the attachment jobs were renamed to add `pm_test.reboot.3.log` and
`pm_test.poweroff.3.log` to its filename to make it more clear for anyone who will debug the submission.

Note that I think there is really an issue with FWTS, since it seems it cannot report the time on newer configs. I've raised an issue about this: https://github.com/ColinIanKing/fwts/issues/8

## Resolved issues

#99 

## Tests

- Run the `stress-suspend-30-cycles-with-reboots-automated` stress test plan

Check that everything runs as expected.

Namely, on a device that fails to provide suspend or resume time, the output of `power-management/suspend-30-cycles-time-check-with-reboots` will look like this, and the job will fail to let users investigate the issue:

```
Warning: Time to resume was only reported 0 times out of 30 iterations!

Average time to sleep: 0.68653

Average time to resume: N/A

ERROR: One or more resume times was not reported correctly:
could not convert string to float: 'N/A'
Warning: Time to resume was only reported 0 times out of 30 iterations!

Average time to sleep: 0.73510

Average time to resume: N/A

ERROR: One or more resume times was not reported correctly:
could not convert string to float: 'N/A'
Warning: Time to resume was only reported 0 times out of 30 iterations!

Average time to sleep: 0.72083

Average time to resume: N/A

ERROR: One or more resume times was not reported correctly:
could not convert string to float: 'N/A'

=================================================
Average time to enter sleep state: 0.7142 seconds
=================================================
```

### DIFF
--- a/checkbox-support/checkbox_support/scripts/fwts_test.py
+++ b/checkbox-support/checkbox_support/scripts/fwts_test.py
@@ -113,21 +113,37 @@ def get_sleep_times(log, start_marker):
 
 
 def average_times(runs):
+    sleep_run_count = 0
     sleep_total = 0.0
+    resume_run_count = 0
     resume_total = 0.0
-    run_count = 0
-    try:
-        for run in runs.keys():
-            run_count += 1
-            sleep_total += runs[run][0]
-            resume_total += runs[run][1]
-        sleep_avg = sleep_total / run_count
-        resume_avg = resume_total / run_count
+    print()
+    print("{} iterations total.".format(len(runs)))
+    for i in runs.values():
+        if type(i[0]) == float:
+            sleep_run_count += 1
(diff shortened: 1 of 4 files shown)


yphus (approved):
LGTM, thanks

PR #305, opened 2023-01-16
Revamp Metabox Source Container (formerly known as Venv Container)

pieqq:
## Description

Metabox can be used to test Checkbox directly from a source code repository. I reworked this feature to make it a little bit better:

- Deploy a systemd service when running Checkbox remote testing, so that if the container running the Checkbox service part is rebooter, Checkbox is restarted as soon as the container is avaible (this makes the `restart` scenario work even in this case, in addition to PPA and Snaps)
- Deploy Checkbox source code directly into the container using pip instead of installing it in a virtual environment: since we do everything in a container, we don't really need a venv

I made some changes in some of the methods (notably `start_service()`, `stop_service()`, `reboot_service()` and `is_service_active()`), but since they are not widely used by any scenario at the moment, I'm not 100% sure of my implementation. I would appreciate feedback on those, especially.

## Resolved issues

Fix CHECKBOX-419

## Tests

I created the following config file (`source-config.py`):

```python
configuration = {
    'local': {
        'origin': 'source',
        'uri': '/home/pieq/path/to/checkbox',
        'releases': ['focal'],
    },
    'remote': {
        'origin': 'source',
        'uri': '/home/pieq/path/to/checkbox',
        'releases': ['focal'],
    },
    'service': {
        'origin': 'source',
        'uri': '/home/pieq/path/to/checkbox',
        'releases': ['focal'],
    },
}
```

and ran Metabox with

```
metabox source-config.py --log TRACE --do-not-dispose
```

Every scenario passed, except the `urwid.testplan.UrwidTestPlanSelection` because the output expected by this scenario has changed with recent commits made in Checkbox, so it will have to be modified at a later date.

### DIFF
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -61,9 +61,10 @@ class LxdMachineProvider:
         try:
             # TODO: Find a suitable timeout value here
             self.client = pylxd.Client(timeout=None)
-        except ClientConnectionFailed as exc:
-            logger.exception("Cannot connect to LXD. Is it installed?")
-            sys.exit()
+        except ClientConnectionFailed:
+            msg = "Cannot connect to LXD. Is it installed?"
+            logger.exception(msg)
+            raise SystemExit(msg)
 
     def setup(self):
         self._create_profiles()
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -61,10 +61,10 @@ class LxdMachineProvider:
         try:
             # TODO: Find a suitable timeout value here
             self.client = pylxd.Client(timeout=None)
-        except ClientConnectionFailed:
-            msg = "Cannot connect to LXD. Is it installed?"
+        except ClientConnectionFailed as exc:
+            msg = f"Cannot connect to LXD. Is it installed? {exc}"
             logger.exception(msg)
-            raise SystemExit(msg)
+            raise SystemExit(msg) from exc
 
     def setup(self):
         self._create_profiles()
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -26,6 +26,7 @@ LXD machines are containers that can run metabox scenarios in them.
 """
 import json
 import os
+import sys
 import time
 import yaml
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


metabox/metabox/core/lxd_provider.py
```diff
@@ -61,7 +62,8 @@ def __init__(self, session_config, effective_machine_config,
             # TODO: Find a suitable timeout value here
             self.client = pylxd.Client(timeout=None)
         except ClientConnectionFailed as exc:
-            raise SystemExit from exc
+            logger.exception("Cannot connect to LXD. Is it installed?")
```
kissiel, line 65: This change hides the information carried by the exception.

metabox/metabox/core/lxd_provider.py
```diff
@@ -61,7 +62,8 @@ def __init__(self, session_config, effective_machine_config,
...
+            sys.exit()
```
kissiel, line 66: This shouldn't exit with a 0 retcode.

What I recommend instead is:
```
msg = f"Cannot connect to LXD. {exc}. Is it installed?"
logger.exception(msg)
raise SystemExit(msg) from exc
```

metabox/metabox/core/machine.py
```diff
@@ -240,57 +240,88 @@ def switch_on_networking(self):
...
-            '''/home/ubuntu/checkbox-ng/venv/bin/activate\\n$@' > '''
-            '''/home/ubuntu/run.sh"''',
-            'chmod +x /home/ubuntu/run.sh',
-            'chmod +x /home/ubuntu/checkbox-ng/setup.py',
-            'chmod +x /home/ubuntu/checkbox-ng/mk-venv',
-            "bash -c 'pushd /home/ubuntu/checkbox-ng ; ./mk-venv venv'",
+        """
+        Installation from source and, if required, creation of systemd service.
+        """
+        service_file = ("[Unit]\\n"
```
kissiel, line 269: How do you feel about having a multiline string here that's dedented afterwards?

kissiel (commented):
Looks good. I'd revert the change that makes it return 0 retcode on exception. See inline.

metabox/metabox/core/lxd_provider.py
pieqq: The original code did not bubble up any error message. Metabox would just stop without any info. But as you said, `sys.exit()` is not any better!

The change I made should bubble up the info and stop the program. [`ClientConnectionFailed`](https://pylxd.readthedocs.io/en/latest/api.html?highlight=ClientConnectionFailed#pylxd.exceptions.ClientConnectionFailed) does not contain any response, so I just `raise SystemExit(msg)`.

metabox/metabox/core/machine.py
pieqq: Solved in commit [2a27407](https://github.com/canonical/checkbox/pull/305/commits/2a27407fff2090a5c22624c67551bed6488df917)

kissiel (approved):
Awesome!
Thanks for the changes! +1

PR #310, opened 2023-01-23
Metabox: Add initial scenarios to check Checkbox config

pieqq:
## Description

- Add support to send files in Metabox scenarios
- By default, scenarios will be run in every modes (local, remote). This can be overridden per scenario.
- Add initial scenarios to check Checkbox config options and precedence (see individual commits messages for more info)

## (Partially) resolved issue

CHECKBOX-425 (still missing stuff for specific config as well as other parts of the launcher)

## Tests

1.  Setup Metabox in a venv
2. Adjust `metabox/configs/source-local-config.py` to match the local Checkbox directory on your computer
3. Run:

```
metabox source-local-config.py --log TRACE --tag config --do-not-dispose
```

Output should look like: https://pastebin.ubuntu.com/p/TKScVVHydj/

### DIFF
--- a/metabox/configs/source-local-config.py
+++ b/metabox/configs/source-local-config.py
@@ -1,0 +1,10 @@
+configuration = {
+    'local': {
+        # Metabox can run tests from a local directory containing a copy of
+        # the Checkbox source code repository.
+        'origin': 'source',
+        # Path to the Checkbox source code repository
+        'uri': '~/dev/work/checkbox',
+        'releases': ['focal'],
+    },
+}
--- a/metabox/configs/source-remote-config.py
+++ b/metabox/configs/source-remote-config.py
@@ -1,0 +1,12 @@
+configuration = {
+    'remote': {
+        'origin': 'source',
+        'uri': '~/dev/work/checkbox',
+        'releases': ['focal'],
+    },
+    'service': {
+        'origin': 'source',
+        'uri': '~/dev/work/checkbox',
+        'releases': ['focal'],
+    },
+}
--- a/metabox/metabox/core/actions.py
+++ b/metabox/metabox/core/actions.py
@@ -25,7 +25,7 @@ __all__ = [
     "Start", "Expect", "Send", "SelectTestPlan",
     "AssertPrinted", "AssertNotPrinted", "AssertRetCode",
     "AssertServiceActive", "Sleep", "RunCmd", "Signal", "Reboot",
-    "NetUp", "NetDown",
+    "NetUp", "NetDown", "Put"
 ]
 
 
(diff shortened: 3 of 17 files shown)


yphus (changes requested):
The local config works well but I don't think you can use the same definitions for remote:

```
16:29:25 | INFO     Starting scenario: config.launcher.CheckboxConfXDG
checkbox-cli remote 10.31.111.195 /home/ubuntu/launcher.checkbox
Using sideloaded provider: checkbox-provider-base, version 2.1.0 from /var/tmp/checkbox-providers/base
Using sideloaded provider: checkbox-provider-certification-client, version 2.1.0 from /var/tmp/checkbox-providers/certification-client
Using sideloaded provider: 2021.com.canonical.certification:metabox, version 1.0 from /var/tmp/checkbox-providers/metabox-provider
Using sideloaded provider: checkbox-provider-resource, version 2.1.0 from /var/tmp/checkbox-providers/resource
/home/ubuntu/launcher.checkbox launcher file was not found!
16:29:26 | ERROR    [remote][('focal', 'focal')] config.launcher.CheckboxConfXDG scenario has failed.
```

I suggest to create remote-specific jobs since they all need a launcher file.
Moreover the "Put" action by default copy files to both remote and service containers. For configs though, I'd prefer to have an explicit arg as I'm not sure doing local only for example works. I'd then separate/duplicate the jobs to tests both (1. copying to local remote only 2. copying to service only).

pieqq:
@yphus I've made a bunch of changes following your comment and the discussion we had yesterday.

Logs of Metabox running both Remote and Local modes:

https://pastebin.ubuntu.com/p/9BT29P7hh3/

The scenarios for Remote mode are passing, because the "home" config file is copied over to the Service's `/root/.config/checkbox.conf` location. This is not clear from the documentation, but this is what Checkbox expects. I will update the docs to make it more clear.

One of the scenario for the Local mode, `CheckboxConfLocalResolutionOrder`, fails. It was written by following [the official documentation on the topic](https://checkbox.readthedocs.io/en/latest/configs.html#invoking-launcher). So either:

1. the documentation is wrong
2. my scenario is wrong
3. Checkbox is wrong :)

yphus (approved):
This first set of config scenarios already proved to be extremely helpful to identify   inconsistencies between the code and the doc.

Indeed the doc says: HOME > LAUNCHER > XDG
But the [code](https://github.com/canonical/checkbox/blob/main/checkbox-ng/checkbox_ng/config.py#L41) (at least for local invocations) does:  HOME  > XDG > LAUNCHER

Let's land the scenario as is and confirm if the config-refactor code follows the existing logic or the documentation.

PR #313, opened 2023-01-25
Add new wireless jobs to check the WPA3 standard

yphus:
## Description

New wireless template jobs to test WPA3 connections
Both NetworkMamanger and Netplan test plans are updated.

The new jobs are based on `connection_wpa_ax` and are sharing the same requirements on WIFI6.
That should prevent this set of jobs to run on incompatibles releases of ubuntu.

Note:

Checkbox configs have to provide two new entries:
- WPA3_AX_SSID
- WPA3_AX_PSK

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-177


jocave:
We need to update the jenkins jobs definitions, but could we also add defaults to the configs_vars files that are shipped in the little snaps? e.g. https://github.com/canonical/checkbox/blob/main/checkbox-snap/series_classic22/config_vars#L11

jocave (approved):
Looks good, thank you

PR #325, opened 2023-02-09
Avoid running the lsblk command in unit tests

yphus:
## Description

A fix to the unit tests of the udev parser.
W/o an extr arg, the output of `lsblk` is used to evaluate if partitions have to be categorized as such by the parser.

## Resolved issues

Recent instability with Tox, see https://github.com/canonical/checkbox/actions/runs/4128463822/attempts/1

## Tests

Tested on python 3.5.2/3.5.10//3.6/3.8/3.10 in LXC


pieqq (approved):
LGTM!

PR #326, opened 2023-02-09
Store DMI ID key in its own field

jocave:
Making this change posthumously as there is use case where we would like to be able to identify processor micro-architectures from machine reports already in the HEXR db. The field would have allowed this to be worked out but it seems to get squashed by the serial key.


yphus (approved):
+1

PR #330, opened 2023-02-14
Remove imagemagick requirement from Debian packages

pieqq:
No tests depend on imagemagick, there does not seem to be a need for it.

This is done after the imagemagick CVEs (https://ubuntu.com/security/notices/USN-5620-1 and https://ubuntu.com/security/notices/USN-5736-2) that cause troubles in some automated tests by the SWE team.


kissiel (approved):
I really hope this won't break anything :)
+1

PR #336, opened 2023-02-23
Update Debian Release README

pieqq:
## Description

Update Debian release process README with a few checkups to do along the way.

(Also modified the markdown of existing paragraphs to fit 80 chars)


yphus (approved):
LGTM

PR #338, opened 2023-02-23
Restore test plan filters in all odm-certification launchers

yphus:
## Description

Add the full list of ODM test plans in odm-certification launchers

## Resolved issues

The SUT can run a different version of ubuntu (Core or classic) than the remote side. Presuming the test plan selection had to match the remote os version was a mistake as it prevents the selection of the correct and targeted OS.


jocave (approved):
Looks good, thanks

PR #349, opened 2023-03-06
Upload snapcraft logs as an artefact if snapcraft build fails

pieqq:
Snapcraft 7.x stopped outputing everything to stdout and instead puts build logs in a `~/.cache/snapcraft/log/` directory. This is not ideal when running snapcraft as part of an automated workflow, because errors are not visible in stdout anymore, instead simply stating:

Logging execution to '/home/runner/.cache/snapcraft/log/xxx.log'

Add a step to capture these log files as artefact and upload them to the Github Action if the build fails.

Fixes CHECKBOX-445


kissiel (approved):
LGTM, +1

pieqq:
Should I also do the same for the other snap builds? (daily builds, checkbox snap...)?

pieqq:
Added the same step for the checkbox core daily builds.

I guess the checkbox snap is smaller and does not really require this, so I haven't added this step for it. Let me know if I should.

PR #350, opened 2023-03-06
Add: EEPROM test

rickwu666666:
## Description

Describe your changes here:

1. To cover EEPROM read and write test. Since some projects came with EEPROM support.

1. Using manifest to determine if the platform supports EEPROM or not.
2. Check if EEPROM is probed as a system device. And return exit code 1 if didn't find any EEPROM device.
3. Generate a 10 byte long random data, and write it into EEPROM
4. Read back the date from EEPROM and check if written data and read-back data are consistent.

## Resolved issues

Note the Jira/Launchpad issue(s) resolved by this PR (`Fixes|Resolves ...`).

## Documentation

- [ ] Automated tests are included for the changed functionality in this PR. If to be merged without tests, please elaborate why.

1. [Side load result](https://pastebin.canonical.com/p/BTRDZrpshq/) with exclude manifest, since manifest will always be "False" if using side load provider.

- [ ] Necessary documentation is provided for the changed functionality in this PR (tests are documentation, too).

## Tests

- [ ] Steps to follow for reviewer to run & manually test are provided.

1. `checkbox.checkbox-cli run com.canonical.certification::eeprom-automated`, and the manifest part should be marked if in the side load environment.

- [ ] A list of what tests were run and on what platform/configuration is provided.
1. [Shiner JACE9000](https://certification.canonical.com/hardware/202211-30889/submission/293753/test/151783/result/30451535/)

### DIFF
--- a/providers/base/units/eeprom/jobs.pxu
+++ b/providers/base/units/eeprom/jobs.pxu
@@ -7,45 +7,4 @@ flags: also-after-suspend
 estimated_duration: 5
 imports: from com.canonical.plainbox import manifest
 requires: manifest.has_eeprom == 'True'
-command:
-  eeprom_path=$(find /sys/devices/platform/ -name eeprom)
-  if [ ! "$eeprom_path" ];
-  then
-    echo "Error: Can't find EEPROM in system!!"
-    exit 1
-  fi
-  status=0
-  # This for loop is just in case multiple EEPROMs are on the platform. 
-  for eeprom in $eeprom_path
-  do
-    # Generates randon data for test. 
-    data=$(echo $RANDOM | md5sum | head -c 10)
-    echo "Write data ${data} into ${eeprom}"
-    if (echo "$data" > "$eeprom")
-    then
-      echo "Write data ${data} successfully"
-      echo "Read data back from EEPROM"
-      # Dump the content of EEPROM as a log.
-      od -c < "$eeprom"
-      # The content of EEPROM should be something like this:
-      # root@ubuntu:~# cat "$eeprom_path"|od -c
-      # 0000000   1   2   e   8   6   0   6   7   0   9  \n 377 377 377 377 377
-      # 0000020 377 377 377 377 377 377 377 377 377 377 377 377 377 377 377 377
-      # *
-      # ...
-      # So if we capture the first line of content for the EEPROM.
-      # It will be 12e8606709, and it is exactly the data we put in.
-      read_data=$(awk 'NR==1{printf $i}' "$eeprom")
-      if [ "$data" == "$read_data" ];
-      then
-        echo "Read back data ${read_data} correct!"
-      else
-        echo "Read back data ${data} fail"
-        status=1
-      fi
-    else
-      echo "Write data ${data} fail"
-      status=1
-    fi
-  done
-  exit "$status"
+command: eeprom_test.sh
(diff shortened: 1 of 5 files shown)


providers/base/units/eeprom/jobs.pxu
```diff
@@ -0,0 +1,39 @@
...
+  status=0
+  for eeprom in $eeprom_path
+  do
+    data=$(echo $RANDOM | md5sum | head -c 10)
+    echo "Write data ${data} into ${eeprom}"
+    if (echo "$data" > "$eeprom")
+    then
+      echo "Write data ${data} successfully"
+      echo "Read data back from EEPROM"
+      read_data=$(cat < "$eeprom" | od -c | awk 'NR==1{for(i=2;i<=11;i++) printf $i}')
```
kissiel, line 26: I bet this was hard to write. Could you add some comment explaining what it does and why?

providers/base/units/eeprom/jobs.pxu
```diff
@@ -0,0 +1,39 @@
...
+_summary: Test EEPROM read and write functions for the system. 
+plugin: shell
+user: root
+category_id: eeprom 
+flags: also-after-suspend
+estimated_duration: 5
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_eeprom == 'True'
+command:
+  eeprom_path=$(find /sys/ -name eeprom)
```
kissiel, line 11: This unnecessarily broad, and will cause some devices like dragonboard to crash the system.
I recommend going for a much narrower search path.

kissiel (changes requested):
See inline.

providers/base/units/eeprom/jobs.pxu
rickwu666666: Hi @kissiel,
The data read from EEPROM is like [this](https://pastebin.canonical.com/p/d3YfgRdVvw/).
And read with `od -c` will be like [this](https://pastebin.canonical.com/p/s8wjMYrNrh/), and this data becomes readable by humans.
Therefore, this command `read_data=$(cat < "$eeprom" | od -c | awk 'NR==1{for(i=2;i<=11;i++) printf $i}')` handle the first line of data and binding then from column 2 to 11 since we write 10 byte long data into EEPROM.
And this part of the code could be simplified to something like `awk 'NR==1{printf $1}' "$eeprom"` and it will get the same result. 
I would like to change this part to something like this.
`cat "$eeprom" | od -c ` dump human-readable data from EEPROM as a log for this job.
And simplify read_data to `read_data=$(awk 'NR==1{printf $1}' "$eeprom")`

providers/base/units/eeprom/jobs.pxu
rickwu666666: I will change the path to "/sys/devices/platform/soc/".

providers/base/units/eeprom/jobs.pxu
```diff
@@ -0,0 +1,51 @@
...
+  if [ ! "$eeprom_path" ];
+  then
+    echo "Error: Can't find EEPROM in system!!"
+    exit 1
+  fi
+  status=0
+  # This for loop is just in case multiple EEPROMs are on the platform. 
+  for eeprom in $eeprom_path
+  do
+    # Generates randon data for test.
```
pieqq, line 21: typo: `random` :P

pieqq (changes requested):
Thanks for this!

One last suggestion (and a little typo).

pieqq (approved):
+1

Thanks a lot Rick!!!

PR #354, opened 2023-03-08
Fix Metabox runs on bionic

pieqq:
## Description

- Fix call to pip when running in bionic containers.
- Comment out sections of the `UrwidTestPlanSelection` Metabox scenario to make it pass, until we can work out a clearer scenario for this use case

## Resolved issues

CHECKBOX-451

## Tests

Tested on 18.04, 20.04 and 22.04 by creating a config such as

```
# e.g. source-22.04.py
configuration = {
    'local': {
        'origin': 'source',
        'releases': ['jammy'],
    },
    'remote': {
        'origin': 'source',
        'releases': ['jammy'],
    },
    'service': {
        'origin': 'source',
        'releases': ['jammy'],
    },
}
```

and calling 

```
metabox configs/source-22.04.py --log TRACE --do-not-dispose
```


yphus (approved):
Good approach, the test plan/urwid scenarios indeed require more work

PR #356, opened 2023-03-09
Upload snapcraft remote-build logs for checkbox-core-snap and checkbox-snap even if no failure raised

pieqq:
Since snapcraft remote-build does not actually fail if only one snap fails to build (see [this example](https://github.com/canonical/checkbox/actions/runs/4370668139/jobs/7645803544#step:6:281)), the logs upload step is never triggered.

Changing that, and also add the same workflow to the checkbox-snap, because why not.


pieqq:
self-approved to run the jobs again and check the results.

PR #362, opened 2023-03-14
Change order of stress-ng-cert-automated nested parts in 22.04 desktop test plan

pieqq:
There was a little misshap and Vincent's changes in commit 57b6c69ceb015af879fbd5459c90edbc9d43ad9e were partially reverted.

stress-ng-cert-automated is moved to the beginning of the stress tests section because it contains the mechanism to modify oom-killer configuration (and restore it at the end) and it does not play well with later parts which may occur after a reboot[1].

[1] https://github.com/canonical/checkbox/pull/297/files#r1067766303


weizhenwu (approved):
LGTM, thanks for the quick fix!

PR #364, opened 2023-03-14
Add section about version bump in Debian release process readme

pieqq:


yphus:
I've updated the release workflows to start supporting a unique version number for both snaps and debs. The `bumpversion` step is now the first and has its own workflow (https://github.com/canonical/checkbox/actions/workflows/bumpversion.yml).

An updated release readme will describe the simplified process.

PR #373, opened 2023-03-17
Run Metabox only for PR affecting checkbox-ng or metabox directories

pieqq:
Metabox should only run for changes in checkbox-ng proper, or in metabox itself.

Add a `metabox_run_required` job to check the difference(s) between the main branch and the current branch's `checkbox-ng` and `metabox` directories.

Make the main Metabox job require `metabox_run_required` job to succeed before it is triggered.

## Resolved issues

Fixes CHECKBOX-469

### DIFF
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -7,6 +7,26 @@ on:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


yphus (approved):
LGTM

PR #377, opened 2023-03-20
Metabox workflow conditional trigger

pieqq:
Previous strategy would mark the whole job as fail and send
notifications.

Instead, watch for some variable to be set to continue the Metabox run.

## Resolved issues

Fix CHECKBOX-469


yphus (dismissed):
LGTM

yphus (approved):
New approval to test the workflow pre-condition step

PR #379, opened 2023-03-20
Fix: Always run job commands with SYSTEMD_IGNORE_CHROOT=1

yphus:
## Description

A follow-up to https://github.com/canonical/checkbox/pull/316

Recent snapd beta (rev 18745) makes all commands relying on systemd to silently fail when called from the checkbox service with the following error msg:

```
----------------------------[ Running job 21 / 21 ]-----------------------------
---------------------------[ Perform warm reboot 2 ]----------------------------
ID: com.canonical.certification::warm-boot-loop-reboot2
Category: Warm-boot Stress Test
Running in chroot, ignoring request.
```
This patch applies the fix already deployed to select the correct restart strategy by setting SYSTEMD_IGNORE_CHROOT=1 for all jobs commands (user: root or not)

### DIFF
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -275,6 +275,9 @@ class UnifiedRunner(IJobRunner):
                                             nest_dir)
             if self._user_provider():
                 env['NORMAL_USER'] = self._user_provider()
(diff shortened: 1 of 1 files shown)


pieqq (approved):
+1

PR #381, opened 2023-03-21
Continue Metabox workflow when the check_diff step returns 1

pieqq:
The check_diff step returns 1 when there are changes in the metabox and checkbox-ng dir. This is expected, and it should trigger the following step (to actually run Metabox).

Adding continue-on-error to make sure the workflow doesn't stop there.

See [Github docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) for more info.


yphus (approved):
+1

PR #419, opened 2023-04-06
add host-enabled bt test

kissiel:


p-gentili (approved):
LGTM. I've also tried it sideloading the provider in our SRU.

PR #425, opened 2023-04-12
Add support for fish shell

pieqq:
Add $PROVIDERPATH to activate.fish so that people using this shell can contribute to Checkbox as well.

This is the (partial) backport of a [patch](https://code.launchpad.net/~pieq/checkbox-ng/+git/checkbox-ng/+merge/426281) I submitted last year on Launchpad that didn't make it, but it's still a pain for fish users to deploy Checkbox locally, so I submit it again :)

### DIFF
--- a/checkbox-ng/mk-venv
+++ b/checkbox-ng/mk-venv
@@ -21,5 +21,8 @@ echo "set -gx PROVIDERPATH $venv_path/share/plainbox-providers-1" >> $venv_path/
 
 echo "virtualenv with an empty provider database created at $venv_path"
 echo "To activate your virtualenv run:"
-echo "$ . $venv_path/bin/activate"
-
+if [ "$(basename "$SHELL")" = "fish" ]; then
+    echo "$ . $venv_path/bin/activate.fish"
+else
+    echo "$ . $venv_path/bin/activate"
+fi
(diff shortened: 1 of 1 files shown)


kissiel:
I think having the activate script check the $SHELL and do the right thing instead of having an additional file is much cleaner.

pieqq:
There is no "additional file", because Python's venv already creates the `activate.fish` file. I think it's better to add a reference to `$PROVIDERPATH` to both `activate` and `activate.fish`.

But I had thought about using $SHELL to display the correct prompt, so I'll work on that.

pieqq:
`mk-venv` now provides the right command to run if you're using fish.

FYI, this is the default `activate.fish` provided by virtualenv:

https://github.com/pypa/virtualenv/blob/main/src/virtualenv/activation/fish/activate.fish

I also fixed stuff suggested by shellcheck in a separate commit.

kissiel:
> There is no "additional file", because Python's venv already creates the `activate.fish` file. I think it's better to add a reference to `$PROVIDERPATH` to both `activate` and `activate.fish`.
> 
> But I had thought about using $SHELL to display the correct prompt, so I'll work on that.

Oh, TIL, thanks!

kissiel (approved):
Awesome!
+1

PR #426, opened 2023-04-13
Require user to provide comment for manually skipped or failed cert-blockers

### LINKED ISSUE #387: Make sure a comment is present when a cert-blocker job is manually skipped/failed
### Enhancement Proposal

In Checkbox, a test plan is made of a series of jobs (test cases). Each job in this a test plan can be set as "certification blocker", which means that if this job fails or is skipped, it might block the issuance of a "compatible with Ubuntu" certificate.

Sometimes, certification blocker jobs are still manually marked as failed or skipped, and sometimes the tester does not provide an explanation as of why it has been marked so. This raises problems down the line, when the Certification team reviews the test results, because it is not clear why such an important job has been manually set to failed or skipped.

Since Checkbox also has a comment feature (allowing testers to add a comment on each manual test, regardless of the outcome), the proposal is to check for a comment to be present when a certification blocker job is set to failed or skipped.

pieqq:
## Description

When issuing certificates for a device, the team reviews the different test reports generated. Jobs that have a certification status set to "blocker" must pass in order to issue a review.

Sometimes, it is not the case, but users do not necessarily provide a comment that would help with the review.

Semi-automated jobs ([`user-interact` and `user-interact-verify`](https://checkbox.readthedocs.io/en/latest/units/job.html#job-fields)) can be skipped before being started or after they've run.

Manual jobs can be skipped after been "run", meaning after their content has been displayed to the user and the user has to pick an outcome.

These are handled in

- checkbox_ng.launcher.stages._run_single_job_with_ui_loop
- checkbox_ng.launcher.stages._interaction_callback

Checkbox local and Checkbox remote call _interaction_callback() differently; Checkbox remote does not provide the current job's state, which is required to know the job's effective certification status (this status can be set in the job definition, but overriden in the test plan). checkbox_ng.launcher.stages._interaction_callback() signature is modified to pass the job state.

Fix #387

Demo screenshot showing a cert-blocker being manually skipped/failed, and Checkbox asking the user to provide a comment

## Resolved issues

CHECKBOX-484

## Tests

Test that any interactive jobs (`manual`, `user-interact` and `user-interact-verify`) marked as `blocker` in a sample test plan cannot be skipped. An error message is displayed until the user adds a comment.

To do so, create a sample test plan such as:

```
id: gh387
_name: Check that interactive cert-blockers cannot be skipped/failed without a comment
unit: test plan
include:
    smoke/user-interact          certification-status=blocker
    smoke/user-interact-verify          certification-status=blocker
    smoke/manual          certification-status=blocker
``` 

Tests need to be run with Checkbox local and Checkbox remote.

**Note:** The sample test plan above only contain `blocker` jobs, but you can try to set some to normal jobs and make sure they can be skipped or failed without requiring a comment.

#### During a session run

Run:

```
checkbox-cli run com.canonical.certification::gh387
```

and try to mark one of the job as skipped or failed. Checkbox should display an error message asking you to add a comment first.

#### When resuming a session

1. Run `checkbox-cli` and select the test plan you created.
2. When prompted, select `q => save the session and quit`
3. Restart checkbox-cli, select `r => resume this session`. A message should be displayed, such as:

```
Previous session run tried to execute job: com.canonical.certification::smoke/manual`.
What do you want to do with that job?
  c => add comment
  s => skip that job
  p => mark it as passed and continue
  f => mark it as failed and continue
  r => run it again
```

4. Try to mark the job as failed or skipped. Checkbox should display an error message and ask you to enter a comment first. 

*(Unrelated: there is a bug with the `run it again` option, see #428)*

### Checkbox Remote

Prepare a launcher:

```
[launcher]
app_id = com.canonical.certification:checkbox
launcher_version = 1
stock_reports = text

[test plan]
unit = com.canonical.certification::gh387
forced = yes

[test selection]
forced = yes
```

Start Checkbox service in a separate process (needs to be done as root):

```
checkbox-cli service
```

then start Checkbox remote:

```
checkbox-cli remote 127.0.0.1 /path/to/your/launcher
```

Try to set some cert-blocker jobs to `fail` or `skip`: Checkbox should display an error and ask you to input a comment in order to continue.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/stages.py
+++ b/checkbox-ng/checkbox_ng/launcher/stages.py
@@ -88,15 +88,15 @@ class MainLoopStage(CheckboxUiStage):
 
 
     def _run_single_job_with_ui_loop(self, job, ui):
+        job_state = self.sa.get_job_state(job.id)
         print(self.C.header(job.tr_summary(), fill='-'))
         print(_("ID: {0}").format(job.id))
         print(_("Category: {0}").format(
-            self.sa.get_job_state(job.id).effective_category_id))
+            job_state.effective_category_id))
         comments = ""
         while True:
             if job.plugin in ('user-interact', 'user-interact-verify',
                               'user-verify', 'manual'):
-                job_state = self.sa.get_job_state(job.id)
                 if (not self.is_interactive and
                         job.plugin in ('user-interact',
                                        'user-interact-verify',
@@ -132,13 +132,30 @@ class MainLoopStage(CheckboxUiStage):
                             comments += new_comment + '\n'
                         continue
                     elif cmd == 'skip':
-                        result_builder = JobResultBuilder(
-                            outcome=IJobResult.OUTCOME_SKIP,
-                            comments=_("Explicitly skipped before"
-                                       " execution"))
-                        if comments != "":
-                            result_builder.comments = comments
-                        break
+                        if (
+                            job_state.effective_certification_status == "blocker"
+                            and comments == ""
+                        ):
+                            print(
+                                self.C.RED(
+                                    _("This job is required to issue a certificate.")
+                                )
+                            )
+                            print(
+                                self.C.RED(
+                                    _(
+                                        "Please add a comment to explain why you want to skip it."
+                                    )
+                                )
+                            )
+                            continue
+                        else:
+                            result_builder = JobResultBuilder(
+                                outcome=IJobResult.OUTCOME_SKIP,
+                                comments=_("Explicitly skipped before execution"))
+                            if comments != "":
+                                result_builder.comments = comments
+                            break
                     elif cmd == 'quit':
                         raise SystemExit()
                 else:
@@ -153,14 +170,14 @@ class MainLoopStage(CheckboxUiStage):
                     if comments != "":
                         result_builder.comments = comments
                     ui.notify_about_verification(job)
-                    self._interaction_callback(job, result_builder)
+                    self._interaction_callback(job, job_state, result_builder)
                 except ReRunJob:
                     self.sa.use_job_result(job.id, result_builder.get_result())
                     continue
             break
         return result_builder
 
-    def _interaction_callback(self, job, result_builder,
+    def _interaction_callback(self, job, job_state, result_builder,
                               prompt=None, allowed_outcome=None):
         result = result_builder.get_result()
         if prompt is None:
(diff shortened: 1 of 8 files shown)


yphus:
Using the `checkbox-cli remote 127.0.0.1` trick to test remote does not allow to validate the restart strategy properly. Ideally you'll have to have a same systemd service to fake a deb install and then trigger the corresponding strategy. Otherwise a new session is created everytime you're killing the service. 

At this point coming up with the same unit you already created for Metabox to get a systemd service running from a local source seems a bit redundant just for a manual verification. I suggest to leverage the metabox `ContainerSourceMachine` to create a nes scenario where the reboot test for example could be turned into a user-verify job (with the two flags `noreturn` and `autorestart`)

checkbox-ng/checkbox_ng/launcher/stages.py
```diff
@@ -132,13 +132,30 @@ def _run_single_job_with_ui_loop(self, job, ui):
...
                         continue
                     elif cmd == 'skip':
-                        result_builder = JobResultBuilder(
-                            outcome=IJobResult.OUTCOME_SKIP,
-                            comments=_("Explicitly skipped before"
-                                       " execution"))
-                        if comments != "":
-                            result_builder.comments = comments
-                        break
+                        if (
```
kissiel, line 135: That piece of code is already a behemoth.
How difficult it would be to outline that function? I know it interacts with the for loop it's in, but the outlined function could returned something like "dealt_with" that could be used to continue/break.

checkbox-ng/checkbox_ng/launcher/stages.py
```diff
@@ -212,9 +229,33 @@ def _interaction_callback(self, job, result_builder,
             if cmd == 'set-pass':
                 result_builder.outcome = IJobResult.OUTCOME_PASS
             elif cmd == 'set-fail':
-                result_builder.outcome = IJobResult.OUTCOME_FAIL
+                if (
+                    job_state.effective_certification_status == "blocker"
+                    and not isinstance(result_builder.comments, str)
```
kissiel, line 234: This is mad. And i'm not talking about this code but the necessity to have it. I'm guessing this is because of the pod madness...

kissiel (approved):
That part of checkbox codebase is nerve wracking for both the author and the reviewer. I guess the code in this PR is the best we can do without having to do a big revolution...
I left some comments below. Good to land as is. +1

PR #431, opened 2023-04-19
Fix: graphics/nvlink-status-check always failed while less than 2 graphic cards

LiaoU3:
## Description
Nowadays, graphics/nvlink-status-check will always failed while there are only 1  graphic card. However, it shouldn’t fail. Instead, it should be skipped. Because nvlink is based on more than 1 Nvidia GPU.
I create a new resource job for nvidia graphic card to count the nv card in the system. And set it as a requirement for `graphics/nvlink-status-check`.

Note: This job only run on sutton.
## Resolved issues
 - This job will failed when there's only 1 graphic card.
## Tests
Run
```
checkbox-cli run com.canonical.certification::graphics/nvlink-status-check
```
## Submissions
 - 1 NV Card: https://certification.canonical.com/hardware/202303-31331/submission/313692/
 - 2 NV Cards: https://certification.canonical.com/hardware/202303-31333/submission/313388/

### DIFF
--- a/providers/base/bin/nvidia_nvlink_check.sh
+++ b/providers/base/bin/nvidia_nvlink_check.sh
@@ -8,7 +8,7 @@ if [ -z "$O" ]; then
     echo "System does not support NVLINK."
     exit 1
 # If any inactive in output that means NVLINK not connected porpery or malfunction, -n use for verify output
-elif echo "$O" | grep -q inactive; then
+elif echo "$O" | grep -iq inactive; then
     echo "NVLINK either the bridge is missing/defective or not configured correctly.";
     exit 1
 else
--- a/providers/base/bin/nvidia_nvlink_check.sh
+++ b/providers/base/bin/nvidia_nvlink_check.sh
@@ -3,16 +3,16 @@
 set -e
 
 O=$(nvidia-smi nvlink -s)
-#Check NVLINK are avaliable first, if only 1 NVIDIA Graphic card, output will be empty.
-if [ -z "$O" ];then
-    echo "System does not support NVLINK or Only 1 NVIDIA graphic card installed";
+# If the system doesn't support NVLINK, the output will be None.
+if [ -z "$O" ]; then
+    echo "System does not support NVLINK."
     exit 1
-#If any inactive in output that means NVLINK not connected porpery or malfunction, -n use for verify output
+# If any inactive in output that means NVLINK not connected porpery or malfunction, -n use for verify output
 elif echo "$O" | grep -q inactive; then
-    echo "NVLINK either the bridge is missing/defective or not configured correctly";
+    echo "NVLINK either the bridge is missing/defective or not configured correctly.";
     exit 1
 else
     #"-e" use for new line character in string"
-    echo -e "NVLINK are Supported!\nStatus as below:\n\n""$O"
+    echo -e "NVLINK Supported!\nStatus as below:\n\n""$O"
     exit 0
 fi
--- a/providers/base/units/graphics/jobs.pxu
+++ b/providers/base/units/graphics/jobs.pxu
@@ -544,6 +544,7 @@ id: graphics/nvlink-status-check
 requires:
  dmi.product in ['Desktop','Low Profile Desktop','Tower','Mini Tower', 'Space-saving']
  graphics_card.driver == 'nvidia'
+ nv_graphics_card.nv_card_count >= "2"
 command: nvidia_nvlink_check.sh
 _summary: Check NVIDIA NVLINK status
 _description: Check NVLINK are supported and NVLINK are connected properly on system, please make sure below items before testing:
--- a/providers/resource/bin/nv_card_resource.sh
+++ b/providers/resource/bin/nv_card_resource.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+echo "nv_card_count: $(lspci | grep NVIDIA | grep VGA -c)"
+exit 0
--- a/providers/resource/jobs/resource.pxu
+++ b/providers/resource/jobs/resource.pxu
@@ -318,6 +318,12 @@ plugin: resource
 _summary: Generate an entry for each graphics card present in the system.
 command: graphics_card_resource.py
 
+id: nv_graphics_card
+estimated_duration: 0.05
+plugin: resource
+_summary: Generate an entry for NVIDIA graphics card present in the system.
+command: nv_card_resource.sh
+
 id: fwts
 estimated_duration: 0.5
 plugin: resource


providers/base/units/graphics/jobs.pxu
```diff
@@ -544,6 +544,7 @@ id: graphics/nvlink-status-check
 requires:
  dmi.product in ['Desktop','Low Profile Desktop','Tower','Mini Tower', 'Space-saving']
  graphics_card.driver == 'nvidia'
+ graphics_card.gpu_count >= "2"
```
yphus, line 547: I don't think it will work as expected. The `gpu_count` provided by this resource job includes both discrete and integrated GPU, like on this platform: https://ubuntu.com/certified/202212-30934

One Nvidia card was present bu the resource reported two GPU:

```
bus: pci
category: VIDEO
driver: i915
gpu_count: 2
index: 1
path: /devices/pci0000:00/0000:00:02.0
prime_gpu_offload: Off
product: PCI ID 0xa7a0
product_id: 42912
product_slug: PCI_ID_0xa7a0
subproduct_id: 3074
subvendor_id: 4136
switch_to_cmd: prime-select intel
vendor: Intel Corporation
vendor_id: 32902
vendor_slug: Intel_Corporation

bus: pci
category: VIDEO
driver: nvidia
gpu_count: 2
index: 2
path: /devices/pci0000:00/0000:00:06.2/0000:01:00.0
prime_gpu_offload: Off
product: PCI ID 0x25bb
product_id: 9659
product_slug: PCI_ID_0x25bb
subproduct_id: 3074
subvendor_id: 4136
switch_to_cmd: prime-select nvidia
vendor: NVIDIA Corporation
vendor_id: 4318
vendor_slug: NVIDIA_Corporation
```

yphus (changes requested):
See my comment below, the gpu count from your submissions was correct because the CPU was a Xeon w/o Intel UHD/Iris. I suggest to drop this extra requirement.

providers/base/units/graphics/jobs.pxu
LiaoU3: @yphus 
You are right. This resource job count all gpu on the DUT, including iGPU and dGPU.
What do you think if I modify this jobs to let it output the count od iGPU and dGPU?

yphus: @LiaoU3 it would be extremely helpful indeed. If you know how to do it please go ahead and add new key/value property to the [resource job](https://github.com/canonical/checkbox/blob/main/providers/resource/bin/graphics_card_resource.py) output.

pieqq: This would be very nice, but it's gonna be tricky I think.

- On a desktop with an AMD graphics card, is this AMD GPU an "integrated" GPU or a "discrete" one?
- You could think that Intel GPUs are always integrated, but there are now Intel GPUs such as [Intel Arc A750](https://www.phoronix.com/review/intel-a750-a770-arc-linux) that are separate graphics cards...

Also, the way Checkbox handles resource jobs might not be what you expect.

In the sample above, the device has one Intel GPU and one Nvidia GPU. `graphics_card.driver == 'nvidia'` does **not** mean "run this job only for the graphics card that uses the nvidia driver", it means "run this job if **any** of the grpahics cards present on this device uses the nvidia driver". It's a subtle difference, but it makes a big difference when you're writing jobs...

LiaoU3: This job is to test the nvlink, so maybe I could have the resource job count all the connected nvidia card. And also it's easy to fulfill,
```
ubuntu@ubuntu:~/oem-qa-checkbox-installer$ nvidia-smi -L
GPU 0: NVIDIA RTX A4500 (UUID: GPU-5b1b2f62-281c-fcda-59a8-65885b5df254)
GPU 1: NVIDIA RTX A4500 (UUID: GPU-405c9741-e9ea-1fd8-ff03-d01cf4d71fc8)
```
```
ubuntu@ubuntu:~/oem-qa-checkbox-installer$ nvidia-smi -L | wc -l
2
```

yphus: is `nvidia-smi` always available if you install the nv drivers or is it a separate tool package? If it's there by default writing a dedicated resource job for graphics/nvlink-status-check running the above command could do it.

stanley31huang: the nvidia-smi is part of the nvidia-utils package, and it would be installed by default while installing the nvidia-driver package.
@yphus Thanks for the suggestion. @LiaoU3  let's do it this way.

LiaoU3: Now I use the command `lspci | grep NVIDIA | grep VGA -c` instead of `nvidia-smi -L | wc -l` to prevent from SW accidentally forget attatching the nvidia utility. Using the first command can solve this kind of problem and also get the correct NVIDIA GPU count.

hanhsuan:
Could the bug fixed in #553 combines with this PR?

hanhsuan:
nvidia_smi_unknow_error

This happens on laptop(AMD iGPU + Nvidia dGPU) while suspend failed. This kind of error message will make this script report the wrong result.

LiaoU3:
@hanhsuan 
I have a [resource job](https://github.com/canonical/checkbox/pull/431/files#diff-fca8f40c6a2cff2d8e042158acb3a7fa7d231e9925d18ce274621b0fb55b8893R3) to detect how many nvidia graphic cards in the system.
Could you try the command in the resource job?

hanhsuan:
@LiaoU3 
I have tested the command, the result as below:
IMG_20230705_115154

While the nvidia driver broken, it could be not working for your command in the resource job too.

providers/resource/bin/nv_card_resource.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo "nv_card_count: $(lspci | grep NVIDIA | grep VGA -c)"
+exit 0
```
kissiel, line 4: do we really need to explicitly return 0 here?

providers/resource/jobs/resource.pxu
```diff
@@ -318,6 +318,12 @@ plugin: resource
 _summary: Generate an entry for each graphics card present in the system.
 command: graphics_card_resource.py
 
+id: nv_graphics_card
+estimated_duration: 0.05
+plugin: resource
+_summary: Generate an entry for NVIDIA graphics card present in the system.
+command: nv_card_resource.sh
```
kissiel, line 325: the echo command from the script go here directly -> less code

providers/base/bin/nvidia_nvlink_check.sh
```diff
@@ -3,16 +3,16 @@
...
-if [ -z "$O" ];then
-    echo "System does not support NVLINK or Only 1 NVIDIA graphic card installed";
+# If the system doesn't support NVLINK, the output will be None.
+if [ -z "$O" ]; then
+    echo "System does not support NVLINK."
     exit 1
-#If any inactive in output that means NVLINK not connected porpery or malfunction, -n use for verify output
-elif echo "$O" | grep -q inactive; then
-    echo "NVLINK either the bridge is missing/defective or not configured correctly";
+# If any inactive in output that means NVLINK not connected porpery or malfunction, -n use for verify output
```
kissiel, line 10: This is very difficult to read. I think better English would be:
"If the word "inactive" is found in the output, it means that NVLINK is not connected properly or it is malfunctioning."
I'd remove the reference to this '-n'.

kissiel (changes requested):
Thank you for all the improvements.
Some comments below.

LiaoU3:
> nvidia_smi_unknow_error
> 
> This happens on laptop(AMD iGPU + Nvidia dGPU) while suspend failed. This kind of error message will make this script report the wrong result.

@hanhsuan
Do you have any idea how do we know the driver is broken when this thing happen?
I am planning to add another job to detect the `driver broken issue`.

hanhsuan:
The `"nv_card_count: $(lspci | grep NVIDIA | grep VGA -c)"` command is not always correct with `grep VGA` as following picture.
nv_driver_error

And I found a python package [pyNVML](https://pypi.org/project/nvidia-ml-py/) provided by NVIDIA that could get the device count and use to check the driver is working or not. The demo code I written as follow

```python
#!/usr/bin/env python3

from pynvml import *

nvmlInit()

print(f"Driver Version: {nvmlSystemGetDriverVersion()}")

deviceCount = nvmlDeviceGetCount()

print(f"Device count: {deviceCount}")

for i in range(deviceCount):
    handle = nvmlDeviceGetHandleByIndex(i)
    print(f"Device {i} : {nvmlDeviceGetName(handle)}")
    info = nvmlDeviceGetMemoryInfo(handle)
    print(f"Total memory: {info.total}")

nvmlShutdown()
```
If the driver is broken, this code will raise `NVMLError_Unknown` while get handle by `nvmlDeviceGetHandleByIndex`.
However, pyNVML isn't included by default. You might need to check with Cert team to find a way to add this.

driver error:

```sh
Driver Version: 525.105.17
Device count: 1
  File "/home/u/nv_driver_test.py", line 14, in <module>
    handle = nvmlDeviceGetHandleByIndex(i)
  File "/home/u/.local/lib/python3.10/site-packages/pynvml.py", line 2123, in nvmlDeviceGetHandleByIndex
    _nvmlCheckReturn(ret)
  File "/home/u/.local/lib/python3.10/site-packages/pynvml.py", line 896, in _nvmlCheckReturn
    raise NVMLError(ret)
pynvml.NVMLError_Unknown: Unknown Error
```

driver correct:

```sh
Driver Version: 525.116.04
Device count: 1
Device 0 : NVIDIA RTX A1000 6GB Laptop GPU
Total memory: 6442450944
```

kissiel:
@LiaoU3 
Heya! What's the plan for this branch? The status is "open", but guessing from the conversation history here, there's still some work to be done, right?

LiaoU3:
@kissiel Hi, can we add the python library `pynvml` @hanhsuan mentioned above? So we could easily detect the status of nvidiva driver.

Hook25:
Hello, we can not have a `pynvml` dependency because there is no python3.5 version on pypi (https://pypi.org/project/pynvml/8.0.1/ seems the oldest version there and it is 3.6+) and we install all dependencies from there or ppas. This may be a very bad idea, but that library states that: 

> This library is a wrapper around https://developer.nvidia.com/nvidia-management-library-nvml

maybe we can use that directly?

hanhsuan:
Yes, it could be and [nvidia-cuda-toolkit](https://launchpad.net/ubuntu/+source/nvidia-cuda-toolkit) package is needed by `nvml` for compiling the code. 

demo Makefile
```Makefile
nvtest: LDLIBS += -lnvidia-ml
```

demo code
```c
#include <stdio.h>

#include <nvml.h>

int main(){
    nvmlReturn_t result;

    unsigned int device_count, i;
    char driver[64];

    // First initialize NVML library
    result = nvmlInit();

    if (NVML_SUCCESS != result)
    {
        printf("Failed to initialize NVML: %s\n", nvmlErrorString(result));
        return 1;
    }

    result = nvmlSystemGetDriverVersion(driver, sizeof(driver)/sizeof(driver[0]));
    if (NVML_SUCCESS != result)
    {
        printf("Failed to query driver version: %s\n", nvmlErrorString(result));
        goto Error;
    }
    printf("Driver Version: %.*s\n", (int)(sizeof(driver)/sizeof(driver[0])), driver);

    result = nvmlDeviceGetCount(&device_count);
    if (NVML_SUCCESS != result)
    {
        printf("Failed to query device count: %s\n", nvmlErrorString(result));
        goto Error;
    }
    printf("Device count: %d\n", device_count);

    for (i = 0; i < device_count; i++)
    {

        nvmlDevice_t device;
        nvmlMemory_t memory;
        char name[64];

        result = nvmlDeviceGetHandleByIndex(i, &device);
        if (NVML_SUCCESS != result)
        {
            printf("Failed to get handle for device %i: %s\n", i, nvmlErrorString(result));
            goto Error;
        }

        result = nvmlDeviceGetName(device, name, sizeof(name)/sizeof(name[0]));
        if (NVML_SUCCESS != result)
        {
            printf("Failed to get name of device %i: %s\n", i, nvmlErrorString(result));
            goto Error;
        }
        printf("Device %d : %.*s\n", i, (int)(sizeof(name)/sizeof(name[0])), name);

        result = nvmlDeviceGetMemoryInfo(device, &memory);
        if (NVML_SUCCESS != result)
        {
            printf("Failed to get memory of device %i: %s\n", i, nvmlErrorString(result));
            goto Error;
        }
        printf("Total memory: %lld\n", memory.total);
    }
Error:
    result = nvmlShutdown();
    if (NVML_SUCCESS != result)
    {
        printf("Failed to shutdown NVML: %s\n", nvmlErrorString(result));
        return 1;
    }
    return 0;
}
```

demo output
```shell
Driver Version: 535.54.03
Device count: 1
Device 0 : NVIDIA RTX A2000 8GB Laptop GPU
Total memory: 8589934592
```

pieqq:
- `python3-pynvml` is available for focal and jammy
- for snaps, it can be included in the checkbox20 and checkbox22's snapcraft.yaml recipes
- for debs, one packaging meta-data per version of Ubuntu can be created, such as

```
unit: packaging meta-data
os-id: ubuntu
os-version-id: 22.04
Depends:
python3-pynvml
```

Just make sure the job(s) that use this pynvml package are not included in test plans that are run as part of 18.04 or 16.04 regression tests in the Cert lab!

LiaoU3:
Closed branch to rewrite the PR.

PR #434, opened 2023-04-26
Fix the release to stable workflow by cloning with full git history

yphus:
Using `fetch-depth: 0`
All tags are available and the following command works as expected:

`$(git describe --tags --abbrev=0 --match v*)`


pieqq (approved):
LGTM! Thanks!

PR #435, opened 2023-05-02
fix: print proper estimated runtime even when some jobs don't provide it

kissiel:
## Description
Drive-by fix for the estimated duration computation that Pierre complained about yesterday.
With this patch, if a job exists that doesn't specify the estimated_duration, Checkbox will still compute the whole estimated runtime but it will add the "at least" clause.

I tested it by creating a provider with two jobs, one with a estimated_duration defined, and one that did not have it.
I've run sessions with varying selections to see if it works properly.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/stages.py
+++ b/checkbox-ng/checkbox_ng/launcher/stages.py
@@ -236,29 +236,31 @@ class MainLoopStage(CheckboxUiStage):
         return CheckboxUI(self.C.c, show_cmd_output=show_out)
 
     def _run_jobs(self, jobs_to_run):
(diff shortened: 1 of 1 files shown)


pieqq (approved):
LGTM

I've tested it by running `com.canonical.certification::client-cert-iot-ubuntucore-22`, which has jobs with no `estimated_duration`, and in this case I get an output like:

```
========[ Running job 5 / 97. Estimated time left (at least): 1:51:00 ]=========
```

If I run checkbox with a test plan where all the jobs have an `estimated_duration`, I get something like this:

```
==============[ Running job 1 / 88. Estimated time left: 0:56:33 ]==============
```

PR #441, opened 2023-05-03
Add debugging info to the Metabox Github Action

pieqq:
The workflow does not seem to be triggered even when it should be.

Adding some debugging info to try to figure out what's wrong.


yphus (approved):
+1

PR #451, opened 2023-05-12
Fix: checkbox crash for non-existent usernames and refactor user handling

kissiel:
## Description

This patch fixes a problem where checkbox would crash when controller provided a username that doesn't exists on the controlled system. It also moves the user handling heuristics into a new module and simplifies the existing logic.

See bug on JIRA and unit tests from this patch for details.

## Resolved issues
Fixes: [CHECKBOX-597](https://warthogs.atlassian.net/browse/CHECKBOX-597)

## Documentation
The functionality doesn't change so no docs.

## Tests
In this patch I provide unit tests for the heuristics.
Moreover the following PR will have a metabox test that will test this as well.

## Testing
I tested the branch using a remote/service combo with launchers specifying random combination of existing and non-existent users.
And of course tox + metabox.

[CHECKBOX-597]: https://warthogs.atlassian.net/browse/CHECKBOX-597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/checkbox-ng/plainbox/impl/session/remote_assistant.py
+++ b/checkbox-ng/plainbox/impl/session/remote_assistant.py
@@ -40,6 +40,9 @@ from plainbox.abc import IJobResult
 
 from checkbox_ng.config import load_configs
 from checkbox_ng.launcher.run import SilentUI
+from checkbox_ng.user_utils import check_user_exists
+from checkbox_ng.user_utils import guess_normal_user
+
 
 import psutil
 
@@ -283,13 +286,17 @@ class RemoteSessionAssistant():
                 session_desc = self._launcher.session_desc
 
         self._sa.use_alternate_configuration(self._launcher)
-
         if configuration['normal_user']:
             self._normal_user = configuration['normal_user']
         else:
             self._normal_user = self._launcher.normal_user
-            if not self._normal_user:
-                self._normal_user = _guess_normal_user()
+        if self._normal_user:
+            if not check_user_exists(self._normal_user):
+                raise RuntimeError(
+                    "User '{}' doesn't exist!".format(self._normal_user)
+                )
+        else:
+            self._normal_user = guess_normal_user()
         runner_kwargs = {
             'normal_user_provider': lambda: self._normal_user,
             'stdin': self._pipe_to_subproc,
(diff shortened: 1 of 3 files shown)


pieqq (approved):
What a beautiful change! We have so little unit tests in `checkbox_ng` proper...!

Thanks for that!

checkbox-ng/plainbox/impl/session/remote_assistant.py
```diff
@@ -732,22 +739,4 @@ def exposed_cache_report(self, exporter_id, options):
...
-            return 'ubuntu'
-    with suppress(KeyError):
-        user = pwd.getpwuid(1000).pw_name
-        _logger.warning("Using `%s` user", user)
-        return user
-    with suppress(KeyError):
-        user = pwd.getpwuid(1001).pw_name
-        _logger.warning("Using `%s` user", user)
-        return user
-    raise RuntimeError(
```
yphus, line 751: The new guess_normal_user() does not raise a RuntimeError and I think just returns none if there's no normal_user in configs and no users with uid 1000/1001. So a user foo with uid 1002 would return None. As before in such case better to abort checkbox

yphus (changes requested):
A corner case not covered I think with the new guess_normal_user() version, a user foo w/ uid 1002.

checkbox-ng/plainbox/impl/session/remote_assistant.py
kissiel: I moved the checking to where the logic can handle it and propagate up. Line 294 of remote_assistant.py. This should cover that case (and an icoming metabox scenario confirms it)

kissiel: > A corner case not covered I think with the new guess_normal_user() version, a user foo w/ uid 1002.

Does it mean we have to cover 1002 too?
Is there a range we should iterate over?

yphus: I took 1002 but it could be 1042 for a user lambda as long as it exists. There's no range, 1000 and 1001 are the most common use cases. In this scenario, guess_normal_user would return None where in the current code it raises a RuntimeError

kissiel: The RuntimeError is still getting raised by the calling code tho, unless I missed something.

yphus: in `check_user_exists` yes but not `guess_normal_user`

pieqq (approved):
LGTM

yphus (approved):
+1 thanks for covering the logic with unit tests. It really helps (as always)

PR #453, opened 2023-05-12
Metabox optimizations

yphus:
## Description

Set of optimizations/fixes for Metabox:
- New Metabox workflow based on pytest (with pytest only the failed scenarios will display the corresponding execution logs.)
- ZFS storage for much faster LXD snapshots (create and restore operations)
- Adopt the certification provider namespace (instead of the 2021 variant)
- Fix the metabox audio test (no dependency on udev resources) 

Tests

Tested and deployed in the staging env: https://github.com/canonical/checkbox-staging/actions/runs/3747735301/jobs/6364279483

## Resolved issues

- The total execution time for that workflow was nearly 45 minutes. The ZFS storage itself brings down the execution time to around 10 minutes.
- Pytest provides all the command line features required to execute metabox w/o having to reinvent them (fail-fast, collect-only, run only a subset of tests, capture or not the test output, test summary, test progress %, etc...)

## Documentation

Several examples of pytest commands added to the Metabox README file.

## Tests

Tested and deployed in the staging env: https://github.com/canonical/checkbox-staging/actions/runs/4957394688

### DIFF
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -51,9 +51,9 @@ jobs:
         uses: actions/checkout@v3
       - name: Setup LXD
         uses: canonical/setup-lxd@90d76101915da56a42a562ba766b1a77019242fd
-      - name: Install pyopenssl
-        run: python3 -m pip install pyopenssl --upgrade
+      - name: Install dependencies
+        run: python3 -m pip install --upgrade pyopenssl pytest
       - name: Install Metabox
         run: python3 -m pip install -e .
       - name: Run Metabox scenarios
-        run: metabox configs/${{ matrix.mode }}-source-${{ matrix.os }}.py --log TRACE
+        run: pytest --config configs/${{ matrix.mode }}-source-${{ matrix.os }}.py --log-lvl TRACE -v --color=yes
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -29,7 +29,7 @@ jobs:
           fi
 
   Metabox:
-    if: (github.event.review.state == 'approved' || github.event_name == 'workflow_dispatch') && needs.metabox_run_required.outputs.required_run == 'true'
+    if: (github.event.review.state == 'approved' && needs.metabox_run_required.outputs.required_run == 'true') || github.event_name == 'workflow_dispatch'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -51,6 +51,14 @@ jobs:
         uses: actions/checkout@v3
       - name: Setup LXD
         uses: canonical/setup-lxd@90d76101915da56a42a562ba766b1a77019242fd
+      - name: Add ZFS storage
+        run: |
+          lxc storage list
+          lxc profile device remove default root
+          lxc storage delete default
+          lxc storage create metabox${{ matrix.os }} zfs
+          lxc profile device add default root disk path=/ pool=metabox${{ matrix.os }}
+          lxc storage list
       - name: Install dependencies
         run: python3 -m pip install --upgrade pyopenssl pytest
       - name: Install Metabox
--- a/metabox/README.md
+++ b/metabox/README.md
@@ -150,6 +150,39 @@ configuration = {
 **Note:** Metabox is always going to check **all possible combinations** of
 `releases`, that means that this example will execute 9 test runs.
 
+# Running Metabox with pytest
+
+Pytest offers a richer command-line and complements nicely the metabox runner.
+A [conftest.py] turns metabox scenarios into a pytest compatible session.
+
+### List the scenarios using keywords expression
+
+An expression is a Python evaluatable expression where all names are
+substring-matched against test names and their parent classes.
+The matching is case-insensitive.
+```
+$ pytest --config configs/source-remote-config.py -k "remote and reboot" --collect-only
+$ pytest --config configs/source-remote-config.py -k "remote and not reboot" --collect-only
+$ pytest --config configs/source-remote-config.py -k "remote and not reboot and not desktop" --collect-only
+$ pytest --config configs/testing-ppa-config.py -k "remote and not jammy" --collect-only
+$ pytest --config configs/source-local-config.py -k CheckboxConfLocalResolutionOrder --collect-only
+```
+### Run scenarios using the TRACE logging level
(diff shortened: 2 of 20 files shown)


kissiel:
quick Q before I jump head first into the code/testing: Are all those changes connected and required to land in one go?

yphus:
> quick Q before I jump head first into the code/testing: Are all those changes connected and required to land in one go?

pytest additions are a bonus but they introduce test capture and % progress.
ZFS storage can't work w/o the systemd changes.

metabox/metabox/core/runner.py
```diff
@@ -40,24 +40,33 @@ def __init__(self, args):
...
+        self.scn_variants = []
         self.machine_provider = None
         self.failed = False
-        self.tags = set(self.args.tags or [])
-        self.exclude_tags = set(self.args.exclude_tags or [])
-        self.hold_on_fail = self.args.hold_on_fail
+        try:
+            self.tags = set(self.args.tags or [])
+        except AttributeError:
+            self.tags = []
```
pieqq, line 61: I'm a bit puzzled by this.

- if no tags have been provided, `self.args.tags` will be `None`, so `self.args.tags or []` will return `[]` and `set(self.args.tags or [])` will therefore be an empty set.
- if tags are provided, then `self.args.tags` is a list of strings and therefore `set(self.args.tags or [])` is a set of strings.

I don't see how this could raise an `AttributeError`.

The other thing is that, if this exception occurs anyway, `self.tags` will be a list instead of a set. Probably not a big deal in Python, but still a bit weird to me :)

metabox/metabox/core/runner.py
```diff
@@ -40,24 +40,33 @@ def __init__(self, args):
...
+        try:
+            self.exclude_tags = set(self.args.exclude_tags or [])
+        except AttributeError:
+            self.exclude_tags = []
```
pieqq, line 65: Same as above.

metabox/metabox/core/runner.py
```diff
@@ -40,24 +40,33 @@ def __init__(self, args):
...
+        try:
+            self.hold_on_fail = self.args.hold_on_fail
+        except AttributeError:
+            self.hold_on_fail = False
```
pieqq, line 69: Again, I'm not sure what conditions are required to raise an `AttributeError` exception here. Either `--hold-on-fail` has been set in the CLI and `self.args.hold_on_fail` equals `True`, either it's not been set in the CLI and `self.args.hold_on_fail` equals False...

metabox/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def repr_failure(self, excinfo):
+        """Called when self.runtest() raises an exception."""
+        if isinstance(excinfo.value, MetaboxException):
+            err_msg = "Scenario execution failed\n    "
+            err_msg += "\n    ".join(excinfo.value.args[1])
+            return err_msg
+
+    def reportinfo(self):
+        return self.path, 0, f"Scenario: {self.name}"
```
pieqq, line 177: [The API](https://docs.pytest.org/en/latest/reference/reference.html#pytest.Item.reportinfo) mentions the default value for the line number is `None`, is there a rationale for using `0`?

metabox/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+        return "<Scenario {}>".format(getattr(self, "name", None))
+
+    def _load(self, mode, release_alias):
+        config = self.config.runner.config[mode].copy()
+        config['alias'] = release_alias
+        config['role'] = mode
+        return self.config.runner.machine_provider.get_machine_by_config(
+            MachineConfig(mode, config))
+
+    def runtest(self):
```
pieqq, line 132: this `runtest()` method and the `_load()` method above look a lot like the `run()` and `_load()` methods from [`metabox.core.runner`](https://github.com/canonical/checkbox/blob/metabox-pytest/metabox/metabox/core/runner.py#L180-L240). There might be a chance for refactoring here...

metabox/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+    elif inspect.isclass(obj) and issubclass(obj, Scenario):
+        return []
+
+
+def pytest_configure(config):
+    """
+    Initialization of the Runner object with the command-line options passed to
+    pytest.
+    """
+    logger = logging.getLogger('ws4py')
```
pieqq, line 91: Why is it the `ws4py` logger that is used?

.github/workflows/metabox.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
           fi
 
   Metabox:
-    if: (github.event.review.state == 'approved' || github.event_name == 'workflow_dispatch') && needs.metabox_run_required.outputs.required_run == 'true'
+    if: (github.event.review.state == 'approved' && needs.metabox_run_required.outputs.required_run == 'true') || github.event_name == 'workflow_dispatch'
```
pieqq, line 32: I'm still not sure if this is gonna be enough to trigger this workflow automatically when a PR impacting metabox or checkbox-ng has been approved, but at least now there is an option to trigger it manually.

metabox/README.md
```diff
@@ -150,10 +150,44 @@ configuration = {
 **Note:** Metabox is always going to check **all possible combinations** of
 `releases`, that means that this example will execute 9 test runs.
 
+# Running Metabox with pytest
```
pieqq, line 153: I'm wondering if this is a bonus to the `metabox` command, or if it should be promoted as the main way of running Metabox...

pieqq (changes requested):
Whoooo! Really great idea to use pytest to run Metabox scenarios. I love seeing the progress in % as well as all the niceties you get with pytest.

The ZFS thing in LXD is Voodoo magic to me so I'll trust you on that one :)

I left a few comments inline.

Thanks!!

kissiel:
I got a few questions about this PR

1. seeing that the scenario run is being considered a pytest item, does it mean that when there's a problem with metabox this will be swallowed by pytest and considered a failing test?
2. what about running pdb inside the logic of metabox? harnesses like pytest often capture calls like that?
3. what's the logical connection between runtime optimizations and pytest - landing this will result in one big commit that's for instance making bisects harder
4. zfs storage - why reuse an action that does something that we delete and do it differently when we should have an action that does the things the right way from the get go

metabox/conftest.py
yphus: I borrowed this `conftest.py` from pytest [official documentation](https://docs.pytest.org/en/latest/example/nonpython.html). default is `None` but how does it renders? I'll check both

PR #461, opened 2023-05-15
Fix: Import checkbox-ng version from the repo not from pypi

yphus:
## Description

Fix the Checkbox documentation version on RTD (always rendered as 1.18.1)

## Resolved issues

checkbox-ng was imported from the version installed via pip, not from source.

## Tests

Tested locally using `make install` and `make run`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,10 +1,12 @@
 import datetime
+import os
 import sys
 
-try:
-    import checkbox_ng
-except ImportError:
-    raise SystemExit("checkbox_ng has to be importable")
+sys.path.append(os.path.abspath(
+    os.path.join(__file__, "../../checkbox-ng")
+))
+
+import checkbox_ng
 
 # Configuration file for the Sphinx documentation builder.
 #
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -1,10 +1,12 @@
...
+import os
 import sys
 
-try:
-    import checkbox_ng
-except ImportError:
-    raise SystemExit("checkbox_ng has to be importable")
+sys.path.append(os.path.abspath(
+    os.path.join(__file__, "../../checkbox-ng")
+))
```
pieqq, line 7: Is the `checkbox_ng` module used for anything else other than retrieving the version number? (It might be used by `sphinx.ext.autodoc`, I'm not familiar with this) 

If not, we could just find a way to grab the version number.

pieqq (approved):
I don't know how you figured this out, but well done!

Can land as is, I just added a comment and a question.

docs/conf.py
yphus: I think it's required by autosummary and co indeed.

PR #463, opened 2023-05-16
Add dry_run config in launcher

petitfou:
Add dry run mode
    
When --dry-run argument is present in the checkbox-cli command, all the tests except attach
and resource will be skipped
    
Some invocation examples:
    
    checkbox-cli --dry-run [<launcher>]
    checkbox-cli run --dry-run <test-plan>
    checkbox-cli remote --dry-run <host> [<launcher>]
    
Tests:
      metabox metabox/configs/local-source-22.04.py --tag TestDryRunOff --tag TestDryRunOn
      metabox metabox/configs/remote-source-22.04.py --tag TestDryRunOff --tag TestDryRunOn
    
[Review#1] Make metabox scenario more robust
[Review#2] Use --dry-run option to checkbox-cli instead of launcher conf field

## Description

Recently, while debugging a SRU testplan, i found myself in a need of quickly list the jobs that might be effectively run/skipped on DUT with a given launcher and configuration, the existing list-bootstrapped command only works on local providers and on a test plan.

One solution would be to use existing (not used) dry_run mode. When the mode is enabled, all the tests (except resource, attach) will be skipped.

## Resolved issues

N/A

## Tests

  metabox metabox/configs/local-source-22.04.py --tag TestDryRunOff --tag TestDryRunOn
  metabox metabox/configs/remote-source-22.04.py --tag TestDryRunOff --tag TestDryRunOn


checkbox-ng/checkbox_ng/launcher/subcommands.py
```diff
@@ -343,6 +343,7 @@ def _start_new_session(self):
         if self.launcher.app_version:
             title += ' {}'.format(self.launcher.app_version)
         runner_kwargs = {
+            'dry_run' : self.launcher.dry_run,
```
kissiel, line 346: if it's a part of configs/launchers then this doesn't have to be pushed to the runner like this. This runner_kwargs is for callables that need to cross the RPyC boundry.

checkbox-ng/plainbox/impl/session/remote_assistant.py
```diff
@@ -291,6 +291,7 @@ def start_session(self, configuration):
             if not self._normal_user:
                 self._normal_user = _guess_normal_user()
         runner_kwargs = {
+            'dry_run' : self._launcher.dry_run,
```
kissiel, line 294: Also here.

kissiel (changes requested):
There's no actual implementation for the dry-run in this PR.
Also metabox tests from this PR do not test whether dry-run actually works.
My first impression is that there should be a job that prints something (like a `command: echo WET RUN`) and an assertion in metabox scenario that it wasn't printed.

I also don't think that having dry-run as a configuration option is a good idea. This option is for debugging/testing only, should never be a part of configuration or a launcher in production. This should be an invocation option `--dry-run`.
Preferably controlled by the controller.

checkbox-ng/plainbox/impl/session/remote_assistant.py
kissiel: Same :)

petitfou:
> There's no actual implementation for the dry-run in this PR. Also metabox tests from this PR do not test whether dry-run actually works. My first impression is that there should be a job that prints something (like a `command: echo WET RUN`) and an assertion in metabox scenario that it wasn't printed.
> 
> I also don't think that having dry-run as a configuration option is a good idea. This option is for debugging/testing only, should never be a part of configuration or a launcher in production. This should be an invocation option `--dry-run`. Preferably controlled by the controller.

Ok

metabox/metabox/core/scenario.py
```diff
@@ -168,7 +169,7 @@ def start_all(self, interactive=False, timeout=0):
 
     def start_remote(self, interactive=False, timeout=0):
         outcome = self.remote_machine.start_remote(
-            self.service_machine.address, self.LAUNCHER_PATH, interactive,
+            self.service_machine.address, self.LAUNCHER_PATH, cmd_args=self.cmd_args, interactive=interactive,
```
kissiel, line 172: Please run `black` to format the code, so we don't get long lines like this.

kissiel (commented):
This is great!
Also awesome Metabox tests.

One thing to consider is that you already noticed that the 'dry-run' should/could be set in the configuration. What this means is that it doesn't have to cross the remote barier (as the whole configuration is transmitted from the controller).
The configuration on main now has an `origin` field that in case of someone using `--dry-run` may be set to "command line arg" (to inform agent about the origin of the setting).

As for where the dry-run should be placed in configs, I think that `[ui]` is a natural fit for this.

petitfou:
The MR sits here for a long time and out of sync wrt main branch, a forced push has been done to sync up this branch, please restart review

metabox/metabox/core/scenario.py
petitfou: how do you run black on checkbox code ? when i run it, i have at the end a bunch of changes ....

petitfou:
> someone

at the first approach, i wanted to put this option in the configuration, but as you suggested at the time, i removed it and use now checkbox command argument instead, do you suggest to put it finally also in the configuration file ?

metabox/metabox/core/scenario.py
kissiel: Because we have so many PR underway we haven't run black on all of the codebase, so we're trying to reformat only the bits we touch. Preferably before we start editing them, as the first commit.

petitfou: done

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -434,6 +436,8 @@ def register_arguments(self, parser):
         parser.add_argument(
             "-u", "--user", help=_("normal user to run non-root jobs")
         )
+        parser.add_argument("-n", "--dry-run", default=False, action="store_true", help=_(
```
kissiel, line 439: - this doesn't fit line length of PEP8
- `default=False` is redundant with `action="store_true"` and can be removed

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -434,6 +436,8 @@ def register_arguments(self, parser):
...
+            "don't actually do anything"))
```
kissiel, line 440: This description doesn't convey what happens. The resource jobs still run, the relationship between jobs are still computed, and interactive jobs run normally.

More adequate description I guess would be "don't run commands associated with normal jobs".

metabox/metabox/scenarios/config/dry_run.py
```diff
@@ -0,0 +1,80 @@
...
+import textwrap
+
+from metabox.core.actions import Start
+from metabox.core.actions import Expect
+from metabox.core.scenario import Scenario
+from metabox.core.actions import AssertPrinted
+from metabox.core.actions import AssertNotPrinted
+
+from metabox.core.utils import _re
+
```
kissiel, line 28: @tag('dry-run') would be welcome.

kissiel (changes requested):
Fantastic stuff! Thank you for all the changes!
This needs a rebase on a fresh main, so we can test it with metabox.

Minor things:
- Please move the MB scenarios to some other directory like "dry-run".
- see inline for other minor things

metabox/metabox/scenarios/config/dry_run.py
petitfou: thanks ! done

petitfou:
This PR has been here for a long time, i do not see a lot of interest from checkbox maintainers, closing ...

kissiel:
There are still things pointed out in the review that you left unaddressed.
There's a clear codecov complaint about test coverage. 

Saying it's lack our interest is disingenuous.

petitfou:
I have addressed all the comments several times, please re-read the PR history, about the codecov, that is something new and i m not aware of any explicit policy on this matter for PRs,

PR #465, opened 2023-05-16
Switch to the ZFS storage driver for LXD

yphus:
## Description

ZFS storage for much faster LXD snapshots (create and restore operations).
See https://linuxcontainers.org/lxd/docs/latest/reference/storage_drivers/#feature-comparison

## Resolved issues

The total execution time for that workflow was nearly 45 minutes. The ZFS storage itself brings down the execution time to around 10 minutes.

## Tests

See https://github.com/canonical/checkbox/pull/453

### DIFF
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -50,9 +50,17 @@ jobs:
       - name: Checkout Checkbox monorepo
         uses: actions/checkout@v3
       - name: Setup LXD
-        uses: canonical/setup-lxd@90d76101915da56a42a562ba766b1a77019242fd
-      - name: Install pyopenssl
-        run: python3 -m pip install pyopenssl --upgrade
+        uses: canonical/setup-lxd@v0.1.1
+      - name: Add ZFS storage
+        run: |
+          lxc storage list
+          lxc profile device remove default root
+          lxc storage delete default
+          lxc storage create metabox${{ matrix.os }} zfs
+          lxc profile device add default root disk path=/ pool=metabox${{ matrix.os }}
+          lxc storage list
+      - name: Install dependencies
+        run: python3 -m pip install --upgrade pyopenssl
       - name: Install Metabox
         run: python3 -m pip install -e .
       - name: Run Metabox scenarios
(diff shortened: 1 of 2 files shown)


metabox/metabox/core/machine.py
```diff
@@ -94,30 +94,28 @@ def rollback_to(self, savepoint):
...
-            attempt += 1
-        else:
-            raise SystemExit(
-                "Rollback to {} failed (systemd not in running state)".format(
-                    savepoint))
+        if self.config.role == 'service':
+            attempt = 0
+            max_attempt = 60
+            old_out = ''
+            while attempt < max_attempt:
```
kissiel, line 101: This can be simplified to:
```
attempts_left = 60
out = ''
while attempts_left and out not in ('starting', 'running', 'degraded'):
   (...)
    attempts_left -= 1
```

kissiel (commented):
Some potential simplifications below.

metabox/metabox/core/machine.py
yphus: I've checked the systemctl man pages and indeed there's nothing more than the values above (except a trailing \n). They can be all checked at once.

yphus:
Re-tested locally to validate the container creation and provisioning steps (but with a local patch to bump the LXD_CREATE_TIMEOUT from 300 to 1500)

kissiel (approved):
I think there's still room for simpler (and faster ZFS-backed LXD action) but this is awesome!

.github/workflows/metabox.yaml
```diff
@@ -50,9 +50,17 @@ jobs:
       - name: Checkout Checkbox monorepo
         uses: actions/checkout@v3
       - name: Setup LXD
-        uses: canonical/setup-lxd@90d76101915da56a42a562ba766b1a77019242fd
-      - name: Install pyopenssl
-        run: python3 -m pip install pyopenssl --upgrade
+        uses: canonical/setup-lxd@v0.1.1
+      - name: Add ZFS storage
+        run: |
```
mz2, line 55: Please the very least comment the rationale for doing this, and the mechanism by which this is crucial for speed. Does the storage pool naming match something in a pre-existing implementation for example? Should the implementation _expect_ that ZFS is used?

mz2 (commented):
As requested earlier, please document the rationale.

PR #469, opened 2023-05-17
Fix: make a warning about bad secure_id more informative

kissiel:
## Description
This adds more context to the InvalidSecureIDError so we can properly test it in metabox.

## Testing:
Tested by running sessions with different secure_ids.


Hook25 (approved):
+1 looks great

PR #470, opened 2023-05-17
fix: metabox exiting on successful rollback

kissiel:
## Description

A hotfix to make metabox not quit after successfully rolling back the state of a container.

Testing was easy - running metabox :)


yphus (approved):
+1

Issue #472, opened 2023-05-17
Metabox crashes on failed Put


Hook25:
### Bug Description

If a scenario contains a `Put` action to a location that can not be written, this will raise an uncaught `pylxd.exceptions.NotFound` crashing metabox

### To Reproduce

1. Create a metabox scenario with the action 
2. Launch it via metabox

### Environment

N/A

### Relevant log output

```shell
Traceback (most recent call last):
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/core/machine.py", line 112, in put
    self._container.files.put(filepath, data, mode, uid, gid)
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/lib/python3.10/site-packages/pylxd/models/instance.py", line 114, in put
    response = self._endpoint.post(
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/lib/python3.10/site-packages/pylxd/client.py", line 216, in post
    self._assert_response(response, allowed_status_codes=(200, 201, 202))
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/lib/python3.10/site-packages/pylxd/client.py", line 144, in _assert_response
    raise exceptions.NotFound(response)
pylxd.exceptions.NotFound: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/bin/metabox", line 33, in <module>
    sys.exit(load_entry_point('metabox', 'console_scripts', 'metabox')())
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/main.py", line 77, in main
    runner.run()
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/core/runner.py", line 195, in run
    scn.run()
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/core/scenario.py", line 89, in run
    step(self)
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/core/actions.py", line 41, in __call__
    getattr(scn, self.handler)(*self.args, **self.kwargs)
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/core/scenario.py", line 242, in put
    self.local_machine.put(filepath, data, mode, uid, gid)
  File "/home/h25/prj/canonical/checkbox/metabox/metabox/core/machine.py", line 118, in put
    self._container.files.mk_dir(dirname, mode, uid, gid)
[...]

PR #473, opened 2023-05-17
Fix put crash metabox

Hook25:
## Description

If a scenario contains a Put action to a location that can not be written, this will raise an uncaught pylxd.exceptions.NotFound crashing metabox

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/472

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/metabox/metabox/core/machine.py
+++ b/metabox/metabox/core/machine.py
@@ -109,28 +109,11 @@ class ContainerBaseMachine:
                 raise SystemExit("Rollback failed (systemd not ready)")
 
     def put(self, filepath, data, mode=None, uid=1000, gid=1000):
-        """
-        Puts the `data` into the `filepath` with the specified `mode`, `uid`, and `gid`.
-
-        This function attempts to create the file (including its parent directory).
-
-        Returns:
-            True if successful, False otherwise.
-        """
-        # Catch LXDAPIException because sometimes it is raised instead of
-        #  NotFound
-        with suppress(pylxd.exceptions.LXDAPIException):
+        try:
             self._container.files.put(filepath, data, mode, uid, gid)
-            return True
-        dirname = os.path.dirname(filepath)
-        logger.debug(("Cannot put {} on container. Trying to create"
-                      " directory {} and put the file again..."), filepath,
-                     dirname)
-        with suppress(pylxd.exceptions.LXDAPIException):
-            self._container.files.mk_dir(dirname, mode, uid, gid)
-            self._container.files.put(filepath, data, mode, uid, gid)
-            return True
-        return False
+        except pylxd.exceptions.LXDAPIException:
+            logger.error("Failed to create {}", filepath)
+            raise
 
     def get_connecting_cmd(self):
         return "lxc exec {} -- sudo --user ubuntu --login".format(
--- a/metabox/metabox/core/machine.py
+++ b/metabox/metabox/core/machine.py
@@ -21,6 +21,7 @@ import os.path
 import time
 from pathlib import Path
 import textwrap
+from contextlib import suppress
 
 import pylxd.exceptions
 from loguru import logger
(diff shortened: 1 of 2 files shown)


pieqq (approved):
Looks good to me!

Just FYI, `pylxd.files.put()` is super limited in what it can do. It's impossible, for instance, to recursively create the directories if they are missing. That's why I added those few lines to create the direct parent at the very least, but it's not ideal...

kissiel:
I'm not so sure about this patch. It hides the fact that there's something very wrong about LXD, and not only about putting the file.
Only the errors related to putting the files should be handled by our, metabox `put` part. All other exceptions should bubble up.
What was the original cause of the problem?

Hook25:
As you can see in this past run [of metabox (link)](https://github.com/canonical/checkbox/actions/runs/5011145012/jobs/8981674293#step:7:513) PyLXD does not seem to deterministically rise `NotFound` if it can't write to a location. Either way, the test runner should not crash if an exception is thrown by the `Put` api, it should just mark the test as failed.

kissiel:
> As you can see in this past run [of metabox (link)](https://github.com/canonical/checkbox/actions/runs/5011145012/jobs/8981674293#step:7:513) PyLXD does not seem to deterministically rise `NotFound` if it can't write to a location. Either way, the test runner should not crash if an exception is thrown by the `Put` api, it should just mark the test as failed.

I disagree. The preparation of a scenario is not scenario's merit. The same would apply if you would be writing normal unit test in python and made a syntax error in the `setUp` method. Test harness would crash and not run any test. This _should_ crash, because the error you're catching here is scenario author's problem, not the thing's that metabox is asked to test.

kissiel:
Or to be more precise can checkbox's bug cause this error?

Hook25:
Short answer: no, but we can be more descriptive. The exception contains the "response" with the error message from lxd, we can log it in debug to make any eventual future problem easier to pin down. What do you think?

The function that ends up being called by the `Put` is [the instance function of the file manager](https://github.com/lxc/pylxd/blob/1f1f903d6512f0fb9ee030d051278c6c90f068b6/pylxd/models/instance.py#L113) that clearly states in its documentation that it is going to raise only `LXDAPIException` errors, and so it does at [line 121](https://github.com/lxc/pylxd/blob/1f1f903d6512f0fb9ee030d051278c6c90f068b6/pylxd/models/instance.py#L121). We get a specific exception instead of the generic one because one "deeper" function in [client.py](https://github.com/lxc/pylxd/blob/1f1f903d6512f0fb9ee030d051278c6c90f068b6/pylxd/client.py#L233) asserts the result it gets back is in [200, 202]. Doing so [this function](https://github.com/lxc/pylxd/blob/1f1f903d6512f0fb9ee030d051278c6c90f068b6/pylxd/client.py#L146) tries to create a specific exception for two cases, one of whom is ours. From the [API documentation](https://linuxcontainers.org/lxd/api/master/#/instances/instance_files_post) the possible error return codes are 4 of whom just 1 is handled by the function, the other 3 will raise a generic `LXDAPIException` leaving little chance for us to separate them into different "categories" (without manually parsing out of the response the `error_code`, but if we do that, when should we crash? 400? 403? 500? I don't know

kissiel:
Yeah, I feel the same way about "parsing" the exception to find the details.
I really would like to somehow differentiate between a problem with the test and problem with what the test is testing... We shouldn't treat test-harness errors the same way as failed assertions of the test. 

ATM failure to put something in a container clearly signifies a problem with the test logic, which should be reported to the test author with all the goodies that come with a traceback.

kissiel:
Some test harnesses have the third outcome for the test - "CRASHED". Maybe that's the way forward.

pieqq:
Following a discussion we had earlier today, the best way to proceed would be:

- let the test harness crash. It's not normal that the test is trying to put a file where it cannot, so there is something fishy here.
- add debugging information to have something more meaninful than just `file not found` or `pylxdAPIException`.

pieqq (approved):
+1, thanks!

PR #474, opened 2023-05-19
Add MIT license for screenoff.sh

KaiChuan-Hsieh:
LP: #1990484
Intel would like to utilize the script to improve their PHM tool, however, it needs the script to be MIT license.


yphus (approved):
+1

Issue #370, opened 2023-03-16
After suspend attachment test cases failed


weizhenwu:
### Bug Description

The following test cases failed due to the attachment log is not exist after running the main cases.
Need to survey why it is not present.
- after-suspend-cpu/scaling_test-log-attach
- 1_xrandr_screens_after_suspend.tar.gz_auto

Suppose to be the effect from changes in PR [304](https://github.com/canonical/checkbox/pull/306) and [306](https://github.com/canonical/checkbox/pull/306)

### To Reproduce

1. checkbox-cli run CPU related test plan
2. checkbox-cli run Graphic related test plan

### Environment

- Checkbox 2.2.0
- Ubuntu 22.04

### Relevant log output

```shell
- Checkbox session(s) (located in `/var/tmp/checkbox-ng/sessions/`, you usually want to select the most recent one)

- logs from the impacted components (e.g. `lsblk` if this is related to an issue when testing a disk...); a safe option is to install and run `sosreport` to gather as much log as possible.
```

weizhenwu:
Quick investigation on the CPU one:
after-suspend-cpu/scaling_test was skipped
due to
```
job cannot be started: resource expression "'userspace' in cpuinfo.governors" evaluates to false
```
but the new attachment cases `after-suspend-cpu/scaling_test-log-attach` didn't  set dependencies on the main case

PR #480, opened 2023-05-22
Fix after-suspend cpu and xrandr attachment test faulty failed

weizhenwu:
## Description

1. after-suspend-cpu/scaling_test-log-attach 
failed due to missing dependency on the main case after-suspend-cpu/scaling_test
fixed by set dependency

2. {index}_xrandr_screens_after_suspend.tar.gz_auto 
Previously it looks for a file name with after_suspend postfix, but the main case store attachment without the postfix. Hence fix the main case to store attachments with the postfix

## Resolved issues
#370 After suspend attachment test cases failed

## Tests
1. after-suspend-cpu/scaling_test-log-attach 
https://certification.canonical.com/hardware/202010-28329/submission/313832/
2. {index}_xrandr_screens_after_suspend.tar.gz_auto 
https://certification.canonical.com/hardware/202010-28329/submission/313831/

### DIFF
--- a/providers/base/bin/graphic_cycle_resolution.sh
+++ b/providers/base/bin/graphic_cycle_resolution.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# shellcheck disable=SC1091
+
+while [ $# -gt 0 ]
+do
+    case "$1" in
+        --index)
+            shift
+            INDEX=$1
+        ;;
+        --after-suspend)
+            shift
+            KEYWORD="after_suspend"
+        ;;
+        *)
+            echo "Not recognize $1"
+            usage
+        ;;
+    esac
+    shift
+done
+
+if [[ $XDG_SESSION_TYPE == "wayland" ]]
+then
+    gnome_randr_cycle.py --keyword="${INDEX}_${KEYWORD}" --screenshot-dir="$PLAINBOX_SESSION_SHARE"
+else
+    xrandr_cycle.py --keyword="${INDEX}_${KEYWORD}" --screenshot-dir="$PLAINBOX_SESSION_SHARE"
+fi
--- a/providers/base/units/cpu/jobs.pxu
+++ b/providers/base/units/cpu/jobs.pxu
@@ -56,7 +56,7 @@ plugin: attachment
 category_id: com.canonical.plainbox::cpu
 id: after-suspend-cpu/scaling_test-log-attach
 estimated_duration: 1.0
-after: after-suspend-cpu/scaling_test
+depends: after-suspend-cpu/scaling_test
 command: [[ -e "${PLAINBOX_SESSION_SHARE}"/scaling_test_after_suspend.log ]] && cat "${PLAINBOX_SESSION_SHARE}"/scaling_test_after_suspend.log
 _summary:
  Attach CPU scaling capabilities log
--- a/providers/base/units/graphics/jobs.pxu
+++ b/providers/base/units/graphics/jobs.pxu
@@ -213,18 +213,10 @@ template-filter: graphics_card.prime_gpu_offload == 'Off'
 plugin: user-interact-verify
 category_id: com.canonical.plainbox::graphics
 id: graphics/{index}_cycle_resolution_{product_slug}
-flags: also-after-suspend
 requires: package.name == 'xorg'
 depends: graphics/VESA_drivers_not_in_use
 command:
- # shellcheck disable=SC1091
- source graphics_env.sh {driver} {index}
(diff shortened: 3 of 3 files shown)


LiaoU3:
The `graphics/{index}_cycle_resolution_{product_slug}` and `after-suspend-graphics/{index}_cycle_resolution_{product_slug}` are very simnilar.
Would it be better if we write this in the with the `_siblings`?

providers/base/units/cpu/jobs.pxu
```diff
@@ -56,7 +56,7 @@ plugin: attachment
 category_id: com.canonical.plainbox::cpu
 id: after-suspend-cpu/scaling_test-log-attach
 estimated_duration: 1.0
-after: after-suspend-cpu/scaling_test
+depends: after-suspend-cpu/scaling_test
```
pieqq, line 59: `after` and `depends` do slightly different things:

- `after` means `after-suspend-cpu/scaling_test-log-attach` should be run after `after-suspend-cpu/scaling_test` regardless of its outcome. It's used so that Checkbox can still do something (for instance retrieve logs) ***even if*** `after-suspend-cpu/scaling_test` (the parent job) failed.
- if you use `depends`, `after-suspend-cpu/scaling_test-log-attach` will be skipped if `after-suspend-cpu/scaling_test` fails.

Now, what I'm not sure is why would this job (`after-suspend-cpu/scaling_test-log-attach`) fail in the first place. Its `command` field (below) says "if there is a log file at that location, attach it.". @weizhenwu would you have a sample submission where the problem you're trying to fix is exhibited?

pieqq (changes requested):
Even though this PR is about a similar topic (some issues with different attachment jobs), the attachment jobs in question are not related at all. It would have been better to propose 2 different PRs, or at least to separate these in different commits.

With that said, I agree with Vincent's comment regarding the xrandr attachment job. I would avoid duplication of code like this. Maybe create a script that calls the stuff currently in the `command` field but use different filenames based on a parameter (so that you can use `siblings` and just replace the `command` from `new_script.sh` to `new_script.sh after-suspend` or something like that).

providers/base/units/cpu/jobs.pxu
weizhenwu: Here is the recent submission:
https://certification.canonical.com/hardware/202303-31378/submission/316037/test-results/?term=cpu%2Fscaling_test

the parent job `after-suspend-cpu/scaling_test` was skipped based on the resource false, so there is no log file existing to attach.
I used `depend` instead of `after` since it should be skipped as well if the parent job is skipped.

weizhenwu:
thanks for all suggestions!

I did the following changes:
1. Divided them into 2 commits. One for CPU and one for xrandr
2. For xrandr job, used `_sibling` instead of duplicated case 

test result for xrandr job:
https://certification.canonical.com/hardware/202010-28329/submission/319316/

providers/base/bin/graphic_cycle_resolution.sh
```diff
@@ -0,0 +1,28 @@
...
+            shift
+            INDEX=$1
+        ;;
+        --after-suspend)
+            shift
+            KEYWORD="after_suspend"
+        ;;
+        *)
+            echo "Not recognize $1"
+            usage
```
kissiel, line 17: undefined

providers/base/bin/graphic_cycle_resolution.sh
```diff
@@ -0,0 +1,28 @@
...
+        ;;
+    esac
+    shift
+done
+
```
kissiel, line 22: so here both: `INDEX` and `KEYWORD` can be undefined.

providers/base/bin/graphic_cycle_resolution.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# shellcheck disable=SC1091
+
+while [ $# -gt 0 ]
+do
+    case "$1" in
+        --index)
+            shift
+            INDEX=$1
```
kissiel, line 9: This fails when there's one argument supplied.

kissiel (changes requested):
I strongly discourage writing shell scripts with any significant logic like this one.
Shell syntax is so brittle is almost always worth jumping straight to python. The proof of that is below, there's a high number of problems already in this small script.

pieqq:
Hey @weizhenwu! This had been in "Draft" mode for months. Any plan for this? If not, we can close it for now.

kissiel:
I guess it's safe to assume it's abandoned by now.

Issue #481, opened 2023-05-22
Metabox provisioning does not carry file permissions


Hook25:
### Bug Description

When provisioning a new machine metabox creates files without carrying the original permission flags. This makes https://github.com/canonical/checkbox/pull/466 fail because the manifest script is not executable

### To Reproduce

1. Create a Metabox container 
2. Run the collect-manifest job

### Environment

N/A

### Relevant log output

```shell
N/A
```

### Additional context

N/A

PR #466, opened 2023-05-16
Checkbox 440/add metabox manifest tests

Hook25:
## Description
This contains metabox tests to check that manifest loading in checkbox works as stated in the documentation. The main features that are under test are:

- Nominal behaviour of loading from configuration/launcher
- Priority of configurations (launcher > disk > cache) 

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-440

## Documentation

N/A

## Tests

This PR can be tested via metabox with the tag `test_manifest`

PR #486, opened 2023-05-22
Metabox provisioning does not carry file permissions

Hook25:
## Description

When provisioning a new machine metabox creates files without carrying the original permission flags. This makes https://github.com/canonical/checkbox/pull/466 fail because the manifest script is not executable.

This changes the file/dir put operations to also propagate the `st_mode`

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/481

## Documentation

N/A

## Tests

I have tested this on both legacy and remote runs, all files seem to now have the correct permissions and the manifest bin now works correctly (once a provider is installed)

### DIFF
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -29,7 +29,9 @@ import os
 import sys
 import time
 import yaml
+import subprocess
 from pathlib import Path
+from contextlib import contextmanager
 
 import pkg_resources
 import pylxd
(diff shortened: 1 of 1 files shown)


Hook25:
Also, given that this should propagate all flags, can we remove manual chmod(s) like: https://github.com/canonical/checkbox/blob/b718be18bce0455c5b1353f4fc76a62263d8882c/metabox/metabox/core/machine.py#L269
and 
https://github.com/canonical/checkbox/blob/b718be18bce0455c5b1353f4fc76a62263d8882c/metabox/metabox/core/machine.py#L270

I have tried to do so locally and it seems to still be working fine

pieqq (approved):
Nice! +1

I learned something with the `yield ...` line :)

The section that makes use of `mkdir -p` could be very useful for our `put()` method, since `pylxd.files.put` does not provide a way to create directories recursively...

Issue #488, opened 2023-05-23
com.canonical.plainbox::networking info_device1_enp1s0 requires ethtool but still passes if it's not installed


djacobs98:
### Bug Description

The informational tests for the Ethernet interfaces relies on ethtool to get some information about them.

However if ethtool is not installed, the test still passes with the result of "ethtool not installed."

If this test requires ethtool, ethtool should be provided by checkbox.

### To Reproduce

1. Select Desktop Manual test plan.
2. Select Informational tests for Ethernet interfaces.

### Environment

OS: Jammy Desktop + RT kernel.
Checkbox22 snap ver. 2.6
Checkbox-iiotg-classic (latest/stable, rev 49)

djacobs98:
Screenshot
![No_Ethtool_Screenshot](https://github.com/canonical/checkbox/assets/79291680/08de0b45-c235-4e4c-8517-bbf4d721f560)

kissiel:
Fix is here:
https://github.com/canonical/checkbox/pull/492

kissiel:
@djacobs98 should the test be skipped or failed if there's no ethtool?

djacobs98:
This is a tangled problem.
There are 2 problems here.
1. ethtool is actually installed, so this error message is incorrect.
2. The script got some sort of error trying to probe this Ethernet interface - which is accurate, it's not a real interface.  So the test should fail for it.

djacobs98:
This is a tangled problem.
There are 2 problems here.
1. ethtool is actually installed, so this error message is incorrect.
2. The script got some sort of error trying to probe this Ethernet interface - which is accurate, it's not a real interface.  So the test should fail for it.
I think it will be easier to open new issues for these.

PR #492, opened 2023-05-24
fix: make network-info jobs require ethtool to be available  (bugfix)

kissiel:
## Description
Make network-info jobs require ethtool.
See bug link for details.

## Resolved issues

Fixes: : [CHECKBOX-626] (https://github.com/canonical/checkbox/issues/488)

## Testing
Tested on a system without the ethtool and now it yields:
```
==============[ Running job 1 / 1. Estimated time left: 0:00:01 ]===============
-------------------[ Network Information of device 1 (eno1) ]-------------------
ID: com.canonical.certification::networking/info_device1_eno1
Category: com.canonical.plainbox::networking
 - resource expression "executable.name == 'ethtool'" evaluates to false
Finalizing session that hasn't been submitted anywhere: checkbox-run-2023-05-24T07.16.45
 ☑ : Enumerate available system executables
 ☑ : Collect information about hardware devices (udev)
 ☐ : Network Information of device 1 (eno1)
```
And with it installed it proceeds to run.

[CHECKBOX-626]: https://warthogs.atlassian.net/browse/CHECKBOX-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/base/units/networking/jobs.pxu
+++ b/providers/base/units/networking/jobs.pxu
@@ -16,6 +16,7 @@ id: networking/info_device{__index__}_{interface}
 flags: also-after-suspend
 _summary: Network Information of device {__index__} ({interface})
 estimated_duration: 1.0
+requires: executable.name == 'ethtool'
 command:
   network_device_info.py info NETWORK --interface {interface}
 _description:
--- a/providers/base/units/networking/packaging.pxu
+++ b/providers/base/units/networking/packaging.pxu
@@ -1,6 +1,6 @@
 unit: packaging meta-data
 os-id: debian
-Depends: ntpdate, net-tools
+Depends: ntpdate, net-tools, ethtool
 
 unit: packaging meta-data
 os-id: debian
--- a/providers/base/units/networking/test-plan.pxu
+++ b/providers/base/units/networking/test-plan.pxu
@@ -15,6 +15,7 @@ include:
     networking/info_device.*                       certification-status=blocker
 bootstrap_include:
     device
+    executable
 
 id: networking-cert-automated
 unit: test plan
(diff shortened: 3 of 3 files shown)


pieqq (commented):
Thanks for that fix. A few comments:

- the bug happens with the snap version of Checkbox, even though it [seems that ethtool is pulled at build time](https://github.com/canonical/checkbox/blob/main/checkbox-core-snap/series22/snap/snapcraft.yaml#L257). Does modifying the packaging meta-data unit helps in that regard?
- the job is skipped (providing zero info about the network) instead of running and providing partial information about the network. Not sure if this is what QA team need.

kissiel:
> Thanks for that fix. A few comments:
> 
> * the bug happens with the snap version of Checkbox, even though it [seems that ethtool is pulled at build time](https://github.com/canonical/checkbox/blob/main/checkbox-core-snap/series22/snap/snapcraft.yaml#L257). Does modifying the packaging meta-data unit helps in that regard?

I'll make sure it is so.

> * the job is skipped (providing zero info about the network) instead of running and providing partial information about the network. Not sure if this is what QA team need.

I asked Doug that @ [the original bug](https://github.com/canonical/checkbox/issues/488#issuecomment-1560637415)

pieqq:
Since the original issue was marked as closed, we might want to close this associated PR.

kissiel:
Yeah, although it still makes sense to not crash on missing ethtool. Recent releases don't include it by default.

pieqq:
Any plan for this? Do we close it or not?

kissiel:
The patch still makes sense IMHO..
The job runs ethtool, but doesn't officially require it.

Hook25 (approved):
+1, this job should require ethtool as it uses it

PR #496, opened 2023-05-24
Show values of GROUP argument for list command help

petitfou:
It can be helpful to have the list of all possible values for GROUP argument

## Tests

Manual test done

$ checkbox-cli list -h
usage: checkbox-cli [-h] [-a] [-f FORMAT] [{category,exporter,job,manifest entry,packaging meta-data,template,test plan,unit}]

positional arguments:
  {category,exporter,job,manifest entry,packaging meta-data,template,test plan,unit}
                        list objects from the specified group

options:
  -a, --attrs           show object attributes
  -f FORMAT, --format FORMAT
                        output format, as passed to print function. Use '?' to list possible values

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/subcommands.py
+++ b/checkbox-ng/checkbox_ng/launcher/subcommands.py
@@ -51,6 +51,7 @@ from plainbox.impl.session.storage import WellKnownDirsHelper
 from plainbox.impl.transport import TransportError
 from plainbox.impl.transport import get_all_transports
 from plainbox.impl.transport import SECURE_ID_PATTERN
+from plainbox.impl.unit import all_units
 
 from checkbox_ng.config import load_configs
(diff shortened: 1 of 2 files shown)


kissiel (changes requested):
There are also "special" group - `all-jobs`. It would be killed with this patch, and I think the feature is still useful to people. I think the easiest way to handle that is just simply adding `all-jobs` to choices here.

petitfou:
> There are also "special" group - `all-jobs`. It would be killed with this patch, and I think the feature is still useful to people. I think the easiest way to handle that is just simply adding `all-jobs` to choices here.

nice catch, thanks

kissiel (approved):
TYVM! +1

pieqq (approved):
+1. Thanks for this QoL improvement!

Issue #498, opened 2023-05-25
Metabox provisioning should install instead of sideload providers


Hook25:
### Bug Description

When provisioning a machine, metabox sideloads all providers instead of installing them. This blocks: https://github.com/canonical/checkbox/pull/466

### To Reproduce

1. Run metabox on any test
2. Start an interactive session with the provisioned lxd containers
3. Run via checkbox-cli the manifest collection job

### Environment

N/A

### Relevant log output

```shell
N/A
```

### Additional context

N/A

PR #500, opened 2023-05-26
Metabox Install instead of sideloading providers

Hook25:
## Description

This removes transfers to sideload location and adds install commands for both the metabox-provider and standard providers in source machines

## Resolved issues

Resolves: https://github.com/canonical/checkbox/issues/498, https://warthogs.atlassian.net/browse/CHECKBOX-624

## Documentation

N/A

## Tests

I have tested remote and legacy for multiple distributions with metabox and all tests are still working as expected

### DIFF
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -221,7 +221,7 @@ class LxdMachineProvider:
         provider_path = pkg_resources.resource_filename(
             'metabox', 'metabox-provider')
         metabox_dir_transfers = machine.get_early_dir_transfer() + [
-            (provider_path, '/var/tmp/checkbox-providers/metabox-provider')]
+            (provider_path, '/var/tmp/metabox-provider')] # avoid sideloading
         for src, dest in metabox_dir_transfers + machine.config.transfer:
             logger.debug("Working on {}", dest)
             with self._mounted_source(machine, src):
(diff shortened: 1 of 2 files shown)


metabox/metabox/core/lxd_provider.py
```diff
@@ -221,7 +221,7 @@ def _run_transfer_commands(self, machine):
         provider_path = pkg_resources.resource_filename(
             'metabox', 'metabox-provider')
         metabox_dir_transfers = machine.get_early_dir_transfer() + [
-            (provider_path, '/var/tmp/checkbox-providers/metabox-provider')]
+            (provider_path, '/var/tmp/metabox-provider')] # avoid sideloading
```
kissiel, line 224: TBH i'd use home dir of the container

metabox/metabox/core/machine.py
```diff
@@ -266,12 +260,13 @@ def get_early_setup(self):
...
 
         commands = [
-            "bash -c 'chmod +x /var/tmp/checkbox-providers/base/bin/*'",
-            "bash -c 'chmod +x /var/tmp/checkbox-providers/resource/bin/*'",
             ("bash -c 'pushd /home/ubuntu/checkbox/checkbox-ng ; "
              "sudo python3 -m pip install -e .'"),
             ("bash -c 'pushd /home/ubuntu/checkbox/checkbox-support ; "
              "sudo python3 -m pip install -e .'"),
+            "bash -c 'sudo mkdir -p /usr/local/share/plainbox-providers-1/'",
+            ("bash -c 'ls /home/ubuntu/checkbox/providers/**/manage.py"
```
kissiel, line 268: I like the smartness, but it's not obvious at the first look what it does, and there's not that many of them.
I'm torn here tbh.

kissiel (commented):
See below

metabox/metabox/core/machine.py
Hook25: I have done it this way because if we create a new one it will be hard to understand why it is not present in the metabox container if either me or you don't recall that we have to update the install command

kissiel: Yeah, if you really want to leave it in, then yeah, write a big fat comment there, please.

kissiel (approved):
Fantastic! +1

PR #505, opened 2023-05-29
Update stress-ng part to v0.15.08 in checkbox-core-snap

zongminl:
stress-ng v0.15.08 contains a fix to stressor af-alg which is needed by
 Limerick project

## Description
- There's an existing problem with af-alg stressor in stress-ng v0.15.07 that stress-ng will fail directly when an errorno 95 (EOPNOTSUPP) is returned during the test.
- With the [upstream commit](https://github.com/ColinIanKing/stress-ng/commit/69ddd9d997cddf8d17a594f8df40f1cfdc3158a8) EOPNOTSUPP has been processed as part of exception handling so this kind of errorno return will no longer causing af-alg test to fail.

## Resolved issues
Launchpad bug [#2017827](https://bugs.launchpad.net/limerick/+bug/2017827) in Limerick project

## Tests
Tested the commit on the device originally used to discover the bug reported, can get passed results with 5 runs in a row (failure rate 0/5)

### DIFF
--- a/checkbox-core-snap/series16/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series16/snap/snapcraft.yaml
@@ -84,7 +84,7 @@ parts:
 ################################################################################
 # Upstream: https://kernel.ubuntu.com/git/cking/stress-ng.git/plain/snap/snapcraft.yaml
   stress-ng:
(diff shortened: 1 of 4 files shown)


pieqq (approved):
+1, thanks!

PR #507, opened 2023-05-29
Checkbox 417/wireless detect

diohe0311:
## Description
The issue is occurs in IoT device while running wireless detect test without WiFi connection.
If there is no WiFi connection, the device didn't turns OFF network module but turns into dormant state, which causing issue.

I add a exception handling to avoid OSError interrupt testing.

(Additionally, there is a indirectly import problem in network_device_info.py ocurrs during testing, this is because the imported module is not the standard module in tox venv.
After discuss with Pierre, Maciej and Massimiliano, since the module is not been used in unit test, the simplest solution is move importing line to where the function is actually needed.)

## Resolved issues
https://warthogs.atlassian.net/jira/software/c/projects/CHECKBOX/boards/590?modal=detail&selectedIssue=CHECKBOX-417

## Tests
$./manage.py test -u (in checkbox/providers/base)
Tested in 10.102.160.211 (Ubuntu 20.04.6) and my laptop.


providers/base/bin/network_device_info.py
```diff
@@ -99,7 +97,13 @@ def get_ipv6_address(interface):
         cmd = ['/sbin/ip', '-6', '-o', 'addr', 'show', 'dev', interface,
                'scope', 'link']
         proc = check_output(cmd, universal_newlines=True)
-        return proc.split()[3].strip()
+        try:
+            ipv6_addr = proc.split()[3].strip()
+        except Exception as e:
```
pieqq, line 102: It's not a good practice to catch the general `Exception`. You can read more about this [here](https://www.learnpython.dev/03-intermediate-python/40-exceptions/70-best-practices/#don-t-catch-exception) and [here](https://python.land/how-not-to-handle-exceptions).

Check what specific exception is thrown when `proc` is not what the script expects, and handle this one.

providers/base/bin/network_device_info.py
```diff
@@ -99,7 +97,13 @@ def get_ipv6_address(interface):
...
+            print("ERROR: getting the IPv6 address for %s: %s" %
```
pieqq, line 103: I'm not a fan of this for two reasons:

1. the [`print()` statement](https://docs.python.org/3/library/functions.html?highlight=print#print) will print something to `stdout`. Checkbox handles `stdout` and `stderr` differently: things printed to `stdout` will show up green, while things printed to `stderr` will show up red. In your case, this error (and the other one for ipv4) will show up as a regular output in the job.
2. because of this, the printed error messages also show up when running the test cases:

```
(...)
test_get_ipv4_address_without_connection (test_network_device_info.GetIpv4AddressTests) ... ERROR: getting the IPv4 address for wlo1: OSError('packed IP wrong length for inet_ntoa')
ok
test_get_ipv6_address_with_connection (test_network_device_info.GetIpv6AddressTests) ... ok
test_get_ipv6_address_without_connection (test_network_device_info.GetIpv6AddressTests) ... ERROR: getting the IPv6 address for wlo1: IndexError('list index out of range')
ok
(...)
```

That said, I'm not sure what's the best way to proceed. Maybe @kissiel or @Hook25 would have an idea?

providers/base/tests/test_network_device_info.py
```diff
@@ -0,0 +1,92 @@
...
+        mock_check_output = Mock(return_value=test_input)
+        with patch("network_device_info.check_output", mock_check_output):
+            interface = "wlo1"
+            addr = Utils.get_ipv6_address(interface)
+            self.assertEqual(addr, "***NOT CONFIGURED***")
+
+    def test_get_ipv6_address_with_connection(self):
+        test_input = "2: wlo1    inet6 fe80::d9eb:3f93:c7b2:86ba/64 scope link noprefixroute \       valid_lft forever preferred_lft forever"
+        mock_check_output = Mock(return_value=test_input)
+        with patch("network_device_info.check_output", mock_check_output): # somehow it don't regonize check_output() if I don't put subprocess in front
```
pieqq, line 86: I'm not sure I understand the comment, since there is no `subprocess` in front of anything :)

But I think I know what you mean: mocking is hard! In this case, you need to point very specifically to what you want to mock. You don't want to mock the `check_output` method from the `subprocess` module, you want to mock the method being imported by `network_device_info`, hence you have to patch `network_device_info.check_output`. It's honestly not very clear...

A while ago, I listened to [this episode](https://talkpython.fm/episodes/show/287/testing-without-dependencies-mocking-in-python) of the "Talk Python to Me" podcast that helped me a lot to better understand testing and mocking. You can check it out.

pieqq (commented):
Glad to see this is finally working!

I'll just leave a few comments below. I think we need @kissiel and/or @Hook25 feedback on that one.

providers/base/bin/network_device_info.py
diohe0311: Yes, this did come to my mind before, I thought there might be some cases I miss if I specify OSError, so I just follow the code of get_ipv4_address().

So would it will be better if we modify the both functions?

providers/base/bin/network_device_info.py
diohe0311: So unit test shouldn't shows error if they pass the test?
I think maybe network_device_info.py shouldn't try to get info if the conditions is not suitable, perhaps adding conditional expressions could avoid this happening.

providers/base/tests/test_network_device_info.py
diohe0311: Ah sorry about that, this comment is me talking to myself while I trying to know how does mock works.
Thanks for link sharing, it really helps.

providers/base/bin/network_device_info.py
kissiel: If getting an ipv6 is part of the test (as in checkbox job) it should definitely fail the test. If this is only for warning purposes, then make it as such, and as Pierre mentioned, dump it on stderr.
Also, the typical logging template of:
`{LEVEL}: {MSG}`  like 'ERROR: no permission' would suggest that the fact that "getting the IPv6 address for" is the error condition, so I'd rephrase it to: "failed to get... " or whatever you prefer, but the message should communicate what went wrong.

As for the final, and weirdest thing: I completely agree that **UNIT TEST** should not invoke a real "backend" utility, like `ip` in our case.

providers/base/bin/network_device_info.py
diohe0311: I modified get_ipv4_addr() & get_ipv6_addr() to check the specific error

providers/base/bin/network_device_info.py
```diff
@@ -82,14 +80,14 @@ def is_iface_connected(cls, iface):
...
-            print("ERROR: getting the IPv4 address for %s: %s" %
-                  (interface, repr(e)))
+
+            buffer = struct.pack('256s', interface[:15].encode())
+            # retrieve the IP address associated with a network interface.
+            data = fcntl.ioctl(s.fileno(), 0x8915, buffer)
+            ipv4_addr = socket.inet_ntoa(data[20:24])
+        except OSError:
+            print("OSError: failed to get the IPv4 address for %s" %
+                  interface, file=sys.stderr)
```
pieqq, line 90: using `file=sys.stderr` is a good idea! It will indeed show up as a red message in the Checkbox output.

I would keep the error message from the exception (`repr(e)`) displayed, because it will make debugging easier.

providers/base/tests/test_network_device_info.py
```diff
@@ -0,0 +1,86 @@
...
+    #    (It pulls out bytes data directly from kernel, so if the given SIOCGIFADDR doesn't exist, it raise a OSError)
+
+    # 2. Connected with WiFi
+    #    It will pulls out a 256 bytes data into buffer, then we get to parse the IP address
+    @patch("fcntl.ioctl")
+    def test_get_ipv4_address_without_connection(self, mock_ioctl):
+        mock_ioctl.side_effect = OSError()
+        interface = "wlo1"
+        addr = Utils.get_ipv4_address(interface)
+        self.assertEqual(addr, "***NOT CONFIGURED***")
```
pieqq, line 35: In order to avoid the error printing as an output of the tests, you could use [Trey Hunter's suggestion](https://www.reddit.com/r/learnpython/comments/5f0vko/comment/dago5w7/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) of using `redirect_stdout` (except you would probably want to catch `stderr` instead using [`redirect_stderr`](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr)).

providers/base/tests/test_network_device_info.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # 3. Connected with WiFi, but we ask the wrong interface name 
+    #    (This would not happen in our case, because the name is given by NetworkManager)
+    def test_get_ipv6_address_without_connection(self):
+        test_input = ""
+        mock_check_output = Mock(return_value=test_input)
+        with patch("network_device_info.check_output", mock_check_output):
+            interface = "wlo1"
+            addr = Utils.get_ipv6_address(interface)
+            self.assertEqual(addr, "***NOT CONFIGURED***")
```
pieqq, line 75: Same comment as above, maybe `redirect_stderr` could help here.

pieqq (changes requested):
We are getting there! A few last comments and I think this should be good to land.

Thanks a lot for your patience and perseverance, @diohe0311 !

providers/base/tests/test_network_device_info.py
diohe0311: Thank you very much for the link!
what do you mean by catch stderr? I thought calling redirect_stderr() to stop from print it out is the solution?

diohe0311: Just to clarify, we only want to know if the function could pass the unit test or not, right?
Or error msg should be printed if the test is failed? (other existing unit tests don't print anything)

pieqq (approved):
Looks good to me, thanks a lot for your time working on this, Dio!

PR #509, opened 2023-05-29
Remove hardcoded version strings in python code

Hook25:
## Description
This changes all `version` and `__version__` variables/kwargs from being hardcoded to be retrieved at runtime

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-614

## Documentation
N/A

## Tests

This was tested running metabox scenarios for both the legacy and the remote version

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -34,10 +34,6 @@ message = Bump version: {current_version} → {new_version}
 search = version="{current_version}"
 replace = version="{new_version}"
 
-[bumpversion:file:checkbox-ng/checkbox_ng/__init__.py]
-search = __version__ = '{current_version}'
-replace = __version__ = '{new_version}'
-
 [bumpversion:file:checkbox-ng/plainbox/__init__.py]
 search = __version__ = '{current_version}'
 replace = __version__ = '{new_version}'
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -38,10 +38,6 @@ replace = version="{new_version}"
 search = version="{current_version}"
 replace = version="{new_version}"
 
-[bumpversion:file:providers/certification-server/manage.py]
-search = version="{current_version}"
-replace = version="{new_version}"
-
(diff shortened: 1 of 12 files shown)


checkbox-ng/checkbox_ng/__init__.py
```diff
@@ -24,5 +24,12 @@
...
 """
 
-__version__ = '2.6'
+from importlib_metadata import version, PackageNotFoundError
+
+try:
+    __version__ = version("checkbox-ng")
+except PackageNotFoundError:
+    import logging
+    logging.error('Failed to retieve checkbox-ng version')
```
pieqq, line 33: Typo here and below (retieve)

.bumpversion.cfg
```diff
@@ -34,46 +34,10 @@ message = Bump version: {current_version} → {new_version}
 search = version="{current_version}"
 replace = version="{new_version}"
 
-[bumpversion:file:checkbox-ng/checkbox_ng/__init__.py]
-search = __version__ = '{current_version}'
-replace = __version__ = '{new_version}'
-
-[bumpversion:file:checkbox-ng/plainbox/__init__.py]
-search = __version__ = '{current_version}'
-replace = __version__ = '{new_version}'
```
pieqq, line 43: If this is removed, then the current release process tools will leave `__version__` in `checkbox-ng/plainbox/__init__.py` as "2.6" forever...

pieqq (dismissed):

.bumpversion.cfg
Hook25: `checkbox-ng/plainbox/__init__.py` should get the version number at runtime now so this line should be safe to delete

yphus (changes requested):
I'd rather use `checkbox_ng.__version__` instead of `plainbox.__version__` in the providers  `manage.py`. It was really confusing when both checkbox and plainbox were used to name checkbox. Let's use the first wherever possible.

yphus (approved):
Thanks for the 16.04 version check. I think we're good, +1

PR #511, opened 2023-05-30
Fixed TP selection empty to be interactive

Hook25:
## Description

Test does not pass in main because checkbox is started non iteractively and fails before the expected failure


kissiel (approved):
Makes sense.
Tested and it works.
Thanks for the quick fix!

PR #513, opened 2023-05-30
Add support to cpuid.py for AMD Bergamo

mreed8855:
## Description
In cpuid.py  support has been added for the AMD Bergamo CPU

## Resolved issues
Test results for cpuid had been failing when servers with the AMD Bergamo cpu were tested.

### DIFF
--- a/providers/base/bin/cpuid.py
+++ b/providers/base/bin/cpuid.py
@@ -111,6 +111,7 @@ CPUIDS = {
         "AMD Milan-X":      ['0xa00f12'],
         "AMD ROME":         ['0x830f10'],
         "AMD Ryzen":        ['0x810f81'],
+        "AMD Bergamo":      ['0xaa0f01'],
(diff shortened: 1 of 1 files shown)


providers/base/bin/cpuid.py
```diff
@@ -111,6 +111,7 @@
         "AMD Milan-X":      ['0xa00f12'],
         "AMD ROME":         ['0x830f10'],
         "AMD Ryzen":        ['0x810f81'],
+        "Bergamo":          ['0xaa0f01'],
```
bladernr, line 114: Since we're at this a second time, can you change the string to "AMD Bergamo" so it's consistent with the other AMD entries?

bladernr (changes requested):
one quick inline comment to resolve...

providers/base/bin/cpuid.py
mreed8855: I have updated the branch to reflect AMD Bergamo

bladernr (approved):
LGTM

Issue #457, opened 2023-05-15
cold/warm boot stress jobs check the BSSID of the wireless AP, not the SSID


yphus:
### Bug Description

After warm or cold reboots, the test  checks that the system re-connects to the same AP.
But it check the BSSID of the AP, not the SSID.
it's basically assuming that there's a single physical access point vs. a typical roaming setup which has more than one physical AP all with the same SSID.

Current command: `nmcli -t -f active,BSSID d w l`
Expected command: `nmcli -t -f active,SSID d w l`

The command above is used multiple time in jobs defined in the [stress/boot.pxu](https://github.com/canonical/checkbox/blob/main/providers/base/units/stress/boot.pxu) file. All of them should be fixed of course.

In addition to the command itself, it would be nice to expand the tests description to actually explain what's being checked (wireless AP, lspci and lsusb after reboots).

### To Reproduce

Start a checkbox remote session and select one of the boot stress test plan:

- Warm boot stress test (id: warm-boot-stress-test)
- Cold boot stress test (id: cold-boot-stress-test)

### Environment

Reported on UC20 but it's not UC specific.

pieqq:
Closed with #517

PR #517, opened 2023-05-31
Modify cold/warm boot stress jobs

diohe0311:
## Description
The cmd has been change BSSID -> SSID, due to:
> it's basically assuming that there's a single physical access point vs. a typical roaming setup which has more than one physical AP all with the same SSID.

Also added the content of "Description".

## Resolved issues
https://github.com/canonical/checkbox/issues/457

### DIFF
--- a/providers/base/units/stress/boot.pxu
+++ b/providers/base/units/stress/boot.pxu
@@ -58,7 +58,7 @@ flags: preserve-locale
 id: cold-boot-loop-reboot1
 category_id: stress-tests/cold-boot
 _summary: Perform cold reboot 1
-_description: Enter sleep mode after 120s.
+_description: Enter sleep mode after a configurable delay.
 unit: job
 plugin: shell
 environ: STRESS_BOOT_WAKEUP_DELAY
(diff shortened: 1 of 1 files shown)


providers/base/units/stress/boot.pxu
```diff
@@ -57,9 +58,7 @@ flags: preserve-locale
 id: cold-boot-loop-reboot1
 category_id: stress-tests/cold-boot
 _summary: Perform cold reboot 1
-_description: This is a template that will be used to generate a stress test
-  of the system boot. Specifically this is how the device will be request a
-  reboot.
+_description: Enter sleep mode after 120s.
```
kissiel, line 61: I think the delay is configurable, so this description may be misleading.

providers/base/units/stress/boot.pxu
```diff
@@ -72,9 +71,7 @@ depends: init-boot-loop-data
 id: cold-boot-loop-reboot{reboot_id}
 category_id: stress-tests/cold-boot
 _summary: Perform cold reboot {reboot_id}
-_description: This is a template that will be used to generate a stress test
-  of the system boot. Specifically this is how the device will be request a
-  reboot.
+_description: Enter sleep mode after 240s.
```
kissiel, line 74: Also configurable with the default being 120s. It should wake up after another 120.

kissiel (commented):
Some discrepancy between the descriptions and actual code below.

providers/base/units/stress/boot.pxu
diohe0311: I changed the description, thanks!

pieqq (approved):
+1

Thanks!

PR #521, opened 2023-05-31
Make sure all the expected snaps are built during the release process

pieqq:
`snapcraft --remote-build` may fail to build all the snaps required (one per architecture), and yet return 0 (no error), resulting in a "passed" GitHub workflow even when not all the snaps have been submitted.

To prevent this, add a step to check the expected number of snaps is present.

fail-fast strategy is set to false so that even if one of the workflow fails, the others will continue. This is useful because it is possible to manually re-run the failed workflow (for instance, the one for checkbox22), so the others (checkbox16, checkbox18, checkbox20) should not be impacted.

[1] https://github.com/snapcore/snapcraft/issues/4142

Fix CHECKBOX-446

### DIFF
--- a/.github/workflows/checkbox-core-snap-beta-release.yml
+++ b/.github/workflows/checkbox-core-snap-beta-release.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   snap:
     strategy:
+      fail-fast: false
       matrix:
         releases: [16, 18, 20, 22]
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/checkbox-core-snap-beta-release.yml
```diff
@@ -50,6 +51,25 @@ jobs:
         with:
           name: series${{ matrix.releases }}
           path: checkbox-core-snap/series${{ matrix.releases }}/*.snap
+      - name: Check that snapcraft built the expected amount of snaps
+        run: |
+          NB_SNAPS=$(ls -l checkbox-core-snap/series${{ matrix.releases }}/*.snap | wc -l)
```
yphus, line 56: You shouldn't parse the output of ls(1) (See http://mywiki.wooledge.org/ParsingLs)

```suggestion
          NB_SNAPS=$(find checkbox-core-snap/series${{ matrix.releases }} -maxdepth 1 -type f -name *.snap | wc -l)
```

yphus (changes requested):
I've proposed a minor fix, see inline (parsing `ls` is a bad idea)

.github/workflows/checkbox-core-snap-beta-release.yml
pieqq: Thanks! I made the modification both for the checkbox-core and the checkbox snap workflows.

yphus (approved):
LGTM

PR #522, opened 2023-05-31
Enable CPU tests for ARM64

rmartin013:
## Description
For Riverside project, I tried to improve CPU test coverage on ARM64 devices, and found out many tests were skipped because of a requirement limitation on the architecture. Using a side loaded provider, I tried to remove those limitations to see if they are still relevant or not, hoping I could increase the test coverage.

## Resolved issues
https://warthogs.atlassian.net/browse/PERI-117

## Tests
As a result, I can observe the following status on the sub tests [submission.tar.gz](https://github.com/canonical/checkbox/files/11615285/submission.tar.gz)
* cpu/arm_vfp_support_aarch64 is passing without any modification
* cpu/clocktest as well
* cpu/maxfreq_test is running fwts but the subsequent test maxfreq isn’t supported on ARM64, as a result, it is skipped, even if we remove the requirement restriction on ARM64 (which makes sense in that regard)
* cpu/offlining_test runs successfully if we remove the requirement restriction on ARM64
`Successfully turned 11 cores off and back on`
* cpu/scaling_test performs some tests, skip some others and fail one qualified with Medium priority, without any change
* cpu/topology relies on providers/base/bin/cpu_topology.py to compare the cpu descriptions obtained with “/proc/cpuinfo” versus “/sys/devices/system/cpu/' + proc + ‘/topology”. While sysfs gives accurate description, procfs don’t display the core id and physical id of each cpu, so the test can neither pass successfully, nor be adapted. This one should stay disabled for ARM64
* cpu/cstates can’t run as well, as it depends on a cstates subset of tests from fwts, not available for ARM64

In conclusion, the only improvement we can make is to enable the cpu/offlining_test

### DIFF
--- a/providers/base/units/cpu/jobs.pxu
+++ b/providers/base/units/cpu/jobs.pxu
@@ -145,7 +145,7 @@ _summary:
  Test offlining of each CPU core
 _description:
  Attempts to offline each core in a multicore system.
-requires: cpuinfo.platform not in ("aarch64", "armv7l")
+requires: cpuinfo.platform not in ("armv7l")
 
 plugin: shell
 category_id: com.canonical.plainbox::cpu
--- a/providers/base/units/cpu/jobs.pxu
+++ b/providers/base/units/cpu/jobs.pxu
@@ -134,18 +134,6 @@ _summary:
 _description:
  Runs a test for clock jitter on SMP machines.
 
-id: cpu_offlining
-estimated_duration: 0.02
-plugin: resource
-command:
- if ls /sys/devices/system/cpu/*/online >& /dev/null
- then
-     echo "state: supported"
- else
-     echo "state: unsupported"
- fi
-_description: Creates resource info for CPU offlining
-
 plugin: shell
 category_id: com.canonical.plainbox::cpu
 id: cpu/offlining_test
--- a/providers/base/units/cpu/jobs.pxu
+++ b/providers/base/units/cpu/jobs.pxu
@@ -134,9 +134,23 @@ _summary:
 _description:
  Runs a test for clock jitter on SMP machines.
 
+id: cpu_offlining
+estimated_duration: 0.02
+plugin: resource
+command:
+ if ls /sys/devices/system/cpu/*/online >& /dev/null
+ then
+     echo "state: supported"
+ else
+     echo "state: unsupported"
+ fi
+_description: Creates resource info for CPU offlining
+
 plugin: shell
 category_id: com.canonical.plainbox::cpu
 id: cpu/offlining_test
+requires:
+  cpu_offlining.state == 'supported'
 flags: also-after-suspend
 estimated_duration: 128.0
 user: root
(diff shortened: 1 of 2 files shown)


zongminl (approved):
+1 to this change, I was planning to submit a similar PR but thank you for doing this.
This change here to cpu/offlining_test is also adapted to one of the CE commercial projects in the project provider and it runs quite well, see submission here: https://certification.canonical.com/hardware/202201-29904/submission/273309/test-results/?term=offlining

kissiel (changes requested):
This unfortunately cannot be enabled for _all_ aarch64 platforms. For instance, AFAIK the most popular ARM computer - Raspberry Pi, doesn't support cpu offlining, so this patch would fail all the testing on RPis.
I can clearly see that you have a pass on riverside, so if you want this test to be included in your testing it has to land in the riverside provider. The easiest way forward is just to copy and paste the definition and offlining script (remember to update the namespace in the `include` field of the test plan)

rmartin013:
@kissiel Thanks for your feedback. Indeed I can copy the test in the riverside provider... I don't want to be disruptive, but I recall a having a conversation with @yphus about a way to limit the scope of a test to a specific device, or maybe in that case, to be excluded for a specific device. You probably know more the details behind this, but we could use this way to enable the test for all aarch64 except PI, since it seems @zongminl is also interested for another platform, and I expect there will be more and more coming in the future...

kissiel:
Very interesting. This opens an important discussion to be had.
The off-lining was not present (at least in a way that this test uses) on almost every aarch64 device that we've tested. It seems that it is going to change in the future.

Let's do proper job of making this sane and let's check if CPU offlining is available in the system and if it's not let's not run the job. AKA cpu offlining should be a resource in Checkbox.

From what I found in kernel docs [here](https://www.kernel.org/doc/Documentation/core-api/cpu_hotplug.rst), [here](https://www.kernel.org/doc/Documentation/admin-guide/cputopology.rst), and [here](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu)
it looks like an easy check is to see whether 'online' exists for cpu0.

Let me know if you need any help with that.

pieqq:
(I've converted this PR to "draft", so that time can be taken to analyze what's required to implement, and other reviewers know this PR is not ready to land yet. Once it is, please mark it as "Ready for review")

rmartin013:
Thanks for the links, I will be happy to work on a patch that can meet everybody's expectation :-)

rmartin013:
I created a resource for CPU offlining and used it as a requirement. I then ran again the test 
* on my Orin board the test is successful
[submission_2023-06-08T13.46.29.203509.tar.gz](https://github.com/canonical/checkbox/files/11690338/submission_2023-06-08T13.46.29.203509.tar.gz)
* on a RPi it is skipped
[submission_2023-06-08T14.39.01.976766.tar.gz](https://github.com/canonical/checkbox/files/11689974/submission_2023-06-08T14.39.01.976766.tar.gz)
The PR is ready for review

kissiel (approved):
Fantastic addition to the Checkbox world! TYVM!
It's landable as is, as a followup I'll move the resource job to the resource.pxu, unless you want to do it.

rmartin013:
Thanks for the review, and happy to help contributing to the project! I moved it to the resource.pxu file

Issue #59, opened 2022-11-28
LP1995454: Graphics/*_driver_version_* failed


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1995454

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Confirmed	| 2022-11-02 07:30:18	| 6	| Critical	| False  	|

## Description
[Summary]
Run Checkbox Graphic test item the following test were fail:
1.Graphics/1_driver_version_PCI_ID_0x4688
2.Graphics/2_driver_version_GA107BM_GeForce_RTX_3050_Ti_Mobie

[`Graphics/1_driver_version_PCI_ID_0x4688` Output]
error: list of process IDs must follow -p

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Traceback (most recent call last):
  File "/tmp/nest-ynqwi8rq.1531dd90874d443ab9a3fbf7a63c668f7a0ab3881a634524bc7811bd83eed721/graphics_driver.py", line 399, in <module>
    sys.exit(main())
  File "/tmp/nest-ynqwi8rq.1531dd90874d443ab9a3fbf7a63c668f7a0ab3881a634524bc7811bd83eed721/graphics_driver.py", line 373, in main
    xorg_owner = check_output("ps -o user= -p $(pidof Xorg)",
  File "/usr/lib/python3.10/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'ps -o user= -p $(pidof Xorg)' returned non-zero exit status 1.

[`Graphics/2_driver_version_GA107BM_GeForce_RTX_3050_Ti_Mobie` Output]
error: list of process IDs must follow -p

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Traceback (most recent call last):
[...]

PR #523, opened 2023-06-01
Fix XDG_SESSION_TYPE is not set.

pseudocc:
This is caused by some timing issue in the suspend/resume process.

In some very rare cases, `XDG_SESSION_TYPE` is not set, which causes `[ "$XDG_SESSION_TYPE" == "wayland" ]` to make the wrong decision.

## Description

Keep polling the value of `XDGSESSION_TYPE` from `systemctl --user show-environment`.

## Resolved issues

Fix this issue: #59

## Tests

ThinkPad P series, Okay.

### DIFF
--- a/providers/base/bin/graphics_env.sh
+++ b/providers/base/bin/graphics_env.sh
@@ -10,6 +10,28 @@
 DRIVER=$1
 INDEX=$2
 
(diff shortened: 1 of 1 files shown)


pieqq (approved):
Finding the root cause for this must have been challenging...

Thanks a lot for providing this fix!

Issue #524, opened 2023-06-01
Docs unknown version due to hack-ish import


Hook25:
### Bug Description

The documentation shows unknown version because of an hack-ish version fetch in the spyhinx config.py

### To Reproduce

N/A

### Environment

N/A

### Relevant log output

```shell
N/A
```

### Additional context

N/A

PR #525, opened 2023-06-01
Fixed version fetching in docs

Hook25:
## Description

This fixes the version fetching in the documentation by reading the setup.py file. This will fail once the file is updated to calculate the version, but will do so yielding an error and (hopefully) not crashing

## Resolved issues

Resolves: https://github.com/canonical/checkbox/issues/524

## Documentation

N//A

## Tests

Tested locally, should not yield any error and fetch the correct version

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,12 +1,10 @@
 import datetime
 import os
+import re
 import sys
 
-sys.path.append(os.path.abspath(
-    os.path.join(__file__, "../../checkbox-ng")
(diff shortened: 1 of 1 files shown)


yphus (commented):
I'd prefer to switch to setuptools-scm rather than landing this hack TBH.

Hook25:
Will fix in https://github.com/canonical/checkbox/pull/534 witha  better solution

PR #526, opened 2023-06-01
Fix version packages stdlib

Hook25:
## Description

This fixes the debian build requirements that the __version__ update broke

Minor: This also alters the import of the importlib-metadata to be required only when the stdlib version is not available

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


yphus (dismissed):
please add `python3-importlib-metadata | python3 (>> 3.8)` in respect of the alphabetical order of both Depends and Builds-Depends fields

yphus (approved):
LGTM

Issue #527, opened 2023-06-01
Metabox tests are failing due to missing parent dir


Hook25:
### Bug Description

Some tests were relying on the old Put behaviour of creating the parent directory of a path if a Put failed. We have removed it in favor of a clear error but now tests are failing because of that. 

### To Reproduce

Launch metabox tests in main

### Environment

N/A

### Relevant log output

```shell
N/A
```

### Additional context

N/A

PR #528, opened 2023-06-01
Fix broken tests mkdir put

Hook25:
## Description

Some tests were relying on the old Put behaviour of creating the parent directory of a path if a Put failed. We have removed it in favor of a clear error but now tests are failing because of that.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/527

## Documentation

N/A

## Tests

All tests are now in the green both legacy and remote


kissiel (approved):
Indeed it fixes a critical bug, let's land it.
Although I'd change a few bits in this, but this is not critical.

I'm taking liberty of landing it so everybody can rebase their work on this.

PR #531, opened 2023-06-04
Add Memory Technology Device test cases

baconYao:
In order to testing MTD, I leverage the mtd-utils which is an open source tool and can be downloaded from Ubuntu packages.

## Description

  - The MTD subsystem has been used in lots of devices widely, so I add MTD related test cases to cover the MTD read and write test

  - Add has_mtd manifest to determine if the platform supports MTD
  - Job mtd/check-total-numbers to make sure the number of MTD partitions in System should match the expectation which is defined by using the TOTAL_MTD_NUM environment variable
  - Resource Jjob mtd/mtd-list shows and generates all of the MTD nodes under /dev directory
  - Job mtd/read-write-test-{MTD_NAME} is a template job which is in charge of the testing for each MTD partition
    - If this MTD is read-only, I only read its content to a file.
    - If this MTD is writable, backup its original content to a file, then do the compare procedure, finally, recover its content back.

## Resolved issues
N/A

## Documentation
N/A

## Tests

CID: 202304-31517
Submission: https://certification.canonical.com/hardware/202304-31517/submission/320303/test-results/
Total MTD number: 17. Start from mt0 to mtd16
Read-only MTD: mtd0, mtd1, mtd6, mtd8, mtd10, mtd11, mtd14

### DIFF
--- a/providers/base/units/mtd/category.pxu
+++ b/providers/base/units/mtd/category.pxu
@@ -1,3 +1,3 @@
 unit: category
 id: mtd 
-_name: Memory Technology Device Test
+_name: Memory Technology Device Tests
--- a/providers/base/units/mtd/category.pxu
+++ b/providers/base/units/mtd/category.pxu
@@ -1,0 +1,3 @@
+unit: category
+id: mtd 
+_name: Memory Technology Device Test
--- a/providers/base/units/mtd/jobs.pxu
+++ b/providers/base/units/mtd/jobs.pxu
@@ -2,17 +2,24 @@ id: mtd/check-total-count
 category_id: mtd
 _summary: Check the total MTD count for the platform
 _description:
-  Check the total MTD count for the platform, relies on the user specifying the number of mtd. 
+  Check the total MTD count for the platform, relies on the user specifying the count of mtd.
   Usage of environment variable: TOTAL_MTD_COUNT={count}
   e.g. TOTAL_MTD_COUNT=2
 plugin: shell
 user: root
 imports: from com.canonical.plainbox import manifest
-requires: manifest.has_mtd == 'True'
+requires:
+  manifest.has_mtd == 'True'
+  package.name == "mtd-utils"
 flags: also-after-suspend
 estimated_duration: 5
 environ: TOTAL_MTD_COUNT
-command: mtd.sh count "$TOTAL_MTD_COUNT"
+command:
+  if [ -z "$TOTAL_MTD_COUNT" ]; then
+    echo "Missing the 'TOTAL_MTD_COUNT' environment variable. Please define it in checkbox.conf"
+    exit 1
+  fi
+  mtd.sh count "$TOTAL_MTD_COUNT"
 
 id: mtd/mtd-list
 plugin: resource
@@ -34,7 +41,9 @@ _description:
 plugin: shell
 user: root
 imports: from com.canonical.plainbox import manifest
-requires: manifest.has_mtd == 'True'
+requires:
+  manifest.has_mtd == 'True'
+  package.name == "mtd-utils"
 flags: also-after-suspend
 estimated_duration: 30
 command:
--- a/providers/base/units/mtd/jobs.pxu
+++ b/providers/base/units/mtd/jobs.pxu
@@ -2,7 +2,7 @@ id: mtd/check-total-count
 category_id: mtd
 _summary: Check the total MTD count for the platform
 _description:
-  Check the total MTD count for the platform, relies on the user specifying the count of mtd.
+  Check the total MTD count for this platform based on the number specified by the user.
   Usage of environment variable: TOTAL_MTD_COUNT={count}
   e.g. TOTAL_MTD_COUNT=2
 plugin: shell
@@ -16,7 +16,7 @@ estimated_duration: 5
 environ: TOTAL_MTD_COUNT
 command:
   if [ -z "$TOTAL_MTD_COUNT" ]; then
-    echo "Missing the 'TOTAL_MTD_COUNT' environment variable. Please define it in checkbox.conf"
+    echo "Missing the 'TOTAL_MTD_COUNT' environment variable. Please define it in the [environment] section of your Checkbox configuration.
     exit 1
   fi
   mtd.sh count "$TOTAL_MTD_COUNT"
--- a/providers/base/units/mtd/jobs.pxu
+++ b/providers/base/units/mtd/jobs.pxu
@@ -1,0 +1,42 @@
+id: mtd/check-total-numbers
+category_id: mtd
+_summary: Check the total MTD numbers for the platform
+_description:
+  Check the total MTD numbers for the platform, relies on the user specifying the number of mtd. 
+  Usage of parameter: {numbers}
+  e.g. TOTAL_MTD_NUM=2
+plugin: shell
+user: root
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_mtd == 'True'
+flags: also-after-suspend
+estimated_duration: 5
+command: mtd.sh count "$TOTAL_MTD_NUM"
+
+id: mtd/mtd-list
+plugin: resource
+_summary: Gather list of MTD on the platform
+_description:
+  Gather list of all Memory Technology Device on the platform
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_mtd == 'True'
+estimated_duration: 1s
+command: mtd.sh list  
+
+unit: template
+template-resource: mtd/mtd-list
+template-unit: job
+id: mtd/read-write-test-{MTD_NAME}
+category_id: mtd
+_summary: Test {MTD_NAME} on the platform
+_description:
+  Write a random value into {MTD_NAME}, and read it back to compare with.
+  To make sure {MTD_NAME} read and write working properly.
+plugin: shell
+user: root
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_mtd == 'True'
+flags: also-after-suspend
+estimated_duration: 30
+command:
+  mtd.sh compare {MTD_NAME}
--- a/providers/base/units/mtd/jobs.pxu
+++ b/providers/base/units/mtd/jobs.pxu
@@ -1,25 +1,24 @@
-id: mtd/check-total-numbers
+id: mtd/check-total-count
 category_id: mtd
-_summary: Check the total MTD numbers for the platform
+_summary: Check the total MTD count for the platform
 _description:
-  Check the total MTD numbers for the platform, relies on the user specifying the number of mtd. 
-  Usage of parameter: {numbers}
-  e.g. TOTAL_MTD_NUM=2
+  Check the total MTD count for the platform, relies on the user specifying the number of mtd. 
+  Usage of environment variable: TOTAL_MTD_COUNT={count}
+  e.g. TOTAL_MTD_COUNT=2
 plugin: shell
 user: root
 imports: from com.canonical.plainbox import manifest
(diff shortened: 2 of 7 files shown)


kissiel:
Quick question: there's a bit of code gathering info about MTD available on the system, yet there's a manifest entry to declare it. Why is it not enough to just have the resource job? is it because this will work only on the 2 last LTSes? If so, I'd go with the `requires` entry.

baconYao:
> Quick question: there's a bit of code gathering info about MTD available on the system, yet there's a manifest entry to declare it. Why is it not enough to just have the resource job? is it because this will work only on the 2 last LTSes? If so, I'd go with the `requires` entry.

Hi @kissiel , thanks your question.

It's not about 2 last LTSes. From my point of view, I thought the resource job no need to be performed if tester already knows this device doesn't support MTD by manifest, therefore, I set it depends on the mtd manifest.

After looking into Checkbox, there are some resource jobs depend on manifest, such as [Blutetooth](https://github.com/canonical/checkbox/blob/main/providers/base/units/bluetooth/resource.pxu#L8) and [TPM](https://github.com/canonical/checkbox/blob/main/providers/base/units/tpm/sysfs.pxu#L20).

But, I can remove the requires dependency from resource job if @kissiel you still has concern.

providers/base/units/mtd/jobs.pxu
```diff
@@ -0,0 +1,42 @@
+id: mtd/check-total-numbers
+category_id: mtd
+_summary: Check the total MTD numbers for the platform
+_description:
+  Check the total MTD numbers for the platform, relies on the user specifying the number of mtd. 
+  Usage of parameter: {numbers}
```
kissiel, line 6: This job is not parameterized and this may be misleading. If you want to express that the job uses environment variable $TOTAL_MTD_NUM to specify what's the right number of MTD devices.

Also there's no environ section in this job, so that variable would be lost when checkbox would elevate the privileges.

kissiel:
Depending on a manifest is absolutely a thing, my interest was caused by the double checking. First the resource job requires the manifest entry to be True, and then the template, on no MTD devices present, would generate 0 jobs.
I don't mind the double check, btw. Just voicing my interest :-)

There's one problem with propagating the envvar, see inline.

providers/base/units/mtd/jobs.pxu
```diff
@@ -0,0 +1,42 @@
+id: mtd/check-total-numbers
```
kissiel, line 1: Only one number is checked, and also...
 Instead of using the term 'number', which can often be associated with a unique identifier or a numerical property of a single item, we might find 'count' more appropriate for this use case. The term 'count' is widely used in coding conventions to signify a total quantity or accumulation of items.

So, for instance, instead of 'bookNumber', we might use 'bookCount' if we are tallying the total number of books. This would immediately clarify to anyone reading the code that the variable is used to keep a count of a certain set of items, rather than representing an individual number or index.

kissiel (changes requested):
there's no environment declaration in the `check-total-numbers` job's declaration - this needs fixing. 

Also see inline for potential clarity improvement.

providers/base/units/mtd/jobs.pxu
baconYao: HI @kissiel,
I revise the description and add the `environ: TOTAL_MTD_COUNT` to mtd/check-total-count job, please help review it again. Many thanks.

baconYao:
> Depending on a manifest is absolutely a thing, my interest was caused by the double checking. First the resource job requires the manifest entry to be True, and then the template, on no MTD devices present, would generate 0 jobs. I don't mind the double check, btw. Just voicing my interest :-)
> 
> There's one problem with propagating the envvar, see inline.

After reading your interest and reference other test cases, I remove the manifest requirement from `mtd/mtd-list` resource job because I think your point does matter.

providers/base/units/mtd/jobs.pxu
```diff
@@ -0,0 +1,41 @@
...
+category_id: mtd
+_summary: Check the total MTD count for the platform
+_description:
+  Check the total MTD count for the platform, relies on the user specifying the number of mtd. 
+  Usage of environment variable: TOTAL_MTD_COUNT={count}
+  e.g. TOTAL_MTD_COUNT=2
+plugin: shell
+user: root
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_mtd == 'True'
```
kissiel, line 11: I think with manifest entry removed this has to be removed as well.

providers/base/units/mtd/category.pxu
```diff
@@ -0,0 +1,3 @@
+unit: category
+id: mtd 
+_name: Memory Technology Device Test
```
kissiel, line 3: plural pls

kissiel (changes requested):
With the removal of the manifest entry, the `requires` field need to also be updated (removed).

providers/base/units/mtd/category.pxu
baconYao: Fixed, thanks

providers/base/units/mtd/manifest.pxu
```diff
@@ -0,0 +1,4 @@
+unit: manifest entry
+id: has_mtd
+_name: Does platfrom support Memory Technology Device?
```
pieqq, line 3: typo: `platform`. Also, currently this appears like that in the Manifest screen:

```
Does this machine have this piece of hardware?

Does platfrom support Memory Technology Device?     ( ) Yes   ( ) No
```

I think it would be better to just write: "Memory Technology Device (MTD) support" or something like this.

pieqq (changes requested):
One comment related to the manifest entry (see below).

Also, I tried to run the test plan on my laptop since it seems to have mtd devices:

```
find /dev -regex '.*mtd[0-9][0-9]?' | awk -F "/" '{print $3}'
mtd1
mtd0
```

I launched checkbox-cli, selected the full mtd test plan, and marked the manifest `has_mtd` to `true`:

```
cat /var/tmp/checkbox-ng/machine-manifest.json 
{
  "com.canonical.certification::has_mtd": true
}
```


But it failed with different errors:

```
mtd.sh: line 18: [: : integer expression expected 

mtd.sh: line 28: mtd_debug: command not found 
```

Please check this submission for more info: https://certification.canonical.com/hardware/202204-30172/submission/327075/

providers/base/units/mtd/manifest.pxu
baconYao: Thanks. I modified the name as your suggestion.

baconYao:
Hi @pieqq, the following links are the test result after fixing something with different conditions.

- The mtd-utils package be installed but no TOTAL_MTD_COUNT be defined
  - https://certification.canonical.com/hardware/202304-31517/submission/329145/
- Without mtd-utils package be installed
  - https://certification.canonical.com/hardware/202304-31517/

providers/base/units/mtd/jobs.pxu
```diff
@@ -2,17 +2,24 @@ id: mtd/check-total-count
 category_id: mtd
 _summary: Check the total MTD count for the platform
```
pieqq, line 3: I'm wondering if "for the platform" is required here, but it's not a big deal :)

providers/base/units/mtd/jobs.pxu
```diff
@@ -2,17 +2,24 @@ id: mtd/check-total-count
...
+requires:
+  manifest.has_mtd == 'True'
+  package.name == "mtd-utils"
 flags: also-after-suspend
 estimated_duration: 5
 environ: TOTAL_MTD_COUNT
-command: mtd.sh count "$TOTAL_MTD_COUNT"
+command:
+  if [ -z "$TOTAL_MTD_COUNT" ]; then
+    echo "Missing the 'TOTAL_MTD_COUNT' environment variable. Please define it in checkbox.conf"
```
pieqq, line 19: It can also be defined in a launcher, so maybe something more generic like

> Please define it in the [environment] section of your Checkbox configuration.

would be better?

providers/base/units/mtd/manifest.pxu
```diff
@@ -1,4 +1,4 @@
 unit: manifest entry
 id: has_mtd
-_name: Does platfrom support Memory Technology Device?
+_name: Memory Technology Device (MTD) support?
```
pieqq, line 3: I think the last part (`support?`) is not required here, because there is already a question above in the Manifest screen:

> Does this machine have this piece of hardware?
> 
> Memory Technology Device (MTD) support?

It feels better if it's just

> Does this machine have this piece of hardware?
> 
> Memory Technology Device (MTD)

pieqq (changes requested):
Thanks for the changes!

I've tested the full test plan on my machine after installing `mtd-utils` and setting the environment variable. Results are here:

https://certification.canonical.com/hardware/202204-30172/submission/329331/

For some reason, one of them fails:

```
##### Erase MTD #####
MEMERASE: Unknown error 524
```

but I guess it's not related to your code.

----------------------

A few minor comments below.

Also, it means `mtd-utils` must be added to the snap packages and to the Debian packages. The problem is this package was [only included since focal](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=mtd-utils), so I'm not sure how to proceed here. @yphus, @kissiel  do you know if it's possible to make use of [packaging meta-data](https://checkbox.readthedocs.io/en/latest/reference/units/packaging-meta-data.html) to target only 20.04 and 22.04 (since these tests would never be executed on 18.04 or older)?

In any case, the `snapcraft.yaml` recipes also need to be modified for checkbox20 and checkbox22.

kissiel:
> Thanks for the changes!
> 
> I've tested the full test plan on my machine after installing `mtd-utils` and setting the environment variable. Results are here:
> 
> https://certification.canonical.com/hardware/202204-30172/submission/329331/
> 
> For some reason, one of them fails:
> 
> ```
> ##### Erase MTD #####
> MEMERASE: Unknown error 524
> ```
> 
> but I guess it's not related to your code.
> 
> A few minor comments below.
> 
> Also, it means `mtd-utils` must be added to the snap packages and to the Debian packages (as a dependency). The problem is this package was [only included since focal](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=mtd-utils), so I'm not sure how to proceed here. @yphus, @kissiel do you know if it's possible to make use of [packaging meta-data](https://checkbox.readthedocs.io/en/latest/reference/units/packaging-meta-data.html) to target only 20.04 and 22.04 (since these tests would never be executed on 18.04 or older)?
> 
> In any case, the `snapcraft.yaml` recipes also need to be modified for checkbox20 and checkbox22.

We use `os-version-id: 22.04` in those units. I guess if you want two, you'd have to create two PMDs.

pieqq:
@zongminl mentioned that this test might be destructive: on one of the project he's working on, the MTD was storing some firmware-related data that was erased when that test was run, leading to a bricked device!

So at the very least, the read-write-test should not be included in test plans, unless it has been confirmed with a given project that it was safe to do so (in which case it could be included in a project-specific test plan).

pieqq:
@zongminl What should be done about this PR? Implement the flag check to decide whether or not to skip these tests? Or merge it as is and leave it up to QA for each project to decide whether or not they want to add `mtd-full` as part of their test plans?

zongminl:
In this change the flag check has been implemented but from my experience there are still devices that don't have readonly flag as a protection to the data inside of it.
I'd suggest not to automatically generate the `mtd/read-write-test-.*` test jobs from checking what's uder `/dev/`, but using a config_var so tester will have to list all the MTD_NAME

kissiel:
There will be improvement landed in another PR making this a bit obsolete. @baconYao confirmed that we can close this one.

Issue #444, opened 2023-05-05
Interrupt menu uses outdated master/slave nomenclature


Hook25:
### Bug Description

The interrupt menu is still using the outdated nomenclature to refer to remote and service

### To Reproduce

1. Launch `checkbox-cli service`
2. Lauch `checkbox-cli remote $SERVICE_IP`
3. Ctrl+C in the checkbox remote CLI

### Environment

N/A

### Relevant log output

```shell
What do you want to interrupt?
    (X) Cancel the interruption and resume the session (ESC)
    ( ) Kill test in progress and move on to next
    ( ) Disconnect the master (Same as CTRL+C)
    ( ) Stop the checkbox slave @127.0.0.1
    ( ) Abandon the session on the slave @127.0.0.1
```

yphus:
Tracked [here](https://warthogs.atlassian.net/browse/CHECKBOX-657)

Issue #129, opened 2022-11-28
LP1937096: checkbox remote ctrl-c prompt is confusing


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/checkbox-ng/+bug/1937096

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Triaged	| 2021-07-21 15:02:36	| 10	| Medium	| False  	|

## Description
when I do ctrl-c when checkbox-remote is running and connected to a session, the options are confusing:

What do you want to interrupt?                                                                                        │
│                                                                                                                                          │
│               ( ) Cancel the interruption and resume the session (ESC)                                                                   │
│               ( ) Kill test in progress and move on to next                                                                              │
│               ( ) Disconnect the master (Same as CTRL+C)                                                                                 │
│               ( ) Stop the checkbox slave @10.245.130.30                                                                                 │
│               (X) Abandon the session on the slave @10.245.130.30                                                            

I can guess that the first option means "I made a mistake and need to resume testing"
I can guess that the second option means exactly what it says.
I "THINK" that the third option is the same as a disconnect in screen, leaving the tests running but disconnecting the remote connection for a bit.
What does the fourth option do?  COmpletely kill "checkbox-cli service" on the remote?
And what does the fifth do? What exactly is "Abandon the session"? Is that the same as "Kill the test run currently in progress"? Or is that the same as disconnect and forget it ever happened?

Basically, which of those is actually "Kill the test run that I just started"? And that could be messaged a bit better.

Additionally, I believe we got away from the Master/Slave terminology in favor of Remote/Service, so there's a bit of a hanger-on there.

Note, this is checkbox-snappy running on the R340, I haven't looked at current checkbox via Debs to see if this affects that too.

## Attachments
No attachments

Tags:
[]

beliaev-maksim:
This thread was migrated from launchpad.net
##### https://launchpad.net/~pieq wrote on 2022-03-15 03:46:14:
Same issue here in the QA team. Sometimes we would like to stop the current session, but it's not clear what option we should use.

For instance, let's say we are running a stress test. After 9 failed reboots, we realize that the test run is going nowhere and there is no point running yet another 91 reboots that are going to fail. We need to stop the current session on the slave, but we still need access to the session data (io-logs, etc.), so we don't want to delete it. Like Jeff, we are not sure what the last option does exactly.

[...]

Issue #19, opened 2022-10-21
LP1898867: Ctrl-c to kill remote job traceback


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/checkbox-ng/+bug/1898867

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Confirmed	| 2020-10-07 11:57:20	| 6	| Critical	| False  	|

## Description
-----------------------------[ Running job 2 / 2 ]------------------------------
---------------------[ Stress of CPUs (very long runtime) ]---------------------
ID: com.canonical.certification::stress/cpu_stress_ng_test
Category: Stress tests
--------------------------------------------------------------------------------
STRESS_NG_CPU_TIME env var is not found, stress_ng cpu running time is default value
stress-ng: info:  [1858] dispatching hogs: 4 af-alg, 4 bsearch, 4 context, 4 cpu, 4 crypt, 4 hsearch, 4 longjmp, 4 lsearch, 4 matrix, 4 qsort, 4 str, 4 stream, 4 tsearch, 4 vecmath, 4 wcs
stress-ng: info:  [1859] stress-ng-af-alg: 23 cryptographic algorithms found in /proc/crypto
stress-ng: info:  [1859] stress-ng-af-alg: 106 cryptographic algorithms max (with defconfigs)
stress-ng: info:  [1913] stress-ng-stream: using built-in defaults as unable to determine cache details
stress-ng: info:  [1913] stress-ng-stream: stressor loosely based on a variant of the STREAM benchmark code
stress-ng: info:  [1913] stress-ng-stream: do NOT submit any of these results to the STREAM benchmark results
stress-ng: info:  [1913] stress-ng-stream: Using CPU cache size of 2048K
^CTraceback (most recent call last):
  File "/usr/local/bin/checkbox-cli", line 11, in <module>
    load_entry_point('checkbox-ng', 'console_scripts', 'checkbox-cli')()
  File "/home/sylvain/lp/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py", line 123, in main
    subcmd.invoked(ctx)
  File "/home/sylvain/lp/checkbox-ng/checkbox_ng/launcher/master.py", line 155, in invoked
    self.connect_and_run(ctx.args.host, port)
  File "/home/sylvain/lp/checkbox-ng/checkbox_ng/launcher/master.py", line 179, in connect_and_run
[...]

PR #540, opened 2023-06-08
Change the menu text to something more clear

Hook25:
## Description
This fixes the text in the interrupt menu to be easier to understand, also removes the outdated nomenclature for remote and service

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/444, https://github.com/canonical/checkbox/issues/129, https://github.com/canonical/checkbox/issues/19, https://github.com/canonical/checkbox/issues/550

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -692,10 +692,10 @@ def interrupt_dialog(host):
     ]
     choices = [
         _("Nothing, continue testing (ESC)"),
-        _("Stop the test in progress and move on to the next"),
+        _("Stop the test case in progress and move on to the next"),
         _("Disconnect but let the test session continue (CTRL+C)"),
         _("Exit and stop the Checkbox service on the testbed at {}".format(host)),
-        _("Finalize this session and launch a new test plan"),
+        _("End this test session preserving its data and launch a new one"),
     ]
     footer_text = [
         ('Press '), ('start', '<Enter>'), (' or '),
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -691,11 +691,11 @@ def interrupt_dialog(host):
         ('start', 'dark green,bold', 'black'),
     ]
     choices = [
-        _("Cancel the interruption and resume the session (ESC)"),
-        _("Kill test in progress and move on to next"),
-        _("Disconnect the master (Same as CTRL+C)"),
-        _("Stop the checkbox slave @{}".format(host)),
-        _("Abandon the session on the slave @{}".format(host)),
+        _("Nothing, continue testing (ESC)"),
+        _("Kill test in progress and move on to the next"),
+        _("Exit but let the service test run continue (CTRL+C)"),
+        _("Exit and kill the service test runer at {}".format(host)),
+        _("Finalize this session and launch a new test plan"),
     ]
     footer_text = [
         ('Press '), ('start', '<Enter>'), (' or '),
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -692,9 +692,9 @@ def interrupt_dialog(host):
     ]
     choices = [
         _("Nothing, continue testing (ESC)"),
-        _("Kill test in progress and move on to the next"),
-        _("Exit but let the service test run continue (CTRL+C)"),
-        _("Exit and kill the service test runer at {}".format(host)),
+        _("Stop the test in progress and move on to the next"),
+        _("Disconnect but let the test session continue (CTRL+C)"),
+        _("Exit and stop the Checkbox service on the testbed at {}".format(host)),
         _("Finalize this session and launch a new test plan"),
     ]
     footer_text = [
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -374,7 +374,9 @@ class UnifiedRunner(IJobRunner):
 
     def send_signal(self, signal, target_user):
         if not target_user:
-            os.kill(self._running_jobs_pid, signal)
+            if self._running_jobs_pid:
+                return os.kill(self._running_jobs_pid, signal)
+            logger.error("No job is currently running")
         else:
             # process used sudo, so sudo is needed to kill it
             in_r, in_w = os.pipe()
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -373,10 +373,13 @@ class UnifiedRunner(IJobRunner):
                             "{}.record.gz".format(slugify(job.id)))
 
     def send_signal(self, signal, target_user):
-        if not target_user:
-            if self._running_jobs_pid:
-                return os.kill(self._running_jobs_pid, signal)
+        if not self._running_jobs_pid:
+            # this can happen because the kill command is issued
+            # just as the job finishes
             logger.error("No job is currently running")
+            return
+        if not target_user:
+            os.kill(self._running_jobs_pid, signal)
         else:
             # process used sudo, so sudo is needed to kill it
             in_r, in_w = os.pipe()


checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -691,11 +691,11 @@ def interrupt_dialog(host):
...
     choices = [
-        _("Cancel the interruption and resume the session (ESC)"),
-        _("Kill test in progress and move on to next"),
-        _("Disconnect the master (Same as CTRL+C)"),
-        _("Stop the checkbox slave @{}".format(host)),
-        _("Abandon the session on the slave @{}".format(host)),
+        _("Nothing, continue testing (ESC)"),
+        _("Kill test in progress and move on to the next"),
+        _("Exit but let the service test run continue (CTRL+C)"),
+        _("Exit and kill the service test runer at {}".format(host)),
```
pieqq, line 697: Maybe `Exit and kill the Checkbox service on the testbed at...`?

checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -691,11 +691,11 @@ def interrupt_dialog(host):
...
     ]
     choices = [
-        _("Cancel the interruption and resume the session (ESC)"),
-        _("Kill test in progress and move on to next"),
-        _("Disconnect the master (Same as CTRL+C)"),
-        _("Stop the checkbox slave @{}".format(host)),
-        _("Abandon the session on the slave @{}".format(host)),
+        _("Nothing, continue testing (ESC)"),
+        _("Kill test in progress and move on to the next"),
+        _("Exit but let the service test run continue (CTRL+C)"),
```
pieqq, line 696: We have [that card](https://warthogs.atlassian.net/browse/CHECKBOX-656) to figure out what naming to use, and it looks like we're gonna go with `controller` and `testbed`.

In this case, using `Disconnect the controller (same as Ctrl+C)` would make sense. What do you think?

checkbox-ng/plainbox/impl/execution.py
```diff
@@ -374,7 +374,9 @@ def get_record_path_for_job(self, job):
 
     def send_signal(self, signal, target_user):
         if not target_user:
-            os.kill(self._running_jobs_pid, signal)
+            if self._running_jobs_pid:
+                return os.kill(self._running_jobs_pid, signal)
```
pieqq, line 378: could this happen in the `else` clause, that is, if there is a `target_user`? Should the existence of `self._running_jobs_pid` be checked regardless of this? (I don't have an answer to this question, but I saw that `self._running_jobs_pid` is being used below when running a `sudo` command to kill the process, so...

pieqq (changes requested):
Thanks for cleaning up the wording in this screen! @bladernr will be very happy about that :D

A few comments below.

checkbox-ng/checkbox_ng/urwid_ui.py
kissiel: +1.
Maybe even: `Disconnect the controller, but let the session continue (same as Ctrl+C)`
Also the original `Exit but let the service test run continue` is difficult to read.

checkbox-ng/checkbox_ng/urwid_ui.py
kissiel: s/kill/stop.
But I think we should remove this option altogether.
In 99%+ cases this would mean that the service is just restarted.

checkbox-ng/plainbox/impl/execution.py
kissiel: in the `else` block we don't crash on failed kills, so we don't have to. But if we want to be courteous then yes.

kissiel (commented):
Added 3 cents in response to Pierre's comments.

pieqq:
FYI, I'm working on rewording the docs (https://warthogs.atlassian.net/browse/CHECKBOX-658) so I'll rework [the section](https://checkbox.readthedocs.io/en/latest/explanation/remote.html#session-control) that explains the screen you are modifying based on whatever will come out from this PR!

checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -691,11 +691,11 @@ def interrupt_dialog(host):
...
-        _("Cancel the interruption and resume the session (ESC)"),
-        _("Kill test in progress and move on to next"),
-        _("Disconnect the master (Same as CTRL+C)"),
-        _("Stop the checkbox slave @{}".format(host)),
-        _("Abandon the session on the slave @{}".format(host)),
+        _("Nothing, continue testing (ESC)"),
+        _("Stop the test in progress and move on to the next"),
+        _("Disconnect but let the test session continue (CTRL+C)"),
+        _("Exit and stop the Checkbox service on the testbed at {}".format(host)),
+        _("Finalize this session and launch a new test plan"),
```
pieqq, line 698: "Finalize" may sound like the session will be wrapped up, and the user will have the opportunity to upload it to C3 or to download it. It's not the case: Checkbox removes the `incomplete` flag from the session `metadata` `flags` section, marking it as "finished" even though nothing has been done with it. By doing so, this session will not be resumable, but since there are probably jobs that have not been run, submitting it to C3 would not be very useful.

So I guess "Abandon" is more accurate.

Moreover, since we are talking about a session, it's better to stick to this word for the second part of the sentence, so you could write "and start a new one" for instance.

checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -691,11 +691,11 @@ def interrupt_dialog(host):
         ('start', 'dark green,bold', 'black'),
     ]
     choices = [
-        _("Cancel the interruption and resume the session (ESC)"),
-        _("Kill test in progress and move on to next"),
-        _("Disconnect the master (Same as CTRL+C)"),
-        _("Stop the checkbox slave @{}".format(host)),
-        _("Abandon the session on the slave @{}".format(host)),
+        _("Nothing, continue testing (ESC)"),
+        _("Stop the test in progress and move on to the next"),
```
pieqq, line 695: If I select this option, what will happen to that test? Will it be marked as `skipped` or `fail`? Or something else? I actually tried and... crashed Checkbox :D → #550

Hook25: This should be fixed by the drive-by fix that I have done in this PR! That `if ... is None: ...`

checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -691,11 +691,11 @@ def interrupt_dialog(host):
...
+        _("Disconnect but let the test session continue (CTRL+C)"),
```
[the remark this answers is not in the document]
Hook25: I disagree on the similar, the keybind is that and when listing switch keys we just mention them in parenthesis. If it is similar it means that the effect it is not the same (or at least, that is what I thought the first time I saw the message), and it left me wondering what was the difference and whether I should use the keybind or the menu entry

checkbox-ng/checkbox_ng/urwid_ui.py
Hook25: Just one caveat to Abandon, the session is still there and downloadable for inspection, I wanted a verb to signal the fact that we are leaving it as finished and not removing it

checkbox-ng/checkbox_ng/urwid_ui.py
pieqq: Awesome! In that case, you should specify this in the commit message. I was not sure why you had changed this part of the code to be honest. Think of commit messages as a help for future archeologists: you in 3 months, me and everyone in the team ;)

Hook25: For reference: https://github.com/canonical/checkbox/pull/540/commits/1f5815c7a41e60527f3848b72983c771cc63272c

checkbox-ng/checkbox_ng/urwid_ui.py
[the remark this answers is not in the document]
pieqq: OK, I see what you mean. I didn't see it like that when I was looking at the git diff, but you're right, if shown like this (`(Esc)`, `(Ctrl+C)`) it looks more like shortcuts that can be used by the user.

checkbox-ng/checkbox_ng/urwid_ui.py
bladernr: `"Stop the test in progress and move on to the next"`
One minor comment, I'd be just bit more specific here and say "test case" rather than "test"  as just "test" feels too generic, is it the test case, the whole test suite? a group of tests (e.g. the CPU tests)? But maybe that's just how I was reading it this morning.

`"Finalize this session and launch a new test plan"`
And with this one... what does "finalize" mean?  That wording always felt awkward to me... 

"End this test session, preserving its data, and launch a new test session" sounds more accurate to me, but that's just IMO.

I feel like these sorts of discussions always feel a bit like bike shedding, despite the fact that getting the words right is so very important, as @pieqq pointed out.

pieqq (approved):
Thanks for this change, and thanks for your patience with all the back and forth renaming strings implies :)

+1

Issue #159, opened 2022-11-28
LP1952975: [remote] Avoid flooding the screen with "Reconnecting..."


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/checkbox-ng/+bug/1952975

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| New	| 2021-12-02 02:33:33	| 8	| Wishlist	| False  	|

## Description
Issue seen with any recent version of Checkbox. The one tested here is checkbox20 snap v1.24 (717).

When using Checkbox remote, if the connection to the DUT is lost, the word "Reconnecting..." is printed out every second or so. When performing something like reboot stress tests on the DUT, it means we get outputs like this:

----------------------------[ Running job 88 / 445 ]----------------------------
---------------------------[ Perform warm reboot 29 ]---------------------------
ID: com.canonical.certification::warm-boot-loop-reboot29
Category: Warm-boot Stress Test
--------------------------------------------------------------------------------
Connection lost!
connection closed by peer
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
Reconnecting...
---------------------------[ Perform warm reboot 29 ]---------------------------
[...]

PR #541, opened 2023-06-08
Avoid flooding stdout with reconnecting

Hook25:
## Description

When disconnected from service, remote prints two reconnecting per second and this is suboptimal because it fills the stdout (and logs) for no reason

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/159

## Documentation

N/A

## Tests

This was tested via metabox on jammy

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/master.py
+++ b/checkbox-ng/checkbox_ng/launcher/master.py
@@ -171,7 +171,7 @@ class RemoteMaster(ReportsStage, MainLoopStage):
         #  check if ever disconnected
         ever_disconnected = False
         #  this to animate the dash
-        spinner = itertools.cycle('-/|\\')
+        spinner = itertools.cycle('-\\|/')
         #  this tracks the disconnection time
         disconnection_time = 0
         while True:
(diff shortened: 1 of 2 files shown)


pieqq (commented):
Thanks for this QoL change.

One question I have is with regards to execution in CI/CD pipelines. The current situation (and the proposal in the issue: `Reconnecting..............`) allows us to have an idea of how long it took to reconnect, because of the actually printed output.

What happens in the case of a Jenkins run for instance?

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -249,8 +266,13 @@ def quitter(msg):
                 if not keep_running:
                     raise
                 # it's reconnecting, so we can ignore refuses
-                print('Reconnecting...')
-                time.sleep(0.5)
+                if not printed_reconnecting:
+                    print('Reconnecting ', end="")
```
pieqq, line 270: You can switch to double quotes since you use them for the `end` parameter.

pieqq (approved):
Minor comments inline, but otherwise LGTM!

Thanks! And the time addition will be useful for sure.

PR #542, opened 2023-06-08
Config refactor load configs

Hook25:
## Description

This includes a new test on config chains and an implementation of `load_configs` that is non-recursive and models the behaviour that we wanted to have from it.

## Resolved issues

N/A

## Documentation

N/A

## Tests

The test was verified and run via metabox on jammy


pieqq:
Unrelated, but don't take the "Link check" problems into account: it's because I removed the Plainbox readthedocs project, so a link was broken. I fix that in #543

checkbox-ng/checkbox_ng/config.py
```diff
@@ -33,51 +33,98 @@
...
+    """
+    if os.path.isabs(expand_all(name)):
+        return [name]
+    to_r = []
+    for sdir in SEARCH_DIRS:
+        config = expand_all(os.path.join(sdir, name))
+        if os.path.exists(config):
+            to_r.append(config)
+        else:
+            _logger.info("Referenced config file doesn't exist: %s", config)
```
kissiel, line 63: We shouldn't INFO about that, maybe DEBUG worth, but even that is a stretch. And the message here is quite strong for a typical case, even more so considering there's no message before the loop.
For instance the following is much less scary:
```
DEBUG: searching for checkbox.conf files...
DEBUG: not found in ~/.config
...
```

metabox/metabox/scenarios/config/chains.py
```diff
@@ -0,0 +1,57 @@
...
+
+    steps = [
+        Put("/etc/xdg/checkbox.conf", etc_checkbox),
+        Put("/etc/xdg/includes.conf", etc_checkbox_includes),
+        Put("/etc/xdg/includes_includes.conf", etc_checkbox_includes_includes),
+        RunCmd("mkdir -p /home/ubuntu/.config"),
+        Put("/home/ubuntu/.config/checkbox.conf", home_checkbox),
+        Put("/home/ubuntu/.config/includes.conf", home_checkbox_includes),
+        Start(cmd="check-config"),
+        AssertPrinted("a\s*=\s*0"),
```
kissiel, line 52: Why fuzzy regex instead of literal string? You're providing the expected output and there are no moving parts. Regexes make everything more complicated :D

metabox/metabox/scenarios/config/config_files/chains/etc_checkbox.conf
```diff
@@ -0,0 +1,10 @@
+[config]
+config_filename=includes.conf
+[manifest]
+# overwrite a set by some.conf
+a = 0
```
kissiel, line 5: 0 doesn't say anything, I'd much more prefer "value from etc checkbox conf".
ofc this applies to all below.

checkbox-ng/checkbox_ng/config.py
```diff
@@ -33,51 +33,98 @@
...
-                cfg.update_from_another(
-                    Configuration.from_path(config),
-                    'config file: {}'.format(config))
-            else:
-                _logger.info(
-                    "Referenced config file doesn't exist: %s", config)
-    new_cfg_filename = cfg.get_value('config', 'config_filename')
-    if new_cfg_filename != previous_cfg_name:
-        load_configs(launcher_file, cfg)
+        # configs to read who may have a redirect in them
```
kissiel, line 96: s/who/which
s/may have a redirect/may reference other configs/

checkbox-ng/checkbox_ng/config.py
```diff
@@ -33,51 +33,98 @@
...
+        to_load_conf_names = _search_configs_by_name(
+            cfg.get_value("config", "config_filename")
+        )
+    # used to avoid "loops"
+    # Note: checkbox.conf is always the default "config_filename"
+    #       so we *always* have a loop eventually
+    already_loaded = {cfg.get_value("config", "config_filename")}
+    loaded_confs_sources = []
+    while to_load_conf_names:
+        to_load = to_load_conf_names.pop(0)
```
kissiel, line 106: AFAIK this will never have more than one item. No need to use a collection.

kissiel (commented):
Some comments below.
I still can't see how introducing this many new things makes it clearer.

checkbox-ng/checkbox_ng/config.py
kissiel: so here's the kicker:
if `~/.config/A specifies config_filename` to be 'B' instead, `/etc/xdg/A` shouldn't be read :D

Hook25: I think that is a bit too much, I'm not saying it can not be done, but I'm saying that it makes the semantic of the config chain too complicated and basically impossible to explain. If it was up to me I would keep this new semantic that is still very complicated, but it is compatible with the old one in the simple case (just one `config_filename` level). Do you agree? Else I can change it to be as you described

Hook25:
Closing as pulled in https://github.com/canonical/checkbox/pull/519

PR #543, opened 2023-06-09
Add link anchors to documentation and fix broken link to removed Plainbox docs

pieqq:
## Description

With the recent removal of Plainbox docs, one link in the Checkbox docs was reported as broken:

https://github.com/canonical/checkbox/actions/runs/5213187789/jobs/9407950316?pr=542#step:4:81

While fixing this, I took the opportunity to rework the anchors to all the fields for all the Checkbox unit types. This will make it easier for us to point users to the right place in the documentation. This is because we list the different fields as a definition list, which does not create HTML anchors. They need to be created manually.

For example, I can now point to the [`command` field of the Job unit](https://canonical-checkbox--543.com.readthedocs.build/en/543/reference/units/job.html#job-command-field).

Edit: this patch also removes an uncessary table of content in the "Understanding" page.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-594

## Tests

Tested by building docs locally:

```
cd docs/
make install
make run 
```

### DIFF
--- a/docs/explanation/understanding.rst
+++ b/docs/explanation/understanding.rst
@@ -1,8 +1,6 @@
 Understanding Checkbox
 ======================
 
-.. contents::
-
 Checkbox by itself doesn't test anything. It uses unit definitions grouped in
 providers to actually run or do something.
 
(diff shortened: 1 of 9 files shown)


diohe0311:
Thanks for rework the docs! really helpful for new starter.
I tried every steps in tutorial, and those pictures really make it easier to understand.

diohe0311 (approved):
Seems great! thanks for the work

PR #544, opened 2023-06-09
Allow reusing containers in metabox provisioning

Hook25:
## Description

When changing the code of anything that is included in the provisioning (providers or checkbox) the only way to test these changes via metabox is re-provisioning the containers. This operation may take a lot of time and it is not always necessary. This PR introduces the option of re-using the already provisioned container by overwriting the source code and re-installing it. The idea is that by doing this we save the time it takes to install all of the dependencies (because they are already there) and creating the "base" machine (setting up the image etc.).

In my tests this saves around 2 minutes per provisioning, cutting the time from 3minutes to 1, double that for remote tests

## Resolved issues

N/A

## Documentation

The option has a standard CLI help prompt 

## Tests

All tests

### DIFF
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -191,11 +191,11 @@ class LxdMachineProvider:
             self._store_config(machine)
             logger.debug("Stopping container {}...", container.name)
             container.stop(wait=True)
-            logger.debug("Creating 'provisioned' snapshot for {}...", container.name)
             if use_existing:
                 with suppress(NotFound):
                     container.snapshots.get('provisioned').delete(wait=True)
-                    logger.debug("Deleted old provisioned")
+                    logger.debug("Deleted old 'provisioned'")
+            logger.debug("Creating 'provisioned' snapshot for {}...", container.name)
             container.snapshots.create(
                 'provisioned', stateful=False, wait=True)
             logger.debug("Starting container {}...", container.name)
@@ -253,6 +253,13 @@ class LxdMachineProvider:
                     machine._container,
                     "sudo mkdir -p {}".format(os.path.dirname(dest)),
                     verbose=self._debug_machine_setup
+                )
+                # Ensure that the location is not already there
+                #  else cp will copy inside it
+                run_or_raise(
+                    machine._container,
+                    "sudo rm -rf {} 2>/dev/null || true".format(dest),
+                    verbose = self._debug_machine_setup
                 )
                 # Copy the mounted dir to the desired location
                 run_or_raise(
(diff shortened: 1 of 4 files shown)


metabox/metabox/core/lxd_provider.py
```diff
@@ -125,7 +133,11 @@ def _create_profiles(self):
                 logger.debug(
                     '{} LXD profile created successfully', profile_name)
 
-    def _create_machine(self, config):
+    def _create_machine(self, config, use_existing=False):
+        assert (
+            not use_existing or
+            (use_existing and config.origin == 'source')
```
kissiel, line 139: This cannot be asserted, therefore don't use assert. Assert is for when you control all of the axioms.
This, speaking very broadly, is a "contradicting input" kind of input error.

metabox/metabox/main.py
```diff
@@ -61,6 +61,9 @@ def main():
     parser.add_argument(
         '--do-not-dispose', action='store_true',
         help="Do not delete LXD containers after the run")
+    parser.add_argument(
+        '--redeploy-existing', action='store_true',
+        help="Re-Deploy and install existing containers")
```
kissiel, line 66: TBH I don't know what do you mean to install a container.

metabox/metabox/core/lxd_provider.py
```diff
@@ -232,6 +254,13 @@ def _run_transfer_commands(self, machine):
                     "sudo mkdir -p {}".format(os.path.dirname(dest)),
                     verbose=self._debug_machine_setup
                 )
+                # Ensure that the location is not already there
+                #  else cp will copy inside it
+                run_or_raise(
+                    machine._container,
+                    "sudo rm -rf {} 2>/dev/null || true".format(dest),
```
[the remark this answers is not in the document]
Hook25: The rationale here is: if the directory is there remove it, if it is not, we do not care. The reason this was done is to avoid copying the files inside it as the command following this would do. I do not know a better way to do such a thing with `cp` alone. I'm piping away the error to "hide" the location does not exist error because I would pollute the log with no benefit

metabox/metabox/main.py
Hook25: Can be made better, this means copy and install everything that needs to be copied/installed (providers + checkbox). Do you have a better description?

Hook25:
> I'm not so sure that the default should be to reuse. I can see multiple ways it can go wrong.

Agree, it is kind of dangerous, but it is useful to have in "developer" mode as a quick and dirty way to test changes according to me

kissiel:
> > I'm not so sure that the default should be to reuse. I can see multiple ways it can go wrong.
> 
> Agree, it is kind of dangerous, but it is useful to have in "developer" mode as a quick and dirty way to test changes according to me

I agree. Having a flag for that doesn't sound unquick :), stuff like `--reuse-containers`

kissiel (approved):
Fantastic changes, TYVM!

PR #545, opened 2023-06-09
Warning for skipped file at provider loading

petitfou:
Currently, if we put a non executable file in the provider bin folder the file will be skipped at provider deployment and can cause the provider malfunction. Tools like checkbox-provider-tools just silently ignores the file at validation and install steps. Printing some warning message would help users to figure out more quickly these kinds of mistake.

## Resolved issues

N/A

## Documentation

N/A

## Tests

Put a non executable python lib file in bin folder, the checkbox-provider-tools will print warning for this skipped file.
 
```
checkbox-provider-kivu$ checkbox-provider-tools install --layout=relocatable --prefix=/providers/checkbox-provider-kivu --root=/tmp/

WARNING plainbox.secure.providers.v1: Skipped file: /home/hector/workspace/src/kivu/conf/checkbox-providers/checkbox-kivu-git/checkbox-provider-kivu/bin/gpu_utils.py
```

### DIFF
--- a/checkbox-ng/plainbox/impl/secure/providers/v1.py
+++ b/checkbox-ng/plainbox/impl/secure/providers/v1.py
@@ -586,12 +586,26 @@ class ProviderContentLoader:
         self.problem_list.extend(self.provider.content_collection.problem_list)
         self.is_loaded = True
 
+    def _warn_ignored_file(self, filename):
+        """
+        Print an warning message for each file that is skipped at loading
+        Do not print warning for all skipped files, do it only for file that is located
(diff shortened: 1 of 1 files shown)


kissiel (approved):
Awesome, thank you!

alexandredosim:
Hello!
It's great that now CheckBox warns about these mistakes, but I think the message should be more descriptive. I found what was wrong by reading the changelog for version 2.8, for the next versions and for the newcomers would be hard to know what the problem is.

PR #547, opened 2023-06-12
CHECKBOX-619: Fan unit test modification

diohe0311:
## Description
After discuss with Pierre, we summarize 4 issues:
1. shouldn’t show msg if test passed
2. unit test shouldn’t create a real instance
3. didn't found any fan but still pass the test (may create a resource job to skip the test or use manifest)
4. possible not finding all the fan on the device, some fan didn't exist in sysfs(might need other approach, TBD)

Current modification only fix the first 2 issues, which is directly effected by test_fan_reaction.py
The other 2 needs more discussion, and might needs to modify fan_reaction_test.py based on discussion result.
(Also need new Jira card for that)

## Resolved issues
https://warthogs.atlassian.net/browse/CHECKBOX-619?atlOrigin=eyJpIjoiNTZmYWFjYzk4MzYwNDE2OWIxOTQxM2VlMmMwYTVkOWUiLCJwIjoiaiJ9

## Tests
Tested locally with:
 `./manage.py test -u`


providers/base/tests/test_fan_reaction.py
```diff
@@ -17,93 +17,84 @@
...
-                    "0000:02:00.0/0000:03:00.0", "foo"]
-            relpath_mock.side_effect = ['hwmon6/fan2_input']
-            # The following call is patching open(pci_class_path, 'r')
-            with patch("builtins.open", mock_open(read_data='0x030000')) as f:
-                fan_mon = FanMonitor()
-                self.assertEqual(len(fan_mon.hwmons), 1)
-                self.assertTrue(fan_mon.hwmons[0].endswith('fan2_input'))
-            self.assertEqual(
-                fan_mon.get_rpm(), {'hwmon6/fan2_input': 412})
+if __name__ == '__main__':
```
kissiel, line 99: This would suggest that this program can act as a standalone thing, yet the import would fail. Is there a real benefit of having this after the test has been developed? (I understand it was easier to iterate by having this here, although one could argue that the same could have been achieved with `python3 -m unittest {module_name}`.

kissiel (commented):
3 things:
1. There's no description for the PR.
2. I don't think that marking one UT module as entry point is useful.
3. I still got:
```
======================================================================
ERROR: test_multiple (test_fan_reaction.FanMonitorTests.test_multiple)
  File "/home/kissiel/checkbox/providers/base/tests/test_fan_reaction.py", line 48, in test_multiple
    fan_monitor = FanMonitor()
                  ^^^^^^^^^^^^
  File "/home/kissiel/checkbox/providers/base/bin/fan_reaction_test.py", line 61, in __init__
    raise SystemExit(0)
SystemExit: 0

======================================================================
ERROR: test_simple (test_fan_reaction.FanMonitorTests.test_simple)
  File "/home/kissiel/checkbox/providers/base/tests/test_fan_reaction.py", line 34, in test_simple
    fan_monitor = FanMonitor()
                  ^^^^^^^^^^^^
  File "/home/kissiel/checkbox/providers/base/bin/fan_reaction_test.py", line 61, in __init__
    raise SystemExit(0)
SystemExit: 0
```

diohe0311:
Hi @kissiel 
What do you mean by marking one UT module as entry point?
Hmmm this is strange, those 2 test case shouldn't enter line 61.
Are you using `./manage.py test -u`? Is this result came from local or in the lab?

I tested it again with 202004-27817, can't reproduce the error you got.

providers/base/tests/test_fan_reaction.py
diohe0311: Thanks! I'll keep in mind

diohe0311:
I think I fixed it now, the error is caused by fan_reaction_test.py line 43: device_path = os.path.realpath(device)
I didn't mock a path for it, and it happens that the first path it finds in your environment contains "pci".

It's been mocked as "foo", should be fine now, please give another try.

kissiel:
Yep, works cleanly!

kissiel (approved):
Awesome!

PR #551, opened 2023-06-13
Add: kernel snap refresh and revert tests

patliuu:
## Description
This pull request is for CHECKBOX-455 to create automated test cases for kernel refresh and revert.
This PR contains 5 test jobs,
1. `kernel_revision_info` is a resource job to gather the kernel name, tracking, and revisions info including original-installed, stable, candidate, beta, and edge revisions for the kernel snap
2. `ubuntucore/kernel-refresh-{kernel}` gets the track name from `kernel_revision_info` and refreshes the kernel to the stable channel. It will only execute when the revision in stable channel exists and the original-installed revision is different from the revision in stable channel.
3. `ubuntucore/kernel-verify-after-refresh-{kernel}` verify the kernel revision is the same as the one in stable channel.
4. `ubuntucore/kernel-revert-{kernel}` waits until the kernel refreshing is `Done` and then reverts the kernel back to the original revision.
5. `ubuntucore/kernel-verify-after-revert-{kernel}` check the revision is back to the original one

The output of `kernel_revision_info` will be like
```
kernel_name: xxxxx-kernel
tracking: latest/stable
tracking_prefix: latest/
stable_rev: 126
candidate_rev: 126
beta_rev: 126
edge_rev: 126
original_installed_rev: 124
```

Note that Tillamook project has different channel names for kernel snap like `stable`, `latest/stable`, and `20/stable`. So the job will capture the current track and then refresh to the corresponding channel. For example, if it's on `20/beta`, then it will refresh to `20/stable` channel.
Also, I noticed sometimes, or on some slow devices, after refreshing the kernel and rebooting, it requires a few seconds to a few minutes to wait for the refreshing process done. Before it is done, we cannot do the kernel revert, so I've added a polling wait for up to 5 minutes on the kernel revert job.

## Resolved issues
CHECKBOX-455
Created kernel refresh and revert test cases for SUV process.

## Tests
* Sideload result on Tillamook device:
https://pastebin.canonical.com/p/sHqfQ7mjbD/
* Sideload result on Nuremberg device:
https://pastebin.canonical.com/p/D57cv8rzsx/
* When the originally installed revision is the same as the revision in the stable channel or there's no revision in the stable channel, the result will be like this:
https://pastebin.canonical.com/p/QzkJpxQFsS/

### DIFF
--- a/checkbox-support/checkbox_support/snap_utils/snapd.py
+++ b/checkbox-support/checkbox_support/snap_utils/snapd.py
@@ -138,14 +138,15 @@ class Snapd():
     def info(self, snap):
         return self.find(snap, exact=True)[0]
 
-    def refresh(self, snap, channel='stable', revision=None):
+    def refresh(self, snap, channel='stable', revision=None, reboot=False):
         path = self._snaps + '/' + snap
         data = {'action': 'refresh', 'channel': channel}
         if revision is not None:
             data['revision'] = revision
         r = self._post(path, json.dumps(data))
-        if r['type'] == 'async' and r['status'] == 'Accepted':
+        if r['type'] == 'async' and r['status'] == 'Accepted' and not reboot:
             self._poll_change(r['change'])
+        return r
 
     def change(self, change_id):
         path = self._changes + '/' + change_id
@@ -157,14 +158,15 @@ class Snapd():
         r = self._get(path)
         return r['result']['tasks']
 
-    def revert(self, snap, channel='stable', revision=None):
+    def revert(self, snap, channel='stable', revision=None, reboot=False):
         path = self._snaps + '/' + snap
         data = {'action': 'revert', 'channel': channel}
         if revision is not None:
             data['revision'] = revision
         r = self._post(path, json.dumps(data))
-        if r['type'] == 'async' and r['status'] == 'Accepted':
+        if r['type'] == 'async' and r['status'] == 'Accepted' and not reboot:
             self._poll_change(r['change'])
+        return r
 
     def get_configuration(self, snap, key):
         path = self._snaps + '/' + snap + '/conf'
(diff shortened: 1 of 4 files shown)


patliuu:
Hi, based on the discussion in the mattermost channel, I've modified these tests in restapi way with python script.

To handle the snap change status of refreshing/reverting kernel snap will not be `Done` unless the reboot is complete, I added a `reboot` parameter for `snapd.refresh()` and `snapd.revert()`. When trying to refresh/revert a snap that will automatically trigger reboot (e.g. core snap and kernel snap), setting `reboot=True` could skip the polling action.

In addition, to check the snap change status after rebooting, the change ID is needed. So I added a return response for these 2 functions.

Here's the new sideload test result:
* On Tillamook device with uc20 image
https://pastebin.canonical.com/p/3wkPmNgnQ7/
* On Nuremberg device with uc20 image
https://pastebin.canonical.com/p/kcz2f2bRtq/
* Testing `snappy-snap-automated` to check snapd.py change doesn't affect current tests
https://pastebin.canonical.com/p/yqXYgg6rgR/

yphus (changes requested):
Amazing work @patliuu ! the ability to skip the test if beta and stable are the same versions is great as it means the tests can safely be added to `ubuntucore-automated`. Support for tracks like 20/stable will also really help for devices like tillamook.

A small improvement/suggestion would be to add the revision number in the logging calls, for example:
- Refreshing kernel snap to stable (from rev YYY to rev XXX)...
- Reverting kernel snap (from rev XXX to rev YYY)...

patliuu:
Hi @yphus ,
Thanks for the suggestions, I've added a commit to log their revision info before refreshing and reverting the kernel.

Sideload results with these changes:
* On Tillamook uc20 device
https://pastebin.canonical.com/p/pQFJMrFYtD/
* On Nuremberg uc20 device
https://pastebin.canonical.com/p/SSztJhXTzd/

yphus (approved):
Thank you for testing the new jobs on Nuremberg and Tillamook devices , LGTM

PR #557, opened 2023-06-14
Rollback to provisioned before installing in reused containers

Hook25:
## Description

This is done because if a test run is interrupted midway through, reusing the container will actually break it (because the old run is still in the container files, leading checkbox to prompt for a restart/drop)

## Resolved issues

N/A

## Documentation

N/A

## Tests

This was tested by running metabox with a container to update, one to create and one that should be updated but it is in an invalid state (missing provisioned)


kissiel (approved):
LGTM!
I like how this also fixes a problem of half-baked containers.

PR #558, opened 2023-06-14
Fix metabox crash on already started containers

Hook25:
## Description

This fixes the crash when metabox is started with a `metabox*` containers already running. Metabox crashes because it is unable to categorize the container but if it has to be used, it tries to recreate it, crashing wih `Failed creating instance record: Add instance info to the database: This "instances" entry already exists`

This also fixes the issue if the already started container is half baked by catching NotFound (config file) and deleting the old container, allowing the recreation to take place

## Resolved issues

N/A

## Documentation

N/A

## Tests

This was tested by running metabox on a running container, an half baked one and a stopped one.


kissiel (approved):
Awesome! +1

PR #559, opened 2023-06-14
Add RPL CPUIDs to the cpuid.py script

mckees:
## Description

<!--
Describe your changes here:

This change includes all RPL CPUIDs from edc.intel.com

## Resolved issues
N/A

## Documentation

This is a simple add of CPUIDs that are documented on Intel's site, and it only impacts functionality if the CPUID matches the machine that it runs on, so adding tests would require hardware that matches those IDs.

## Tests

I ran the script on a RPL device matching the ID "0xB06A2"


bladernr (approved):
Looks great, thanks for updating the script!

PR #560, opened 2023-06-14
Add ability to get a numerical Intel gen number to cpuid.py

mckees:
## Description
DEPENDS ON: pull/559

Some feautures we test for may be heavily dependent on the CPU
    generation. For instance, a media stress test for Chromium seems to
    max out VECS on Gen12+ while it maxes out VCS on older generations.

-->

    Adding a command line argument to print a simple number allows tests to
be more flexible around different device capabilities at runtime rather than parsing this script's output and deciding which codenames have certain behaviors associate with newer features. 

## Resolved issues
N/A

## Documentation

There is no documentation on this script, as far as I can tell, so there is nothing to update. So I added some command line documentation while I was at it.

## Tests

I ran the script standalone to check the version that comes out (with and without the flag I added to make sure I didn't break the original purpose of the script)

### DIFF
--- a/providers/base/bin/cpuid.py
+++ b/providers/base/bin/cpuid.py
@@ -25,6 +25,7 @@
 # Modifications: 2019 Jeffrey Lane (jeffrey.lane@canonical.com)
 
 import argparse
+from enum import Enum
 import ctypes
 import platform
 import sys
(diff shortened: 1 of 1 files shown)


providers/base/bin/cpuid.py
```diff
@@ -139,6 +140,26 @@
         "Whisky Lake":      ['0x806eb', '0x806ec'],
         }
 
+# Server chips use different generations, so skip them
```
bladernr, line 143: What do Server chips use then? Especially since the IDs for Westmere, Sandy Bridge, Ivy Bridge, Haswell, Broadwell, and Skylake all express server CPUs (Xeons of various flavors).

Also, one project I had for this cycle is adding in information like kernel version where each is supported initially.  The reason I mention that is the idea I had was to change the CPUID dict to use a dict of dicts like this:

```
"Sapphire Rapids":  {"kernel": ['5.15']; "cpuid": ['0x806f3', '0x806f6', '0x806f7', '0x806f8']},
```
So we could do things like run this to determine if the kernel being used is the minimal required kernel for the CPU identified (and thus fail if someone runs a test on Sapphire Rapids using 5.4, for example).

Would it be better to use your changes here and create the dict of dicts mentioned above so that it's all together rather than in separate lookup tables?

providers/base/bin/cpuid.py
```diff
@@ -139,6 +140,26 @@
...
+        "Rocket Lake":      11,
+        "Tiger Lake":       11,
+        "Comet Lake":       10,
+        "Amber Lake":       10,
+        "Ice Lake":         10,
+        "Whiskey Lake":     10,
+        "Canon Lake":       8,
+        "Coffee Lake":      8,
+        "Kaby Lake":        7,
+        "Skylake":          6,
```
bladernr, line 156: If this is truly NOT meant for Server CPUs, you'll need to also break out the ID look up to identify "Skylake" vs "Skylake Xeon", look up the IDs and break them out accordingly.  

And then what about things like Coffee Lake, Kaby Lake, Rocket Lake, etc Xeon-E,D,W chips?

bladernr (commented):
I'm not against adding additional features to this... my questions are more about how the lookup table is implemented (with the context that you didn't have that I had plans to modify the CPUID lookup table to add additional data for each CPU family already).

providers/base/bin/cpuid.py
mckees: The server chips started a new generation branding a few years back: 
Skylake (1st Generation Intel® Xeon® Scalable Processors)
Cascade Lake (2nd Generation Intel® Xeon® Scalable Processors)
Ice Lake (3rd Generation Intel® Xeon® Scalable Processors)
Sapphire Rapids (4th Generation Intel® Xeon® Scalable Processors)

Someone with a bit deeper history on generation numbering of Xeon products might be able to add to this, but I'm not familiar with a version numbering system before that. I specifically didn't add Xeon CPUs because I'm not sure how relevant this is for servers. For desktop/mobile chips, the generation is definitely referenced a lot to determine whether or not certain features get enabled.

For informational purposes, this script was probably doing enough to just give a less specific generation (i.e. Ice Lake instead of Ice Lake-D or even Ice Lake Xeon), but adding something more functional seems like it might cause some problems with the two types of being intermingled. I didn't realize that these would need to be disentangled, but it seems like a good idea before merging this. It sounds like you'd like to see this for Xeon too, but my thinking is that it should be a different flag so that people get what they expect. For instance, if someone intends to get something Ice Lake Xeon ("3" for servers) or higher, and they get "6" back from the script for Skylake, that test will not be running as intended. I think I'd rather their test fail because they used the desktop CPU gen flag and no desktop CPU was found. 

As for the kernel version addition, that feels like a nice change vs separate lookup tables. I can switch it around a bit

providers/base/bin/cpuid.py
mckees: Ya, this is why I planned to not address server chips. I don't know the requirements of people testing Xeon products, and I figure they'd be able to piggyback on my changes if the need exists.

mckees:
@bladernr It's all organized into a new dict, but I don't know where to find most of these CPU IDs without decoding them by hand, which I don't have time to do at the moment. Without that, I'm not sure which are server chips and which are "desktop" chips (open to a change of terminology here, like "client").

providers/base/bin/cpuid.py
bladernr: Yeah, I guess the question is where does this get run?  So the way this is, if you run this against a skylake, for example, on a skylake server (Xeon Scalable 1st Gen), that will then incorrectly report your generation of 6 since the skylake entry has codes for both:
"Skylake":          ['0x406e3', '0x506e3', '0x50654', '0x50652'],
** IIRC the first two are Skylake Y and Skylake DT/H/S (https://en.wikichip.org/wiki/intel/microarchitectures/skylake_(client)#CPUID)

also, Some of those ONLY list server CPUs since I wrote it for server, expecting that if anyone needed any additional client CPUs, they'd add those themselves... and I didn't differentiate Client vs Server per platform, as I said...

I guess I'm just trying to imagine the cleanest way to do this, because for now, matching any server cpu platform on your lookup table will return the client cpu generation incorrectly.

So Maybe it's a good idea to also break the CPUID table down as part of this?

bladernr:
> @bladernr It's all organized into a new dict, but I don't know where to find most of these CPU IDs without decoding them by hand, which I don't have time to do at the moment. Without that, I'm not sure which are server chips and which are "desktop" chips (open to a change of terminology here, like "client").

That's what I did ;) the Server ones I got because I ran this on a ton of servers and recorded the variuos ID strings the CPU returned to build up the table.  But I also filled in gaps and added some of the desktop stuff based on info at https://en.wikichip.org/wiki/WikiChip...

it should be at least easy enough to compare what is in the CPUID table with the listed IDs on wikichip for most of the client platofrms... still a bit time consuming, but nto as bad as decoding by hand.  (see my skylake client link in my other comment)

providers/base/bin/cpuid.py
bladernr: The server chips started a new generation branding a few years back:

Even older than that:
Ivy Bridge = Intel Xeon v2
Haswell = Intel Xeon v3
Broadwell = Intel Xeon v4

Then you get to skylake which is confusing because
Skylake Xeon E3 = v6
Skylake Scalable = Xeon Scalable 1st Gen

Heh... that was a transition one though, thankfully, after that they stopped versioning the smaller Xeons.

providers/base/bin/cpuid.py
```diff
@@ -189,23 +372,41 @@ def __del__(self):
...
 
     my_id = (hex(cpu[0]))
     complete = False
-    for key in CPUIDS.keys():
-        for value in CPUIDS[key]:
-            if value in my_id:
-                print("CPUID: %s which appears to be a %s processor" %
-                      (my_id, key))
+    for key in CPUDICT.keys():
+        platform = CPUDICT[key]
```
kissiel, line 395: This overwrites the module `platform` imported at the top.

providers/base/bin/cpuid.py
```diff
@@ -189,23 +372,41 @@ def __del__(self):
...
+                    if platform['chiptype'] == ChipType.CLIENT and \
+                       platform['generation'] > 0:
+                        print(platform['generation'])
+                    else:
+                        print("%s is not supported for conversion to Intel \
+                              generation number")
+                else:
+                    print("CPUID: %s which appears to be a %s processor" %
+                          (my_id, key))
+
```
kissiel, line 409: With this complexity this whole function is not just a simple entry point to the program. There's a lot of complex logic that is implemented here and that logic needs to be outlined (possibly to more than one function). The aforementioned decomposition will also facilitate unit testing the new branches of logic that are proposed here which are missing and necessary.

kissiel (changes requested):
See inline for things that need fixing.

kissiel:
Ah, we've also compiled a list of guidelines to help avoid some common problems.
[You can find it here](https://docs.google.com/document/d/1aMw23nplzvSFYI-uJh8_VB6zox2Bbewz3HCTrV820eA/edit)

pieqq:
Any update on this, @mckees ?

mckees:
@pieqq I have not had the time to work on this, and I don't see myself working on it any time soon. I'm just going to close it. If someone wants it, feel free to rework it (with @kissiel's suggestions) and submit your own PR

PR #561, opened 2023-06-15
CHECKBOX-345: No longer ignore virtual watchdog

diohe0311:
## Description
The problem is that watchdog/detect failed to find any watchdog, any yet passed the subsequent trigger-system-reset test.
This is because udevadm ignores the virtual watchdog by default (the reason behind it is that there used to be a machine with a large number of virtual watchdogs: https://code.launchpad.net/~jocave/checkbox-support/+git/checkbox-support/+merge/395442)

After discussions with Pierre, Jonathan and Slyvain, since the situation is different, the virtual watchdog can no longer be ignored (checked several latest IoT and PC projects, some machines do not have physical watchdog, which using virtual watchdog to run the trigger-system-reset test). And it is necessary to use the udevadm log of 202109-29473 as a unit test to verify whether the virtual watchdog will be included in the list after modification.

This change will affect the expected results of the udevadm_data unit tests (e.g. Lenovo E431, need to increase the number of expected devices from 68 to 100, because the virtual watchdog is now also included in the calculation)

## Resolved issues
https://warthogs.atlassian.net/browse/CHECKBOX-345

## Tests
Run `./setup.py test `under checkbox/checkbox-support/


pieqq (approved):
Very good job!

I was curious as to why we needed

```python
        if device.category == "WATCHDOG":
            return False
```

but it's because below, there is a generic condition that ignores all the devices without any `vendor` or `product` info, which is the case for the watchdog devices...

Thanks for adding the unit tests and modifying the existing ones!

PR #566, opened 2023-06-15
Remote api bump

kissiel:
## Description

The config refactor branch changed some bits how Controller talks to the Testbed. This means it should have come with a Remote API bump.
Here it is.

## Resolved issues

Resolves crashes if someone would try to use mismatching versions of checkbox.

## Tests

Metabox run is happy.


Hook25 (approved):
+1, needed, tyvm

PR #567, opened 2023-06-15
Add launcher option and refactor args

Hook25:
## Description

In order to deterministically pass to a checkbox-cli instance a launcher, one has to wonder if the specific sub-command that is about to be launched does actually read it and put it as the last positional argument. This is not great for automatic testing (namely, one has to parse the start command to see if the launcher can be provided or not) and it is not great for programmatic usage as well (for the same reason).  Additionally, some options that should read the launcher do not (most notably: `check-config` and `run`). This aims to fix these two things while also upgrading the cli arg parser to use sub-parsers and avoid hacking sys.argv

After this patch the following will be possible
```
# Added in this patch, support for launcher in check config
checkbox-cli check-config launcher.conf
# Added in this patch, support for launcher in run
checkbox-cli run ... launcher.conf
# Added in this patch, support for launcher positional arg
checkbox-cli --launcher abc.conf check-config
```

Also, although not fully portable
```
#!/usr/bin/env -S checkbox-cli --launcher
```
is now a more powerful shebang for launchers, allowing a launcher to be used with any checkbox command like this:
```
./launcher.conf check-config
./launcher.conf remote 127.0.0.1
```

## Resolved issues

N/A

## Documentation

The new option is documented in the cli interface

## Tests

This was tested via metabox in legacy and remote runs.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
+++ b/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
@@ -22,9 +22,13 @@ Checkbox Launcher Interpreter Application
 
 import gettext
 import logging
+import argparse
 import os
 import subprocess
 import sys
+import itertools
+import contextlib
+import functools
 
 from plainbox.impl.jobcache import ResourceJobCache
 from plainbox.impl.session.assistant import SessionAssistant
@@ -58,44 +62,7 @@ class Context:
         self.sa = sa
 
 
-def main():
-    import argparse
-
-    commands = {
-        "check-config": CheckConfig,
-        "launcher": Launcher,
-        "list": List,
-        "run": Run,
-        "startprovider": StartProvider,
-        "submit": Submit,
-        "show": Show,
-        "list-bootstrapped": ListBootstrapped,
-        "merge-reports": MergeReports,
-        "merge-submissions": MergeSubmissions,
-        "tp-export": TestPlanExport,
-        "service": RemoteSlave,
-        "remote": RemoteMaster,
-    }
-    deprecated_commands = {
-        "slave": "service",
-        "master": "remote",
-    }
-
-    known_cmds = list(commands.keys())
-    known_cmds += list(deprecated_commands.keys())
-    known_cmds += ["-h", "--help"]
-    if not (set(known_cmds) & set(sys.argv[1:])):
-        sys.argv.insert(1, "launcher")
-
-    for i, arg in enumerate(sys.argv):
-        if arg in deprecated_commands:
-            sys.argv[i] = deprecated_commands[arg]
-            logging.warning(
-                "%s is deprecated. Please use %s instead",
-                arg,
-                deprecated_commands[arg],
-            )
-
+def parse_args(default_command, commands, deprecated_commands={}):
     top_parser = argparse.ArgumentParser()
     top_parser.add_argument(
         "-v",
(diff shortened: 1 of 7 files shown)


checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
```diff
@@ -58,45 +64,51 @@ def __init__(self, args, sa):
...
+            return parser.parse_args(args=sys_argv[1:])
+    # -> typo is some arg
+    parser.print_usage()
+    raise SystemExit(str(error))
 
-    top_parser = argparse.ArgumentParser()
+
+def setup_and_parse_args(default_command, commands, deprecated_commands={}):
+    top_parser = argparse.ArgumentParser(
+        conflict_handler="resolve", exit_on_error=False
```
yphus, line 110: [`exit_on_error`](https://docs.python.org/3/library/argparse.html#argumentparser-objects) is python3.9 only unfortunately

yphus (changes requested):
spotted a 3.9 syntax only, see below

checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
Hook25: Thank you for catching that, fixed!

pieqq (changes requested):
I've tested this in a venv, and it brings many nice updates! The most interesting is that you can now use `check-config` in conjunction to a launcher.

I didn't find any problem when running it, just a little inline comment that you might want to change, and after rebase we should be able to land this!

pieqq (approved):
+1

Thanks a lot!

kissiel (changes requested):
>In order to deterministically pass to a checkbox-cli instance a launcher, one has to wonder if the specific sub-command that is about to be launched does actually read it and put it as the last positional argument. 

There is a subcommand "launcher" that does that and takes the launcher file as the value.

This PR is advertised as introducing clarity, yet with it `launcher` becomes an option **and** a positional arg, to a point where you had to have logic to check for that corner case.

There's a lot of unpredictable behavior with using --launcher for other sub-commands. For instance `checkbox-cli --launcher launcher_with_preselected_TP run some-bootstrapping-job-that-overwrites-resource`

I really would like to see the problem that this PR fixes in simple words.
Please do dump what's being difficult right now to do, or examples of yucky invocations, possibly with the replacement you're proposing.
The description provided with this PR is a bit lacklustre.
Also the proposed invocation `./launcher.conf remote 127.0.0.1` is just crazy :exploding_head: 

On top of that this introduces 100cloc and some crazy shenanigans with subclassing ArgumentParser.

There are also a few problems with the code below, but I don't know if the general idea for this PR will go through. (what if someone called their launcher `master`? they'll get a misleading deprecation warning for instance).

I can't approve this PR, not in this state.

Hook25:
Do you want me to create a subset pr for each of the changes? At least the check-config arg I want to merge asap, it is so useful.

kissiel:
> Do you want me to create a subset pr for each of the changes? At least the check-config arg I want to merge asap, it is so useful.

Let's first decide which pieces of this PR are indeed needed.

pieqq:
Having the `check-config` option be able to parse a launcher would be super helpful indeed. Launchers define some (if not all) of the configuration used by Checkbox for most of our projects now, so having a way to take it into account and make some basic checks on it would be great.

Issue #520, opened 2023-05-31
"Using sideloaded provider" warning is missing when using Checkbox remote


p-gentili:
### Bug Description

When using a sideloaded provider Checkbox displays a warning message like:
`Using sideloaded provider: checkbox-provider-base, version 2.4 from /var/tmp/checkbox-providers/base`

The same warning is not displayed when using Checkbox remote.

### To Reproduce

1. Sideload a Checkbox provider on DUT
2. Launch a Checkbox testplan using Checkbox remote
3. No warning message is displayed on the controller side

### Environment

DUT:
- 20.04
- checkbox 2.6 from archives

Controller:
- 23.04
- checkbox 2.6 snap

### Relevant log output

```shell
When launched locally on DUT:

ubuntu@ubuntu-HP-ProBook-440-14-inch-G9-Notebook-PC:~$ checkbox-cli ./checkbox-launcher-stock 
Using sideloaded provider: checkbox-provider-base, version 2.4 from /var/tmp/checkbox-providers/base
```

When launched remotely:
```
~ checkbox.checkbox-cli remote 10.102.158.105 checkbox-launcher-stock
$PROVIDERPATH is defined, so following provider sources are ignored ['/home/pgentili/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop'] 
Connecting to 10.102.158.105:18871. Timeout: 600s
```
```


### Additional context

_No response_

PR #569, opened 2023-06-16
Warn if any provider is sideloaded

Hook25:
## Description

When using a sideloaded provider Checkbox displays a warning message like:
Using sideloaded provider: checkbox-provider-base, version 2.4 from /var/tmp/checkbox-providers/base

The same warning is not displayed when using Checkbox remote.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/520

## Documentation

N/A

## Tests

This was tested locally and checked that nothing breaks via metabox

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/master.py
+++ b/checkbox-ng/checkbox_ng/launcher/master.py
@@ -312,6 +312,8 @@ class RemoteMaster(ReportsStage, MainLoopStage):
 
         try:
             tps = self.sa.start_session(configuration)
+            if self.sa.sideloaded_providers:
+                _logger.warning("Using sideloaded providers")
         except RuntimeError as exc:
             raise SystemExit(exc.args[0]) from exc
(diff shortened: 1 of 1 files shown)


checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -312,6 +312,8 @@ def new_session(self):
 
         try:
             tps = self.sa.start_session(configuration)
+            if self.sa.sideloaded_providers:
+                _logger.warning("Using sideloaded providers")
```
kissiel, line 316: To ensure that someone doesn't assume it's the controller that has side-loaded stuff, I'd add:
"Agent is using..."

kissiel (commented):
I guess we can dodge future ambiguity bullet by adding "Agent " in front of the warning.

kissiel (approved):
Awesome, thanks!

PR #577, opened 2023-06-22
Fix errors for some type hints on python<3.9

plars:
For python versions prior to 3.9, type hints that look like "list[str]" will get errors like this:
...
  File "/usr/lib/python3/dist-packages/checkbox_ng/config.py", line 50, in <module>
    def _search_configs_by_name(name: str) -> list[str]:                                                                                                        
TypeError: 'type' object is not subscriptable


Hook25 (approved):
+1, well done

PR #581, opened 2023-06-23
Disable buffering in deb jobs

Hook25:
## Description

This disables output buffering in deb daily and beta jobs to (hopefully) allow `ppa-dev-tools` to output continuously even in github jobs.

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-702

## Documentation

N/A

## Tests

N/A


kissiel (approved):
Makes sense.
Any performance repercussions we should be aware?

Hook25:
There should not be any performance repercussion because the script that prints in the chain (that is ppa-dev-tool) is configured to print every ~1minute, so it should be ok https://github.com/canonical/checkbox/blob/092effef6361b41097f65cb592cf91212c41c5c2/tools/daily-builds/deb_daily_builds.py#L28-L34

Issue #491, opened 2023-05-24
graphics_card_resource.py couldn't define index for iGPU and dGPU correctly


hanhsuan:
### Bug Description

graphics_card_resource.py will use [PCI ID](https://github.com/canonical/checkbox/blob/main/providers/resource/bin/graphics_card_resource.py#L95) to sort the GPU order. However we get one new [DUT](https://certification.canonical.com/hardware/202304-31460/) with `AMD(APU)` + `NV(dGPU)`, the index will be set to `2` for `AMD(APU)` and `1` for `NV(dGPU)`. 
* Advanced Micro Devices, Inc. [AMD/ATI] - 1002:15bf
* nVidia - 10de:28a1

Finally create the wrong test case
* [graphics/1_glxgears_PCI_ID_0x28a1](https://certification.canonical.com/hardware/202304-31460/submission/313703/test/198597/result/33368789/)
* [graphics/2_glxgears_PCI_ID_0x15bf](https://certification.canonical.com/hardware/202304-31460/submission/313703/test/198698/result/33368797/)

and doing the wrong prime-offload to AMD(APU).
```
Setting up PRIME GPU offloading for AMD discrete GPU
Running synchronized to the vertical refresh.  The framerate should be
approximately the same as the monitor refresh rate.
2038 frames in 5.0 seconds = 406.658 FPS
1387 frames in 5.0 seconds = 276.908 FPS
```

In the future, there might be more combination of GPU, such as:
* Intel(iGPU) +  Intel(dGPU)
* AMD(APU) +  NV(dGPU)
* AMD(APU) + Intel(dGPU)

Therefore, we need a way to define the index correctly.

### To Reproduce

Run graphic test

### Environment

- OS: Ubuntu 22.04.2 LTS
- Checkbox type: Debian
- Checkbox version: 2.5
- CPU: AMD APU
- GPU: Nvidia

pieqq:
Before addressing this bug, I think we should have a discussion to decide how we should test grpahics card with more recent versions of Ubuntu. The current system used by Checkbox was developed at different time, back when

- nvidia GPU did not have on-demande mode
- there was no AMD APUs being certified
- there was no Intel dGPUs on the market
- etc.


I'll prepare a Jira card to track this.

pieqq:
Fixed with #586 and #942

PR #586, opened 2023-06-28
Change to new graphic test strategy (BugFix)

hanhsuan:
## Description

Change the GPU test to prime/revert prime strategy to make test jobs to meet real usage, and remove redundant jobs.

## Resolved issues

Tracking in this [Jira card](https://warthogs.atlassian.net/browse/CQT-2567)
Github issue #491 

## Submissions

I + I
* [remote auto](https://certification.canonical.com/hardware/202306-31654/submission/321049/)
* [auto](https://certification.canonical.com/hardware/202306-31654/submission/321050/)
* [graphic](https://certification.canonical.com/hardware/202306-31654/submission/321051/)

I + A
* [auto](https://certification.canonical.com/hardware/202203-30094/submission/321056/)
* [graphic](https://certification.canonical.com/hardware/202203-30094/submission/321057/)

A + N (This DUT will make nvidia driver error while suspend/resume)
* [auto](https://certification.canonical.com/hardware/202304-31460/submission/321054/)
* [graphic](https://certification.canonical.com/hardware/202304-31460/submission/321055/)

A only
* [auto](https://certification.canonical.com/hardware/202304-31463/submission/321061/)
* [graphic](https://certification.canonical.com/hardware/202304-31463/submission/321062/)


providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,317 @@
...
+            "-p", "--pci", type=str, default='0000:00:02.0',
+            help='pci name in NNNN:NN:NN.N format (default: %(default)s)'
+        )
+        parser.add_argument(
+            "-d", "--driver", type=str, default='i915',
+            help='Type of GPU driver (default: %(default)s)'
+        )
+        parser.add_argument(
+            "-t", "--timeout", type=int, default=0,
+            help='executing command duration in second (default: %(default)s)'
```
pieqq, line 292: From the code, it looks like 0 means no timeout. Maybe it should be explained in the help text.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,317 @@
...
+            infos = json.loads(card_infos.stdout)
+            for info in infos:
+                if pci_name in info['businfo']:
+                    return info['product']
+        except (ValueError, IndexError):
+            return None
+
+    def check_offload(self, cmd, card_id, card_name):
+        """
+        checking command is running
```
pieqq, line 124: This is a bit too vague of a description. This function checks that a given command is being executed by a specific GPU using the kernel debug interface, right? Could you add some more info for our future selves? :)

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+template-resource: graphics_card
+plugin: shell
+category_id: com.canonical.plainbox::graphics
+id: graphics/{index}_glxgears_auto_{product_slug}
+flags: also-after-suspend
+requires: executable.name == 'glxgears'
+command:
+ # shellcheck disable=SC1091
+ prime_offload_tester.py -c glxgears -p {pci_device_name} -d {driver} -t 30
+_summary: Test that glxgears works for {vendor} {product}
```
pieqq, line 302: Since this job is gonna be running for 30 seconds, I would add an `estimated_duration: 30s` to these jobs.

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+unit: template
+template-resource: graphics_card
+plugin: user-interact-verify
+category_id: com.canonical.plainbox::graphics
+id: graphics/{index}_glxgears_valid_{product_slug}
+flags: also-after-suspend
+requires: executable.name == 'glxgears'
+command:
+ # shellcheck disable=SC1091
+ prime_offload_tester.py -c glxgears -p {pci_device_name} -d {driver} -t 30
```
pieqq, line 313: Same comment as above: I would add an `estimated_duration: 30s`.

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+_summary: Test that glxgears works for {vendor} {product}
+_description:
+ PURPOSE:
+     This test tests the basic 3D capabilities of your {vendor} {product} video card
+ STEPS:
+     1. Click "Test" to execute an OpenGL demo. Press ESC at any time to close.
+     2. Verify that the animation is not jerky or slow.
+ VERIFICATION:
+     1. Did the 3d animation appear?
+     2. Was the animation free from slowness/jerkiness?
```
pieqq, line 323: `description` is deprecated. It's better to use `purpose`, `steps` and `verification` fields directly. See https://checkbox.readthedocs.io/en/latest/reference/units/job.html#job-fields for more info.

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+unit: template
+template-resource: graphics_card
+plugin: shell
+category_id: com.canonical.plainbox::graphics
+id: graphics/{index}_glxgears_auto_fullscreen_{product_slug}
+flags: also-after-suspend
+requires: executable.name == 'glxgears'
+command:
+ # shellcheck disable=SC1091
+ prime_offload_tester.py -c "glxgears -fullscreen" -p {pci_device_name} -d {driver} -t 30
```
pieqq, line 334: Add an `estimated_duration: 30s` for these jobs.

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+template-resource: graphics_card
+plugin: user-interact-verify
+category_id: com.canonical.plainbox::graphics
+id: graphics/{index}_glxgears_valid_fullscreen_{product_slug}
+flags: also-after-suspend
+requires: executable.name == 'glxgears'
+command:
+ # shellcheck disable=SC1091
+ prime_offload_tester.py -c "glxgears -fullscreen" -p {pci_device_name} -d {driver} -t 30
+_summary: Test that glxgears works on fullscreen for {vendor} {product}
```
pieqq, line 347: This is the same summary as for the `glxgears_auto_fullscreen` job. Is it expected?

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+_description:
+ PURPOSE:
+     This test tests the basic fullscreen 3D capabilities of your {vendor} {product} video card
+ STEPS:
+     1. Click "Test" to execute an OpenGL demo. Press ESC at any time to close.
+     2. Verify that the animation is not jerky or slow.
+ VERIFICATION:
+     1. Did the 3d animation appear?
+     2. Was the animation free from slowness/jerkiness?
```
pieqq, line 356: Same comment as above: use `steps`, `purpose` and `verification` instead of `description`.

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+unit: template
+template-resource: graphics_card
+plugin: user-interact-verify
+category_id: com.canonical.plainbox::graphics
+id: graphics/{index}_glxgears_valid_fullscreen_{product_slug}
+flags: also-after-suspend
+requires: executable.name == 'glxgears'
+command:
+ # shellcheck disable=SC1091
+ prime_offload_tester.py -c "glxgears -fullscreen" -p {pci_device_name} -d {driver} -t 30
```
pieqq, line 346: Add an estimated duration, like for the other jobs above.

providers/base/units/graphics/jobs.pxu
```diff
@@ -289,6 +289,72 @@ command:
...
+_summary: Test that glxgears works on fullscreen for {vendor} {product}
+_description:
+ PURPOSE:
+     This test tests the basic fullscreen 3D capabilities of your {vendor} {product} video card
+ STEPS:
+     1. Click "Test" to execute an OpenGL demo. Press ESC at any time to close.
```
pieqq, line 352: If the user presses `Esc`, what will be the impact on your script? Will the thread still be running for 30 seconds?

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,317 @@
...
+import argparse
+import logging
+from enum import IntEnum
+
+
+class PrimeOffloaderError(IntEnum):
+    """
+    A class used to define PrimeOffloader Error code
+
+    Attributes
```
pieqq, line 19: I like to use the Sphinx way of writing docstrings, such as the one used in checkbox internals, for instance:

https://github.com/canonical/checkbox/blob/main/checkbox-ng/plainbox/impl/session/state.py#L229

You can find more info here: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,317 @@
...
+        find card name by pci name
+
+    check_offload(cmd, card_id, card_name)
+        check offload status
+
+    check_nv_offload_env()
+        check the environment is ok for prime offload
+
+    run_offload_cmd(cmd, pci_name, driver)
+        exectue the command that you would like to offload to specific GPU
```
pieqq, line 70: I'm not sure if this part of the docstring is required. As long as you provide a good description docstring for each of the functions below, you don't need to repeat yourself in the class docstring.

pieqq (changes requested):
Thank you for this big update! You've led a great work, first with some researches and documentation on the prime and reverse prime process, then with this implementation!

I made some suggestions inline, please check them and let me know if you have any question.

I've seen that you've created new test plans (for instance `graphics-gpu-cert-automated`), but it's not immediately clear why this is so. My understanding is that they are only to be used with 22.04+ test plans, in order to keep backward compatibility with previous OSes and jobs? It would be good to have some kind of explanation somewhere (I like to put this kind of info in the git commit message, because there is a higher change to find this info back when doing code archeology later on).

hanhsuan:
@pieqq  I've fixed the code and description as your suggestion. If there is something I missed, please let me know.

diohe0311:
Hi @hanhsuan, thanks for well explanation!
I've leave some comments based on our discussion.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,311 @@
...
+        :type cmd: str
+
+        :param card_id: card id of dri device
+        :type card_id: str
+
+        :param card_name: card name of dri device
+        :type card_name: str
+        """
+        cmd_without_args = cmd.split(' ')[0]
+        for index in range(11):
```
diohe0311, line 126: If index represents the number of times, starts from 1 will make more sense.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,311 @@
+#!/usr/bin/env python3
+"""prime offload test module."""
```
diohe0311, line 2: Please add author info like this:
'''
Copyright (C) 2023 Canonical Ltd.

Authors
  Hanhsuan Lee <hanhsuan.lee@canonical.com>
'''

diohe0311 (commented):
Hi @hanhsuan, thanks for well explanation!
I'll run test on DUTs based on our discussion.

diohe0311:
Hey @hanhsuan, as discussed, we ran `ODM Client Certification for Desktop 22.04 - (1/2) Manual tests` and found automated test jobs under it, could you please check on that?

hanhsuan:
> Hey @hanhsuan, as discussed, we ran `ODM Client Certification for Desktop 22.04 - (1/2) Manual tests` and found automated test jobs under it, could you please check on that?

@diohe0311 Thanks for you help to noticed me some issues in my PR, I have modified to make `ODM Client Certification for Desktop 22.04 - (1/2) Manual tests` include correct test-plan.

kissiel:
/canonical/self-hosted-runners/run-workflows 28e1ba434be1fe35cced8b485971804b3716d907

pieqq:
/canonical/self-hosted-runners/run-workflows 79599f81d991aab1397bc16de343899a07c49fb7

diohe0311:
@hanhsuan can you sign your commits and resubmit? Please see our contrib doc about this: https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#signed-commits-required

diohe0311:
/canonical/self-hosted-runners/run-workflows [93c5634](https://github.com/canonical/checkbox/pull/586/commits/93c56348983bbdf1859004f583fd2fa7ca987347)

diohe0311 (approved):
LGTM+1

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+        cmd = 'sudo lshw -c display -json'
+        try:
+            card_infos = subprocess.run(cmd, shell=True,
+                                        stdout=subprocess.PIPE,
+                                        universal_newlines=True)
+            infos = json.loads(card_infos.stdout)
+            for info in infos:
+                if pci_name in info['businfo']:
+                    return info['product']
+        except (ValueError, IndexError, TypeError):
```
kissiel, line 118: 1. I guess the ValueError here is only for catching JSONDecodeErrors, so I recommend using the more precise type.
2. Question: how can we end up with TypeError here?
3. There are bunch of exceptions that can be raised from the subprocess, since the handling of all problems happens here, maybe it's worth considering catching them too.
4. If a problem happened maybe it's worth printing it out?

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+            return None
```
kissiel, line 119: This makes the return type become a union (str|None).
Is not obtaining the name a normal case or an exceptional case? If it's an exception, maybe it's worth making it as such.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+    check_result = PrimeOffloaderError.OFFLOAD_FAIL
+
+    def find_card_id(self, pci_name):
+        """
+        use pci name to find card id under /sys/kernel/debug/dri
+
+        :param pci_name: pci device name in NNNN:NN:NN.N format
+        :type card_name: str
+        """
+        cmd = "sudo grep -lr --include=name \"{}\""  \
```
kissiel, line 88: 1. Please don't escape line endings. Use inline string concatenations instead.
2. The preferred way of calling subprocesses is to use a list of arguments, and considering this cmd is getting interpolated with other strings, it will be unsafe to run with `shell=True` as there is no lexical check.
3. If the program is meant to be run as root, don't invoke sudo, as it leads to unpredictable results.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+              " /sys/kernel/debug/dri 2>/dev/null".format(pci_name)
```
kissiel, line 89: Please don't use shell's facilities to pipe the stderr to dev null. Use `stderr=subprocess.DEVNULL`. I'd also consider if we really want to ignore all the messages printed there.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+
+        :param card_id: card id of dri device
+        :type card_id: str
+
+        :param card_name: card name of dri device
+        :type card_name: str
+        """
+        cmd_without_args = cmd.split(' ')[0]
+        for index in range(1, 11):
+            time.sleep(2)
```
kissiel, line 144: What is this sleep for?
Why two seconds?
If we need to wait for some condition to occur, let's check for that in a busy loop with a tiny sleep (like 10ms) instead of arbitrary waiting for some time.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+                    self.logger.info("Checking success:")
+                    self.logger.info("  Offload process:[{}]".format(cmd))
+                    self.logger.info("  Card ID:[{}]".format(card_id))
+                    self.logger.info("  Device Name:[{}]".format(card_name))
+                    self.check_result = PrimeOffloaderError.NO_ERROR
+                    return
+                self.check_result = PrimeOffloaderError.OFFLOAD_FAIL
+            except (OSError, TypeError):
+                # Missing file or permissions?
+                self.logger.info("Couldn't open file for"
```
kissiel, line 161: It's better to forward the message from the exception instead of guessing :)

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+        :type card_id: str
+
+        :param card_name: card name of dri device
+        :type card_name: str
+        """
+        cmd_without_args = cmd.split(' ')[0]
+        for index in range(1, 11):
+            time.sleep(2)
+            try:
+                read_clients_cmd = \
```
kissiel, line 146: please don't escape newlines

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+        :param cmd: command that running under prime offload
+        :type cmd: str
+
+        :param card_id: card id of dri device
+        :type card_id: str
+
+        :param card_name: card name of dri device
+        :type card_name: str
+        """
+        cmd_without_args = cmd.split(' ')[0]
```
kissiel, line 142: This is why cmd should be structured.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+    def check_nv_offload_env(self):
+        """
+        prime offload of nvidia driver is limited.
+        Only on-demand mode is supported.
+        """
+        # nvidia-smi ship with NVIDIA GPU display drivers on Linux
+        # https://developer.nvidia.com/nvidia-system-management-interface
+        # check prime-select to make sure system with nv driver.
+        # If there is no nv driver, prime offload is fine for other drivers.
+        ps = subprocess.run("whereis prime-select | cut -d ':' -f 2",
```
kissiel, line 179: 1. doing piping like that in the process invocation can yield to many problems. It's better, more clear, and easier to do it in python `str.split()`
2. `whereis` yields you a mapping that you need a parse and potentially many records. `which` follows PATH and is more in line what you need. And you don't have to parse it.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+import subprocess
+import time
+import re
+import json
+import argparse
+import logging
+from enum import IntEnum
+
+
+class PrimeOffloaderError(IntEnum):
```
kissiel, line 32: This is not how error handling is done in python.
Here's a good resource on it: https://docs.python.org/3/tutorial/errors.html

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+        # nvidia-smi ship with NVIDIA GPU display drivers on Linux
+        # https://developer.nvidia.com/nvidia-system-management-interface
+        # check prime-select to make sure system with nv driver.
+        # If there is no nv driver, prime offload is fine for other drivers.
+        ps = subprocess.run("whereis prime-select | cut -d ':' -f 2",
+                            shell=True,
+                            stdout=subprocess.PIPE,
+                            universal_newlines=True)
+        if 'prime-select' not in ps.stdout:
+            return PrimeOffloaderError.NO_ERROR
```
kissiel, line 184: Are you sure that not having `prime-select` available in the system means "NO ERROR"? Feels like the opposite.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+            self.logger.info("System isn't on-demand mode")
+            return PrimeOffloaderError.NOT_SUPPORT_NV_PRIME
+
+        # prime offload couldn't running on nvlink active or inactive
+        # Therefore, only return empty string is supported environment.
+        nvlink = subprocess.run("nvidia-smi nvlink -s",
+                                shell=True,
+                                stdout=subprocess.PIPE,
+                                universal_newlines=True)
+        if len(nvlink.stdout) != 0:
```
kissiel, line 201: for checks like this `if nvlink.stdout:` is enough

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+            if re.search('error', nvlink.stdout, re.IGNORECASE):
```
kissiel, line 202: no need for heavy weapons like RE 
`if 'error' in nvlink.stdout.lower()`

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+                             " please check your pci name")
+            return PrimeOffloaderError.NO_CARD_NAME
+
+        # run offload command in other process
+        dri_pci_name_format = re.sub("[:.]", "_", pci_name)
+
+        if timeout > 0:
+            tmp_cmd = "timeout {} {}".format(timeout, cmd)
+        else:
+            tmp_cmd = cmd
```
kissiel, line 241: No need to create another variable, just prepend it with timeout if needed.
```
if timeout > 0:
    cmd = "timeout {} {}".format(timeout, cmd)
```
or even better, when using a lexically structured command:
```
if timeout:
    cmd = ['timeout', timeout] + cmd
```
another benefit of structured cmds

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+        if len(nvlink.stdout) != 0:
+            if re.search('error', nvlink.stdout, re.IGNORECASE):
+                self.logger.info("nvidia driver error")
+                return PrimeOffloaderError.NV_DRIVER_ERROR
+            self.logger.info("NVLINK detected")
+            return PrimeOffloaderError.NOT_SUPPORT_NV_PRIME
+
+        return PrimeOffloaderError.NO_ERROR
+
+    def run_offload_cmd(self, cmd, pci_name, driver, timeout):
```
kissiel, line 210: if the callers now about supplying timeouts, they as well may provide the right cmd (with timeout).

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+
+        :param cmd: command that running under prime offload
+        :type cmd: str
+
+        :param pci_name: pci device name in NNNN:NN:NN.N format
+        :type pci_name: str
+
+        :param driver: GPU driver, such as i915, amdgpu, nvidia
+        :type driver: str
+        """
```
kissiel, line 222: timeout argument left undocumented

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
...
+        card_id = self.find_card_id(pci_name)
+        if card_id == PrimeOffloaderError.NO_CARD_ID:
+            self.logger.info("Couldn't find card id,"
+                             " please check your pci name")
+            return PrimeOffloaderError.NO_CARD_ID
```
kissiel, line 227: This is a good example why in python C-like error handling is not great.
If the find_card_id couldn't find a card and it threw an exception, it would also bubble up from here and could be all handled in one place.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,334 @@
+#!/usr/bin/env python3
+"""prime offload test module."""
```
kissiel, line 2: This should go after the header, and overall brings zero value. Same information could be deduced from the file name.

kissiel (changes requested):
I've hit 20 comments mark.
Overall comments are:
* I think the logic regarding not find card is wrong (and test also reflects the wrong logic)
* the error handling is unnecessarily complicated, and not following python

The branch should have been easily split into smaller chunks. For instance the checkbox job definitions don't have to be here. Let's start with the testing program.

providers/base/bin/prime_offload_tester.py
hanhsuan: I have to use original cmd for checking this command is running.
```python
# use other thread to check offload is correctly or not
check_thread = threading.Thread(target=self.check_offload,
                                                      args=(cmd, card_id, card_name))
check_thread.start()
```

providers/base/bin/prime_offload_tester.py
hanhsuan: This sleep could make sure it won't use too much CPU resources. If we use a busy loop and the timeout is very long, it might make some issues.

providers/base/bin/prime_offload_tester.py
hanhsuan: For prime offloading, it only works on non Nvidia mode. Therefore, without `prime-select` (means no Nvidia driver installed).

providers/base/bin/prime_offload_tester.py
hanhsuan: For 2.
If the `info` is not an object that could be extracted, it will raise `TypeError`.

hanhsuan:
> * I think the logic regarding not find card is wrong (and test also reflects the wrong logic)

I couldn't figure out where is the logic error, cloud you explain more about this?

hanhsuan:
The submissions for the new commit:
[Intel only](https://certification.canonical.com/hardware/202312-33251/submission/349172/)
[Intel + Nvidia](https://certification.canonical.com/hardware/202312-33252/submission/349173/)
[Intel + Nvidia remote test](https://certification.canonical.com/hardware/202312-33252/submission/349179/)

hanhsuan:
The [bug of 6.5 kernel](https://bugs.launchpad.net/ubuntu/+source/linux-oem-6.5/+bug/2047461) is a public bug now.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        use pci name to find card id under /sys/kernel/debug/dri
+
+        :param pci_name: pci device name in NNNN:NN:NN.N format
+        :type card_name: str
+
+        : returns : card id
+        : rtype : str
+        """
+        if not re.match("[0-9]{4}:[0-9]{2}:[0-9]{2}.[0-9]", pci_name):
+            raise RuntimeError("pci name format error")
```
kissiel, line 58: Do we want the whole test to crash with a backtrace? Or do we want to exit the test program with a tidy error message and an error code? 
If it's the latter then I recommend raising `SystemExit`.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+                   "-lr",
+                   "--include=name",
+                   pci_name,
+                   "/sys/kernel/debug/dri"]
+
+            card_path = subprocess.check_output(cmd,
+                                                universal_newlines=True)
+            return card_path.split('/')[5]
+        except (IndexError, AttributeError) as e:
+            raise RuntimeError("return value format error {}".format(e))
```
kissiel, line 71: the inner raise creates an exception raising context while already inside a raising context. What it means the program crashes with the double traceback, like so:
```
  File "/home/kissiel/tmp/exc.py", line 7, in <module>
    foo()
  File "/home/kissiel/tmp/exc.py", line 3, in foo
    return a[5]
           ~^^^
IndexError: list index out of range

During handling of the above exception, another exception occurred:

  File "/home/kissiel/tmp/exc.py", line 9, in <module>
    raise RuntimeError("error msg {}".format(e))
RuntimeError: error msg list index out of range
```
The code that caused that:
```
def foo():
    a = [1]
    return a[5]


try:
    foo()
except IndexError as e:
    raise SystemExit("error msg {}".format(e)) 
```

If you're fine with crashing the program here, the original exception is fine, the double exception IMHO doesn't introduce anything of value and just obfuscates the code.

Of course gracefully exiting with SystemExit would be a much better solution.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+            cmd = ["grep",
+                   "-lr",
+                   "--include=name",
+                   pci_name,
+                   "/sys/kernel/debug/dri"]
+
+            card_path = subprocess.check_output(cmd,
+                                                universal_newlines=True)
+            return card_path.split('/')[5]
+        except (IndexError, AttributeError) as e:
```
kissiel, line 70: Why are there a catch for AttributeError? I don't see an object of which the attribute is called that it has the potential of not having it.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        line_b = b.strip().splitlines()
+        line_a = a.strip().splitlines()
+
+        len_b = len(line_b)
+        len_a = len(line_a)
+
+        if (len_b == (len_a - 1)
+           and any(x in line_a[len_a - 1] for x in ('Xorg', 'Xwayland'))):
+            return True
+        return False
```
kissiel, line 123: so the if can have two results, True and False.

So this block equivalent to:
```
if True:
   return True
else:
   return False
```

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        https://warthogs.atlassian.net/browse/OEMQA-3459
+
+        :param b: clients before test
+        :type b: str
+
+        :param a: clients after test
+        :type a: str
+
+        : returns : check result
+        : rtype : Boolean
```
kissiel, line 112: ```NameError: name 'Boolean' is not defined```

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        before that, we have to check the difference
+        https://warthogs.atlassian.net/browse/OEMQA-3459
+
+        :param b: clients before test
+        :type b: str
+
+        :param a: clients after test
+        :type a: str
+
+        : returns : check result
```
kissiel, line 111: inconsistent wrapping of blocks with colons ( `:`).

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+            for info in infos:
+                if pci_name in info['businfo']:
+                    return info['product']
+            raise RuntimeError("Card name not found")
+        except (KeyError, TypeError, json.decoder.JSONDecodeError) as e:
+            raise RuntimeError("return value format error {}".format(e))
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(e)
+
+    def check_for_kernel_bug(self, b, a):
```
kissiel, line 99: This function has nothing to do with kernel bugs.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        for 6.5 kernel, a bug of dir debugfs should be fixed.
```
kissiel, line 101: This should state what the function does.
From what I can tell by looking a the code it checks if the second string is longer by one line and checks whether the "after" one has "Xorg" or "Xwayland" in the last line.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        before that, we have to check the difference
+        https://warthogs.atlassian.net/browse/OEMQA-3459
+
+        :param b: clients before test
```
kissiel, line 105: why not call it "before"?

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        :type b: str
+
+        :param a: clients after test
```
kissiel, line 108: and "after"

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        : rtype : Boolean
+        """
+        line_b = b.strip().splitlines()
+        line_a = a.strip().splitlines()
+
+        len_b = len(line_b)
+        len_a = len(line_a)
+
+        if (len_b == (len_a - 1)
+           and any(x in line_a[len_a - 1] for x in ('Xorg', 'Xwayland'))):
```
kissiel, line 121: line_a[len_a - 1] in python is `len_a[-1]`

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+        :param b: clients before test
+        :type b: str
+
+        :param a: clients after test
+        :type a: str
+
+        : returns : check result
+        : rtype : Boolean
+        """
+        line_b = b.strip().splitlines()
```
kissiel, line 114: the point of splitlines is to split a string into line**s**. So the variable name is very misleading.

providers/base/bin/prime_offload_tester.py
```diff
@@ -0,0 +1,357 @@
...
+            raise RuntimeError(e)
+
+    def check_for_kernel_bug(self, b, a):
+        """
+        for 6.5 kernel, a bug of dir debugfs should be fixed.
+        before that, we have to check the difference
+        https://warthogs.atlassian.net/browse/OEMQA-3459
+
+        :param b: clients before test
+        :type b: str
```
kissiel, line 106: Type hints are even better for expressing types as they can be automatically checked.

providers/base/bin/prime_offload_tester.py
kissiel: The body of the function could be summarized to:
```
dri_clients_before = before.strip().splitlines()
dri_clients_after = after.strip().splitlines()
is_after_longer = len(dri_clients_after) == (len(dri_clients_before) + 1)

return is_after_longer and ("Xorg" in dri_clients_after[-1] or "Xwayland" in dri_clients_after[-1])
```
Please notice the readability differences.

kissiel (changes requested):
1. There are `time.sleep`()s being run when executing unit tests from this PR (it takes 2 minutes to run unit tests)
2. The code is unnecessarily complicated - see below.
3. There are unnecessary double exception contexts.
4. The `if __name__ == "__main__":` is abused, should only call main
5. The PR is too big. Please split it into smaller chunks (like one function with unit tests that does one thing).
6. The tests provided here are not unit tests. They also suffer from having to monkeypatch a lot of thing in the god object because of the suboptimal problem decomposition.

Overall recommendation: try isolating one problem at a time, write a function and that solves that one problem together with tests for that function and file a PR.
Then another one, another one, and so on. After 4-5PRs, you'll be able to compose a small change to a Checkbox job that uses all those functions.

This cannot land in this state.

hanhsuan:
1. The patch for 6.5 kernel has been released to proposed (6.5.0.16). Therefore the workaround isn't needed anymore.

**submission**
[submission](https://certification.canonical.com/hardware/202001-27667/submission/351906/) of this script without workaround that runs on 6.5.0.16

2. Move the changes of job and test plan to another PR, and will modify 24.04 test plan only to reduce the impact on SRU.

kissiel (changes requested):
Function decomposition proposed here introduces unnecessary complexity, and makes both, the logic, and the tests less readable and harder to maintain.

There are some problem around 

In line I'm providing comments and suggestions.

hanhsuan:
@kissiel I have fixed the code. Please help me to review, while you have time. Thanks.

kissiel (approved):
There are a few tweaks that could be applied to this PR, but it's not critical, so we can land this.

Thank you for the very extensive work on this!

PR #587, opened 2023-06-28
Fix: jobs.pxu in the wrong name

LiaoU3:
The jobs.pxu is wrong in iiotg/units/tsn. It's `jobs.pxu` not `jobs.psu`.


pieqq (approved):
Thanks for this!

PR #588, opened 2023-06-28
Setuptools scm deb daily

yphus:
## Description

Build debian packages using the setuptools_scm version (daily builds) and fully deprecate bumpversion

## Resolved issues

deb daily builds are using build recipes which originally computed the version using the following pattern:

`{debupstream}+{revtime:monorepo}+git{git-commit:monorepo}`

The recipe text is now updated with the version string computed by setuptools_scm

## Documentation

Release readme updated to remove all references to bumpversion and explains how setuptools_scm generates the version (using a small code snippet)

## Tests

Only tested the recipe update: https://code.launchpad.net/~checkbox-dev/+recipe/staging-checkbox-support-daily , not the full workflow which now depends on python3-setuptools-scm of course.


Hook25 (approved):
+1

PR #590, opened 2023-06-29
Add comments/context to explain why and how python-distro was patched on s16/18

yphus:
## Description

python3-distro was broken until the 1.6 release.
See https://github.com/python-distro/distro/issues/260

## Resolved issues

this `sed` oneliner was added in 2019 w/o any context/info

## Documentation

Inline comments added for future readers/maintainers

## Tests

snapcraft clean and build commands


mz2 (approved):
Makes sense 👍

PR #592, opened 2023-06-29
remove broken tests that fail provider validation

kissiel:
## Description

This patch **removes** two broken jobs from the IIOTG provider.

I tried making any sense of those jobs.
The ptp4l one just runs a command that lists status of the ptp4l servvice. This succeds even when the service is inactive. The summary says the test checks if the service is running.

The other one has two glaring problems:
1) it's a templated job and there's no resource specified that should be
   used to fill the values
2) the binary it calls is nowhere to be seen

In other words it's completely broken and probably hasn't been tested by anyone.

I checked if they are included in any of the test plans, and they're not meaning I didn't have to touch anything else.

## Tests
**Now** they pass :)


pieqq:
I added some github workflow to run tox tests, similar to what we do with other providers, to make sure this kind of problem does not happen again.

yphus (approved):
+1

PR #598, opened 2023-06-30
Fix remote builds snaps

Hook25:
## Description
SNAPCRAFT_REPO_ROOT seems to be empty in remote builds while the version file is staged and present in staging. This (should) fix snap daily and non dailiy builds.

## Resolved issues

N/A

## Documentation

N/A

## Tests

series22 and series16 were both ran. They implement the same thing and are two very different recipe, so this should work.

### DIFF
--- a/checkbox-core-snap/series16/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series16/snap/snapcraft.yaml
@@ -61,8 +61,9 @@ parts:
     plugin: dump
     source: .
     override-pull: |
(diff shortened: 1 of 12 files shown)


yphus (commented):
Can you really use a subshell in the in the build-environment section

Hook25:
Yes and it works! Cleared up quite a lot of clutter in the build steps

yphus (approved):
> Yes and it works! Cleared up quite a lot of clutter in the build steps

Fantastic, +1

PR #603, opened 2023-07-04
Fix EDID cycling test case

p-gentili:
## Description

EDID cycling testcase has always failed in our lab. The reason is that on laptops (and in general extended monitors) the resolution retrieved using `xdpyinfo` is considering the whole setup, not the single outputs.

With this PR, tools like `xrandr` and `gnome-randr` are used instead: they display for each available outputs the possible resolutions, the preferred one (+) and the one in use (*).

Also, since I've experimented some noise moving directly from one EDID to another, I added an "unplug" step between every new EDID. 

## Resolved issues

Fixes [ZAP-516](https://warthogs.atlassian.net/browse/ZAP-516)

## Documentation

The script was quite poor in terms of docstrings, I added a couple of them.

## Tests

Successful runs:
- X11: http://10.102.156.15:8080/job/cert-hp-probook-440-g9/370
- Wayland (locally, on NUC): 
```
$ XDG_SESSION_TYPE="wayland" DISPLAY=:0 checkbox-cli run com.canonical.certification::monitor/edid
Using sideloaded provider: checkbox-provider-base, version 2.7 from /var/tmp/checkbox-providers/base
[...]
changing EDID to 2560x1440
checking resolution... PASS
changing EDID to 1920x1080
checking resolution... PASS
changing EDID to 1280x1024
checking resolution... PASS
[...]
 ☑ : Check if the system automatically changes the resolution based on EDID
```

[ZAP-516]: https://warthogs.atlassian.net/browse/ZAP-516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/base/bin/edid_cycle.py
+++ b/providers/base/bin/edid_cycle.py
@@ -20,11 +20,69 @@ from checkbox_support.scripts.zapper_proxy import (             # noqa: E402
 PATTERN = r"^HDMI-1.*\n.*^\s+(\d+x\d+).*\*"
 
 
+def _match(output):
+    """
+    Match the given randr output with a pattern to grab
+    the current resolution on HDMI-1.
+
+    Both gnome-randr and xrandr highlight the currently
+    selected resolution for each monitor with a `*`.
+
+    xrandr
+    >>> _match("Screen 0: minimum 320 x 200, current 1920 x 1080,"
+    ... " maximum 16384 x 16384i\\n"
+    ... "DPI-1 connected primary 1920x1080+0+0 (normal left inverted"
+    ... "   1920x1080     60.00*+  59.97    59.96    59.93    40.00\\n"
+    ... "HDMI-1 connected primary 1920x1080+0+0 (normal left inverted"
+    ... " right x axis y axis) 309mm x 174mm\\n"
+    ... "   1920x1080     60.00+  59.97    59.96    59.93    40.00\\n"
+    ... "   1680x1050     59.95    59.88\\n"
+    ... "   1600x1024     60.17*\\n")
+    '1600x1024'
+
+    xrandr, no output on HDMI-1
+    >>> _match("Screen 0: minimum 320 x 200, current 1920 x 1080,"
+    ... " maximum 16384 x 16384i\\n"
+    ... "DPI-1 connected primary 1920x1080+0+0 (normal left inverted"
+    ... "   1920x1080     60.00*+  59.97    59.96    59.93    40.00\\n"
+    ... "HDMI-1 connected primary 1920x1080+0+0 (normal left inverted"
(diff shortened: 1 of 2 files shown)


providers/base/bin/edid_cycle.py
```diff
@@ -17,17 +17,34 @@
...
+    match = re.search(PATTERN, randr_output, re.MULTILINE | re.DOTALL)
+    if match:
+        return match.group(1)
+    return None
 
 
 def change_edid(host, edid_file):
+    """Clear EDID and then 'plug' back a new monitor."""
+    zapper_run(host, "change_edid", None)
+    time.sleep(1)
```
kissiel: I'll be thinking loud here:
If there's a delay that we have to wait in order for EDID to land, it seems like the better place to do that would be on controller side. Cause right now it "maybe changes" :)
Second thing is why 1s? IIRC we can check whether they got propagated (`v4l2 (something, something) query timings`), so I guess a busy loop would be better, that after 5seconds returns an error if it failed to do appropriate operation.
What do you think?

providers/base/bin/edid_cycle.py
```diff
@@ -17,17 +17,34 @@
 from checkbox_support.scripts.zapper_proxy import (             # noqa: E402
     zapper_run)
 
+PATTERN = "^HDMI-1.*\n.* (\\d+x\\d+).*\\*"
```
kissiel, line 20: use r-string literals, this way the whole string is taken as and you don't have to escape characters.
also the space before the parenthesis is redundant.
AKA:
`r"^HDMI-1.*\n.*(\d+x\d+).*\*'

Comment with explanation would be awesome, preferably with examples (**ehem** tests :P )

kissiel (changes requested):
See below.

providers/base/bin/edid_cycle.py
p-gentili: Thanks for the suggestion. The space is actually required because it matches just the last digit of the first number otherwise. Like "0x1080" instead of 1920x1080. I modified it a little bit to make it clearer.
Also, added doctests.

providers/base/bin/edid_cycle.py
p-gentili: Good point, those manual delays are not nice :D
It could be done on both sides actually. On controller side there's a problem though: the EDID change is immediate but the timings in use are available (and credible) only if there's a monitor connected. The command is `v4l2-ctl --query-dv-timings`.
Since I don't want the service to hang and timeout only because there's no monitor connected I'd prefer to have the busy loop on the checkbox side. Done in the last commit.

kissiel (approved):
Nice!
+1

PR #607, opened 2023-07-05
fix: sleep for 10s between connection retries in BT A2DP test

kissiel:
## Description
Previously it was only a second which meant that the temporary radio blackouts may have been responsible for failing the test. This patch makes the testing script wait for quite a bit longer between retries.

Landing this for the second time. After the first time it got reverted because of unrelated issue.
Let's hope this will stick better :-)

### DIFF
--- a/providers/base/bin/bt_a2dp.py
+++ b/providers/base/bin/bt_a2dp.py
@@ -60,6 +60,10 @@ def main():
                 break
             except bluetooth.btcommon.BluetoothError as exc:
                 print("Failed to connect. {}".format(exc))
+                # the sleep below lets the Bluetooth stack react to the
+                # changing environment. Without it the retries may happen
+                # too quickly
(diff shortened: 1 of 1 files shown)


Hook25 (approved):
+2, 1 for change, 1 for useful comment

kissiel:
Oh man, this branch is unlucky, I closed it by accident.

p-gentili:
Sorry guys but stuff like this should be tested manually (I should have done it in the first place as well). I sideloaded it in the lab and, unfortunately, it doesn't fix the problem.

After rebooting the host (and not the laptop) it started working again, for 100+ runs in sequence. This issue is so frustrating to reproduce :)

kissiel:
Oh I tested it manually.
The problem is it doesn't solve _all_ the problems, but helps with some of them.

PR #616, opened 2023-07-13
Fix the metabox trigger so that it is run to completion

Hook25:
## Description

This fixes the metabox github action by catching the return code instead of letting github killing the job (namely, the `if` was never run when `$? != 0` because github would kill the job)

## Resolved issues

N/A

## Documentation

Added comment to document why/what the new thing does

## Tests

This was tested offline via [act](https://github.com/nektos/act) with the following simplified workflow:
```yaml
name: Run Metabox when PR is approved

on:
  pull_request_review:
    types: [submitted]
  # Allow manual trigger
  workflow_dispatch:

jobs:
  metabox_run_required:
    runs-on: ubuntu-latest
    name: Check for changes in metabox and checkbox-ng dirs
    outputs:
      required_run: ${{ steps.check_diff.outputs.required_run }}
    steps:
      - name: Use git diff to see if there are any changes in the metabox and checkbox-ng directories
        id: check_diff
        run: |
          #DIFF_LENGTH=`true | wc -l`
          DIFF_LENGTH=`echo | wc -l`
          if [[  $DIFF_LENGTH -eq 0 ]]
            then
              echo "No Metabox run required."
              echo "required_run=false" >> $GITHUB_OUTPUT
            else
              echo "Metabox run required!"
              echo "required_run=true" >> $GITHUB_OUTPUT
          fi

  Metabox:
    if: true == fromJSON(needs.metabox_run_required.outputs.required_run)
    runs-on: ubuntu-latest
    needs: metabox_run_required
    steps:
      - name: Run Metabox scenarios
        run: |
          echo "running"
```

### DIFF
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -10,16 +10,19 @@ jobs:
   metabox_run_required:
     runs-on: ubuntu-latest
     name: Check for changes in metabox and checkbox-ng dirs
+    outputs:
+      required_run: ${{ steps.check_diff.outputs.required_run }}
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Use git diff to see if there are any changes in the metabox and checkbox-ng directories
         id: check_diff
-        continue-on-error: true
         run: |
-          git diff --exit-code HEAD origin/main -- checkbox-ng metabox
-          if [[ $? -eq 0 ]]
+          # catch the return code in subshell to avoid github stopping the job 
+          # due to it being non-0, non-0 means: "found some diff"
+          SHOULD=$(git diff --exit-code HEAD origin/main -- checkbox-ng metabox 1>/dev/null 2>/dev/null; echo $?)
+          if [[ $SHOULD -eq 0 ]]
             then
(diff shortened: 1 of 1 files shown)


.github/workflows/metabox.yaml
```diff
@@ -10,16 +10,19 @@ jobs:
...
           fetch-depth: 0
       - name: Use git diff to see if there are any changes in the metabox and checkbox-ng directories
         id: check_diff
-        continue-on-error: true
         run: |
-          git diff --exit-code HEAD origin/main -- checkbox-ng metabox
-          if [[ $? -eq 0 ]]
+          # catch the return code in subshell to avoid github stopping the job 
+          # due to it being non-0, non-0 means: "found some diff"
+          SHOULD=$(git diff --exit-code HEAD origin/main -- checkbox-ng metabox 1>/dev/null 2>/dev/null; echo $?)
```
kissiel, line 24: This introduces a complexity on top of the previous bad choice.
Why not do something simple like `DIFF_LENGTH=git diff (...) |wc -n`

kissiel (approved):
looks good!
Let's see if it'll actually work this time :D

PR #617, opened 2023-07-13
Move the error to the exception to give non-0 return

Hook25:
## Description
This replaces all logger critical and empty SystemExit w/ a SystemExit with the logged message so that we get it AND a non-0 return code. This was done because else, for example, github actions pass due to metabox returning 0 if it fails to start.

## Resolved issues

N/A

## Documentation

N/A

## Tests

I have run metabox locally, it will fail in github due to another problem related to https://github.com/canonical/checkbox/actions/runs/5543583811/


yphus (approved):
LGTM, thanks for improving metabox error handling

PR #619, opened 2023-07-14
Fix Metabox fails to run in github action

Hook25:
## Description

Metabox can not run in github actions because it fails to dry install checkbox-ng. This fixes this problem

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


Hook25:
Tests are running now: https://github.com/canonical/checkbox/actions/runs/5573778669/jobs/10181634823

They do still fail (Audio and ConfigChains are failing) but we should fix them in a separate PR

yphus (approved):
thanks for the pyproject support, the pip upgrade and the missing dirs. +1

PR #620, opened 2023-07-17
Remove old metabox scenario AudioPlayback

Hook25:
## Description

This was removed because its dependencies are heavy and we have decided to remove them a while back. The scenario is still here but it can not run (missing deps) and will make the test fail

## Resolved issues

N/A

## Documentation

N/A

## Tests

This was tested locally running metabox, the test is indeed gone


kissiel (approved):
+1

PR #621, opened 2023-07-17
Fix start remote cmd arg

Hook25:
## Description

This fixes the start cmd using it as intended (similarly to the other two kwargs).

This also fixes the ConfigChains scenario, limiting it to local because we do not yet support `check-config` for remote

## Resolved issues

N/A

## Documentation

This also adds a comment to explain how cmd is used/augmented by the remote start

## Tests

Metabox run with remote22 and local22 config

### DIFF
--- a/metabox/metabox/core/machine.py
+++ b/metabox/metabox/core/machine.py
@@ -22,6 +22,7 @@ import re
 import time
 from pathlib import Path
 import textwrap
+import shlex
 from contextlib import suppress
 
 import pylxd.exceptions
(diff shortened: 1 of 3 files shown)


yphus (approved):
LGTM, thanks for adding more flexibility to the start command

metabox/metabox/core/machine.py
```diff
@@ -165,19 +166,25 @@ def get_late_setup(self):
...
         return []
 
-    def start_remote(self, host, launcher, interactive=False, timeout=0):
+    def start_remote(self, host, launcher, cmd="remote", interactive=False, timeout=0):
+        """
+        Start remote
+        Note: This method will inject host and launcher at the end of cmd, if this is
+              undesirable, use run_cmd
+        """
+        if not cmd:
```
kissiel, line 175: The whole function signature is changed, and now it accepts a "cmd" , even the default is supplied, so I really don't understand why we have this check.

Looking at invocations of this function below AFAICT it's never called with anything other than cmd="", so why is the default value there in the first place.

metabox/metabox/core/machine.py
```diff
@@ -165,19 +166,25 @@ def get_late_setup(self):
...
+            cmd = "remote"
         assert self.config.role == "remote"
+        cmd_arr = shlex.split(cmd)
```
kissiel, line 178: you introduced the cmd with this patch, but here you change the type of it. If you want an list, just make it so.
Also what's with the piratey arr? Is it short for array, which kinda this is not.

kissiel (changes requested):
I don't see the problem this PR solves. The description lacklustre. Aside from that the new code is just weird.

Hook25:
Problem: Start action has a cmd parameter that is:
- Not used by remote runs in any way
- Overwritten by local if the action has a launcher, used else

We either use it always or we remove it.

metabox/metabox/core/machine.py
Hook25: The default value documents what the default behaviour is, the if is there to "enforce' it because else one would have to check in calling functions what the value of cmd is and if it is "" then change it (this because Start is the same action for remote and non remote)

metabox/metabox/core/machine.py
Hook25: I want them all to have the same type, the original one is a str and I did not change it, so it remain str everywhere for uniformity. The arr stands for array, I can change it to lst given that there is an array module, but I don't think that anyone reading it would have any doubt on what the type is, I just didnt want to alias the one above

Hook25:
Closing this as it is stale

PR #622, opened 2023-07-18
Checkbox 691/missing audio device (New)

diohe0311:
## Description
The problem is that the audio device cannot be detected in the RPi4. The root cause is udev_resource.py, which filters out the audio devices without a vendor ID. However, the audio devices in RPi4 do not have any vendor IDs.

After discussing with Maciej, we concluded that udev_resource is too complicated, with too many exception cases. It would be challenging to maintain and understand if we were to add more exceptions to it. Therefore, it's better to allow the audio card to have its own resource script, similar to the graphics card. Additionally, we already have alsa_pcm_info.py, which can gather information about the audio card.

Also add a unittest to testify if audio_card_resource.py can process data from file.

Changes:
1. Replace alsa_pcm_info.py with audio_card_resource.py.
2. Update all the test plans and jobs that required alsa_pcm_info.py.
3. Add a unittest for audio_card_resource

Update:
At first, we thought, 'Hey, why not do a better job by (1) extracting audio resources from the udev resources? This way, udev wouldn't become more complicated. And (2) adding automated detection for playback and capture, while also removing them from the manifest, so users don't have to answer it every time.' 
After serveral meetings and discussions with Pierre, Hanhsuan, Patrick and Kent, turns out manifest has great usage to QA team and OEM, its convenience and flexibility allow users to adjust the hw info efficiently. So we decide to undo the changes related to manifest and jobs `requires`.

## Resolved issues
https://warthogs.atlassian.net/browse/CHECKBOX-691

## Tests
Check flake8 and unit test
`./manage.py test -f`
`./manage.py test -u`

Check RPi4
`cd checkbox/providers/`
`scp -r resource ubuntu@10.102.163.11:/var/tmp/checkbox-providers/`
`ssh ubuntu@10.102.163.11`
`cd /var/tmp/checkbox-providers/resource/bin`
`checkbox.shell`
`./audio_card_resource.py`

Test the following jobs
`checkbox-cli run com.canonical.certification::audio/[JOB_NAME]`

- audio/detect-capture-devices
- audio/detect-playback-devices

Test with checkbox remote
- audio/detect_sources
- audio/detect_sinks
- audio/alsa_record_playback_automated

failed due to known issue: https://github.com/canonical/checkbox/issues/655
- audio/alsa-loopback
- audio/alsa-loopback-automated
- audio/alsa-playback


providers/base/units/audio/jobs.pxu
```diff
@@ -388,7 +388,7 @@ estimated_duration: 1.0
 imports: from com.canonical.plainbox import manifest
 requires:
   package.name == 'pulseaudio-utils'
-  manifest.has_audio_capture == 'True'
+  audio_card.Capture == 'supported'
```
Hook25, line 391: same as above, also, I think that resource/manifest entity names are usually lower case

providers/resource/bin/audio_card_resource.py
```diff
@@ -0,0 +1,70 @@
...
+                card_id = ids[0]
+                device_id = ids[1]
+                device_name = info[1]
+                capabilities = info[3:]
+                audio_cards.append({
+                    'Card': card_id,
+                    'Device': device_id,
+                    'Name': device_name,
+                    'Playback': ("supported" if any(cap.startswith('playback')
+                                 for cap in capabilities) else "unsupported"),
```
Hook25, line 42: make this into a small function like: (comment about `supported` and `unsupported` applies here as well)
```python
def has_capability(capability_prefix : str, capabilities : list) -> bool:
  return any(capability.startswith(capability_prefix) for capability in capabilities)
```

providers/resource/bin/audio_card_resource.py
```diff
@@ -0,0 +1,70 @@
...
+                    'Capture': ("supported" if any(cap.startswith('capture')
+                                for cap in capabilities) else "unsupported")
```
Hook25, line 44: consider making this a namedtuple, it is a bit cleaner according to me, also, upper case names here are a bit weird I think

providers/base/units/audio/jobs.pxu
```diff
@@ -695,32 +695,30 @@ _summary: Check that at least one audio playback device exits
 plugin: shell
 category_id: com.canonical.plainbox::audio
 flags: also-after-suspend
-imports: from com.canonical.plainbox import manifest
 requires:
-  manifest.has_audio_playback == 'True'
+  audio_card.Playback == 'supported'
```
Hook25, line 699: I may be very wrong here, but doesn't the resource job output `playback : 1`? Genuine question, how does `Playback` arrive here?

providers/resource/bin/audio_card_resource.py
```diff
@@ -0,0 +1,70 @@
...
+def print_audio_cards(cards):
+    """Print audio card information."""
+    for card in cards:
+        print("card: {}".format(card['Card']))
+        print("device: {}".format(card['Device']))
+        print("name: {}".format(card['Name']))
+        if card['Playback'] == "supported":
+            print("playback: 1")
+        if card['Capture'] == "supported":
+            print("capture: 1")
```
Hook25, line 59: This makes me even more convinced that they should be `True/False`, also, I'm not sure but I think that `playback: true` is a valid output here. Also, `has_playback_support` suggestion applies here as well!

Hook25 (changes requested):
The naming of the resource job is a bit unfamiliar to me but the overall functionality seems very nice!

providers/base/units/audio/jobs.pxu
```diff
@@ -374,7 +374,7 @@ estimated_duration: 1.0
 imports: from com.canonical.plainbox import manifest
 requires:
   package.name == 'pulseaudio-utils'
-  manifest.has_audio_playback == 'True'
+  audio_card.Playback == 'supported'
```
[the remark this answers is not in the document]
diohe0311: This change is to avoid using manifest(more automated), and Pierre suggests using `supported` or `unsupported`, so QA could easier to understand the meaning without confusion.

providers/resource/bin/audio_card_resource.py
```diff
@@ -0,0 +1,70 @@
...
+import os
+
+
+def get_audio_cards():
+    """Retrieve audio card information."""
+    audio_cards = []
+    PCM_FILE = '/proc/asound/pcm'
+    if os.path.exists(PCM_FILE):
+        with open(PCM_FILE, 'r') as f:
+            data = f.readlines()
```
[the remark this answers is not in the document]
diohe0311: You're right! thanks!

providers/base/units/audio/jobs.pxu
diohe0311: fixed!

providers/resource/bin/audio_card_resource.py
diohe0311: fixed!

providers/base/units/audio/jobs.pxu
diohe0311: Checkbox could create an entry for a resource job, so other jobs can call variables from the entry:
1. Define an ID in resource.pxu.
```
id: audio_card
estimated_duration: 0.05
plugin: resource
_summary: Collect information about the audio card
_description: Gets audio resource info from /proc/asound/pcm
command: audio_card_resource.py
```

2. Create the entry in test-plan.pxu (so before the jobs in the test-plan start, audio_card will run first and gather audio card info).
```
bootstrap_include:
 audio_card
```
Then, in job.pxu, we can use `requires` to set preconditions. Which variable should be set to what value to have a go.
```
requires:
  audio_card.playback == 'supported'
```

Hook25: the thing that puzzles me a bit is that I always thought that it is the printer that defines what is inside the resources, and the printer that you have created prints either 1 or 0, not supported and unsupported. Also I thought that resource entities had to be newline separated. I am not up to the task of saying if this is correct or not though. @pieqq may know more here

providers/base/units/audio/manifest.pxu
```diff
@@ -4,16 +4,6 @@
...
 
-unit: manifest entry
-id: has_audio_playback
-_name: Audio playback
-value-type: bool
-
-unit: manifest entry
-id: has_audio_capture
-_name: Audio capture
-value-type: bool
```
pieqq, line 15: As discussed above, it might be better to keep manual manifest entries to let the tester tell Checkbox if the device being tested has sound input/output or not.

providers/resource/bin/audio_card_resource.py
```diff
@@ -0,0 +1,81 @@
...
+def print_audio_cards(cards):
+    """Print audio card information."""
+    for card in cards:
+        print("Card: {}".format(card.card))
+        print("Device: {}".format(card.device))
+        print("Name: {}".format(card.name))
+        if card.playback == "supported":
+            print("Playback: 1")
+        if card.capture == "supported":
+            print("Capture: 1")
```
pieqq, line 70: There should be a last empty line being printed (so basically adding `print()`), so that each device is separated from each other.

I just tried running this script on my computer and I get this output:

```
$ ./audio_card_resource.py
Card: 00
Device: 00
Name: ALC887-VD Analog
Playback: 1
Capture: 1
Card: 00
Device: 01
Name: ALC887-VD Digital
Playback: 1
Card: 00
Device: 02
Name: ALC887-VD Alt Analog
Capture: 1
Card: 01
Device: 00
Name: USB Audio
Playback: 1
Capture: 1
Card: 02
Device: 03
Name: HDMI 0
Playback: 1
Card: 02
Device: 07
Name: HDMI 1
Playback: 1
Card: 02
Device: 08
Name: HDMI 2
Playback: 1
Card: 02
Device: 09
Name: HDMI 3
Playback: 1
Card: 02
Device: 10
Name: HDMI 4
Playback: 1
Card: 02
Device: 11
Name: HDMI 5
Playback: 1
Card: 03
Device: 00
Name: USB Audio
Capture: 1
```

This will not work as intended.

pieqq (changes requested):
As discussed in a meeting earlier with @diohe0311 , I think it's risky to remove the manifest entries (`has_audio_capture` and `has_audio_playback`). These entries are manual to allow for the tester to let Checkbox know if this device has, indeed, a sound output and/or a sound input.

If we rely solely on automated jobs, like with the `audio_card` resource created here, if, for any reason, the audio device(s) are not detected by Ubuntu (because of a driver issue, because of a kernel issue, because of a UEFI issue, etc.), then Checkbox will assume there are no sound input/output and will proceed with skipping the jobs. This is bad because QA and/or PM may not see that some audio jobs were skipped, and they may deem the device tests as complete whereas there should be bugs raised about the missing sound interfaces...

@kissiel, any opinion about this?

providers/resource/bin/audio_card_resource.py
Hook25: Thsi responds to my question above, at best it is missing newline, but also, idk how "supported" is getting out of this

providers/base/units/audio/jobs.pxu
pieqq: I agree with @Hook25, there is a problem here.

If I run the `audio/detect_sources` job, for instance, which has been modified to use the `audio_card` resource requirements, I get this:

```
$ checkbox-cli run com.canonical.certification::audio/detect_sources
(...)
==============[ Running job 3 / 3. Estimated time left: 0:00:01 ]===============
----------------------------[ audio/detect_sources ]----------------------------
ID: com.canonical.certification::audio/detect_sources
Category: com.canonical.plainbox::audio
 - resource expression "audio_card.capture == 'supported'" evaluates to false
```

And this is expected, because if I run the `audio_card` resource job itself, I get that:

```
$ checkbox-cli run com.canonical.certification::audio_card
------------------[ Collect information about the audio card ]------------------
ID: com.canonical.certification::audio_card
Category: com.canonical.plainbox::uncategorised
Card: 00
Device: 00
Name: ALC887-VD Analog
Playback: 1
Capture: 1

(...)

Card: 03
Device: 00
Name: USB Audio
Capture: 1

Outcome: job passed

```

Two issues here:

1. In the `audio_card` resource job, there is no `capture` field (but there is a `Capture` field!)
2. In the `audio_card` resource job, the `Capture` field is set to `1` when capture is supported, not to `supported`

I have a few recommendations:

- If you look at the scripts used in other resource jobs (for instance `cpuinfo_resource.py`), you will see that the fields they provide are all lowercases, and the values for (un)supported status is usually `supported`/`unsupported`:

```
$ ./cpuinfo_resource.py 
bogomips: 6385
cache: 6291456
count: 4
model: Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz
model_number: 6
model_revision: 3
model_version: 60
other: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm cpuid_fault epb invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid xsaveopt dtherm ida arat pln pts md_clear flush_l1d
platform: x86_64
speed: 3600
type: GenuineIntel
governors: conservative ondemand userspace powersave performance schedutil
scaling: supported
```

I would recommend you use the same convention for clarity.

(I'm gonna add a few other comments in a separate review)

providers/base/units/audio/jobs.pxu
```diff
@@ -374,7 +374,7 @@ estimated_duration: 1.0
 imports: from com.canonical.plainbox import manifest
 requires:
   package.name == 'pulseaudio-utils'
-  manifest.has_audio_playback == 'True'
+  audio_card.playback == 'supported'
```
pieqq, line 377: It's OK to add this, but I would keep the manifest requirement as well (the same remark applies to the other places where you removed the requirement on `manifest.xxx` below).

Just for your information, if you add the `fail-on-resource` flag to a [job unit](https://checkbox.readthedocs.io/en/latest/reference/units/job.html), the job will fail instead of being skipped if one of the requirements is not met. But in our case, it's a bit annoying, because I guess we would like to raise an alarm if the tester had said "yes, this device has audio playback" (thanks to the `manifest.has_audio_playback == 'True'` bit) and yet the `audio_card.playback` value came back as `unsupported`... but we don't have a way to do that, cause `fail-on-resource` will mark the job as failed if **any** resource requirement is not met.

pieqq (changes requested):
A few more comments inline.

Also, I would recommend that you test your work by calling the jobs and/or the test plans you have modified. In your venv, you can run a job with, for example:

```
(venv) $ checkbox-cli run com.canonical.certification::audio/detect-playback-devices

(...)

==============[ Running job 2 / 2. Estimated time left: 0:00:00 ]===============
------------[ Check that at least one audio playback device exits ]-------------
ID: com.canonical.certification::audio/detect-playback-devices
Category: com.canonical.plainbox::audio
 - resource expression "audio_card.playback == 'supported'" evaluates to false
```

kissiel:
> As discussed in a meeting earlier with @diohe0311 , I think it's risky to remove the manifest entries (`has_audio_capture` and `has_audio_playback`). These entries are manual to allow for the tester to let Checkbox know if this device has, indeed, a sound output and/or a sound input.
> 
> If we rely solely on automated jobs, like with the `audio_card` resource created here, if, for any reason, the audio device(s) are not detected by Ubuntu (because of a driver issue, because of a kernel issue, because of a UEFI issue, etc.), then Checkbox will assume there are no sound input/output and will proceed with skipping the jobs. This is bad because QA and/or PM may not see that some audio jobs were skipped, and they may deem the device tests as complete whereas there should be bugs raised about the missing sound interfaces...
> 
> @kissiel, any opinion about this?

If the skip is not enough, let's mark the job with `fail-on-resource` flag. This way it will be very red if checkbox fails to detect _any_ device of the type.

The problem I have right now is that we're extending general manifest questionnaire to cover some extreme corner cases that don't affect 99% of the systems (how many systems don't have **any** audio dev?). So for systems  that are in fact audio-less, it's up to the test plan author to make sure to exclude all the audio jobs so they don't pop up as skipped/failed.

kissiel (commented):
There's one unnecessary shebang (marked inline).

scotthu27:
Agree with @pieqq it's risky to remove the manifest entries (has_audio_capture and has_audio_playback). it usually observed the audio device not detected in develop phase. 
On the other hand, there are too many different configs with desktop/ workstation in PC project, lots problem we observed by auto detection like `audio_card ` resource here. That's why we need the manifest entries to exclude related cases in the beginning.

has_audio_capture & has_audio_playback in the manifest entries would've been for including/excluding the audio jobs. 
Is it better to exclude the jobs one by one or check all jobs with `fail-on-resource` flag manually rather than input these 2 values in few secs?

providers/base/units/audio/jobs.pxu
```diff
@@ -369,26 +369,28 @@ _description:
 plugin: shell
 category_id: com.canonical.plainbox::audio
 id: audio/detect_sinks
-flags: also-after-suspend
+flags: also-after-suspend fail-on-resource
```
pieqq, line 372: After the long discussion we had on mattermost, I think it's better to remove this `fail-on-resource` flag. The reason is that if we put it, the audio jobs are going to fail for all the devices where the tester specified `manifest.has_audio_playback: false`

providers/base/units/audio/jobs.pxu
```diff
@@ -369,26 +369,28 @@ _description:
...
+  audio_card.playback == 'supported'
 command:
   pactl_list.sh sinks
 _description:
   Test to detect if there's available sinks
 
 plugin: shell
 category_id: com.canonical.plainbox::audio
-flags: also-after-suspend
+flags: also-after-suspend fail-on-resource
```
pieqq, line 386: As explained above, remove `fail-on-resource`.

providers/base/units/audio/jobs.pxu
```diff
@@ -694,33 +696,35 @@ id: audio/detect-playback-devices
 _summary: Check that at least one audio playback device exits
 plugin: shell
 category_id: com.canonical.plainbox::audio
-flags: also-after-suspend
+flags: also-after-suspend fail-on-resource
```
pieqq, line 699: As explained above, remove `fail-on-resource`.

providers/base/units/audio/jobs.pxu
```diff
@@ -694,33 +696,35 @@ id: audio/detect-playback-devices
...
+    exit 1
+  fi
+estimated_duration: 0.5
 
 id: audio/detect-capture-devices
 _summary: Check that at least one audio capture device exists
 plugin: shell
 category_id: com.canonical.plainbox::audio
-flags: also-after-suspend
+flags: also-after-suspend fail-on-resource
```
pieqq, line 716: As explained above, remove `fail-on-resource`.

pieqq (changes requested):
After discussing with QA on Mattermost on that topic, it's better to

- leave the manual manifest
- not use `fail-on-resource` flag, otherwise the jobs would be marked as failed if tester specified device has no audio input/output.

Once you've done these changes, please create a snap from your branch and try to run the following jobs (the ones that are impacted by the manual manifest) to make sure they work as expected:

- audio/detect-capture-devices
- audio/detect_sources
- audio/detect_sinks
- audio/detect-playback-devices
- audio/alsa-loopback
- audio/alsa-loopback-automated
- audio/alsa_record_playback_automated
- audio/alsa-playback

Try setting the required manifest to `false`, then `true`.

To make a snap:

1. go to `checkbox-core-snap`
2. run `./prepare.sh series22`
3. go to `series22`
4. run `snapcraft`

If you want to create a snap for the Raspberry Pi, you need to build the arm64 version of the snap. To do that, in step 4, run `snapcraft remote-build --build-for=arm64 --launchpad-accept-public-upload`.

Once the snap is available, you can install it by running `sudo snap install <snap-name>.snap --dangerous`.

diohe0311:
> snapcraft remote-build --build-for=arm64 --launchpad-accept-public-upload

Thanks! Pierre, I tested those jobs you list on the snap built from my branch, all passed except for:
- audio/alsa-loopback
- audio/alsa-loopback-automated
- audio/alsa-playback
They failed because of https://github.com/canonical/checkbox/issues/655

providers/base/units/audio/jobs.pxu
```diff
@@ -698,13 +700,14 @@ flags: also-after-suspend
 imports: from com.canonical.plainbox import manifest
 requires:
   manifest.has_audio_playback == 'True'
+  audio_card.playback == 'supported'
```
patliuu, line 703: Hi, I followed the discussion of this PR a few weeks ago, so I just took a look and am a little bit curious about this `requires: audio_card.playback == 'supported'`.

If I understand correctly, considering the following situation,
When `manifest` is set to `True` (the device should have an audio playback feature), but the resource job `audio_card` **cannot** detect the playback devices so is NOT `supported`. This `audio/detect-playback-devices` job should fail. However, adding this `requires: audio_card.playback == 'supported'` will make this job be skipped. 

Same as the `audio/detect-capture-devices` below.

providers/resource/bin/audio_card_resource.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def has_capability(capability_prefix: str, capabilities: list) -> bool:
+    return any(capability.startswith(capability_prefix)
+               for capability in capabilities)
+
+
+def print_audio_cards(cards):
+    """Print audio card information."""
+    print(cards)
```
patliuu, line 59: This `print()` here will make the resource job output like,
```
[{'card': '00', 'device': '00', 'name': 'f130000.audio-controller-sgtl5000 sgtl5000-0', 'playback': True, 'capture': True}]
card: 00
device: 00
name: f130000.audio-controller-sgtl5000 sgtl5000-0
playback: supported
capture: supported
```
I'm wondering if this should be removed.

pieqq (approved):
LGTM! Thanks :)

Hook25 (approved):
+1, this came before the coverage requirement, I think it can be merged

PR #625, opened 2023-07-19
Fix docs makefile to use local copy of checkbox-ng for version info

pieqq:
## Description

Install checkbox-ng from local repository rather than relying on outdated version in PyPI.

## Tests

```
cd docs/
make install
make run
```

This works and return a live version of the documentation with the correct Checkbox version in the header.


Hook25 (approved):
+1

Issue #611, opened 2023-07-07
miscellanea/cpuid doesn't recognize 0xaa0f02 as code for AMD EPYC 9754


rodwsmith:
### Bug Description

The miscellanea/cpuid test is failing when presented with an AMD EPYC 9754 CPU with an ID of 0xaa0f02. This code should be added to the script.

### To Reproduce

Observed in the following certificate requests:

https://certification.canonical.com/certificates/2307-14800/
https://certification.canonical.com/certificates/2307-14817/

### Environment

Lenovo ThinkSystem SR645 V3 (oahu) under Ubuntu 20.04 and 22.04

### Relevant log output

```shell
CPU Model: AMD EPYC 9754 128-Core Processor
Unable to determine CPU Family for this CPUID: 0xaa0f02
```

Hook25:
Closing as solved in #612

PR #633, opened 2023-07-21
Providers/base/bin/cpuid.py: added another bergamo ID fixes: #611 SERVCERT-1230

bladernr:
## Description

Adds another missing bergamo ID and reformats the CPUID list a bit.

## Resolved issues

#611 
SERVCERT-1230


bladernr:
Closing / withdrawing this, I fired it off automatically as I was doing cert reviews and only realized later this is the missing ID that Michael has already submitted a patch for.

PR #640, opened 2023-07-26
Check the installed package instead of from ubuntu-archive (BugFix)

os369510:
…e instead of from ubuntu-archive (LP: #2027655)

## Description

The current test case checks the recommended packages of ubuntu-desktop which is the version from ubuntu-archive.
However, we need to check the package we currently installed on the system.

## Resolved issues

Fixes: the tool part from https://bugs.launchpad.net/stella/+bug/2027655

## Tests

Tested-by: [andy.chi@canonical.com](mailto:andy.chi@canonical.com)

### DIFF
--- a/providers/base/bin/check-ubuntu-desktop-recommends.sh
+++ b/providers/base/bin/check-ubuntu-desktop-recommends.sh
@@ -3,13 +3,16 @@ set -euo pipefail
 IFS=$'\n\t'
 
 noninstalled=()
+target_version="$(dpkg-query --showformat='${Version}' --show ubuntu-desktop)"
+apt_show_ud="$(apt-cache show ubuntu-desktop="$target_version")"
+recommends="$(echo "${apt_show_ud}"| grep ^Recommends | head -n 1)"
 while read -r pkg; do
     # libreoffice-impress provides libreoffice-ogltrans, and libreoffice-ogltrans becomes a transitional package on Ubuntu 20.04.
     # shellcheck disable=SC2016
     if ! dpkg-query -W -f='${Status}\n' "$pkg" 2>&1 | grep "install ok installed" >/dev/null 2>&1 && [ "$pkg" != "libreoffice-ogltrans" ]; then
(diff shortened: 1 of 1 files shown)


providers/base/bin/check-ubuntu-desktop-recommends.sh
```diff
@@ -9,7 +9,7 @@ while read -r pkg; do
     if ! dpkg-query -W -f='${Status}\n' "$pkg" 2>&1 | grep "install ok installed" >/dev/null 2>&1 && [ "$pkg" != "libreoffice-ogltrans" ]; then
         noninstalled+=("$pkg")
     fi
-done < <(apt-cache show ubuntu-desktop | grep ^Recommends | head -n 1 | cut -d : -f 2- | xargs | sed 's/ //g' | tr , $'\n')
+done < <(apt-cache show ubuntu-desktop="$(dpkg-query --showformat='${Version}' --show ubuntu-desktop)" | grep ^Recommends | head -n 1 | cut -d : -f 2- | xargs | sed 's/ //g' | tr , $'\n')
```
pieqq, line 12: This line is starting to get really big. I would compute the version number first, then use it in the `apt-cache...` command.

pieqq (changes requested):
Thanks for catching this problem! Just one suggestion inline.

providers/base/bin/check-ubuntu-desktop-recommends.sh
os369510: @pieqq would you please review it again?

pieqq (approved):
Looks good, thanks!

pieqq:
@os369510 can you sign your commits and resubmit? Please see our contrib doc about this: https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#signed-commits-required

Then we'll be good to land this!

kissiel:
I've done the rebase in #914, so it may actually land quickly.
Let's close this one.

PR #643, opened 2023-07-27
Fix list split exporter

Hook25:
## Description

The exporter options parameter in config is now a list and no longer a string (this was introduced in the config refactor). This split is done before now and causes an obvious problem when it is done here (list can not be split)

Also, found a small bug when providing secure_id, ctx is no longer available in that function, now the secure_id is fetched differently

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-776

## Documentation

N/A

## Tests

Metabox run locally and remotely

### DIFF
--- a/metabox/metabox/scenarios/config/exporters.py
+++ b/metabox/metabox/scenarios/config/exporters.py
@@ -2,12 +2,11 @@
 #
 # Copyright 2023 Canonical Ltd.
 # Written by:
-#   Pierre Equoy <pierre.equoy@canonical.com>
+#   Massimiliano Girardi <massimiliano.girardi@canonical.com>
 #
 # Checkbox is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3,
 # as published by the Free Software Foundation.
-
 #
 # Checkbox is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


metabox/metabox/scenarios/config/exporters.py
```diff
@@ -0,0 +1,44 @@
+# This file is part of Checkbox.
+#
+# Copyright 2023 Canonical Ltd.
+# Written by:
+#   Pierre Equoy <pierre.equoy@canonical.com>
```
kissiel, line 5: copy pasta problem?

metabox/metabox/scenarios/config/exporters.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# Checkbox is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3,
+# as published by the Free Software Foundation.
+
```
kissiel, line 10: delete me

kissiel (commented):
TYVM!
The only problem I see is with the file header comment.

kissiel (approved):
TYVM!
+1

PR #644, opened 2023-07-27
Readable metabox output in CI

Hook25:
## Description

Right now when a metabox action fails on github it is very tedious to recover why it did. This is due to one fact: the amount of output metabox produces. Metabox is very verbose in CI because we need to launch it with log level TRACE, else it will not print any stdout/stderr of the scenarios that it is testing and that would be useless (we would know that a test failed but not why). 

This PR does two things:
**Raises the log level to INFO:** This should make the log informative (giving testing progress for example via `Starting scenario` prompts) but way more concise (skipping all debug messages and all outputs produced by successful jobs)
**stdout/stderr in log in failure:** When a scenario fails, metabox will now always log at the error level the fact that it did

## Resolved issues

N/A

## Documentation

N/A

## Tests

I have forced a local(manual + auto) and a remote(manual + auto) test to fail to see the new output

### DIFF
--- a/metabox/metabox/core/scenario.py
+++ b/metabox/metabox/core/scenario.py
@@ -72,12 +72,13 @@ class Scenario:
         self._ret_code = None
         self._stdout = ""
         self._stderr = ""
+        self._oudstr_full = ""
         self._pts = None
 
     def get_output_streams(self):
         if self._pts:
             return self._pts.stdout_data_full.decode("utf-8")
-        return self._stdout + self._stderr
+        return self._outstr_full
 
     def has_passed(self):
         """Check whether all the assertions passed."""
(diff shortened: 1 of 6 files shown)


metabox/metabox/core/scenario.py
```diff
@@ -63,10 +70,15 @@ def __init__(self, mode, *releases, remote_revision="current", service_revision=
...
-        self._stdout = ''
-        self._stderr = ''
+        self._stdout = ""
+        self._stderr = ""
         self._pts = None
 
+    def get_output_streams(self):
+        if self._pts:
+            return self._pts.stdout_data_full.decode("utf-8")
+        return self._stdout + self._stderr
```
kissiel, line 80: This is almost never a good idea. Those two streams have their own timeline, and concatenating them like that can be very misleading. Either:
1. when grabbing the streams, create a mixed one (while preserving the discrete ones, or not :) )
2. yield them separately.

kissiel (commented):
We really need to plough the whole codebase with black, cause those formatting tweaks are making reviewing process much more time consuming than what it should be.

I love the general idea, and TYVM for a very clear description!

I was kind of surprised that by:

>Raises the log level to INFO: 

you meant the workflow will call metabox with --log INFO.
If that granularity is the best, I think we should make it the default level throughout the code. It's a very dev-facing tool, so it won't hurt anyone.

metabox/metabox/core/scenario.py
Hook25: I was being lazy here because I did not want to overhaul the `run_or_raise` api. You are right, I have changed this. Also: we should really look into that function and at least change the name

Hook25:
Agree, INFO is the best granularity -> should be the default

I have changed that

kissiel (approved):
TYVM for the combined log!
All looks great, +1

PR #645, opened 2023-07-27
Add tutorial provider

kissiel:
## Description

This PR introduces a Checkbox Tutorial Provider.
The included test plan contains various jobs that yield different outcomes, including outcomes that are results of failing resources or dependent jobs.

Tested throughout manually :man_mechanic: 

Resolves: [CHECKBOX-750](https://warthogs.atlassian.net/browse/CHECKBOX-750)


providers/tutorial/manage.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+from plainbox.provider_manager import setup, N_
+
+setup(
+    name='com.canonical.certification:tutorial',
```
pieqq, line 5: All the other providers in our codebase have something like this:

```
    name='checkbox-provider-base',
    namespace='com.canonical.certification',
```

Is prefixing the `name` field with the `namespace` the way to run now?

providers/tutorial/units/jobs.pxu
```diff
@@ -0,0 +1,52 @@
...
+ (tutorial_resources.name == 'first-resource' and tutorial_resources.has_required_thing == 'true')
+_description: A job that requires a resource and it is available
+command:
+ true  # this will run!
+
+
+id: tutorial/no_resource
+plugin: shell
+requires:
+ (tutorial_resources.name == 'second-resource' and tutorial_resources.has_required_thing == 'true')
```
pieqq, line 33: I was very surprised to see this works. I always assumed Checkbox would check that **any** `tutorial_resources.name` is set to `second-resource` and **any** `tutorial_resources.has_required_thing` is set to `true`, but apparently it's smart enough to understand we want the value of `tutorial_resources.has_required_thing` for the resource named `second-resource`... Interesting!

pieqq (changes requested):
One question inline, and also one request: as I pointed out in Jira, I need a job that uses an environment variable (with a default set), because of [this section in the tutorial](https://github.com/canonical/checkbox/blob/main/docs/tutorial/using-checkbox/basic-launcher.rst?plain=1#L143).

So basically a job that runs something like

```
echo ${TUTO:-default}
```

so that it still outputs something if no env var is set, but behaves differently if an env var is set (in launcher, in my case).

providers/tutorial/units/jobs.pxu
kissiel: Yep, that's one of the "features" of checkbox. And one that is extremely taxing when you use too many resources :)

providers/tutorial/manage.py
kissiel: Interesting comment... I've tracked this down, and [found this](https://github.com/canonical/checkbox/blob/main/checkbox-ng/plainbox/impl/secure/providers/v1.py#L700C30-L700C30).
And I got even more interested in what this "_has_dedicated_namespace" is used for, and I found only one thing this affects is printing `namespace: {}` instead of `namespace: {} (derived from name)` when `./manage.py info` is run.


*slaps the roof of checkbox* this bad boy can fit so much over engineering in it.

kissiel: Let's proceed with having it separate so the `info` command prints only the unqualified (cleaner) part.

pieqq:
Another request for this PR: the tutorial needs to be available from an installed snap, so the snapcraft.yaml recipes need to package it, the same way we package `provider-sru` for instance.

pieqq:
Revisiting this:

- yes, `description` should be replaced by `summary`, otherwise user will see job ids in the test selection screen
- a manual job is needed to show the user how to proceed with them
- probably also need a user-interact and a user-interact-verify job as well

pieqq:
I've done the following:

- Add 3 interactive jobs (manual, user-interact and user-interact-verify) to demonstrate them when running the Tutorial test plan
- Add a Tutorial category so that our Tutorial jobs are not in the `Uncategorised` section of the test plan (it looked lame in the Tutorial page)
- Add some output to the passing and failing job for demonstration purposes (see #705)
- Change `description` to `summary` so that we see a nice list of jobs summaries in the job selection screen

tang-mm (approved):
Changes in docs/ look good to me

checkbox-core-snap/series22/snap/snapcraft.yaml
```diff
@@ -438,6 +442,18 @@ parts:
       python3 manage.py build
       python3 manage.py install --layout=relocatable --prefix=/providers/checkbox-provider-certification-server --root="$SNAPCRAFT_PART_INSTALL"
     after: [checkbox-provider-certification-client]
+################################################################################
+  checkbox-provider-tutorial:
```
kissiel, line 446: Interesting. I haven't seen the value in *delivering* the tutorial provider with the rest of the stack. I thought it's more of a development thing that's meant to be run venv.

kissiel (commented):
Thanks for picking up the branch!
One question below

checkbox-core-snap/series22/snap/snapcraft.yaml
pieqq: One of the first sections of the tutorial, after installing Checkbox via snaps, is about running the tutorial test plan:

https://checkbox.readthedocs.io/en/latest/tutorial/using-checkbox/basic-running.html

So it's required (or the base tutorial needs to be re-written :))

kissiel: Right. Then it makes total sense :)

kissiel (commented):
I can't officially +1 this PR as I opened it, but big +1 from me!

pieqq (approved):
Yay, I have to approve my own changes! Thank you GitHub!

PR #646, opened 2023-07-31
Fix flake8 error

diohe0311:
## Description
Flake8 has been updated over the weekend: https://pypi.org/project/flake8/#history. 
As a result, some scripts in providers/base failed the Flake8 test. 
I'm not sure if specifying the flake version to run the tests is a good solution. So I changed those scripts under providers/base/bin/ to pass flake8 tests.

## Tests
Create this PR see if provider-base could pass tox test in py38 env


Hook25 (approved):
+1, this is not equivalent but it seems to match the meaning that the code was trying to express, nice

PR #652, opened 2023-08-04
Fix metabox snap testing

Hook25:
## Description

Metabox has a mechanism to test the checkbox runtime/frontend snap but it is outdated. This is due to the fact that the metabox provider was sideloaded (changed to installed in [this PR](https://github.com/canonical/checkbox/pull/500)) but this can not work with snaps because:
- The install process needs the `plainbox` module, that is not in the system modules 
- The install process can not write in the `provider` location of the snap (it is readonly, snaps are a squashfs)
-  We have to avoid sideloading due to https://warthogs.atlassian.net/browse/CHECKBOX-624

This installs the metabox provider in an overlay that is auto-mounted to the checkbox snap at boot. This is done to avoid sideloading but also having an install of the metabox provider persistent across reboots. Once mounted checkbox will find the `metabox-provider` directory correctly located in `/snap/checkboxXX/current/providers`

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-777

## Documentation

The documentation was updated to explain how to do snap testing (in the metabox README.md) 

## Tests

Running the following test I was able to test the store snap for `focal`:
```python
configuration = {
    "local": {
        "origin": "classic-snap",
        "checkbox_core_snap": {"risk": "edge"},
        "checkbox_snap": {"risk": "edge"},
        "releases": ["focal"],
    },
    "remote": {
        "origin": "classic-snap",
        "checkbox_core_snap": {"risk": "edge"},
        "checkbox_snap": {"risk": "edge"},
        "releases": ["focal"],
    },
    "service": {
        "origin": "classic-snap",
        "checkbox_core_snap": {"risk": "edge"},
        "checkbox_snap": {"risk": "edge"},
        "releases": ["focal"],
    },
}
```

## Caveats

- This feature does not seem work if the host machine of the lxd deployment runs a linux kernel version < `5.19` because the needed overlay support is not there. This is a problem mainly for WSL2 users

### DIFF
--- a/metabox/metabox/core/machine.py
+++ b/metabox/metabox/core/machine.py
@@ -534,21 +534,21 @@ class ContainerSnapMachine(ContainerBaseMachine):
         service_content = (
             textwrap.dedent(
                 """
-            [Unit]
-            Description=Checkbox Overlay Service
-            Wants=network.target
-            Before=snap.{name}.service.service
+                [Unit]
+                Description=Checkbox Overlay Service
+                Wants=network.target
+                Before=snap.{name}.service.service
 
-            [Service]
-            ExecStart={cmd}
-            SyslogIdentifier=overlay-checkbox.service
-            Restart=on-failure
-            TimeoutStopSec=30
-            Type=simple
+                [Service]
+                ExecStart={cmd}
+                SyslogIdentifier=overlay-checkbox.service
+                Restart=on-failure
+                TimeoutStopSec=30
+                Type=simple
 
-            [Install]
-            WantedBy=multi-user.target
-            """
+                [Install]
+                WantedBy=multi-user.target
+                """
             )
             .lstrip()
             .format(name=self._snap_name, cmd=service_cmd)
--- a/metabox/metabox/core/machine.py
+++ b/metabox/metabox/core/machine.py
@@ -16,10 +16,11 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
-import itertools
 import time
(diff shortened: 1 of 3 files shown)


metabox/metabox/core/machine.py
```diff
@@ -495,6 +504,84 @@ def get_file_transfer(self):
...
+        ).format(
+            snap_runtime_loc=snap_runtime_loc,
+            overlay_dir=overlay_dir,
+            work_dir=work_dir,
+        )
+
+        service_content = (
+            textwrap.dedent(
+                """
+            [Unit]
```
kissiel, line 537: this (and the rest) should be indented too I think. Or is it a black thing?

metabox/metabox/core/machine.py
```diff
@@ -495,6 +504,84 @@ def get_file_transfer(self):
...
+            SyslogIdentifier=overlay-checkbox.service
+            Restart=on-failure
+            TimeoutStopSec=30
+            Type=simple
+
+            [Install]
+            WantedBy=multi-user.target
+            """
+            )
+            .lstrip()
```
kissiel, line 553: the lstrip shouldn't be needed (I guess it's the indent problem).

kissiel (commented):
This is an awesome hack. I really hope no one will replicate this is their Checkbox journeys :P

See below for indentation problem

metabox/metabox/core/machine.py
Hook25: black thing I think, I can try to re-run it

kissiel (approved):
awesome! +1

Issue #651, opened 2023-08-04
"environment" section is not consumed (but it worked with checkbog-ng version 2.6~ppa~ubuntu18.04.1)


tai271828:
### Bug Description

"environment" section in a launcher is not consumed by checkbox-ng `2.9.dev52+gc71db97b4~ubuntu18.04.1 `. However it would work with checkbog-ng version `2.6~ppa~ubuntu18.04.1`.

### To Reproduce

1. Install canonical-certification-server from Checkbox dev PPA
2. Run the example launcher by simple `./server-bug-reproducer`[1].

[1] The example launcher, `server-bug-reproducer`:
```
#!/usr/bin/env checkbox-cli

[launcher]
app_id = com.canonical.certification:testlauncher
launcher_version = 1
stock_reports = submission_files

[test plan]
unit = com.canonical.certification::server-regression
forced = yes

[test selection]
exclude = com.canonical.certification::cpu/clocktest com.canonical.certification::cpu/cpufreq_test-server com.canonical.certification::disk/.* com.canonical.certification::memory/.*
forced = yes

[ui]
type = silent
#output = hide

[environment]
#STRESS_NG_DISK_TIME = 30
STRESS_NG_CPU_TIME = 30
# not implemented yet
#STRESS_NG_MEMORY_TIME = 30
```

### Environment

- OS: Ubuntu Bionic
- Checkbox Type: Debian
- Checkbox Version: 2.6~ppa~ubuntu18.04.1
- Hardware being tested: this is a userspace issue and should be across platforms.

### Relevant log output

```shell
I expected log output like this if the variable `STRESS_NG_CPU_TIME` from the environment section of the example launcher is consumed correctly (the the corresponding job will be much shorter):


---------------------[ Stress of CPUs (very long runtime) ]---------------------
ID: com.canonical.certification::stress/cpu_stress_ng_test
Category: com.canonical.plainbox::stress                                                                              
... 8< -------------------------------------------------------------------------
Found STRESS_NG_CPU_TIME env var, stress_ng cpu running time is now: 30 seconds
Estimated total run time is 0 minutes 
                                                           
04 Aug 08:51: Running multiple stress-ng stressors in parallel for 30
seconds...                 
stress-ng: info:  [40498] setting to a 30 second run per stressor                                                                                                                                                                            
[...]

PR #653, opened 2023-08-04
Fix config options are now case insensitive

Hook25:
## Description

The config refactor made all config options case insensitive. This is not a huge problem for most of them but environment variables can not work with this change (given that all options are flattened to lower case and envvars are case sensitive).

#### Example:
GIven the following launcher:
```
[environment]
ABC=123
```
This test case will fail:
```
id: config-environ-case
flags: simple
command: [ $ABC ]
environ: ABC
```

The reason is that the launcher will try to supply a variable named `abc` and not `ABC`

## Resolved issues

Resolves: https://github.com/canonical/checkbox/issues/651

## Documentation

The change seems a bit strange so I have added a small comment above it. Note that it is what the configparser documentation suggests: [ConfigParser.optionxform](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.optionxform)

## Tests

This also adds a metabox test to verify/non-regress this behaviour. Note that this was not caught before because all `environment` section used in tests defined lower case variables! 

To run the new test use the following:
```
(metabox) $ metabox --tag CheckboxConfEnvvarCaseSensitive -- remote-source-22.04.py
```

### DIFF
--- a/checkbox-ng/plainbox/impl/config.py
+++ b/checkbox-ng/plainbox/impl/config.py
@@ -240,6 +240,9 @@ class Configuration:
         """
         cfg = Configuration(origin)
         parser = ConfigParser(delimiters="=")
+        # make the option names case sensitive
+        # else envvars are broken
+        parser.optionxform = str
         parser.read_string(ini_file.read())
         for sect_name, section in parser.items():
             if sect_name == "DEFAULT":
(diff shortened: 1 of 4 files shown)


pieqq (approved):
+1

Thanks a lot for the fix, for the detailed explanation and for adding a Metabox test to prevent any regressions in the future!

PR #656, opened 2023-08-09
Add section in the release process about sending release e-mails

pieqq:
Part of the release process is about announcing the new release to the right people. Since we have a public mailing list and an internal one, both should receive an e-mail when a new release is available.

### DIFF
--- a/tools/release/README.md
+++ b/tools/release/README.md
@@ -25,7 +25,7 @@ Then, it's time to build the new beta version.
   ```
 - Tag the release
   ```
-  git tag -s "v2.4" -m "Bump version: 2.3 → 2.4"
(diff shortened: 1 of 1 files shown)


Hook25:
I remember when we went over this: one should edit the changelog to make new features more visible and minor changes less visible (as not everything as the same weight). I would add also that to the description

pieqq:
> I would add also that to the description

Done.

I also added some information about the Snap Update Verification process, and promotion to Candidate. @jocave I would appreciate if you could have a look at [this commit specifically](https://github.com/canonical/checkbox/pull/656/commits/98beed1509a335838899c2848753ea0e27b02811?short_path=16b686e#diff-16b686e8573494fa8c9b648df12db661c64e83e9c997448ec2b9a83119fd7d05).

Hook25 (approved):
+1, nice

Issue #648, opened 2023-07-31
Cpuid.py  Error When Detecting Raptor Lake cpu


mreed8855:
### Bug Description

Currently the cpuid.py script does not account for the lowercase listing of the Raptor Lake CPU ID.

Cpuid entry raptor lake
"Raptor lake": ['0xB0671', '0xB06F2', '0xB06F5', '0xB06A2'],

HPE cpuid output
CPU Model: Intel(R) Xeon(R) E E-2488
Unable to determine CPU Family for this CPUID: 0xb0671

### To Reproduce

Run the certification testsuite

### Environment

-22.04.2

### Relevant log output

```shell
CPU Model: Intel(R) Xeon(R) E E-2488
Unable to determine CPU Family for this CPUID: 0xb0671
```

PR #660, opened 2023-08-10
Fix for issue #648

mreed8855:
https://github.com/canonical/checkbox/issues/648

On some Raptor Lake CPU's the cpuid is lower case.

## Resolved issues

Fixes SERVCERT-1243


bladernr (approved):
LGTM

PR #661, opened 2023-08-11
Force series16 to install dependencies

Hook25:
## Description

Right now the series16 snap does not install the dependencies. The reason seems to be that the latest version that support python3.5 does not support reading dependencies from pyproject. 

This forces the install of dependencies in the snapcraft.yaml.

Also, in order to detect any funny business in the future with dependencies, this adds a very simple test and metabox scenario that tries to import every dependency. 

## Resolved issues

Checkbox16 2.9.1 is broken
Metabox coverage of dependencies was lacking

## Documentation

This also documents why every step of the install is done in more detail with comments

## Tests

This adds a new metabox test. To run it call metabox with the `VerifyInstallation` tag like this:
```
metabox --tag VerifyInstallation local-source-16.04.py
```

To test a custom built snap create a metabox configuration (snap-16.04.py):
```
configuration = {
    "local": {
        "origin": "classic-snap",
        "checkbox_snap": {"uri": "~/checkbox16.snap"},
        "checkbox_core_snap": {"risk": "edge"},
        "releases": ["xenial"],
    },
}
```
Build the checkbox16 snap and put it in the location you specified above, then run:
```
metabox --tag VerifyInstallation snap-16.04.py
```

### DIFF
--- a/checkbox-core-snap/series16/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series16/snap/snapcraft.yaml
@@ -173,12 +173,21 @@ parts:
     override-build: |
       # this is a bodge to be backward compatible
       snapcraftctl build
-      # also use build to ensure install (pip is not compinat to pyproject)
-      # on this version
+      # setup.py is necessary for this old pip verision
       echo "from setuptools import setup; setup()" > setup.py
+      # update pip and setuptools (+deps) because the directive
+      # is ignored in pyproject (and pip on xenial is preinstalled 8.x)
       $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install "pip<21" "setuptools<48" "setuptools_scm[toml]>=3.4" "importlib_metadata==1.0.0" "zipp<2"
-      # also, call pip again, this installs the module itself
+      # install the dependencies of checkbox-support itself because this
+      # pip version  pyproject.toml dependencies
+      $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install "pyparsing<3.0.0" "requests<2.26.0" "distro<1.7.0" "requests_unixsocket<=0.3.0" "importlib_metadata<=1.0.0"
+      # call pip again to actually install the module itself
       $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install .
+      # fix shebangs in bins.This is necessary because by calling pip via the
+      # $SNAPCRAFT_PART_INSTALL path (to get the correct one) this path gets
+      # propagated into bins. This path will be invalid once the snap is installed
+      # so it must be changed to the appropriate one (/usr/bin/env python3) that
+      # we could not use here
       sed -iE 's|\#\!.*/usr/bin/python3$|#!/usr/bin/env python3|' $SNAPCRAFT_PART_INSTALL/usr/local/bin/*
 ################################################################################
   checkbox-ng:
@@ -197,6 +206,7 @@ parts:
       - python3-urwid
       - python3-xlsxwriter
       - python3-setuptools
+      - python3-dev
     python-packages:
       - tqdm
     after: [checkbox-support]
@@ -206,12 +216,15 @@ parts:
     override-build: |
       # this is a bodge to be backward compatible
       snapcraftctl build
-      # also use build to ensure install (pip is not compinat to pyproject)
-      # on this version
+      # setup.py is necessary for this old pip verision
       echo "from setuptools import setup; setup()" > setup.py
-      # setup the correct pip and related tools version
+      # update pip and setuptools (+deps) because the directive
+      # is ignored in pyproject (and pip on xenial is preinstalled 8.x)
       $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install "pip<21" "setuptools<48" "setuptools_scm[toml]>=3.4" "importlib_metadata==1.0.0" "zipp<2"
-      # install checkbox-ng
+      # install the dependencies of checkbox-support itself because this
+      # pip version  pyproject.toml dependencies
+      $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install "packaging<21.0" "psutil<=5.9.5" "requests<2.26.0" "urwid<=2.1.2" "Jinja2<=2.11.3" "XlsxWriter<=3.0.3" "tqdm<4.65.0" "importlib_metadata<=1.0.0"
+      # call pip again to actually install the module itself
       $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install .
       # fix shebangs in bins.This is necessary because by calling pip via the
       # $SNAPCRAFT_PART_INSTALL path (to get the correct one) this path gets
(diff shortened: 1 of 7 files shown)


metabox/metabox/metabox-provider/units/basic-tps.pxu
```diff
@@ -45,3 +45,9 @@ include:
  config-environ-source
  config-environ-vars
  config-environ-case
+
+unit: test plan
+id: installation_verify
```
kissiel, line 50: I don't like how the thing evolves on each layer.
All starts with "dependency-import" then it becomes a "dependency-test.py", then it becomes "dependency_installation" job and is included in "Verify Checkbox installation" test plan that has an id "installation-verify".  Everything that's done is the original "dependency-import" test. Introducing those intermediate names is an unnecessary burden for the reader I think.

metabox/metabox/core/machine.py
```diff
@@ -286,6 +286,17 @@ def _get_install_dependencies_cmds(self):
             # Use <21 to get the latest 20 as 21+ is not supported here
             return [
                 "bash -c 'sudo python3 -m pip install -U \"pip<21\"'",
+                (
+                    # This is here because this pip version does not support
+                    # installing dependencies from pyproject.toml
+                    "bash -c 'sudo python3 -m pip install "
```
kissiel, line 292: 1. This cannot be done this way. I know that for you it'll be easy to understand why it's here and how to extend this with time, but for anybody else it will be a hassle.
2. The list is different to the one in pyproject.toml, kind of defeating the whole point of this PR. We want to verify if all the deps are there in "production" so for this test to make sense this should do the same thing as "production" build would do. I think this is not the case here.

The first two ways to solve this that popped up in my brain are:
1. metabox has it's own python environment, that is not ancient, thus it knows how to read the deps from the toml. This can be forwarded by MB itself.
2.
 ```import toml; print('\n'.join(toml.load("pyproject.toml").get('project', {}).get('dependencies', {})))```

metabox/metabox/scenarios/installation/validation.py
```diff
@@ -0,0 +1,41 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+import textwrap
+
+from metabox.core.actions import AssertPrinted, Start
+from metabox.core.scenario import Scenario
+
+
```
kissiel, line 24: @tag("dependencies", "installation")

kissiel (changes requested):
The whole premise is great! I absolutely love it and we **NEED** this. Having said that this IMHO requires some rework.
All the deps in this test should be handled the same way as there are when building checkbox, so we actually test the thing. This also (heavily) applies to transitive dependencies. And the version locks.

I think embedding this test in a test plan is unnecessary. This can be *just a job* and I'd use `checkbox-cli -run`
 I also think those import from within a job run by Checkbox is not the way the modules would be used. The jobs don't import Checkbox's logic. They do import `checkbox-support`'s logic.

To test if **Checkbox** can import modules it should have this need to be tested in the same environment that checkbox would run. We facilitated debugging of similar issues in the past by providing the `shell` entry point to the snap. It has (or at least should have :) ) the same environment as `checkbox-cli` would have.

Lastly, does the provided example `metabox --tag VerifyInstallation local-source-16.04.py` even work? `VerifyInstallaiton` is not a tag. Is the scenario class name automatically appended to the list of tags the scenario has?

checkbox-core-snap/series16/snap/snapcraft.yaml
```diff
@@ -173,12 +173,21 @@ parts:
     override-build: |
       # this is a bodge to be backward compatible
       snapcraftctl build
-      # also use build to ensure install (pip is not compinat to pyproject)
-      # on this version
+      # setup.py is necessary for this old pip verision
```
pieqq, line 176: typo: `verision`

checkbox-core-snap/series16/snap/snapcraft.yaml
```diff
@@ -206,20 +216,22 @@ parts:
...
```
pieqq, line 219: typo: `verision`.

pieqq (changes requested):
Maciek already covered a lot. I'm just the annoying typo guy here. :D

metabox/metabox/metabox-provider/units/basic-tps.pxu
Hook25: Ok the dependency dance (import/verify etc.) is not ok, changing that.

The test plan is there so that if we come up with some new way of verifying the installation we can just add to that test plan and the scenario will automatically run any new verification. We may want to verify the environment of snap for example? Something like that.

metabox/metabox/metabox-provider/bin/dependency_test.py
```diff
@@ -0,0 +1,64 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+"""
+This module imports all dependencies of checkbox-ng and checkbox-support.
```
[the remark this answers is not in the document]
Hook25: This was done this way because I did not want to create a weird machinery to read the pyproject file and extract the dependencies at runtime. That would be possibly harder to read although possible for sure. The manually updating the file part is due to this design, if you want I can make it automatic but it may be more trouble than it is worth

metabox/metabox/scenarios/installation/validation.py
Hook25: ok, to clarify your comment above, yes, every test is automatically "tagged" with its name

metabox/metabox/core/machine.py
Hook25: The list content is (should, I have double checked this but I may still be wrong) the same as the pyproject.toml files (I have merged them into one, checkbox-support + checkbox-ng).

The forced versions are sadly here because there is no good automated (sane) way of getting something like this out of pip. If you tell pip to install a package it will always install the latest version that is compatible with your constraint. For example we say `Jinja2>=2.7` and that will make pip install any version above the one we have asked (including `3.1.2`, which is not compatible with python3.5).

kissiel: Right! So let's use python environment that perfectly replicates this, i.e. python3.5-pip.
By doing any intelligent work expanding this list yourself we're cutting away the value of the whole story.

metabox/metabox/metabox-provider/units/basic-tps.pxu
Hook25: but the modification to use the test plan is to edit the scenario and create the test plan, while with this we have 2 lines more and we get the same result, it is a minuscule amount of code for a pretty nice future result

metabox/metabox/metabox-provider/bin/dependency_test.py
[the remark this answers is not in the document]
Hook25: Reporting Meet discussion: We agreed that the situation is suboptimal, but making this code clean and orderly is well outside the scope ot this pr and very hard, so this will have to suffice for the time being

metabox/metabox/core/machine.py
Hook25: Reporting Meet discussion: As above, we agreed that cleaning this up is possibly very complicated and outside the scope of this pr

pieqq (approved):
I understand that you had a long discussion with Maciek, so we should be all good now. Thanks!

pieqq:
(we have to fix those chain scenarios...)

PR #663, opened 2023-08-14
Add: snaps confinement test (new)

patliuu:
## Resolved issues
Jira task https://warthogs.atlassian.net/browse/CQT-2907

## Rational
Murcia project's customer asked us to include 2 snaps in the GM image. One of the snap is `strict` confined and installed in devmode, the other one is `devmode` confined and also installed in devmode.[1]

Recently this problem was found. For security reason, this should not be allowed on the image we delivered.
So we were asked to create a test case to catch this in the future.

[1] snap confinement document https://snapcraft.io/docs/snap-confinement

## Information
We originally have a test case called `snappy/test-snap-confinement-mode`. It checks the confinement and sandbox features of the system, if the system confinement is not `strict`, the test case will print out the missing requirements (e.g. sandbox features or cgroup version). Therefore, this test is testing the confinement of the system. We don't actually test the snap confinements of each installed snap.

In the problematic image of Murcia project, there are 2 snaps sideloaded at the last stage of image installation.

Name | Version | Revision |   | Confinement | Devmode |   |  
-- | -- | -- | -- | -- | -- | -- | --
gcap | 1.0.0 | x1 |   | strict | True |   | July 14, 2023
litebmc-fwupd | 2.6 | x1 |   | devmode | True |   | July 14, 2023

## Requirement
After checking with OEM SWE team, the requriments of this test are checking the following things for each installed snap,
1. Check the revision should NOT be sideloaded like `x1`.
2. Check the confinement should be `strict`. `classic` and `devmode` confinement are not allowed.
3. Check the `installed in devmode` should be `False`.
4. This test should be executed on all desktop, server, and core images.

## What are implemented in this pull request?
1. Refactor the original script. Splitting them to `SystemConfinement` class and `SnapsConfinement` class corresponding to the system confinement test and snap confinement test(the new test created)
2. Whitelisted the snaps that are installed in devmode for testing. Including `bugit`, `checkboxXX`, `mir-test-tools`, `graphic-test-tools`.
3. Implement the requirements above.
4. Add `snappy/test-system-confinement` and `snappy/test-snaps-confinement` to snappy-snap-automated test plan.
5. Add snappy-snap-automated test plan to desktop test plans. As we are thinking desktop image should also test the snap features including install, refresh, revert ...etc.

## Sideload test results
* Tested on a Ubuntu Desktop 22.04 laptop
https://pastebin.canonical.com/p/hsyFyRVtVQ/
* Tested on a Ubuntu Core 22 device with sideloading stress-ng snap
https://pastebin.canonical.com/p/CngKh9fNyD/
* Tested on that problematic machine (Desktop image on IoT device)
https://pastebin.canonical.com/p/gT8WNzbjbt/

### DIFF
--- a/providers/base/bin/snap_confinement_test.py
+++ b/providers/base/bin/snap_confinement_test.py
@@ -105,15 +105,21 @@ class SnapsConfinement:
     Test the confinement status of all installed snaps.
 
     Attributes:
-        whitelist_snaps (list): A list of snap names or regex patterns
-            that are exempted from the confinement check.
+        allowlist_snaps (list): A list of snap names or regex patterns
+            that are exempted from the confinement check. To match the
+            entire snap name, use the pattern "^<snap_name>$". For
+            example, "bugit" matches only "bugit". To match multiple
+            snaps with similar names, use ".*" suffixes. For instance,
+            "checkbox.*" matches all snap names starting with "checkbox".
+            Customize this list to exclude specific snaps from the
+            confinement checks based on their names or patterns.
     """
 
-    whitelist_snaps = [
-        "bugit",
+    allowlist_snaps = [
+        r"^bugit$",
         r"checkbox.*",
-        "mir-test-tools",
-        "graphics-test-tools",
+        r"^mir-test-tools$",
+        r"^graphics-test-tools$",
     ]
 
     def invoked(self):
@@ -138,10 +144,11 @@ class SnapsConfinement:
             if snap_name is None:
                 logging.error("Snap 'name' not found in the snap data.")
                 exit_code = 1
+                continue  # Skipping following checks if snap_name not found
 
             if any(
                 re.match(pattern, snap_name)
-                for pattern in self.whitelist_snaps
+                for pattern in self.allowlist_snaps
             ):
                 print("Skipping whitelisted snap: {}".format(snap_name))
                 continue
@@ -182,7 +189,7 @@ def main():
     }
     parser = argparse.ArgumentParser()
     parser.add_argument("subcommand", type=str, choices=sub_commands)
-    args = parser.parse_args(sys.argv[1:2])
+    args = parser.parse_args()
     return sub_commands[args.subcommand]().invoked()
 
 
--- a/providers/base/bin/snap_confinement_test.py
+++ b/providers/base/bin/snap_confinement_test.py
@@ -20,6 +20,7 @@
 import argparse
 import json
 import logging
+import re
 import sys
 from checkbox_support.snap_utils.snapd import Snapd
 
@@ -99,11 +100,85 @@ class SystemConfinement:
         return sandbox_features_output
 
 
+class SnapsConfinement:
+    """
+    Test the confinement status of all installed snaps.
+
+    Attributes:
+        whitelist_snaps (list): A list of snap names or regex patterns
+            that are exempted from the confinement check.
+    """
+
+    whitelist_snaps = [
+        "bugit",
+        r"checkbox.*",
+        "mir-test-tools",
+        "graphics-test-tools",
+    ]
+
+    def invoked(self):
+        """
+        Check the confinement of all snaps installed in the system.
+
+        A snap confinement should be 'strict', devmode should be False,
+        and should not have a sideloded revision starts with 'x'.
+
+        Returns:
+            int: Exit code. 0 if the test passes for all snaps,
+            otherwise 1.
+        """
+        data = Snapd().list()
+        exit_code = 0
+        for snap in data:
+            snap_name = snap.get("name")
+            snap_confinement = snap.get("confinement")
+            snap_devmode = snap.get("devmode")
+            snap_revision = snap.get("revision")
+
+            if snap_name is None:
+                logging.error("Snap 'name' not found in the snap data.")
+                exit_code = 1
+
+            if any(
+                re.match(pattern, snap_name)
+                for pattern in self.whitelist_snaps
+            ):
+                print("Skipping whitelisted snap: {}".format(snap_name))
+                continue
+
+            if snap_confinement != "strict":
+                exit_code = 1
+                logging.error(
+                    "Snap '%s' confinement is expected to be 'strict' "
+                    "but got '%s'", snap_name, snap_confinement,
+                )
+
+            if snap_devmode is not False:
+                exit_code = 1
+                logging.error(
+                    "Snap '%s' devmode is expected to be False but "
+                    "got '%s'", snap_name, snap_devmode,
+                )
+
+            if snap_revision and snap_revision.startswith("x"):
+                exit_code = 1
+                logging.error(
+                    "Snap '%s' has sideloaded revision '%s', which "
+                    "is not allowed", snap_name, snap_revision,
+                )
+            elif snap_revision is None:
+                exit_code = 1
+                logging.error(
+                    "'revision' not found in snap '%s'", snap_name,
+                )
+        return exit_code
+
+
 def main():
     logging.basicConfig(format='%(levelname)s: %(message)s')
     sub_commands = {
         "system": SystemConfinement,
-        # "snaps": SnapsConfinement,
+        "snaps": SnapsConfinement,
     }
     parser = argparse.ArgumentParser()
     parser.add_argument("subcommand", type=str, choices=sub_commands)
--- a/providers/base/bin/snap_confinement_test.py
+++ b/providers/base/bin/snap_confinement_test.py
@@ -56,7 +56,7 @@ class SystemConfinement:
 
         Returns:
             str: A detailed output of the system's confinement and
-                sandbox features in JSON format.
+                 sandbox features in JSON format.
         """
         data = Snapd().get_system_info()
 
@@ -125,7 +125,7 @@ class SnapsConfinement:
 
         Returns:
             int: Exit code. 0 if the test passes for all snaps,
-            otherwise 1.
+                 otherwise 1.
         """
         data = Snapd().list()
         exit_code = 0
--- a/providers/base/bin/snap_confinement_test.py
+++ b/providers/base/bin/snap_confinement_test.py
@@ -25,13 +25,20 @@ import sys
 from checkbox_support.snap_utils.snapd import Snapd
 
 
-class SystemConfinement:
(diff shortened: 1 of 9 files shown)


providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+class SnapsConfinement:
+    """
+    Test the confinement status of all installed snaps.
+
+    Attributes:
+        whitelist_snaps (list): A list of snap names or regex patterns
+            that are exempted from the confinement check.
+    """
+
+    whitelist_snaps = [
```
kissiel, line 112: we're following modern naming schemes in which "allowlist" is preferred.

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+
+        return sandbox_features_output
+
+
+class SnapsConfinement:
+    """
+    Test the confinement status of all installed snaps.
+
+    Attributes:
+        whitelist_snaps (list): A list of snap names or regex patterns
```
kissiel, line 108: we're following modern naming schemes in which "allowlist" is preferred.

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+        """
+        data = Snapd().list()
+        exit_code = 0
+        for snap in data:
+            snap_name = snap.get("name")
+            snap_confinement = snap.get("confinement")
+            snap_devmode = snap.get("devmode")
+            snap_revision = snap.get("revision")
+
+            if snap_name is None:
```
kissiel, line 138: I'm thinking that if we can't get the snap name, there's no point in proceeding with the other `if`s.

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+                logging.error("Snap 'name' not found in the snap data.")
+                exit_code = 1
```
kissiel, line 140: so shouldn't we `continue` iterating over snaps here?

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+
+            if any(
+                re.match(pattern, snap_name)
```
kissiel, line 143: RE always introduces some difficulty and complexity. In this case `snap_name.startswith(pattern)` would suffice (granted that the `.*` is removed from the patttern)

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
-    "kernel:file",
-    "kernel:mount",
-    "kernel:namespaces",
-    "kernel:network",
-    "kernel:ptrace",
-    "kernel:signal",
-    "parser:unsafe",
-]
 
+class SystemConfinement:
```
kissiel, line 28: Both classes in this testing program have only one method and they don't benefit from being a class. Simple function would be cleaner (for instance the indentation would be better allowing cleaner code).

Furthermore the function could be decomposed further to facilitate unit testing. For instance the "check_snap_confinment" could accept snap's data as an argument, and do the checks. Then in the main scope in the  iterating over snaps could be done and this function could be invoked for each.
In turn the unit test becomes simple - the snap data would have to be prepared with the conditions you want, and so on...

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+
+def main():
+    logging.basicConfig(format='%(levelname)s: %(message)s')
+    sub_commands = {
+        "system": SystemConfinement,
+        "snaps": SnapsConfinement,
+    }
+    parser = argparse.ArgumentParser()
+    parser.add_argument("subcommand", type=str, choices=sub_commands)
+    args = parser.parse_args(sys.argv[1:2])
```
kissiel, line 185: Any specific reason for parsing only one string here?
I'd guess that doing a general `parser.parse_args()` would achieve the same in normal cases **plus** it would complain when someone would provide some extra things. Those extra things would be silently ignored with the `[1:2]` approach.

providers/base/bin/snap_confinement_test.py
```diff
@@ -17,64 +17,173 @@
...
+                exit_code = 1
+                logging.error(
+                    "'revision' not found in snap '%s'", snap_name,
+                )
+        return exit_code
+
+
+def main():
+    logging.basicConfig(format='%(levelname)s: %(message)s')
+    sub_commands = {
```
kissiel, line 179: Fantastic, very Pythonic approach, <3

kissiel (changes requested):
Fantastic stuff!
The code below is quite testable, but there's no tests. I added some comments below that could be helpful with that.
See other comments for other tweaks.

providers/base/bin/snap_confinement_test.py
patliuu: Good suggestion! I'll modify the naming in the next commit.

providers/base/bin/snap_confinement_test.py
patliuu: You're right. I'll apply what you said in the next commit.

providers/base/bin/snap_confinement_test.py
patliuu: The reason that I didn't use `startswith()` was because I was thinking there might be a snap named `bugit_abc`, this kind of name will also match the `startswith()` and the checking will be skipped.

Only checkboxXXXX will be allowed since we have lots of checkbox-{projects, providers} names. Others should match the whole word.

providers/base/bin/snap_confinement_test.py
patliuu: Indeed, I'll apply what you suggest in the next commit :smile:

providers/base/bin/snap_confinement_test.py
patliuu: Honestly, I was referring to the structure [snappy tests](https://github.com/canonical/checkbox/blob/main/providers/base/bin/snap_tests.py).

I think the idea is to separate the test cases by classes, and the data like `features_should_include` or `allowlist_snaps` could be located in that class. Re-write it to only functions are OK, but I think both ways have their own advantages.

Regarding decomposing the functions to apply unit testing, there are 2 things I'd like to discuss with you.
1. For this test case, I feel the unit tests could be implemented for functions in [snapd.py](https://github.com/canonical/checkbox/blob/main/checkbox-support/checkbox_support/snap_utils/snapd.py). As this test is calling the functions in snapd.py and get data in the dictionaries, it seems to me that unit tests for this test case are not necessary.
2. In my opinion, wrapping those `if conditions` to functions or not are fine. If the functions will not be reused or extended, keeping these simple `if conditions` will be easier to understand the test steps. If we are going to wrap them in functions like `check_snap_confinement()` or `check_snap_demode()`, it will be better to keep the class structure and implement these functions in classes. Something like,
```python3
class SnapsConfinement:
    def __init__(self):
        data = Snapd.list()

    def check_snap_confinement(self, snap, snap_name):
        confinement = snap.get("confinement")
        if confinement != "strict":
            logging.error(
                "Snap '%s' confinement is expected to be 'strict' "
                "but got '%s'", snap_name, confinement,
            )
            return False
        return True
```
And iterate the snaps `self.data` and get the snap name in `invoked()` functions.

Which way do you prefer?

providers/base/bin/snap_confinement_test.py
kissiel: but if i'm not mistaken, checking for `bugit` in a string `bugit_abc` will also match...

kissiel: I checked.
```
python3 -c 'import re; print(re.match("foo", "foobar"))'
<re.Match object; span=(0, 3), match='foo'>
```

kissiel (commented):
Quick answer about the RE below. I'll answer more soon.

providers/base/bin/snap_confinement_test.py
patliuu: Ahhh, you're right. Thanks for correcting me.
Then, would it be OK to modify the `allowlist_snaps` to
```
whitelist_snaps = [
    r"^bugit$",
    r"checkbox.*",
    r"^mir-test-tools$",
    r"^graphics-test-tools$",
]
```
To check the whole string matches? or do you still prefer using `startswith()`?

kissiel: I think your idea of being strict about matching the names is great. I just noted in the previous version that what was really happening, was equivalent to .startswith().
This is much better, I'm just gonna ask you, if I may, to write a comment with an explanation on |what it does/why we do this|.

patliuu: Sure! I'll add comments to explain why these regex are added.

patliuu: Added explanation into the docstring.

patliuu:
Hi @kissiel ,
I've applied the suggestion except for the discussion of `using classes or not` and `wapping "if conditions" into other functions or not`.

providers/base/bin/snap_confinement_test.py
kissiel: >keeping these simple if conditions will be easier to understand the test steps.

I agree with this statement.

If what you want to have is a class that reflects one idea of a test, like check "check snaps' confinement", then I think natural decomposition would be to iterate over the snaps in `invoked` and call the `is_snap_confined` check for each (btw. I changed the name on purpose - for checks that end in booleans the is_something is clearer tbh).
I hope this answers your question.

patliuu: Thank you for the clarification.

I rethought it a little bit. Since these 2 test classes are simple, I would prefer just to flatten them into functions `test_system_confinement()` and `test_snaps_confinement()`. And the attributes `features_should_include` and `allowlist_snaps` for each class become a local variable in each function.

To keep the test steps easy to understand, I would leave the `if conditions` as they are.
Feel free to give any comments in case I misunderstand anything.

kissiel (approved):
Looks great! +1
Thank you for all the changes!

I really hope that the CI will not choke on the `test_` function prefix (it may look like a unit test ;) )

PR #672, opened 2023-08-21
Enforce coverage requirement and reporting via codecov

Hook25:
## Description

This upgrades our coverage reporting by uploading to codecov.io. Using the features of said service, this also starts to enforce a patch based coverage requirement (for now just hardcoded at 80%).

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-799

## Documentation

This adds a new section to the contributing guide that explains how to get a local coverage report, why we want a coverage constraint and how we enforce it. This section also points to where the coverage constraint is specified and how it is reported io PRs (the pointer is there because in the future we may want to change the requirement, we don't want to have to update this file with the new one!)

## Tests

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ Note that every part of this repository has a `.coveragerc` file, they should
 already include anything you may want to see in the report. If something is
 missing you can edit it but please, consult with the team before doing so.
 Tests are intentionally excluded from the coverage report, this is because
-test files tend to inflationate the coverage with no real benefit, so don't
+test files tend to inflate the coverage with no real benefit, so don't
 worry if you can not spot yours in the report.
 
 Of course, you may only be interested in the coverage of your patch (for
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,15 +72,15 @@ you have to activate the `checkbox-cli service` on the Machine under test:
 ```bash
 (venv) # checkbox-cli service
 ```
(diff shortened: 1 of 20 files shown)


Hook25:
The codecov message works :) the issue it is pointing at is due to the fact that he does not know how much coverage we have in main yet

CONTRIBUTING.md
```diff
@@ -103,6 +103,40 @@ tests for providers:
...
+```
+$ python -m pip install coverage pytest pytest-cov
+# cd to where your test is
+$ python -m coverage run -m pytest .
+```
+Note that every part of this repository has a `.coveragerc` file, they should
+already include anything you may want to see in the report. If something is
+missing you can edit it but please, consult with the team before doing so.
+Tests are intentionally excluded from the coverage report, this is because
+test files tend to inflationate the coverage with no real benefit, so don't
```
kissiel, line 128: s/inflationate/inflate

codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  status:
+    patch:
+      default:
+        target: 80%
```
kissiel, line 5: What's the source of this value?

kissiel (commented):
One suggestion and one question below.

codecov.yml
Hook25: uhm seems like the card was edited? pretty sure I took the card suggestion on the coverage and copy-pasted the snippet. I will raise it to 90%

kissiel (approved):
Looks good!
+1

PR #675, opened 2023-08-22
New plainbox unittests (>0%)

Hook25:
## Description

This PR aims to make every module in plainbox at least imported by the unittests. This allows tox, pytest and any other static analysis tool we may want to start using to analyze them.

I have used this occasion to try to produce some reasonable unittests, so that we get also some value more than just a plain include. 

## Resolved issues

This is part of: https://warthogs.atlassian.net/browse/CHECKBOX-740

## Documentation

N/A

## Tests

To run the tests run the following
```
$ cd checkbox/checkbox-ng/plainbox
$ pytest . -vvv
```

### DIFF
--- a/checkbox-ng/plainbox/impl/commands/test_cmd_session.py
+++ b/checkbox-ng/plainbox/impl/commands/test_cmd_session.py
@@ -1,0 +1,34 @@
+# This file is part of Checkbox.
+#
+# Copyright 2023 Canonical Ltd.
+# Written by:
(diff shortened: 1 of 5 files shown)


kissiel (approved):
Awesome stuff! +1
Is this really everything that was needed to import all modules while testing?

Hook25:
this imports everything in plainbox but `plainbox/impl/box.py` because it is more trouble than it is worth fixing it I'm worried.

PR #676, opened 2023-08-23
Add daily process to tag edge releases

pieqq:
## Description

`edge_tag.py` generates an edge release tag based on the latest release tag and the number of commits since, and pushes it.

For instance, if the last release is "v2.9.1", and there have been 6 new commits since, the script will compute tag "v2.9.2-edge+6", tag the latest commit in HEAD with it, and push the tags.

## Resolved issues

Fix CHECKBOX-807

## Documentation

N/A

## Tests

Added a very basic doctest for one of the functions. The scripts in `/tools/` are not reached by pytest I believe...

### DIFF
--- a/.github/workflows/daily-edge-tag.yml
+++ b/.github/workflows/daily-edge-tag.yml
@@ -1,0 +1,19 @@
+name: Daily tag commit for edge release
+
+on:
+  schedule:
+    - cron: '00 04 * * *'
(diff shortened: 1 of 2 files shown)


pieqq:
After discussing with the rest of the team, and analyzing the workflow we want and the triggers and tooling we already have in place, this change is not required anymore.

PR #678, opened 2023-08-23
New checkbox-ng unittests (>0%)

Hook25:
## Description

There are some files in checkbox-ng that are not even invoked in test runs via tox. This leads to several problems, like files with unsupported syntax going through the tox check due to them not being imported anywhere.

This creates a unittest per file who was missing any coverage. Additionally, this greately increases the overall coverage of the checkbox-ng module, given that the new unit tests actually cover the most important function of all the files (mostly `invoke` or `main`).

> Note: After this PR all files in checkbox-ng must have a coverage >0%

## Resolved issues

Along with https://github.com/canonical/checkbox/pull/675 this
Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-740

## Documentation

N/A

## Tests

To run the new tests do the following:
```
$ cd checkbox/checkbox-ng/checkbox_ng
$ pytest . -vvv
```

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/test_check_config.py
+++ b/checkbox-ng/checkbox_ng/launcher/test_check_config.py
@@ -1,0 +1,51 @@
+# This file is part of Checkbox.
+#
+# Copyright 2023 Canonical Ltd.
+# Written by:
+#   Massimiliano Girardi <massimiliano.girardi@canonical.com>
+#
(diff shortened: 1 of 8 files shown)


kissiel (approved):
Those chad-size patch list is a clear signal that the code needs a refactor or two (billions).
The thing that's here looks good! +1

PR #680, opened 2023-08-23
Carryover codecov coverage

Hook25:
## Description

Given that we have a monorepo, we have several "parts" that are independent in the same spot. Every time a change is pushed to a pr or main, if a part was changed, it is tested. Once a part is tested the coverage results are pushed to codecov.io. 

This fixes this issue by enabling carryover and flagging the parts. Carryover basically means that if a flag (part) is not updated in a submission, the previous value is used instead. Flags are how codecov identifies the parts, this creates one per part + one for all `unittests`

## Resolved issues

The issue that we are having is that if a part is not tested in a submission (for example, because it was not changed), codecov "forgets" it existed right now, leading to the coverage values that are not reflecting of reality.

## Documentation

This adds a new flag per part, this should make the output more readable without any need for additional documentation.

## Tests

See the codecov bot comment to get to the dashboard, it should now have a "flags" section that lists the coverage of each part

### DIFF
--- a/.github/workflows/tox-checkbox-ng.yaml
+++ b/.github/workflows/tox-checkbox-ng.yaml
@@ -32,3 +32,6 @@ jobs:
         run: tox
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          flags: unittests,checkbox-ng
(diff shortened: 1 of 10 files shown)


Hook25:
The drop in coverage of this PR is "fake", last run on main only executed checkbox-ng tests, and due to the missing carryover, now codecov thinks that we have a 5x% coverage on main

kissiel (changes requested):
I think that for providers the "provider-base" would be much better than just "base".

kissiel (approved):
Awesome, +1

Issue #659, opened 2023-08-10
Confusing error message: "Remote API version mismatch. Service uses: 11. Remote uses: 12"


Saviq:
### Bug Description

When your CLI and remote use incompatible API versions, Checkbox complains:

```
Remote API version mismatch. Service uses: 11. Remote uses: 12
```

But what is Service? What is Remote here?

### To Reproduce

Install checkbox22 2.9 on the host; and 2.8 on the… remote… DUT.

`checkbox-cli remote <IP>`

### Environment

N/A

### Relevant log output

```shell
Remote API version mismatch. Service uses: 11. Remote uses: 12
```

Hook25:
We agree that the error message is unclear, we are working on making it clearer! thank you for your report. 

Remote is the checkbox program that you are using (the one you start with checkbox-cli remote ...), service is the one that is running on the device that you are testing. This error basically states, "I have detected that the two versions of checkbox that you are using are incompatible (Remote is newer than Service)"

Also note: we are changing the nomenclature of remote/service to something better soon, that should also make the error message more comprehensible overall

PR #687, opened 2023-08-28
Make the remote api mismatch message clearer

Hook25:
## Description

As a step to make the release (aftermath) smoother, we have to improve how errors are communicated to the users. Checkbox remote does not start if both agent and controller are compatible but the current message is not very clear. 

This updates that message with hopefully clearer wording and a link to the docs.

The new error message looks like this:
```
The controller that you are using is newer than the agent you are trying to connect to.
To solve this, upgrade the agent to the controller version.
If you are unsure about the nomenclature see:
https://checkbox.readthedocs.io/en/latest/reference/glossary.html

Error: (Agent version: 11, Controller version 13)
```

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/659
Part of: https://warthogs.atlassian.net/browse/CHECKBOX-820

## Documentation

This separates the check into a function with a new docstring to explain what the purpose of it is

## Tests

This move the checking into a new function and tests it via two new unit tests. To run them:
```
$ cd checkbox_ng/launcher
$ pytest test_master.py
```

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/master.py
+++ b/checkbox-ng/checkbox_ng/launcher/master.py
@@ -180,7 +180,7 @@ class RemoteMaster(ReportsStage, MainLoopStage):
         #  check if ever disconnected
         ever_disconnected = False
         #  this to animate the dash
-        spinner = itertools.cycle('-\\|/')
+        spinner = itertools.cycle("-\\|/")
         #  this tracks the disconnection time
         disconnection_time = 0
         while True:
@@ -235,12 +235,22 @@ class RemoteMaster(ReportsStage, MainLoopStage):
                     )
                 master_api_version = RemoteSessionAssistant.REMOTE_API_VERSION
                 if slave_api_version != master_api_version:
-                    raise SystemExit(
-                        _(
-                            "Remote API version mismatch. Service "
-                            "uses: {}. Remote uses: {}"
-                        ).format(slave_api_version, master_api_version)
+                    problem_msg = (
+                        "You are trying to connect to an incompatible "
+                        "checkbox agent! To solve this, upgrade the {} to the "
+                        "{} version. (Agent version: {}, Controller "
+                        "version {})"
                     )
+                    if master_api_version > slave_api_version:
+                        solution = ("agent", "controller")
+                    else:
+                        solution = ("controller", "agent")
+
+                    problem_msg = problem_msg.format(
+                        *solution, slave_api_version, master_api_version
+                    )
+
+                    raise SystemExit(_(problem_msg))
                 state, payload = self.sa.whats_up()
                 _logger.info("remote: Main dispatch with state: %s", state)
                 if printed_reconnecting and ever_disconnected:
@@ -297,7 +307,7 @@ class RemoteMaster(ReportsStage, MainLoopStage):
                     ever_disconnected = True
                     printed_reconnecting = True
                 print(next(spinner), end="\b", flush=True)
-                time.sleep(.25)
+                time.sleep(0.25)
             except KeyboardInterrupt:
                 interrupted = True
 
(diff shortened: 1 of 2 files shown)


kissiel (changes requested):

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -166,6 +166,41 @@ def invoked(self, ctx):
...
+    def check_remote_api_match(self):
+        """
+        Check that agent and controller are running on the same
+        REMOTE_API_VERSION else exit checkbox with an error
+        """
+        agent_api_version = self.sa.get_remote_api_version()
+        controller_api_version = RemoteSessionAssistant.REMOTE_API_VERSION
+        if agent_api_version == controller_api_version:
+            return
+        template_msg = (
```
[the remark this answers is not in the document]
Hook25: I think it is reasonable, it is quite a long message with just small variations, it would be very ugly to have two separate messages.

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -166,6 +166,41 @@ def invoked(self, ctx):
...
+            "The controller that you are using is {} than the agent "
+            "you are trying to connect to.\n"
+            "To solve this, upgrade the {} to the "
+            "{} version.\n"
+            "If you are unsure about the nomenclature see:\n"
```
mz2, line 183:
```suggestion
            "If you are unsure about the nomenclature or what any of this means, see:\n"
```

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -166,6 +166,41 @@ def invoked(self, ctx):
...
+        )
+        if controller_api_version > agent_api_version:
+            problem = "newer"
+            solution = ("agent", "controller")
+        else:
+            problem = "older"
+            solution = ("controller", "agent")
+        problem_msg = template_msg.format(
+            problem,
+            *solution,
```
kissiel, line 196: so this implicitly has to match the template in number of argument, and the position of the args is magical.

checkbox-ng/checkbox_ng/launcher/master.py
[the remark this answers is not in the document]
kissiel: The problem is that the person reading this will have to redo the work of an interpreter. As for argument for having two strings means you have to keep them in sync -> this template has a certain number of placeholders that have to be kept in sync with the format call.

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -166,6 +166,41 @@ def invoked(self, ctx):
...
+            agent_api_version,
+            controller_api_version
+        )
+
+        raise SystemExit(_(problem_msg))
```
kissiel, line 201: This is not how `_` works.

kissiel (commented):
added some more notes

checkbox-ng/checkbox_ng/launcher/test_master.py
```diff
@@ -60,3 +60,51 @@ def test_invoked_ok(
...
+        self_mock = mock.MagicMock()
+        session_assistant_mock = mock.MagicMock()
+        self_mock.sa = session_assistant_mock
+
+        remote_assistant_mock.REMOTE_API_VERSION = 0
+        session_assistant_mock.get_remote_api_version.return_value = 0
+
+        RemoteMaster.check_remote_api_match(self_mock)
+        # assert this runs with no problem, the two versions are the same
+        self.assertTrue(True)
```
kissiel, line 82: This only checks whether python works.

checkbox-ng/checkbox_ng/launcher/test_master.py
```diff
@@ -60,3 +60,51 @@ def test_invoked_ok(
             RemoteMaster.invoked(self_mock, ctx_mock)
 
         self.assertTrue(self_mock.connect_and_run.called)
+
+    @mock.patch("checkbox_ng.launcher.master.RemoteSessionAssistant")
+    @mock.patch("checkbox_ng.launcher.master._")
```
kissiel, line 65: This is not needed AFAICT

kissiel (commented):
some comments @ unit tests.

Hook25:
The patch coverage here is skewed by single line strings chained together (that is weird) and black changes, I think this is safe to land @kissiel

kissiel (approved):
Thanks for all the changes!

PR #688, opened 2023-08-29
Always allow manual triggering of daily builds

Hook25:
## Description

The manual triggering of daily builds for snaps still checks for changes in the previous 24h. 

This makes it do it either way, given that we are manually triggering it, we want it to run. 

## Resolved issues

Today we want to re-run daily builds because one of them failed and the changes are not detected by the workflow anymore. (This is another thing that needs fixing, but it is outside of this PR scope)

## Documentation

N/A (arguably, this is the behavior one would expect from a manual run of the workflow)

## Tests

N/A


kissiel (approved):
awesome!

PR #691, opened 2023-08-29
make the config chains scenario in metabox local-only

kissiel:
This should never have run via remote because:
- the configs on the controller are not taken into account
- the agent runs as root, so configs for the `ubuntu` user are pointless

There is more that we can improve in this test, but this patch at least fixes the metabox runs.


Hook25 (approved):
+1

PR #690, opened 2023-08-29
Fix a few typos

dannf:

PR #693, opened 2023-08-30
Removed category flags (Infra)

Hook25:
## Description

Codecov.io carryover does not seem to support partial carry over. What this would do is make our tests fail due to dropping coverage on "categories" flags (for example, unittest) due to the fact that if any coverage for a category is submitted, it is assumed as the overall coverage of the flag. 

This removes category flags leaving just the leaf flags for each part on the monorepo

## Resolved issues

See: https://github.com/canonical/checkbox/pull/690, unittest coverage is marked as -58% but this is not actually the case, `unittest` is just a category that includes one of the tests that were run

## Documentation

N/A

## Tests

N/A


kissiel (approved):
And the actual +1

PR #696, opened 2023-09-01
Fix clocktest.c complie warning and add more error handling (BugFix)

hanhsuan:
## Description

1.`_BSD_SOURCE` is deprecated when [glibc 2.20](https://sourceware.org/legacy-ml/libc-alpha/2014-09/msg00088.html) was released at 2014/09/08.
It should be fine to change `_BSD_SOURCE` to `_DEFAULT_SOURCE`.

2.If `num_cpus` is [`-1`](https://www.gnu.org/software/libc/manual/html_node/Sysconf-Definition.html), it will make garbage value for `time` to calculate `jitter`.
Adding to check `num_cpus` shouldn't be `-1`.

3.malloc won't be always return success.
Adding to check the return value shouldn't be `NULL`.

4.After `malloc` memory for `time`, it is never released.
Adding `free` before return.

5.Change `tab` to `space`

## Resolved issues

This commit makes the code more robust only.

## Tests

This change has been tested to compile and run on Ubuntu 22.04 and 16.04 without any issue.

```sh
u@u-Inspiron-5594:/dev/shm$ ./clocktest
Testing for clock jitter on 8 cpus
PASSED: largest jitter seen was 0.000620

Testing clock direction for 5 minutes...
PASSED: Iteration 0 delta: 0.000466
PASSED: Iteration 1 delta: 0.000514
PASSED: Iteration 2 delta: 0.000463
PASSED: Iteration 3 delta: 0.000298
PASSED: Iteration 4 delta: 0.000404
clock direction test: sleeptime 60 sec per iteration, failed iterations: 0
u@u-Inspiron-5594:/dev/shm$ ./clocktest-m 
Testing for clock jitter on 8 cpus
PASSED: largest jitter seen was 0.000605

Testing clock direction for 5 minutes...
PASSED: Iteration 0 delta: 0.000163
PASSED: Iteration 1 delta: 0.000096
PASSED: Iteration 2 delta: 0.001384
PASSED: Iteration 3 delta: 0.000346
PASSED: Iteration 4 delta: 0.000537
clock direction test: sleeptime 60 sec per iteration, failed iterations: 0
```

### DIFF
--- a/providers/base/src/clocktest.c
+++ b/providers/base/src/clocktest.c
@@ -36,24 +36,36 @@ int test_clock_jitter(){
         printf("Single CPU detected. No clock jitter testing necessary.\n");
         return 0;
     }
+    else if (num_cpus < 1) {
+        printf("Get number of cpu failed.\n");
+        return 1;
+    }
 
     printf ("Testing for clock jitter on %u cpus\n", num_cpus);
 
-    time=malloc(num_cpus * sizeof(struct timespec));
+    time = malloc(num_cpus * sizeof(struct timespec));
+    if (time == NULL) {
+        printf("Malloc memory failed.\n");
+        return 1;
+    }
 
     for (iter=0; iter<ITERATIONS; iter++){
         for (cpu=0; cpu < num_cpus; cpu++) {
             CPU_ZERO(&cpumask); CPU_SET(cpu,&cpumask);
-	        if (setaffinity(cpumask) < 0){
-    	        perror ("sched_setaffinity"); return 1;
-    	    }
-    	    /*
-    	     * by yielding this process should get scheduled on the cpu
-        	 * specified by setaffinity
-	         */
-        	sched_yield();
+            if (setaffinity(cpumask) < 0){
+                perror ("sched_setaffinity");
+                free(time);
+                return 1;
+            }
+            /*
+             * by yielding this process should get scheduled on the cpu
+             * specified by setaffinity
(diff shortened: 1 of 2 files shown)


providers/base/src/clocktest.c
```diff
@@ -36,24 +36,36 @@ int test_clock_jitter(){
         printf("Single CPU detected. No clock jitter testing necessary.\n");
         return 0;
     }
+    else if (num_cpus < 1) {
+        printf("Get number of cpu failed.\n");
```
bladernr, line 40: Is this just adding an error trap for a potential issue, or have folks seen cases where this actually happens? (just curious)

bladernr (approved):
Looks good to me. Thanks for making this better. That program has existed for ... well over a decade at this point, largely unchanged, and these are definitely some welcome improvements.

Note, I only know enough C to be dangerous, so there may be some C intricacies I'm missing here, but as far as I can tell this looks good and you've tested it out.

hanhsuan:
I've noticed that you are curious about the issue about `num_cpus` equal `-1`.
For me, I haven't noticed this happen before.
However, I use clocktest.c to know how to put C code into checkbox and I have a habit to check my  C code by [Clang Static Analyzer](https://clang-analyzer.llvm.org/) that will tell me some logical errors.

kissiel (approved):
Asome stuff!
TYVM for the changes

PR #701, opened 2023-09-05
Fix retcode on fail

kissiel:
## Description
This patch makes Checkbox return a valid return code depending on the outcome of the test plan being run.
Previously Checkbox remote always returned 0, even when some of the jobs failed. This patch makes checkbox return 0 only when **everything** passed.

## Tests

I'm providing a metabox test in this PR that checks this mechanic.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/master.py
+++ b/checkbox-ng/checkbox_ng/launcher/master.py
@@ -130,6 +130,7 @@ class RemoteMaster(ReportsStage, MainLoopStage):
         self._C = Colorizer()
         self._override_exporting(self.local_export)
         self._launcher_text = ""
+        self._has_anything_failed = False
         self._is_bootstrapping = False
         self._target_host = ctx.args.host
         self._normal_user = ""
@@ -340,6 +341,7 @@ class RemoteMaster(ReportsStage, MainLoopStage):
 
             if not keep_running:
                 break
+        return self._has_anything_failed
 
     def new_session(self):
         _logger.info("remote: Starting new session.")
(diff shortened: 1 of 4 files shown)


checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -503,6 +505,14 @@ def finish_session(self):
                 tmp_sig = signal.signal(signal.SIGINT, signal.SIG_IGN)
                 stack.callback(signal.signal, signal.SIGINT, tmp_sig)
                 self._export_results()
+        # let's see if any of the jobs failed, if so, let's return an error code of 1
+        job_state_map = (
+            self._sa.manager.default_device_context._state._job_state_map
+        )
+        self._has_anything_failed = any(
+            job.result.outcome == "fail" for job in job_state_map.values()
```
pieqq, line 513: what about crashed jobs? (i know that somewhere in checkbox there is a mapping that is done to map every possible state to one of `pass/skip/fail`, I'm just not sure if it's the case here)

pieqq: OK I checked by running a test plan that only has one job that's crashing, and checkbox retrurns 0. Oops :)

pieqq (changes requested):
Totally makes sense!

I tried both running local and remote. For local I tried going through the selection screens (OK) and running `checkbox-cli run testplan-with-a-job-that-fails`. Works as expected in both cases.

However, this patch does not seem to take other failing jobs (e.g. crashed) into account. See inline comments.

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -158,7 +159,7 @@ def invoked(self, ctx):
             )
         while time.time() < deadline:
             try:
-                self.connect_and_run(ctx.args.host, port)
+                return self.connect_and_run(ctx.args.host, port)
                 break
```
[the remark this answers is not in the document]
kissiel: Of course not! removed in 591fcb7c11f47d616708a2db633df3644868121b

checkbox-ng/checkbox_ng/launcher/master.py
kissiel: Yeah, very valid concern.
I think it can be done as a followup as the code provided here is critical to land fast.

pieqq (approved):
+1 provided that we actually follow-up on this to implement returning 1 for any bad job outcome.

For a list of the possible outcomes, see [checkbox-ng/plainbox/abc.py#L194-L219](https://github.com/canonical/checkbox/blob/main/checkbox-ng/plainbox/abc.py#L194-L219).

PR #702, opened 2023-09-05
sanitize output printed by loguru so it doesn't crash on backtraces

kissiel:
## Description

This patch makes sure that we don't send unescaped `<` characters to loguru (the other end is not needed to be sanitized).
This helps develop metabox scenarios (that may crash during development).

### DIFF
--- a/metabox/metabox/core/runner.py
+++ b/metabox/metabox/core/runner.py
@@ -276,9 +276,12 @@ class Runner:
             if not scn.has_passed():
                 self.failed = True
                 logger.error(scenario_description + " scenario has failed.")
-                logger.error(
-                    "Scenario output:\n" + scn.get_output_streams().strip()
-                )
+
+                # let's strip < from the output to avoid confusing loguru
+                # loguru assumes that <> is used for colorizing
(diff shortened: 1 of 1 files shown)


metabox/metabox/core/runner.py
```diff
@@ -276,9 +276,12 @@ def run(self):
             if not scn.has_passed():
                 self.failed = True
                 logger.error(scenario_description + " scenario has failed.")
-                logger.error(
-                    "Scenario output:\n" + scn.get_output_streams().strip()
-                )
+
+                # let's strip < from the output to avoid confusing loguru
```
pieqq, line 280: Looks like the comment does not match what's being done, cause in the code you escape `<` instead of stripping it from the output.

pieqq (changes requested):
Minor comment with the comment in that commit, otherwise good to go.

Hook25 (approved):
+1, this is blocking my testing and it is very much needed

pieqq (approved):
+1 provided the comment I mentioned is modified.

PR #703, opened 2023-09-05
return a "1" return code on job crashes

kissiel:
## Description

Similarly to https://github.com/canonical/checkbox/pull/701 this makes Checkbox return an actual error also when a job has crashed.
It also uses the enums instead of strings for identifying the outcomes.

## Tests
As with the previous one metabox scenario is included for the crash.


checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -508,8 +509,16 @@ def finish_session(self):
         job_state_map = (
             self._sa.manager.default_device_context._state._job_state_map
         )
+        failing_outcomes = (
```
pieqq, line 512: Looks like this variable is not used anywhere?

checkbox-ng/checkbox_ng/launcher/master.py
```diff
@@ -508,8 +509,16 @@ def finish_session(self):
...
+            IJobResult.OUTCOME_FAIL,
+            IJobResult.OUTCOME_CRASH,
+            # Outcome may be still set to `None`. It may  happen if the job
+            # was interrupted or crashed hasn't been properly registered (yet)
+            None,
+        )
         self._has_anything_failed = any(
-            job.result.outcome == "fail" for job in job_state_map.values()
+            job.result.outcome in (None, "fail", "crashed")
```
pieqq, line 520: Did you mean to use `failing_outcomes` here? (I'm not sure, cause `IJobResult.OUTCOME_CRASH` is set to `crash`, not `crashed`, so I don't know if this is intentional)

pieqq (commented):
I just glanced at the changes, didn't try it; just one comment below.

kissiel:
I've tested the "crashed" value vs "crash" and the "crashed" one worked because there was another job that yielded "None", so this is why metabox scenario passed.
This is why in https://github.com/canonical/checkbox/pull/703/commits/8360f752f2b990edcb6421cc5bd775e78da07039 I removed the None as a failing outcome, and made proper values (the enums) be used.

pieqq (approved):
LGTM! Thanks!

PR #706, opened 2023-09-06
Fix metabox test RunFailingTestplan (infra)

Hook25:
## Description

Fixing the return code of the run command makes the metabox scenario that tested it (somewhat wrongly) fail. This fixes it

## Resolved issues

N/A

## Documentation

I think this is the expected behavior (test fails, `exit(1)`), we don't need any new documentation

## Tests

To run the test you can run the following:
```
$ metabox local-source-16.04.py --tag RunFailingTestplan
$ metabox local-source-18.04.py --tag RunFailingTestplan
$ metabox local-source-20.04.py --tag RunFailingTestplan
$ metabox local-source-22.04.py --tag RunFailingTestplan
```

### DIFF
--- a/metabox/metabox/scenarios/basic/run-invocation.py
+++ b/metabox/metabox/scenarios/basic/run-invocation.py
@@ -1,14 +1,14 @@
 # This file is part of Checkbox.
 #
-# Copyright 2021 Canonical Ltd.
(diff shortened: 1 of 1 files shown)


kissiel (approved):
Obvious +1.
I'm actually glad that metabox failed this test today, shows that it actually works :)

PR #707, opened 2023-09-06
keyboard test (New)

p-gentili:
## Description
This PR introduces a new test job based on the latest HID feature which let us check in an automated way if the system correctly recognizes a USB keyboard. 

## Resolved issues

 Resolves [ZAP-555](https://warthogs.atlassian.net/jira/software/c/projects/ZAP/issues/ZAP-555)
 
## Documentation

Tests are included and can be seen as documentation in this case: the usage is still limited and under development.

[ZAP-555]: https://warthogs.atlassian.net/browse/ZAP-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/base/bin/zapper_keyboard_test.py
+++ b/providers/base/bin/zapper_keyboard_test.py
@@ -57,7 +57,9 @@ def read_keyboard_events(event, callback):
 
 
 def assert_key_combo(host, events):
-    """Request to press a key combination and expect the corresponding events."""
+    """
+    Request to press a key combination and expect the corresponding events.
+    """
 
     events.clear()
     zapper_run(host, "robot_run", ROBOT_TESTCASE_COMBO.encode(), {}, {})
@@ -102,7 +104,9 @@ def main(argv):
 
     events = []
     threading.Thread(
-        target=_listen_keyboard_events, args=(event_file, events.append), daemon=True
+        target=_listen_keyboard_events,
+        args=(event_file, events.append),
+        daemon=True,
     ).start()
 
     try:
--- a/providers/base/bin/zapper_keyboard_test.py
+++ b/providers/base/bin/zapper_keyboard_test.py
@@ -94,10 +94,11 @@ def main(argv):
     if len(argv) != 2:
         raise SystemExit("Usage: {} <zapper-ip>".format(argv[0]))
 
-    try:
-        event_file = os.path.realpath(f"/dev/input/by-id/{ZAPPER_KBD}", strict=True)
-    except FileNotFoundError as exc:
-        raise SystemExit("Cannot find Zapper Keyboard.") from exc
+    zapper_kbd_path = f"/dev/input/by-id/{ZAPPER_KBD}"
+    if os.path.exists(zapper_kbd_path):
+        event_file = os.path.realpath(zapper_kbd_path)
+    else:
+        raise SystemExit("Cannot find Zapper Keyboard.")
 
     events = []
     threading.Thread(
--- a/providers/base/bin/zapper_keyboard_test.py
+++ b/providers/base/bin/zapper_keyboard_test.py
@@ -6,7 +6,6 @@ USB keyboard.
 To run the test you need Zapper board connected and set up.
 """
 import os
-import select
 import struct
 import sys
 import threading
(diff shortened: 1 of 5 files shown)


providers/base/bin/zapper_keyboard_test.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+import os
+import struct
+import sys
+import threading
+from enum import Enum
+
+from checkbox_support.scripts.zapper_proxy import zapper_run  # noqa: E402
+
+ZAPPER_KBD = "usb-Canonical_Zapper_main_board_123456-event-kbd"
```
kissiel, line 16: Will this be used by some external entity?
If not, maybe it's better to keep it local to a place where it's used (as there's only one place it's used).

providers/base/bin/zapper_keyboard_test.py
```diff
@@ -0,0 +1,116 @@
...
+EVENT_BIN_FORMAT = "llHHI"
```
kissiel, line 17: It's a standalone test that doesn't or shouldn't require any prior knowledge. Could you write a comment for what it is and what it's for?

providers/base/bin/zapper_keyboard_test.py
```diff
@@ -0,0 +1,116 @@
...
+    ]
+
+
+def main(argv):
+    if len(argv) != 2:
+        raise SystemExit("Usage: {} <zapper-ip>".format(argv[0]))
+
+    zapper_kbd_path = f"/dev/input/by-id/{ZAPPER_KBD}"
+    if os.path.exists(zapper_kbd_path):
+        event_file = os.path.realpath(zapper_kbd_path)
```
kissiel, line 101: What's the thinking behind this?
From what I know `open()` doesn't care if it's opening a symlink.

providers/base/bin/zapper_keyboard_test.py
```diff
@@ -0,0 +1,116 @@
...
+        (KeyEvent.DOWN, 24),
+        (KeyEvent.UP, 24),
+    ]
+
+
+def main(argv):
+    if len(argv) != 2:
+        raise SystemExit("Usage: {} <zapper-ip>".format(argv[0]))
+
+    zapper_kbd_path = f"/dev/input/by-id/{ZAPPER_KBD}"
```
kissiel, line 99: This dodges a problem where the FS node doesn't exist, but will still yield exceptions when there's a problem when opening the file (permissions or other IO errors).
So following a more pythonic EAFP instead of LBYL can give you better results here.
Namely surrounding the actual opening code with a try/except block would potentially capture more classes of errors and also you could use `raise SystemExit(...) from exc` that facilitates debugging.

providers/base/bin/zapper_keyboard_test.py
```diff
@@ -0,0 +1,116 @@
...
+    if os.path.exists(zapper_kbd_path):
+        event_file = os.path.realpath(zapper_kbd_path)
+    else:
+        raise SystemExit("Cannot find Zapper Keyboard.")
+
+    events = []
+    threading.Thread(
+        target=_listen_keyboard_events, args=(event_file, events.append), daemon=True
+    ).start()
```
kissiel, line 110: Since there's no explicit finish to this thing I'm guessing you count on the thread being killed when the process finishes. That's a bit like not calling `free()` after `malloc()` because it'll get freed regardless.

I'm guessing you've probably considered subclassing thread and making the listening thread **object** handle this, right? Why have you decided to do this anonymous thread object?

where I'm getting at is that the ListenerThread could have a property like `keep_running` that could be used by the infinite[-ish] loop above. This way, in `main` you could set this to false, and the thread would finish in a graceful way.

kissiel (commented):
Nothing critical, just some consideration below.

providers/base/bin/zapper_keyboard_test.py
p-gentili: done in https://github.com/canonical/checkbox/pull/707/commits/600ea83d429edd36a78330f525d340610084e5d6

providers/base/bin/zapper_keyboard_test.py
p-gentili: moved in https://github.com/canonical/checkbox/pull/707/commits/600ea83d429edd36a78330f525d340610084e5d6

providers/base/bin/zapper_keyboard_test.py
p-gentili: brain fart :)

providers/base/bin/zapper_keyboard_test.py
p-gentili: I want to open this file in the listener instead, it makes the main() function simpler. I'm now using `os.access` to check for file existence and permission.

providers/base/bin/zapper_keyboard_test.py
p-gentili: The main blocker here was not the named thread as you imagined but the blocking `read()` function. In https://github.com/canonical/checkbox/pull/707/commits/f5f1d4c8f9201dcc3132b470d3f1d30d5e683140 + https://github.com/canonical/checkbox/pull/707/commits/9d239cd978b1e8ff967f23267841e9c11e16661c I created a new listener class as you suggested and took actions to make the _reading_ section non blocking.

kissiel: Blocker? TBH my cache is thoroughly flushed at this point. But my comment was more about encapsulating the condition on which the thread continues to read() (even if it's blocking). 
It looks great now.

kissiel (approved):
Fantastic stuff, +1

Issue #570, opened 2023-06-16
Setting incorrect CPU affinity in network.py can cause CPU load to spike


rodwsmith:
### Bug Description

On some systems, the network.py script can set an incorrect CPU affinity (via the -A option to iperf3) -- typically to core 0 on devices that don't support NUMA. This can cause CPU load to spike -- for instance, this has been observed to spike to over 90% on a Socionext Developerbox with an ARM Cortex-A53 CPU using a 1 Gbps network test, whereas omitting the -A option to iperf3 in the script produces CPU loads of around 30-35%. On the Socionext system, the throughput dropped from 92% (without the -A option) to about 82% (with the incorrect CPU affinity set).

### To Reproduce

1. Run network.py manually (for instance, "sudo /usr/lib/checkbox-provider-base/bin/network.py test -i {interface} -t iperf --iperf3 --scan-timeout 36 --fail-threshold 80 --cpu-load-fail-threshold 90 --runtime 90 --num_runs 4 --target {ipaddr}"
2. Monitor CPU load with top; or wait for the test to finish and look at the reported CPU load.

### Environment

My own tests were done on a Socionext Developerbox running Ubuntu 23.04; however, Rick Wu (rickwu4444 on MM) reported similar issues on an imx8-based platform.

rickwu666666:
This is my log for running network.py on imx8MP platform
`root@ubuntu:/snap/checkbox20/current/providers/checkbox-provider-base/bin# ./network.py test -i etheqos -t iperf --iperf3 --scan-timeout 3600 --fail-threshold 80 --cpu-load-fail-threshold 90 --runtime 90 --num_runs 1 --target 10.102.88.220
INFO:root:Testing etheqos against 10.102.88.220
INFO:root:Have successfully pinged 10.102.88.220 on etheqos
INFO:root:-------------------- Test Run Number 1 --------------------
INFO:root:Using 1 thread.
WARNING:root:WARNING: Could not find the NUMA node associated with etheqos!
WARNING:root:Setting the association to NUMA node 0, which may not be optimal!
INFO:root:NUMA node of etheqos is 0....
INFO:root:Connecting to port 5201 on server....
INFO:root:Avg Transfer speed: 938.2555555555556 Mb/s
INFO:root:93.83% of theoretical max 1000 Mb/s
INFO:root:Average CPU utilization: 97.4%
WARNING:root:The network test against 10.102.88.220 failed because:
ERROR:root:  CPU load: 97.4%
[...]

PR #711, opened 2023-09-08
Don't use '-A' in network.py on non-numa systems (BugFix)

rodwsmith:
## Description
This PR fixes bug #570, which can cause the CPU load to spike if the CPU affinity is set incorrectly on systems that don't support NUMA features.

The fix involves identifying systems that lack NUMA support and _not_ passing the -A {nodes} option to iperf3 on such systems.

## Resolved issues

Resolves bug #570

Jira card: https://warthogs.atlassian.net/browse/SERVCERT-1183

## Documentation

No documentation changes required.

## Tests

Tested on lloobee (my own ARM64-based development system), which exhibited the bug. The new version drops the CPU load from 90%+ to about 35-40%, and increases the throughput by about 5% (from 88% to 93%, give or take a percent or two).

I've also tested on two servers in the Server Certification pool, torchtusk and whomp, to be sure the script doesn't create problems on systems that DO support NUMA. No unexpected problems arose. The change should result in exactly the same iperf3 command being used on such systems as the previous version of the script. Such systems were unaffected by the bug to begin with.

### DIFF
--- a/providers/base/bin/network.py
+++ b/providers/base/bin/network.py
@@ -174,16 +174,15 @@ class IPerfPerformanceTest(object):
             with open(filename, "r") as file:
                 node_num = int(file.read())
         except FileNotFoundError:
+            node_num = -1
+        # Some systems (that don't support NUMA?) produce a node_num of -1.
+        # Later in the script, this will be interpreted to omit the -A option
+        # to iperf3, thus disabling NUMA features.
+        if node_num == -1:
             logging.warning("WARNING: Could not find the NUMA node "
                             "associated with {}!".format(device))
-            logging.warning("Setting the association to NUMA node 0, "
-                            "which may not be optimal!")
-            node_num = 0
-        # Some systems (that don't support NUMA?) produce a node_num of -1.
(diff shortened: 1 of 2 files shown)


bladernr (dismissed):
Looks good to me... thanks

Hook25 (changes requested):
I have to ask a few changes here:
- codecov is complaining because we need a new test here, `sumarize_cpu` is untested
- it is kind of weird to return `-1` instead to throwing an exception. Would you consider changing this to raise an exception both if the file is not found or if it is found and contains a `-1`? That way anyone calling the function will not have to check if `-1` is a valid numa number or what is going on, if something is returned, it is valid, else an exception is thrown
- Minor: Change the title of the PR after rebasing on main, this is a bugfix so add a (bugfix) postfix to the title!

Ty

kissiel:
I added tests and fixed the PR title.
The -1 as the node index is in line with what sysfs would report on a non-numa systems, so I left it in.

@hook please take another look

Hook25 (approved):
+1, if this causes a problem we can change in the future I guess

PR #534, opened 2023-06-06
Calculate version using setuptools scm

Hook25:
## Description

Since bumpversion modifies the version history of the repository as part of setting version metadata in place, and since our CI needs are made complex by lab automation being done by Jenkins jobs that can be executed once there is a package available that are used to validate the state of the package for promoting it to the next stability level (edge → beta → stable kind of flow), we are best off migrating away from bumpversion to a tool that does not modify the state of source code in the repository but can operate with tags

## Resolved issues

Resolves: [CHECKBOX-613](https://warthogs.atlassian.net/browse/CHECKBOX-613)
Fixes: https://github.com/canonical/checkbox/issues/524

## Documentation

N/A

## Tests

This was tested offline via metabox both in legacy an remote runs on jammy and kinetic

[CHECKBOX-613]: https://warthogs.atlassian.net/browse/CHECKBOX-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #712, opened 2023-09-11
Fix PIP dependencies installation on bionic and older (Infra)

kissiel:
## Description

This patch changes the `requires` to `install_requires` to be compliant with: [The Offical Python Packaging User Guide](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements)

Which also fixes installation behaviour on bionic.

This got (accidentally?) broken with this PR: https://github.com/canonical/checkbox/pull/534
IIRC the `requires` is obsolete for a long time

## Resolved issues
Partially fixes: [Checkbox-840](https://warthogs.atlassian.net/browse/CHECKBOX-841)

## Tests
I've tested deployment of checkbox on: Bionic, Focal, Jammy, and Lunar.

### DIFF
--- a/checkbox-ng/setup.cfg
+++ b/checkbox-ng/setup.cfg
@@ -3,7 +3,7 @@ sign=true
 [options]
 packages=find_namespace:
 include_package_data=true
-requires=
+install_requires=
   packaging
   requests >= 1.0
   urwid >= 1.1.1
--- a/checkbox-support/setup.cfg
+++ b/checkbox-support/setup.cfg
@@ -3,7 +3,7 @@ sign = True
 [options]
 packages=find_namespace:
 include_package_data=true
-requires=
+install_requires=
   pyparsing >= 2.2.0
   requests >= 1.0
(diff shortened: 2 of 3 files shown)


Hook25 (changes requested):
Initially it was done that way https://github.com/canonical/checkbox/pull/534/commits/cbc94af0ec1f12377415d009bfa0e56bbd967597#diff-3912142dffe89489011cd217741ced726bf3a1fd848b42a1357fd9f65e6496faR33 but I can not recally why it did not work. Possibly there were other issues with it that were fixed afterward, I am not sure. If this fixes the installation of dependencies (I really hope it does, that would be so great!) we have to remove the bodges in snaps (core16) and metabox.

pieqq (approved):
It looks OK to me with the documentation provided.

Not sure of the implications mentioned by @Hook25

Hook25 (approved):
+1, we can finally remove some of those hacks from machine.py :)

PR #713, opened 2023-09-11
fix `pip install -e .` for focal (older pip) (Infra)

kissiel:
## Description
PIP, when run in editable mode requires the projects to have `setup.py`, it cannot do it just with setup.cfg and pyproject.toml.

This patch makes it so checkbox can be deployed and used from source.

## Resolved issues
This partially fixes [Checkbox-841](https://warthogs.atlassian.net/browse/CHECKBOX-841)

## Tests
Tested on: Bionic, Focal, Jammy, and Lunar

### DIFF
--- a/checkbox-core-snap/series16/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series16/snap/snapcraft.yaml
@@ -173,8 +173,6 @@ parts:
     override-build: |
       # this is a bodge to be backward compatible
       snapcraftctl build
-      # setup.py is necessary for this old pip verision
-      echo "from setuptools import setup; setup()" > setup.py
       # update pip and setuptools (+deps) because the directive
       # is ignored in pyproject (and pip on xenial is preinstalled 8.x)
       $SNAPCRAFT_PART_INSTALL/usr/bin/python3 -m pip install "pip<21" "setuptools<48" "setuptools_scm[toml]>=3.4" "importlib_metadata==1.0.0" "zipp<2"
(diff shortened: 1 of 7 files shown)


Hook25 (changes requested):
If we want to re-introduce `setup.py` I have a few changes to request. Note that we explicitly decided with @yphus to not include in the repository any `setup.py`, rather create them in the recipe for the various parts. I can not recall why we settled on that decision but we may re-evaluate that.

For now, this PR should also remove those creations:
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/checkbox-core-snap/series16/snap/snapcraft.yaml#L220
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/checkbox-core-snap/series16/snap/snapcraft.yaml#L177
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/checkbox-core-snap/series18/snap/snapcraft.yaml#L181
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/checkbox-core-snap/series18/snap/snapcraft.yaml#L210
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/checkbox-core-snap/series20/snap/snapcraft.yaml#L236
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/checkbox-core-snap/series20/snap/snapcraft.yaml#L190
- https://github.com/canonical/checkbox/blob/b725268f3dbabb4351a7adae70fdf22930ef08b4/metabox/metabox/core/machine.py#L312

Minor: Change your PR title with the correct postfix so that the merge message is correct automatically and we don't have to remember to include it. (Add an (infra) postfix)

kissiel:
I've removed the dynamic creation of setup.py as @Hook25 suggested. There was also one in series22.

Hook25 (approved):
+1 ty for the change

PR #714, opened 2023-09-11
Fix deb version to adhere to LP naming (New)

Hook25:
## Description

The version generator that we use (`setuptools_scm`) creates a version string that is not correctly interpreted by launchpad. The reason is that, when marking a non-tagged commit, it will include a `.devXX[...]` postfix. Launchpad supports this kind of postfixes but they are not delimited by a `.`, rather a `~`. 

This is the reason why deb builds are currently broken, as `2.9.dev...` is `>=` to `2.9.1.dev[...]` (because `d >= 1`). That would not be the case if we used the correct naming from the beginning.

## Resolved issues

N/A

## Documentation

This change is introduced with a explanation above it so that next time we will know why it is there

## Tests

To test if a version is upgradable to another we can use `dpkg --compare-versions`. 

Note how:
```
$ dpkg --compare-versions 2.9.1~dev12 gt 2.9~dev123 && echo Can release || echo Rejected 
Can release
```
While on the contrary:
```
$ dpkg --compare-versions 2.9.1.dev12 gt 2.9.dev123 && echo Can release || echo Rejected 
```

This PR changes the second into the first (replaces `.dev` into `~dev`)


kissiel (changes requested):
1. The ladder of `re.sub()`s is unacceptable.
2. The thing implemented here is somewhere in the middle of other code instead of being a function with decent docstring and everything that comes with it, like...
4. Unit tests. Because this is so functional, writing tests for this would be a breeze.

If this is redoing the version numbering smarter way would be to parse the original and regenerate one you want instead of doing this re.sub-cement mixer.

And my zeroth point - what is stopping us from having one version string template?

kissiel:
Oh gosh, I just noticed that this is not your creation (the devilish ladder of re.subs and everything around it).
So this has to be fixed, please file an appropriate tickets so we don't forget to actually fix this. After the ticket is filed I'll be ok to land this.

kissiel (approved):
with CHECKBOX-857 filed I think we can land this as it still improves,
It doesn't look good to me, but gets the job done and is not wrong. +1

PR #726, opened 2023-09-14
Limit tox run to the env of the current matrix (infra)

Hook25:
## Description

Currently tox is run on a machine without limiting the environment it is meant to cover. When this is done, tox selects any environment that it is able to run so, if an image contains more than a version of python, tox will in a single action for multiple versions of python.

See for example this [action run](https://github.com/canonical/checkbox/actions/runs/6184293747/job/16787663978), the python3.6 environment contains both python3.6 and python3.8. 

This makes the output very long and hard to read, the failure here is at line [1399](https://github.com/canonical/checkbox/actions/runs/6184293747/job/16787663978#step:5:1400), that would be at the end of the log, but it is not because right after 3.6, 3.8 is tested. This also means that tests are run multiple times with the same python, wasting a lot of resources/time.

## Resolved issues

Fixes: [CHECKBOX-870](https://warthogs.atlassian.net/browse/CHECKBOX-870)

## Documentation

N/A

## Tests

N/A

[CHECKBOX-870]: https://warthogs.atlassian.net/browse/CHECKBOX-870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/tox-checkbox-ng.yaml
+++ b/.github/workflows/tox-checkbox-ng.yaml
@@ -20,6 +20,15 @@ jobs:
     strategy:
       matrix:
         python: ["3.5", "3.6", "3.8", "3.10"]
+        include:
+          - python: "3.5"
+            tox_env_name: "py35"
+          - python: "3.6"
+            tox_env_name: "py36"
+          - python: "3.8"
+            tox_env_name: "py38"
+          - python: "3.10"
+            tox_env_name: "py310"
     steps:
       - uses: actions/checkout@v3
       - name: Setup Python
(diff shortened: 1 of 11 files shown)


kissiel (approved):
It seems this is the most elegant way to do it.

the setup-python action requires a dotted version string, so we can't use one version string type, and we have to have this mapping.

Thanks for the patch and clarification.
+1

PR #729, opened 2023-09-15
Make ppa-dev-tools actually look for the correct package (infra)

Hook25:
## Description

ppa-dev-tools does not notice builds in the beta workflow but it does in the daily one. The reason is that the name it is looking for is wrong (it is looking for the name of the recipe instead of the package!)

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-701

## Documentation

N/A

## Tests

Comparing this with the beta workflow (`tools/deb_daily_builds.py`) it is clear that this is the only difference


kissiel (approved):
lol, +1

PR #732, opened 2023-09-18
Validate pr title for traceability markers (infra)

Hook25:
## Description

We have a new traceability system for PR titles but we have no way to actually enforce it right now beside manually checking. 

This changes that by adding a new GH Action that validates the title to check that it actually ends with a traceability marker. To make them not too invasive and hard on contributors, these markers are case insensitive and optionally enclosed in brackets.

## Resolved issues

N/A

## Documentation

N/A

## Tests

This PR was created with a title that does not contain any marker. The job should fail. Once that happens it will be edited to the correct title and it should pass.

Failure: https://github.com/canonical/checkbox/actions/runs/6223347828/job/16889096967?pr=732
Passing: https://github.com/canonical/checkbox/actions/runs/6223329257/job/16889036066?pr=732

### DIFF
--- a/.github/workflows/pr_validation.yaml
+++ b/.github/workflows/pr_validation.yaml
@@ -14,4 +14,15 @@ jobs:
     - name: Checking the presence of the Traceability Marker in the PR title
       env:
         PR_TITLE: ${{ github.event.pull_request.title }}
-      run: echo "$PR_TITLE" | grep -iP "(infra|bugfix|new|breaking)\)?$"
+      run: |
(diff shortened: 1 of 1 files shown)


Hook25:
I have made the brackets compulsory.

As for (fix) no, it is not in our convention right now, we could map it to bugfix and accept it but I think it would become a bit weird. Afterall a fix may be backward compatible, non backward compatible or relative to infra. I'm more on the side of not accepting (fix) as a valid marker

kissiel (approved):
Oh, sorry, my mistake. Please ignore my input on "fix", "bugfix" is ok.
I've used "fix" so many times before that it's inprinted in my brain.

Overall +1

PR #733, opened 2023-09-18
Exclude python main starting code from coverage report (infra)

Hook25:
## Description

Most python scripts call main via the idiom 
```
if __name__ == "__main__":
  exit(main())
```
or something similar. 

These lines are not significant most of the times and can be safely ignored in the coverage report else they will (sometimes heavily) make the coverage requirement harder to hit for new small scripts.

## Resolved issues

N/A

## Documentation

N/A

## Tests

The snippet was taken from the coverage official documentation [(here)](https://coverage.readthedocs.io/en/stable/excluding.html#advanced-exclusion). Note that we use `exclude_lines` instead of `exclude_also` because that keyword in not supported for older releases of coverage.py (anything <3.8)

### DIFF
--- a/checkbox-ng/.coveragerc
+++ b/checkbox-ng/.coveragerc
@@ -23,4 +23,5 @@ exclude_lines =
     @public
     pragma: no cover
     raise NotImplementedError
+    if __name__ == .__main__.:
 show_missing = True
--- a/checkbox-support/.coveragerc
+++ b/checkbox-support/.coveragerc
@@ -18,4 +18,5 @@ exclude_lines =
     @public
     pragma: no cover
     raise NotImplementedError
+    if __name__ == .__main__.:
 show_missing = True
--- a/providers/base/.coveragerc
+++ b/providers/base/.coveragerc
@@ -13,4 +13,5 @@ exclude_lines =
     @public
     pragma: no cover
     raise NotImplementedError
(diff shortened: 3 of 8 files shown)


kissiel (commented):
The line following the `if` is not being ignored. I'm starting to think it may be as easy as adding:
`    *main()*` to the excludes

kissiel:
> [(everything in the sub-block of ignore matches should be ignored as well, this is why this is dangerous a bit)](https://coverage.readthedocs.io/en/7.3.0/config.html#report-exclude-lines). Adding `*main()*` will make coverage ignore... the main function!
> 
> From the coverage doc:
> 
> > You can exclude lines introducing blocks, and the entire block is excluded. If you exclude a def line or decorator line, the entire function is excluded.

but the definition of a main function wouldn't be indented, so this clause wouldn't apply [maybe?]

Hook25:
I think what that means is that if you `exclude_lines` the header of any block like this
```
block_start:
  indented_code
  more_code

```
and you exclude `block_start`, you exclude them all. So the `if __name__ == .__main__.:" is exactly what we want here

kissiel (approved):
>I think what that means is that if you exclude_lines the header of any block like this

Oh, that's nice, big +1
TIL

PR #735, opened 2023-09-18
Add tracers tests (New)

ahresse:
This is an attempt to add `ptrace` and `ftrace` tests in checkbox.

## Description

For a specific project at Canonical we are using checkbox and we need to check for the availability of user-space `ptrace` and `ftrace` tools in our images. Thus, we want to add a dedicated test to check this and since it could be relevant to other, this might be into base checkbox test plan.

This, PR:
 - Introduces a new dedicated unit for tracers.
 - Introduces a specific binary tool: `ptrace_test` (source code in C) that aims to test the `ptrace()` system call availability.
 - Introduces the associated test with `ptrace_test`.
 - Introduces a new test to check `ftrace` availability (based on the availability of `tracefs`).

A potential alternative to the usage of specific `ptrace_test` binary: parsing `/proc/kallsyms` got identified but not developed.

## Resolved issues

[Associated Jira ticket](https://warthogs.atlassian.net/browse/ERL-358)

### DIFF
--- a/providers/base/units/tracers/jobs.pxu
+++ b/providers/base/units/tracers/jobs.pxu
@@ -29,6 +29,8 @@ _summary:
  Tests ftrace kernel tracing infrastructure is enable
 _description:
  Check if kernel configuration item CONFIG_FTRACE bool is set to 'y'.
+requires:
+  kernel_config_file.detected == 'true'
 
 id: tracers/tracefs
 plugin: shell
--- a/providers/base/units/tracers/jobs.pxu
+++ b/providers/base/units/tracers/jobs.pxu
@@ -13,3 +13,19 @@ _summary:
  Tests the ptrace tracer availability
 _description:
  The test initiates tracing using the ptrace syscall with ptrace_test binary (c source code).
+
+id: tracers/ftrace
+plugin: shell
+command: 
+  if grep -q -e "^CONFIG_FTRACE=y" /boot/config-"$(uname -r)"
+  then
+      echo "ftrace: supported"
+      exit 0
+  else
+      echo "ftrace: unsupported"
+      exit 1
+  fi
+_summary:
+ Tests ftrace kernel tracing infrastructure is enable
+_description:
+ Check if kernel configuration item CONFIG_FTRACE bool is set to 'y'.
--- a/providers/base/units/tracers/jobs.pxu
+++ b/providers/base/units/tracers/jobs.pxu
@@ -29,3 +29,19 @@ _summary:
  Tests ftrace kernel tracing infrastructure is enable
 _description:
  Check if kernel configuration item CONFIG_FTRACE bool is set to 'y'.
+
+id: tracers/tracefs
+plugin: shell
+command: 
+  if mount | grep -q tracefs
+  then
+      echo "tracefs: supported"
+      exit 0
+  else
+      echo "ftracefs: unsupported"
+      exit 1
+  fi
+_summary:
+ Tests the tracefs file system availability
+_description:
+ Check if the tracefs is mounted on a system using the mount command.
--- a/providers/base/units/tracers/jobs.pxu
+++ b/providers/base/units/tracers/jobs.pxu
@@ -1,0 +1,15 @@
+id: tracers/ptrace
+plugin: shell
+command:
+  if ptrace_test
+  then
+      echo "ptrace: supported"
+      exit 0
+  else
+      echo "ptrace: unsupported"
+      exit 1
+  fi
+_summary:
+ Tests the ptrace tracer availability
+_description:
+ The test initiates tracing using the ptrace syscall with ptrace_test binary (c source code).
--- a/providers/base/units/tracers/resource.pxu
+++ b/providers/base/units/tracers/resource.pxu
@@ -1,0 +1,12 @@
+id: kernel_config_file
+plugin: resource
+_summary:
+ Kernel config file resource
+command:
+ if [[ -f "/boot/config-$(uname -r)" ]]
+ then
+     echo "detected: true"
+ else
+     echo "detected: false"
+ fi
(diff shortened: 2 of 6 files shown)


ahresse:
Can't assign reviewer but @zongminl should be in it.

yphus:
What are the reasons why `ptrace`/`ftrace` would not be available on the base image? is it via kernel config?

I'd suggest turning the two proposed jobs into a resource job that would output something like that:

ptrace: supported
ftrace: unsupported

More tracers could be added in the future to this resource.
and have a new job that checks their availability (using the fail-on-resource flag):

```
id: tracers/check
category_id: tracer
_summary: ptrace and ftrace availability
_description:
  Check availability of ptrace syscall and ftrace mount (tracefs)
flags: simple fail-on-resource
command: echo "ptrace and ftrace found in the image"
requires:
  tracers.ptrace == 'supported'
  tracers.ftrace == 'supported'
```

providers/base/units/tracers/jobs.pxu
```diff
@@ -0,0 +1,15 @@
+plugin: shell
+id: tracers/ptrace
+command: ptrace_test
+_summary:
+ Tests the ptrace tracer availability
+_description:
+ Runs a test for ptrace tracer availability.
```
kissiel, line 7: Similarly to the one below, the description could be more specific. For instance:
The test initiates tracing using the `ptrace` syscall from the c code.

kissiel (commented):
Looks great. One comment about descriptions below.

providers/base/units/tracers/resource.pxu
```diff
@@ -0,0 +1,12 @@
+id: kernel_config_file
+plugin: resource
+_summary:
+ Kernel config file resource
+command:
+ if [[ -f "/boot/config-$(uname -r)" ]]
+ then
+     echo "state: supported"
```
kissiel, line 8: What's the 'support' in this case?
Looking at what this job does, it seems like it checks whether kernel config file is available in the system.

The id here would suggest, that this resource provides the configuration options of the kernel.

providers/base/units/tracers/resource.pxu
```diff
@@ -0,0 +1,12 @@
...
+ else
+     echo "state: unsupported"
+ fi
+_description: Creates resource info for kernel config file
```
kissiel, line 12: More truthful description would be to state that it checks for kernel config file existence.

providers/base/units/tracers/resource.pxu
kissiel: Two ways forward from here is either:
1) make the job do what it promises, which is produce all the configuration options with their value, so for instance the ftrace job could require: `kernel_config_file.CONFIG_FTRACE == "y"`, or, simpler:
2) adjust description of the job with what it actually does: check for existence/availability of the kernel config.

ahresse: Description should be more clear now: `_description: Check existence of kernel config file`

providers/base/units/tracers/jobs.pxu
```diff
@@ -0,0 +1,43 @@
...
+ The test initiates tracing using the ptrace syscall with ptrace_test binary (c source code).
+
+id: tracers/ftrace
+plugin: shell
+command: 
+  if grep -q -e "^CONFIG_FTRACE=y" /boot/config-"$(uname -r)"
+  then
+      echo "ftrace: supported"
+  else
+      echo "ftrace: unsupported"
```
kissiel, line 22: Without `exit 1` here the test will always pass.

providers/base/units/tracers/jobs.pxu
```diff
@@ -0,0 +1,43 @@
+id: tracers/ptrace
+plugin: shell
+command:
+  if ptrace_test
+  then
+      echo "ptrace: supported"
+  else
+      echo "ptrace: unsupported"
```
kissiel, line 8: Without `exit 1` here the test will always pass.

providers/base/units/tracers/resource.pxu
```diff
@@ -0,0 +1,12 @@
+id: kernel_config_file
+plugin: resource
+_summary:
+ Kernel config file resource
+command:
+ if [[ -f "/boot/config-$(uname -r)" ]]
+ then
+     echo "detected: true"
+ else
+     echo "detected: false"
```
kissiel, line 10: Without `exit 1` here the test will always pass.

kissiel (commented):
when any of those tests fail the error code is not propagated, so in the end this may happen:
```
=========[ Running job 2 / 3. Estimated time left (at least): 0:00:00 ]=========
--------------------[ Tests the ptrace tracer availability ]--------------------
ID: com.canonical.certification::tracers/ptrace                                 
Category: com.canonical.plainbox::uncategorised                                 
ptrace: unsupported                                                                            
Outcome: job passed

providers/base/units/tracers/resource.pxu
kissiel: Oh, this shouldn't exit with 1. Sorry, one copy pasta too many.

ahresse: I though about it maybe you wanted to talk about the tracefs test which I patched too

ahresse:
/canonical/self-hosted-runners/run-workflows 42863d9be27a7afb4505ae94e346c93a1eee06c4

kissiel:
/canonical/self-hosted-runners/run-workflows ca010399c43153cf52c6c932e6c0d92d680365a8

kissiel:
/canonical/self-hosted-runners/run-workflows 0ef5bdcc31c93d6a4d4201dbbecfe5d5c8284be6

kissiel (approved):
Fantastic work, TYVM

PR #740, opened 2023-09-21
Add ADL-N production key (new)

xanthein:
Update ADL-N production key from sof upstream
https://github.com/thesofproject/sof/commit/afb2d913ca66c57b2b0aab58cf89d23a73f7df6f

## Description
Add ADL-N production key


Hook25 (changes requested):
Thank you for this update. Given that this file is basically a [vendorized version of this tool](https://github.com/thesofproject/sof/blob/main/tools/sof_ri_info/sof_ri_info.py), can you please add it to the `omit` section in the coveragerc file in `checkbox/providers/base/.coveragerc` like this:
```
omit=
  tests/*
  # vendorized https://github.com/thesofproject/sof/blob/main/tools/sof_ri_info/sof_ri_info.py
  bin/sof_ri_info.py
```

This should stop codecov from complaining

Hook25 (approved):
+1, tyvm for this update

PR #742, opened 2023-09-21
Changed pyotherside to qml-module-io-thp-pyotherside (bugfix)

Hook25:
## Description

pyotherside was just a transitional dummy package to install qml-module-io-thp-pyotherside. It was removed in mantic so we must update the dependency now to match what is actually in the repos.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


kissiel (commented):
I've been checking this ;)
```
qml-module-io-thp-pyotherside/xenial 1.4.0-2 amd64
  asynchronous Python 3 Bindings for Qt 5 (QML plugin)
```

Issue #695, opened 2023-08-31
virtualization/verify_lxd_vm should wait longer before trying to access the VM


bladernr:
### Bug Description

The test works fine, but is a bit quick to start trying to ssh into the VM, causing several retries.  This makes the log noisy.  

Simple fix, it just needs a longer pause before trying to connect to the VM, and maybe longer between retries too, to avoid unnecessary repetition.

### To Reproduce

Run checkbox and run the virtualization/verify_lxd_vm job
Note in the logs all the retries (captured below).

### Environment

Current checkbox version

### Relevant log output

```shell
DEBUG:root:Wait for vm to boot
ERROR:root:Command lxc exec testbed -- lsb_release -a returnd a code of 1
ERROR:root: STDOUT: 
ERROR:root: STDERR: Error: Failed to connect to lxd-agent

DEBUG:root:Re-verify VM booted
ERROR:root:Command lxc exec testbed -- lsb_release -a returnd a code of 1
ERROR:root: STDOUT: 
ERROR:root: STDERR: Error: Instance is not running

DEBUG:root:Re-verify VM booted
ERROR:root:Command lxc exec testbed -- lsb_release -a returnd a code of 1
ERROR:root: STDOUT: 
ERROR:root: STDERR: Error: Failed to connect to lxd-agent

DEBUG:root:Re-verify VM booted
ERROR:root:Command lxc exec testbed -- lsb_release -a returnd a code of 1
ERROR:root: STDOUT: 
ERROR:root: STDERR: Error: Failed to connect to lxd-agent

DEBUG:root:Re-verify VM booted
DEBUG:root:Command lxc exec testbed -- lsb_release -a:
DEBUG:root: STDOUT: Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```

### Additional context

Simple fix to clear up a minor annoyance and make the test output cleaner.

mreed8855:
The logs from the updated pull request now look like the following:

DEBUG:root:Wait for vm to boot
ERROR:root:Command lxc exec testbed -- lsb_release -a returned a code of 255
ERROR:root: STDOUT: 
DEBUG:root:Re-verify VM booted
ERROR:root:Command lxc exec testbed -- lsb_release -a returned a code of 255
ERROR:root: STDOUT: 
DEBUG:root:Re-verify VM booted
DEBUG:root:Command lxc exec testbed -- lsb_release -a:
[...]

PR #743, opened 2023-09-21
Update the Virtualization Test to Clean up Logs (Bugfix)

mreed8855:
I have added an option to reduce the error messages shown in the logs while the script checks to see if the virtual machine has booted.

This fixes this issue
https://github.com/canonical/checkbox/issues/695

Jira Card
https://warthogs.atlassian.net/browse/SERVCERT-1259

This PR will replace  PR https://github.com/canonical/checkbox/pull/720

## Resolved issues
https://github.com/canonical/checkbox/issues/695

## Tests
Run checkbox and run the virtualization/verify_lxd_vm job

### DIFF
--- a/providers/base/bin/virtualization.py
+++ b/providers/base/bin/virtualization.py
@@ -737,20 +737,23 @@ class LXDTest_vm(object):
     def run_command(self, cmd, log_stderr=True):
         task = RunCommand(cmd)
         if task.returncode != 0:
-            logging.error('Command {} returned a code of {}'.format(
-                task.cmd, task.returncode))
-            logging.error(' STDOUT: {}'.format(task.stdout))
+            logging.error(
+                "Command {} returned a code of {}".format(
+                    task.cmd, task.returncode
+                )
+            )
+            logging.error(" STDOUT: {}".format(task.stdout))
             if log_stderr:
-                logging.error(' STDERR: {}'.format(task.stderr))
+                logging.error(" STDERR: {}".format(task.stderr))
             return False
         else:
-            logging.debug('Command {}:'.format(task.cmd))
-            if task.stdout != '':
-                logging.debug(' STDOUT: {}'.format(task.stdout))
+            logging.debug("Command {}:".format(task.cmd))
+            if task.stdout != "":
+                logging.debug(" STDOUT: {}".format(task.stdout))
             if task.stderr and log_stderr:
-                logging.debug(' STDERR: {}'.format(task.stderr))
+                logging.debug(" STDERR: {}".format(task.stderr))
             if not (task.stderr or task.stdout):
-                logging.debug(' Command returned no output')
+                logging.debug(" Command returned no output")
             return True
 
     def setup(self):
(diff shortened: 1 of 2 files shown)


providers/base/bin/virtualization.py
```diff
@@ -734,19 +734,20 @@ def __init__(self, template=None, image=None):
         self.default_remote = "ubuntu:"
         self.os_version = get_release_to_test()
 
-    def run_command(self, cmd):
+    def run_command(self, cmd, error_msgs=True):
```
Hook25, line 737: I would change this parameter to `log_stderr` because it avoids the `logger.error` call but also the `logger.debug` further below

providers/base/bin/virtualization.py
```diff
@@ -734,19 +734,20 @@ def __init__(self, template=None, image=None):
...
         task = RunCommand(cmd)
         if task.returncode != 0:
             logging.error('Command {} returned a code of {}'.format(
                 task.cmd, task.returncode))
             logging.error(' STDOUT: {}'.format(task.stdout))
```
Hook25, line 742: Minor: maybe we want the same pattern as below? `if task.stdout: ...`

providers/base/bin/virtualization.py
```diff
@@ -734,19 +734,20 @@ def __init__(self, template=None, image=None):
...
-            logging.error(' STDERR: {}'.format(task.stderr))
+            if error_msgs:
+                logging.error(' STDERR: {}'.format(task.stderr))
             return False
         else:
             logging.debug('Command {}:'.format(task.cmd))
             if task.stdout != '':
                 logging.debug(' STDOUT: {}'.format(task.stdout))
-            elif task.stderr != '':
+            elif task.stderr != '' and error_msgs:
```
Hook25, line 750: this `elif` is a bit weird, a task may have both `stdout and stderr`. I would change this to:
```python
if task.stdout:
  logging.debug(' STDOUT: {}'.format(task.stdout))
if task.stderr and error_msgs: # <- if you change the name, change it here as well
  logging.debug(' STDERR: {}'.format(task.stderr))
if not (task.stderr or task.stdout): 
  # we could be smart here and do elif not task.stdout
  # but I think this is a bit cleaner
  logging.debug(' Command returned no output')
```

Minor: `if string_var != ''` == `if string_var`

providers/base/bin/virtualization.py
```diff
@@ -842,8 +843,8 @@ def cleanup(self):
         Clean up test files an Virtual Machines created
         """
         logging.debug('Cleaning up images and VMs created during test')
-        self.run_command('lxc image delete {}'.format(self.image_alias))
-        self.run_command('lxc delete --force {}'.format(self.name))
+        self.run_command('lxc image delete {}'.format(self.image_alias), False)
+        self.run_command('lxc delete --force {}'.format(self.name), False)
```
Hook25, line 847: Minor and possibly out of the scope of this PR: it is a very strange behavior to use strings everywhere and then split them with `shlex.split`, this should be changed in the future to use arrays everywhere IMO

Hook25 (changes requested):
I have suggested a couple of ideas here and there, overall nice contribution, thank you.

To consider this ready I will need to ask you to write a few unit tests to improve the overall [coverage of this contribution](https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#coverage). Given that we have a 90% target on patches, covering `cleanup` and `run_command` should be enough but if you want/can, try to reach 100% (covering `start_vm` is going to be a little bit more complicated but we really need to improve the overall coverage of provider scripts and checkbox itself!).

kissiel (approved):
Awesome, +1

PR #745, opened 2023-09-25
Add miscellanea/ubuntu-desktop-minimal-recommends job (New)

baconYao:
## Description

For certain IoT Desktop images, the default installation is the ubuntu-desktop-minimal package. Therefore, it is necessary to check whether its recommended packages have been installed.

Ref: https://packages.ubuntu.com/jammy/ubuntu-desktop-minimal

## Resolved issues

N/A

## Documentation

N/A

## Tests

Sideload result on Baoshan Project and the information of `apt-cache show ubuntu-desktop-minimal`:

- https://pastebin.canonical.com/p/Kh94grNfYr/

CID: 202309-32022
Image: iot-baoshan-classic-desktop-2204-genio-desktop-09-01-20230908-116

### DIFF
--- a/providers/base/bin/check-ubuntu-desktop-recommends.sh
+++ b/providers/base/bin/check-ubuntu-desktop-recommends.sh
@@ -2,6 +2,8 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+target_package=${1:-ubuntu-desktop}
+
 noninstalled=()
 while read -r pkg; do
     # libreoffice-impress provides libreoffice-ogltrans, and libreoffice-ogltrans becomes a transitional package on Ubuntu 20.04.
@@ -9,7 +11,7 @@ while read -r pkg; do
     if ! dpkg-query -W -f='${Status}\n' "$pkg" 2>&1 | grep "install ok installed" >/dev/null 2>&1 && [ "$pkg" != "libreoffice-ogltrans" ]; then
         noninstalled+=("$pkg")
     fi
-done < <(apt-cache show ubuntu-desktop | grep ^Recommends | head -n 1 | cut -d : -f 2- | xargs | sed 's/ //g' | tr , $'\n')
+done < <(apt-cache show "${target_package}" | grep ^Recommends | head -n 1 | cut -d : -f 2- | xargs | sed 's/ //g' | tr , $'\n')
 
 if [ -n "${noninstalled[*]}" ]; then
     IFS=' '
@@ -17,5 +19,5 @@ if [ -n "${noninstalled[*]}" ]; then
     exit 1
 fi
 
-echo "All packages in Recommends of ubuntu-desktop are installed."
+echo "All packages in Recommends of ${target_package} are installed."
 exit 0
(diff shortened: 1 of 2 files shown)


providers/base/bin/check-ubuntu-desktop-recommends.sh
```diff
@@ -2,20 +2,22 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+target_package=${1}
```
pieqq, line 5: Since the default package to test is `ubuntu-desktop`, I would use that as a default in the bash variable. This way, you don't need to modify the job definition. Something like:

```
target_package=${1:-ubuntu-desktop}
```

providers/base/units/miscellanea/jobs.pxu
```diff
@@ -489,10 +489,19 @@ category_id: com.canonical.plainbox::miscellanea
 estimated_duration: 60
 id: miscellanea/ubuntu-desktop-recommends
 requires: package.name == 'ubuntu-desktop'
-command: check-ubuntu-desktop-recommends.sh
+command: check-ubuntu-desktop-recommends.sh ubuntu-desktop
```
pieqq, line 492: If you use a default shell variable value, you don't need to modify this.

providers/base/units/miscellanea/jobs.pxu
```diff
@@ -489,10 +489,19 @@ category_id: com.canonical.plainbox::miscellanea
...
 _summary: Check if Debian packages in Recommends of ubuntu-desktop are installed
 _description: Check if Debian packages in Recommends of ubuntu-desktop are installed
 
+plugin: shell
+category_id: com.canonical.plainbox::miscellanea
+estimated_duration: 60
```
pieqq, line 498: Does it really take 60 seconds to run this script?

providers/base/units/miscellanea/jobs.pxu
```diff
@@ -489,10 +489,19 @@ category_id: com.canonical.plainbox::miscellanea
...
+id: miscellanea/ubuntu-desktop-minimal-recommends
+requires: package.name == 'ubuntu-desktop-minimal'
+command: check-ubuntu-desktop-recommends.sh ubuntu-desktop-minimal
+_summary: Check if Debian packages in Recommends of ubuntu-desktop-minimal are installed
```
pieqq, line 502: This summary could be improved. What about `Check that all the recommended packages for ubuntu-desktop-minimal are installed`?

(Maybe it would be good to modify the summary of the `miscellanea/ubuntu-desktop-recommends` as well to match)

providers/base/units/miscellanea/jobs.pxu
```diff
@@ -489,10 +489,19 @@ category_id: com.canonical.plainbox::miscellanea
...
+_description: Check if Debian packages in Recommends of ubuntu-desktop-minimal are installed
```
pieqq, line 503: Since it's a copy/paste of the summary, it's not really useful. Please consider removing it (and the same should apply to `miscellanea/ubuntu-desktop-recommends` as well)

pieqq (changes requested):
Thanks for this submission!

I have a suggestion (using a default value for the shell var), as well as some wording suggestions. Please check below.

diohe0311:
/canonical/self-hosted-runners/run-workflows 66ef2d98aae2fad2719d8ff7b292d30cb2fc7c04

pieqq (approved):
All good! Thanks a lot @baconYao !

Issue #748, opened 2023-09-27
change cuda to cuda-toolkit in gpu-setup and as the requirement for the gpgpu test


bladernr:
### Bug Description

Currently the gpgpu test has a requirement for having CUDA installed but looks for the "cuda" metapackage.  The problem with this is that installing the cuda metapackage has a dependency deep in the chain that removes our packaged NVIDIA drivers and installs the ones from the nvidia developer repos.  We want testing to be done using both the  linux-nvidia kernel and our packaged nvidia drivers.

To fix this, the setup script has to change to not install cuda, but rather cuda-toolkit which does not appear to remove our driver packages.  So we need to change this requirement from cuda to cuda-toolkit in the job description.

### To Reproduce

Install canonical-certification-server and checkbox-provider-gpgpu
install our GPU drivers (ubuntu-drivers --gpgpu install nvidia:versionnumber-server)
run the gpu-setup script
(*the current installation of the cuda meta-package will show it removing our driver packages in the process of installing the cuda stack)
run test-gpgpu

### Environment

Any ubuntu server

PR #749, opened 2023-09-27
Install and require cuda-toolkit metapackage instead of the cuda metapackage fixes: #748 (BugFix)

bladernr:
## Description

previously we just installed the cuda metapacakge. That works fine until you want to test using Canonical packaged NVIDIA drivers.  one of the dependencies in the cuda chain removes our drivers and force installs the ones from the NVIDIA repos and there's no way to reverse that without breaking other things.  however, all we need for this test is the cuda-toolkit to compile GPU burn.  After some testing, we find that installing just the cuda-toolkit metapackage will get us the toolkit and libraries needed without forcing the NVIDIA driver to also be installed.  Thus we can now test linux-nvidia and the nvidia drivers installed using the ubuntu-drivers tool safely.

## Resolved issues

#748 
SERVCERT-517

## Documentation

No docs impact
## Tests

install ubuntu server
install NVIDIA drivers using ubuntu-drivers
install canonical-certification-server and checkbox-provider-gpgpu
reboot to ensure drivers are loaded
run gpu-setup
observe that the drivers installed by ubuntu-drivers are not removed when installing the cuda toolkit.

### DIFF
--- a/providers/gpgpu/tools/gpu-setup
+++ b/providers/gpgpu/tools/gpu-setup
@@ -59,7 +59,7 @@ add-apt-repository -y "deb http://developer.download.nvidia.com/compute/cuda/rep
 echo "*  Installing necessary pacakges"
 apt install -y build-essential git
 ## need to break this out to fix issue where cuda installs gdm3
-apt install -y --no-install-recommends cuda
+apt install -y --no-install-recommends cuda-toolkit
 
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
+1, the package is there in the nvidia developer ppa. ty

PR #751, opened 2023-09-28
Replace hardcoded path with find for distro.py (infra)

Hook25:
## Description

This change is due to the distro module changing location every time the snap recipe is modified. The reason for this behaviour is currently unknown but it is happening in most builds/branches we do

## Resolved issues

As described, snaps builds fail seldom due to this changing location

## Documentation

The comment above this change is good enough to still encapsulate what it does, this does not change the what, it changes the how

## Tests

Manually built via `remote-bulds` both the runtime22 and runtime16 snaps


yphus (approved):
Yet another motivation to only maintain the latest base checkbox-runtime snap. +1 for the fix

Issue #752, opened 2023-09-28
checkbox produces stack trace on malformed launcher file


jocave:
### Bug Description

Given that setup of a launcher file is a "user" action it might be nice to provide a user-friendly response when a badly formatted launcher file is supplied.

### To Reproduce

Run checkbox as below was launcher file as indicated (but without redactions):

```
17:08:42 + checkbox-cli remote 10.102.163.153 checkbox-launcher
17:08:42 Traceback (most recent call last):
17:08:42   File "/usr/bin/checkbox-cli", line 11, in <module>
17:08:42     load_entry_point('checkbox-ng==1.19.0.dev0', 'console_scripts', 'checkbox-cli')()
17:08:42   File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/checkbox_cli.py", line 140, in main
17:08:42     subcmd.invoked(ctx)
17:08:42   File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 140, in invoked
17:08:42     self.launcher.read_string(self._launcher_text)
17:08:42   File "/usr/lib/python3/dist-packages/plainbox/impl/secure/config.py", line 589, in read_string
17:08:42     self._read_commit(parser)
17:08:42   File "/usr/lib/python3/dist-packages/plainbox/impl/secure/config.py", line 658, in _read_commit
17:08:42     value = reader_fn[variable.kind](
17:08:42   File "/usr/lib/python3/dist-packages/plainbox/impl/secure/config.py", line 443, in getlist
17:08:42     return self._get(section, self._convert_to_list, option,  **kwargs)
17:08:42   File "/usr/lib/python3.8/configparser.py", line 803, in _get
17:08:42     return conv(self.get(section, option, **kwargs))
17:08:42   File "/usr/lib/python3/dist-packages/plainbox/impl/secure/config.py", line 450, in _convert_to_list
17:08:42     return shlex.split(value.replace(',', ' '))
17:08:42   File "/usr/lib/python3.8/shlex.py", line 311, in split
17:08:42     return list(lex)
17:08:42   File "/usr/lib/python3.8/shlex.py", line 300, in __next__
17:08:42     token = self.get_token()
17:08:42   File "/usr/lib/python3.8/shlex.py", line 109, in get_token
17:08:42     raw = self.read_token()
17:08:42   File "/usr/lib/python3.8/shlex.py", line 191, in read_token
17:08:42     raise ValueError("No closing quotation")
[...]

PR #753, opened 2023-09-28
Catch error of kind/type parsers and not crash (Bugfix)

Hook25:
## Description

From the linked issue:
> Given that setup of a launcher file is a "user" action it might be nice to provide a user-friendly response when a badly formatted launcher file is supplied. 

We already have such a mechanism but it fails to notice when a value that it is trying to parse is not parsable (for example "abc" for an int/float value or a non-valid `shlex.split` input as in the linked issue).

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/752

## Documentation

N/A

## Tests

I have run tox and metabox on this branch


kissiel (approved):
lovely, +1

PR #757, opened 2023-10-06
Add launcher cli arg to check_config (New)

Hook25:
## Description

We have a `check-config` command that is very useful to debug a config/check it before using it. This command does not support checking a launcher directly. 

## Resolved issues

N/A

## Documentation

Added the documentation for this option in the correct place of the tutorial (where it is originally introduced) with an example run

## Tests

Added a new test that covers the addition of the parameter, the other part was already well tested.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/check_config.py
+++ b/checkbox-ng/checkbox_ng/launcher/check_config.py
@@ -24,9 +24,9 @@ class CheckConfig:
     """Implementation of the `check-config` sub-command."""
 
     @staticmethod
-    def invoked(_):
+    def invoked(context):
         """Function that's run with `check-config` invocation."""
-        config = load_configs()
+        config = load_configs(context.args.launcher)
         print("Configuration files:")
         for source in config.sources:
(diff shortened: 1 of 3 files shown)


kissiel:
> We have a `check-config` command that is very useful to debug a config/check it before using it. This command does not support checking a launcher directly.

Cause it's not meant to do that. The original intent was that Checkbox would start up as it normally would and print all the configs instead (and where they come from).

This feels like it should be an option to the launcher subcmd (implied or not).

The way it's here would be: `checkbox-cli check-config ./some.launcher` right?

Hook25:
It is pretty nice as `checkbox-cli check-config ./some.launcher` because it shows everything, origin of config values, problems in the launcher or other configs etc. I would not introduce a new command, the launcher is a special config, you use it by putting it at the end of a cli invocation, it should work with `check-config` I think

kissiel (approved):
Good stuff, let's land this, but if we decide to make this also work for remote we need to rework it a bit, so we don't replicate the remote pipeline in order to get the values.

PR #759, opened 2023-10-11
Added camera quality test using brisque (New)

fernando79513:
## Description

We want to have in place a basic test to check if the webcam signal is properly retrieved by the camera. The idea is to use an implementation of the BRISQUE algorithm to assess if the quality of the image is adequate and if we don't have issues with the decoding.

## Resolved issues

Resolves "https://warthogs.atlassian.net/browse/CHECKBOX-456"

## Documentation

So far, this PR does not include any changes in the documentation

## Tests

The test is included in the `Camera tests (automated)`. Alternatively, you can run the Python script under `providers/base/bin/` called `camera_quality_test.py`


providers/base/units/camera/test-plan.pxu
```diff
@@ -68,6 +69,7 @@ _description: Camera tests After Suspend (automated)
 include:
  after-suspend-camera/detect                                     certification-status=blocker
  after-suspend-camera/multiple-resolution-images_.*              certification-status=blocker
+ camera/camera-quality_*                                         certification-status=non-blocker
```
yphus, line 72: Prefix our job with `after-suspend-` here. This version of your checkbox camera job is automatically instantiated if the flag `after-suspend-` is used

fernando79513: I have added the after-suspend flag

PR #760, opened 2023-10-12
Collect system information (New)

Hook25:
## Description

We have decided in [CR041](https://docs.google.com/document/d/1padyHQBDuz4sVBYWlG3aWk0sE9j55mPSzDq510x3-b8/edit) that we are going to rely, among others, on inxi to fetch the hardware of a system. The best place to inject this new step is the bootstrap sequence before any resource job is collected and executed. 

This should result in the collection being included as a map in the session storage and dumped at the end of the execution under a new key of the submission.json file in this form:

If a tool runs successfully
```
"system_information" : {
  "tool_name" : { 
    “tool_version” :  tool_version,
     "success" : True,
     “outputs” : { 
       "payload" : tool_json_output,
       "stderr" : tool_stderr
    }
  } 
}
```

If a tool fails or produced an invalid json output:
```
"system_information" : {
  "tool_name" : { 
    “tool_version” :  tool_version, 
    "success" : False,
    “outputs” : { 
      “stdout” : tool_stdout,
      “stderr” : tool_stderr
    }
  } 
}
```

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-900

## Documentation

WIP

## Tests

The change was tested via metabox both in snaps and source runs


kissiel (changes requested):
Many things need changing, most of them posted below.
Generally to much unnecessary tight coupling, for something that's practically static.
Now for the whole keeping the state of the session in the state object -> why?
The sysinfo is gathered at the start of the session, and kept there and submitted with the submission. Awesome. But it may be done at the end of the session as well, and wouldn't require any change to the suspend/resume in Checkbox. "what if sysinfo changes during the session", well, with the code submitted here we also wouldn't notice.

checkbox-ng/plainbox/impl/session/assistant.py
```diff
@@ -500,6 +501,9 @@ def start_new_session(
         self._metadata.app_id = self._app_id
         self._metadata.title = title
         self._metadata.flags = {SessionMetaData.FLAG_BOOTSTRAPPING}
+
+        self._manager.state.update_system_information(system_information.collect())
```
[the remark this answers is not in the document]
Hook25: If I understood your comment correctly: this should not be done manually here but automatically by `SessionManager.create`. I agree

checkbox-ng/plainbox/impl/session/state.py
```diff
@@ -815,6 +816,15 @@ def trim_job_list(self, qualifier):
                 self.on_job_removed(job)
                 self.on_unit_removed(job)
 
+    def update_system_information(self, system_information):
```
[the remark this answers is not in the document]
Hook25: This is done this way rather than by directly assigning the variable because I wanted to make it similar to the other properties. I'm not a fan of this pattern of having empty setters and getters in python but the code in this file seems to call for that.

checkbox-ng/plainbox/vendor/system_information/__init__.py
```diff
@@ -0,0 +1,3 @@
+from pathlib import Path
+
+INXI_PATH = str((Path(__file__).parent / "inxi").resolve())
```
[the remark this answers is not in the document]
Hook25: (python 3.5 does not support paths in subprocess calls, I dropped the type here rather than in the collector code to not write strange `str(INXI_PATH)` or wonkly `[str(x) for x in cmd]` like statements. Makes me sad as well to be honest. Either way we have to reconsider if we want to go this route, bundling inxi this way or not, the type thing we can fix once we decide)

checkbox-ng/plainbox/impl/session/assistant.py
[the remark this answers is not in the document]
Hook25: As per discussion: the responsibility has to be as local as possible -> I have moved all of the logic within the setter/getter

yphus:
> The sysinfo is gathered at the start of the session

Getting hw info at the start of the session is IMHO a good idea, to reuse the same data for resources expression (whatever the form of such expression) or to combine one collector with another, or even with the hw manifest. Needless to call again inxi if the data collected at session start can be turned into something jobs requirements can consume.

checkbox-ng/plainbox/impl/session/system_information.py
```diff
@@ -0,0 +1,241 @@
...
+            version_cmd=[str(system_information.INXI_PATH), "--vs"],
+        )
+
+
+if __name__ == "__main__":
+    collection = collect()
+    to_dump = {
+        name: collected.to_dict() for (name, collected) in collection.items()
+    }
+    print(json.dumps(to_dump, indent=4))
```
kissiel, line 241: I'm guessing this is a debugging leftover :)

kissiel (changes requested):
I think I found something in need of deletion

kissiel:
Have you tested how it affects other subcommands that create temporary sessions?
I remember other subcommands creating a "throw-away" sessions, and if this collection is run when, for instance, listing units using `checkbox-cli list` it would become extremely slow.

checkbox-ng/plainbox/impl/session/system_information.py
Hook25: This is intentional, it is pretty nice to have and not expensive at all! (I think this also came out of a discussion we had before)

checkbox-ng/plainbox/impl/session/README.md
```diff
@@ -0,0 +1,78 @@
...
+}
+```
+
+The outputs field's format depends on the success of the collection.
+
+If it ran successfully, the output field will have the follwing structure:
+
+```
+  "outputs" : {
+    "json_output" : collector_json_output,
```
kissiel, line 33: I still struggle to understand why we need this dynamic schema.

The whole "outputs" shenanigans looks unnecessary. Two streams are produced by a collector, stdout, and stderr. If the collector run successfully, the stdout has the collection json. It's not like the collector will open different streams if it's unsuccessful.

kissiel (changes requested):
The code looks great. The problem I've got is I think its complexity is unnecessary in places.
It may be best if we sit on it together, so one of two outcomes happen:
* I see what I'm missing, and why the things included here are necessary
* the outputs gets simpler

checkbox-ng/plainbox/impl/session/README.md
```diff
@@ -0,0 +1,78 @@
+# System Information Collection
+
+Checkbox collects every information on the system it can using Collectors.
+A collector is a tool that can provide a JSON-serializable set of information.
+When Checkbox starts, it will leverage these collectors and collect/store
+all information it can gather in the session storage before the first
+session checkpoint is created. From then onward these information will remain
+in memory (and on disk in the checkpoint). When generating a submission report
+checkbox will include all information in a top level field of the json called
+"system_information".
```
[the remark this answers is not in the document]
Hook25: Mostly changed as you suggested, mixed and matched the last phrase because both now read a bit weird to me. Hope this new version is clear

checkbox-ng/plainbox/impl/session/README.md
```diff
@@ -0,0 +1,78 @@
+# System Information Collection
+
+Checkbox collects every information on the system it can using Collectors.
+A collector is a tool that can provide a JSON-serializable set of information.
```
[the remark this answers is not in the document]
Hook25: Almost, fixed the phrasing, its a class that calls a tool, I hope it is now clear

checkbox-ng/plainbox/impl/session/README.md
```diff
@@ -0,0 +1,78 @@
+# System Information Collection
+
+Checkbox collects every information on the system it can using Collectors.
```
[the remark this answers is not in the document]
Hook25: Mostly adopted your version ty

checkbox-ng/plainbox/impl/session/README.md
Hook25: I think we are going back to the initial design of always reporting `stdout` and `stderr`. I don't want that. I am convinced that we want the output json to be interpreted so that we know it is valid and we can actually use it in checkbox as well in the future. We went back and fourth on this a few times already. Can you write an example output that you would like to see? Something like
```json
"system_information" : {
    "tool_success" : {
        "version" : "1",
        "success" : true,
        "outputs" : { $SUGGESTION_SUCCESS }
    },
    "tool_failure" : {
        "version" : "1",
        "success" : false,
        "outputs" : { $SUGGESTION_FAILURE }
    }
}
```

kissiel: Isn't the information replicated. I mean "success: true" is redundant with "tool_success".

Hook25: (tool success is the name of the tool, fake it like tool success is "inxi" and tool failure is "false")

kissiel: Gotcha. So like a "successful_tool".

My example:
```json
"system_information" : {
    "happy_inxi" : {
        "version" : "1",
        "retcode" : 0,
        "stdout": "{\"Computer\": \"yes\", \"Brand\": \"ACME\", \"RAM\": \"A lot\"}"
        "stderr": ""
    },
    "unhappy_lshw" : {
        "version" : "1",
        "retcode" : 1,
        "stdout": ""
        "stderr": "Can't open /sys/some/info. Operation not permited"
    },
    "unhappy_another": {
        "version": "1",
        "retcode": 2,
        "stdout" : "{\"first field\": \"crashed while printing, bad json",
        "stderr": "Ran out of memory, omg"
    }
}
```

Hook25: This way the json is unparsed and it is one of the things we want to avoid in the happy case because it makes the schema less descriptive (in case success is true, you can parse stdout and it will be a valid json). This was one of the first requirement that was put in place, when we are in the happy case, there is no string parsing to be done, the data is already parsed in a usable form. 

That is the reason why I originally proposed to have a stdout field that could either be a string or a dict, that was struck down because we don't want a union there. The final iteration of this loop was to have a `json_output` in the happy case that contained the parsed result and a failure output (in whatever form, just for debug/presentation purposes) in the unhappy case

Hook25: As per our dicussione we have decided to (via meet):
- Change `json_output` -> `payload`, as it is a better name that both signals the meaning and the type better
- Change the junja template to contain a version
- Change the jinja template to not encode the json manually
- Change the "system_information" field to have a version

kissiel (approved):
Monumental work!
Thanks for all the tweaks and improvements!

checkbox-ng/plainbox/impl/session/README.md
Hook25: Implemented all changes

Issue #697, opened 2023-09-01
fan_stress_reaction is marginal pass/fail


seankingyang:
### Bug Description

Not sure if this test case designed is only need to check the fan rpm will drop down after colling.
Actually we found some dut on this test case is marginal pass/fail.

Marginal PASS:
https://certification.canonical.com/hardware/202308-31965/submission/328555/test-results/pass/?term=fan

Marginal FAIL:
https://certification.canonical.com/hardware/202308-31965/submission/329067/test-results/fail/?term=fan

Also do some experiment on two different platforms [google sheet](https://docs.google.com/spreadsheets/d/1WBjx2kU5xbtNIs1K4-8tePpCHkYjObgUSY1jIjNqKH8/edit?usp=sharing).
There are two config's fan will not increase (obvious, most of time is keep almost the same range) fan speed.

I suggest to remove this test case from the cert plan, because:
1. This test case can't correctly identify  the true pass/fail on each machine.
2. The fan is controlled by EC/ME on PC.
3. The cooling mechanism on INTEL platform not using the increasing fan speed at the first priority.

If we need to test if the DUT haing the cooling mechanism, we should find another way~~

### To Reproduce

1. Boot in OS
5. checkbox-cli run com.canonical.certification::miscellanea/fan_stress_reaction

### Environment

1. Checkbox dev version 2.9.1
2. kernel: 6.5.0-9002-oem

hanhsuan:
Other submissions

[Marginal PASS](https://certification.canonical.com/hardware/202307-31889/submission/328088/test/83456/result/35536077/)
[Marginal FAIL](https://certification.canonical.com/hardware/202307-31898/submission/328089/test/83456/result/35536214/)

seankingyang:
Could we remove this test case from the cert plan, because:
1. This test case can't correctly identify the true pass/fail on each machine.
2. The fan is controlled by EC/ME on PC.
3. The cooling mechanism on INTEL platform not using the increasing fan speed at the first priority.

diohe0311:
Based on the discussion in MM: https://chat.canonical.com/canonical/pl/frak6f8rp7yq3q8cgmfn4den7o
Sylvain agreed to remove it for now since it's not reliable.

Created a card: https://warthogs.atlassian.net/browse/CHECKBOX-940

diohe0311:
Hi @seankingyang, I've created a PR for this: https://github.com/canonical/checkbox/pull/763

PR #763, opened 2023-10-13
Remove fan stress test from test plan(BugFix)

diohe0311:
## Description
This change is to fix: https://github.com/canonical/checkbox/issues/697

QA team notice fan stress test is marginal pass/fail, Issac has conducted experiments to prove that fan RPM is controlled by the EC not the OS. They found that CPU loading is not the only factor affecting fan RPM, but also the thermal design of the form factor. Therefore, the QA team has proposed to remove the fan test from the test plan, and Sylvain has approved it.

## Resolved issues
https://warthogs.atlassian.net/browse/CHECKBOX-940

### DIFF
--- a/providers/base/units/miscellanea/test-plan.pxu
+++ b/providers/base/units/miscellanea/test-plan.pxu
@@ -31,7 +31,6 @@ include:
     firmware/fwts_uefirtvariable.*                 certification-status=blocker
     miscellanea/oops                               certification-status=blocker
     miscellanea/oops_results.log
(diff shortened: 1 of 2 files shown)


pieqq (approved):
LGTM! Thanks for pointing to all the relevant documentation (discussion on Mattermost, etc.) as a rationale for this in [your comment on the related issue](https://github.com/canonical/checkbox/issues/697#issuecomment-1760716179).

PR #768, opened 2023-10-18
fix reboot metabox scenario (BugFix)

kissiel:
## Description
The reboot metabox scenario asserted that the controller would write `Reconnecting` when the reboot happens, but on a very fast machine, the reboot happens so fast, that the controller doesn't have to wait even one sleep iteration making it not print out `Reconnecting` and just connecting immediately.

This patch removes the assertion for the reconnecting text.

The patch also improves the formatting and introduces a tag for the test.

## Tests
This patch fixes the test.


Hook25 (approved):
+1, its incredible how fast your machine is!

PR #769, opened 2023-10-18
Fix UnboundLocalError in subcommands.py (bugfix)

Hook25:
## Description

The pybuild tester fails to test every package because an `_` assignment is shadowing the global  `_`. 
Marking the variable as `nonlocal` or `global` is not enough in the context of the exception handler because that leads the first assignment to fail, I decided to just use the function directly to fix the issue.

Additionally, once that is fixed, the tests fail (and this is a difference between pytest and uniitest). I have fixed this by mocking also the function call that causes this to happen

## Resolved issues

Fixes: https://launchpadlibrarian.net/692605124/buildlog_ubuntu-mantic-amd64.checkbox-ng_2.10.2~dev27+g1aa81f06a~ubuntu23.10.1_BUILDING.txt.gz

## Documentation

N/A

## Tests

I have verified this via `pytest`, `unittest` and by locally re-building the package using `dpkg-buildpackage` on jammy

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/subcommands.py
+++ b/checkbox-ng/checkbox_ng/launcher/subcommands.py
@@ -295,7 +295,7 @@ class Launcher(MainLoopStage, ReportsStage):
             _ = detect_restart_strategy(session_type="local")
         except LookupError as exc:
             _logger.warning(exc)
-            _logger.warning(_("Automatic restart disabled!"))
+            _logger.warning(gettext.gettext("Automatic restart disabled!"))
             return
 
         snap_name = os.getenv("SNAP_NAME")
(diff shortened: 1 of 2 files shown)


kissiel (approved):
Thanks for the fix!

checkbox-ng/checkbox_ng/launcher/subcommands.py
```diff
@@ -295,7 +295,7 @@ def _configure_restart(self, ctx):
             _ = detect_restart_strategy(session_type="local")
         except LookupError as exc:
             _logger.warning(exc)
-            _logger.warning(_("Automatic restart disabled!"))
+            _logger.warning(gettext.gettext("Automatic restart disabled!"))
```
yphus, line 298: I'd not even try to call gettext at all. Translations files for checkbox are completely outdated and i18n support will be dropped soon (next cycle).

yphus (changes requested):

checkbox-ng/checkbox_ng/launcher/subcommands.py
Hook25: I think we can let this be for now and once we remove them, we remove them all

yphus: ok for me.

PR #770, opened 2023-10-19
add submission schema and tools to generate it

kissiel:
## Description
This PR introduces the schema of submission.json that's uploaded to C3 from Checkbox.
Alongside there's an instruction on how to generate that.

## Resolved issues
Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-927


submission-schema/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Generating Python loader out of the schema
+
+Quicktype can generate a valid Python class parallel to the object described by the generated schema. It can also be used to generate schema, though it's significantly slower than the genson method.
+
+Getting quicktype
+
+```bash
+sudo apt install npm  # warning - a lot of packages will be pulled
+sudo sudo npm install -g quicktype
```
mz2, line 74: Better than to a dvise globally installing the package would be to have a `package.json` in this directory with an exact quicktype version reference, and a "script" entry with which you then run quicktype (which just does the equivalent of `quicktype --lang json-schema ...` below when you call `npm run json-schema`).

This way you're a) not polluting the OS with the globally installed package and b) you're robust to `quicktype` eventually becoming incompatible with the version of `node` you installed.

You should also state what version of Ubuntu this step has been tested, since `node` and `npm` installed from the Ubuntu archive may or may not be compatible with the package `quicktype`.

kissiel: My lack of experience using JS as "programs" is showing up. 

The way I understand you suggestion:
* save-exact quicktype
* store it in package.json
* create a "scripts" entry like `"checkbox-json-schema": "quicktype --lang json-schema (...)"`
* update the instruction in doc

submission-schema/README.md
```diff
@@ -0,0 +1,102 @@
...
+./build_schema.py ../jsons -o ./schema.json
+```
+
+### Generating Python loader out of the schema
+
+As an alternative to `genson`, `quicktype` can be used to generate the schema, as well as Python types using classes.
+
+#### Getting quicktype
+
+1. Install `nodejs` [using the nodesource.com provided Node.js binary distribution](https://github.com/nodesource/distributions#nodejs).
```
mz2, line 74: @kissiel I took the liberty to change these steps in a commit I pushed to the branch directly, was easier explained and tested as correct steps that way.

Basically, `nodejs` from nodesource is really the better way even for this usecase, and the package stays local with this workflow to the directory (no polluting the system with a globally installed package).

submission-schema/README.md
```diff
@@ -0,0 +1,102 @@
...
+  # Get the base name of the file without the extension.
+  base_name="$(basename "$file" .tar.xz)"
+
+  # Extract only the submission.json file.
+  tar -xvf "$file" --wildcards --no-anchored 'submission.json' -O > "${TARGET_DIR}/${base_name}.json"
+done
+
+```
+
+### Running genson
```
mz2, line 50: Hmm, I would just drop the `genson` instructions entirely: the project has last seen commits in March, has 8 contributors through its time to quicktype's 167. We're not really after performance, but correctness and a long term maintained tool, and quicktype is a familiar choice.

submission-schema/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+With genson installed you can now run the `build_schema.py` program.
+
+```bash
+./build_schema.py ../jsons -o ./schema.json
+```
+
+### Generating Python loader out of the schema
+
+As an alternative to `genson`, `quicktype` can be used to generate the schema, as well as Python types using classes.
```
mz2, line 70:
```suggestion

`quicktype` can be used to generate the schema, as well as Python classes representing submission data.
```

submission-schema/build_schema.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
```
mz2, line 1: I'd prefer to just drop this (we need just one method, to get us to the point where we'll start validating against the schema).

submission-schema/README.md
```diff
@@ -0,0 +1,102 @@
...
+#### Generating Python class with Quicktype using the existing schema
+
+```bash
+npm run generate-python-types-from-schema
+```
+
+
+#### Generating Python class with Quicktype using submissions
+
+Python classes can be created directly from input JSONs.
```
mz2, line 98: I removed here a comment about input JSONs being more "reliable" since I don't think that's really true: it's a moving window of the latest 2K submissions, which is a few weeks worth submissions at best actually (16K submissions in a quarter). We'll aim for reliability by sampling widely enough to support some sensible time range of submission data, and then iterating from there, whilst maintaining compatibility with a known set of submissions.  We probably need a bunch of these representative examples to housed in the repo itself, though not to run into thousands is itself a little bit of a challenge (perhaps a clustering approach of some kind would find us a maximally divergent set of examples?)

mz2 (changes requested):
I'd prefer dropping the `genson` instructions and build script dependent on it, since I'd prefer we relied on quicktype which is a very commonly used tool for the purpose (that it's slower just makes me think it does something more thoroughly :slightly_smiling_face: ).

submission-schema/README.md
mz2: Yep, I did all this in a follow-up commit to the branch.

mz2: I also switched the installation instructions from Ubuntu archive to nodesource

Hook25 (changes requested):
Lets keep one way to do this: quicktype.

How was the schema.json file generated in this PR?  Did we use genson or quicktype? If we used genson please double check that the result is the same. Also, the `build_schema.py` script should be updated to reflect this tool change (if we still need it)

kissiel:
I'll rerun the checks with submissions I used.

kissiel:
The one in the PR is the genson one.
I'm attaching the one generated with QT.
[schema-qt.json](https://github.com/canonical/checkbox/files/13513125/schema-qt.json)

.github/workflows/daily-builds.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  checkbox-core-snap-daily:
+    uses: ./.github/workflows/checkbox-core-snap-daily-builds.yml
+  checkbox-snap-daily:
+    uses: ./.github/workflows/checkbox-snap-daily-builds.yml
+  checkbox-deb-daily:
+    uses: ./.github/workflows/deb-daily-builds.yml
```
Hook25, line 12: Remove this please, accidentally committed?

submission-schema/README.md
Hook25: commit this!

submission-schema/build_schema.py
Hook25, line 1: Possibly due to removal given the discussion

Hook25 (changes requested):
Few changes to get this landed

.github/workflows/daily-builds.yml
mz2: Oops, not sure how that made it there. Fixed.

mz2 (approved):
Looks ready now, thanks for putting this together @kissiel and for the help @Hook25

PR #772, opened 2023-10-20
A daily release pipeline to trigger the package specific dialy pipelines (Infra)

mz2:
## Description

A `workflow_dispatch` activated release pipeline for dailies following instructions from https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow to the best of my understanding. This should call the three necessary build pipelines at the same git ref as the pipeline calling is at.

## Resolved issues

Does not resolve any tracked issue, associated with making release build debugging (https://warthogs.atlassian.net/browse/CHECKBOX-945) more convenient, so we get all the builds with one button press built at the same git ref predictably.

## Documentation

A reference to this being available needed in the release process docs.

## Tests

I have not dared pressed the button, please finish this to a workable state ok baiii

### DIFF
--- a/.github/workflows/checkbox-core-snap-daily-builds.yml
+++ b/.github/workflows/checkbox-core-snap-daily-builds.yml
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: '00 04 * * *'
   workflow_dispatch:
+  workflow_call:
 
 jobs:
   check_history:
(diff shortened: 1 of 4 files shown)


yphus (changes requested):
I'd leverage even more this umbrella workflow to also initiate the dailies automatically by moving the crontab from the 3 individual daily workflows to the one you proposed:

```
  schedule:
    - cron: '00 04 * * *'
```

Issue #75, opened 2022-11-28
LP1898024: remote get_result stack trace


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/checkbox-ng/+bug/1898024

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Confirmed	| 2020-10-01 09:50:32	| 6	| High	| False  	|

## Description
```
---------------------------[ Running job 158 / 195 ]----------------------------
--------------------------[ Attach C-States test log ]--------------------------
ID: com.canonical.certification::cpu/cstates_results.log
Category: CPU tests
--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/bin/checkbox-cli", line 11, in <module>
    load_entry_point('checkbox-ng==1.10.0.dev0', 'console_scripts', 'checkbox-cli')()
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/checkbox_cli.py", line 123, in main
    subcmd.invoked(ctx)
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 155, in invoked
    self.connect_and_run(ctx.args.host, port)
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 219, in connect_and_run
    keep_running = {
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 435, in resume_interacting
    self.run_jobs()
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 420, in run_jobs
    self._run_jobs(jobs_repr, total_num)
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 607, in _run_jobs
    self.wait_for_job()
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 461, in wait_for_job
    self.finish_job()
  File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/master.py", line 466, in finish_job
    job_result = self.sa.finish_job(result)
  File "/usr/lib/python3/dist-packages/plainbox/vendor/rpyc/core/netref.py", line 253, in __call__
    return syncreq(_self, consts.HANDLE_CALL, args, kwargs)
  File "/usr/lib/python3/dist-packages/plainbox/vendor/rpyc/core/netref.py", line 76, in syncreq
[...]

PR #774, opened 2023-10-20
Fix the race condition when resuming via remote (BugFix)

kissiel:
## Description
There was a chance that when the controller reconnected to the agent, it would get an information that something is running, so naturally it would proceed with waiting and finishing, but this "wait and finish" assumed the job was running, which may have finished by that point. This could happen only on resume, cause otherwise the controller would control the execution, so there would be no race.

## Resolved issues
Fixes: https://github.com/canonical/checkbox/issues/75
## Documentation

No need for documentation changes.
## Tests
Tests covering the mix of logic that caused the crash are in the PR.

I'm proposing no bespoke regression metabox scenario, as there's no reliable way to reproduce the problem automatically.

PS I also removed the 20 second sleep when auto resuming and running silent session. Silent sessions are not meant to be interactive, so there's no need to give the operator a window to kill the session. And it interfered significantly in my debugging.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -533,9 +533,7 @@ class RemoteController(ReportsStage, MainLoopStage):
         self.run_jobs()
 
     def _handle_last_job_after_resume(self, resumed_session_info):
-        if self.launcher.get_value("ui", "type") == "silent":
-            time.sleep(20)
-        else:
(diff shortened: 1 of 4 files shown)


Hook25 (approved):
+1, this is ready to be merged and very well documented (ty). If you have a second to spare add a test for `_handle_last_job_after_resume` to check that `resume_dialog` is called on non-silent uis, else this is fine by me

PR #775, opened 2023-10-20
fix mocking of Remote Session Assistant (BugFix)

kissiel:
## Description

Fix the unit tests not fully mocking parts of the remote assistant object, which made it run subprocess in the host system.


kissiel:
not from a fresh main with unsquashed commits... let's kill it and do the right one

PR #776, opened 2023-10-20
mock the RemoteSessionAssistant entirely (BugFix)

kissiel:
# Description
Fix the unit tests not fully mocking parts of the remote assistant object, which made it run subprocess in the host system.


Hook25 (approved):
+1, strange this didn't fail in CI

PR #780, opened 2023-10-23
Move deb commit checking to the edge workflow (infra)

Hook25:
## Description

Right now we can not force a deb build, additionally, the workflow will only try to build packages that did change in the last24 hours. This leads to the pipeline being impossible to retry if nothing landed in the last 24 hours and version numbers of packages being out of sync. 

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


yphus (approved):
LGTM, thanks

PR #782, opened 2023-10-24
replace generic GH runner with the self-hosted (Infra)

kissiel:
This PR replaces the use of "ubuntu-latest" github runners in favour of the self-hosted ones.
This targets only the ubuntu-latest as this matches what we offer with self-hosted (we have Jammy runners only).
All the actions for which I switched the runners are just orchestrating work done by other systems (for instance calling readthedocs to build the docs, or calling remote-build on LP), this is why I opted to use the `large` instance for those.

### DIFF
--- a/.github/workflows/checkbox-core-snap-beta-release.yml
+++ b/.github/workflows/checkbox-core-snap-beta-release.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         releases: [16, 18, 20, 22]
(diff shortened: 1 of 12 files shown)


Hook25 (approved):
+1, this is fantastic! well done

kissiel:
The metabox workflow uses a `pull_request_review` event, which is not supported by the hosted-runners, I rolled it back.

Hook25 (approved):
+1, we should ask for that rule to be supported if possible

PR #783, opened 2023-10-24
An attempt at fixing a retry fail at the snap upload step (Infra)

mz2:
# Description

An attempted fix to https://github.com/canonical/checkbox/actions/runs/6621575266/job/17986706086#step:10:246

The command is not really multiline, it is collapsed to one line. This means that we need to delimit the subcommands via `;` to make them work

# Testing

I have executed a slightly modified version of the commands in a `sh` shell (that is what the workflow uses). 

To replicate: 
- Edit the conditions with either `[ ]` for false or `[ true ]` for true to check the branches
- Remove the interpolation `{{ }}` and replace it with working expandable paths where needed (in `for`)
- replace `snapcraft` with `echo`

### DIFF
--- a/.github/workflows/checkbox-core-snap-daily-builds.yml
+++ b/.github/workflows/checkbox-core-snap-daily-builds.yml
@@ -82,8 +82,8 @@ jobs:
           attempt_delay: 600000 # 10min
           attempt_limit: 3
           command: |
-            for snap in checkbox-core-snap/series${{ matrix.releases }}/*.snap \
+            for snap in checkbox-core-snap/series${{ matrix.releases }}/*.snap ; \
             do \
-              echo "Uploading $snap..." \
-              snapcraft upload $snap --release edge \
+              echo "Uploading $snap..." ; \
+              snapcraft upload $snap --release edge ; \
             done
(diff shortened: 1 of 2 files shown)


yphus (approved):
the bash section works once loaded in a shell, tested the checkbox-snap variant wit two fake files to iterate over generated snaps:

````
$ for snap in checkbox-snap/series_classic22/*.snap ; \
>             do \
>               echo "Uploading $snap..." ; \
>               if [[ 'classic' == 'classic' ]]; then \
>                 if [[ '22' == '22' ]]; then \
>                   echo upload $snap --release 22.04/edge,latest/edge ; \
>                 else \
>                   echo upload $snap --release 22.04/edge ; \
>                 fi \
>               else \
>                 echo upload $snap --release uc22/edge ; \
>               fi ; \
>             done
Uploading checkbox-snap/series_classic22/1.snap...
upload checkbox-snap/series_classic22/1.snap --release 22.04/edge,latest/edge
Uploading checkbox-snap/series_classic22/2.snap...
upload checkbox-snap/series_classic22/2.snap --release 22.04/edge,latest/edge
````

PR #786, opened 2023-10-24
Another try at hoster runners

kissiel:
DO NOT LAND
This PR tests another theory about why self-hosted runners fail, namely the parenthesis in the pr name.


kissiel:
This can be deleted now.

PR #787, opened 2023-10-24
Build one arch per matrix expansion (infra)

Hook25:
## Description
Building multiple arches per matrix expansion leads the workflow to hang waiting for slower archs even if a build failed and makes us unable to re-run the build for a specific arch if it failed. With this modification each arch is built with its own command. 

Minor: This also fixes sh compatibility
Minor: build-for is now the correct way to call build-on

## Tests

I have re-checked the `[[` vs `[` thing and it should now work. The matrix expansion should be ok but I'm unable to run act at the moment.

### DIFF
--- a/.github/workflows/checkbox-core-snap-daily-builds.yml
+++ b/.github/workflows/checkbox-core-snap-daily-builds.yml
@@ -32,6 +32,7 @@ jobs:
     strategy:
       matrix:
         releases: [16, 18, 20, 22]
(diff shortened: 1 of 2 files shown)


Hook25:
This doens't work because snapcraft doesn't differentiate between arches when creating the repo on LP, so doing them in parallel simply results in the first one being run and all the others racing to upload whatever the first one produces

PR #788, opened 2023-10-24
use strict POSIX shell syntax (Infra)

kissiel:
## Description
This patch fixes the daily snap builds workflow by switching from bash syntax to a more compliant strict POSIX syntax for the shell code.


Hook25 (approved):
Ty +1

PR #792, opened 2023-10-25
Move to forked snapcraft build action (infra)

Hook25:
## Description

The current version of the action doesn't fail when snapcraft produces less than it expects (if snapcraft produces something, that is ok). This is not enough for us because that way we will never retry to build the snaps when some fail to build. 

I have forked the upstream version of the action and introduced a new check to monitor if snapcraft produced all that we need, else it will yield an error and fail. This moves our repository to my forked version of the action till a new release of the upstream action that contains this patch is released.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


kissiel (approved):
The action on your fork with the hash looks good, and this looks good too!

PR #798, opened 2023-10-26
move toxing to self-hosted runners

kissiel:
## Description
This PR switches the runners that run tox testing from the GH provided 20.04 ones to the ones we host.


kissiel:
This can't land, even though docs say the old images should be there, they are not, and the verification failed:
https://github.com/canonical/checkbox/actions/runs/6652352398/job/18076152063

PR #799, opened 2023-10-26
Sync beta edge workflow (infra)

Hook25:
## Description

We have upgraded the daily build workflow significantly in these days, including separated builds, retries and new descriptions. This PR ports these improvements over to the beta workflow.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/.github/workflows/checkbox-snap-beta-release.yml
+++ b/.github/workflows/checkbox-snap-beta-release.yml
@@ -61,7 +61,7 @@ jobs:
             /home/runner/.local/state/snapcraft/log/
             checkbox-snap/series_${{ matrix.type }}${{ matrix.releases }}/checkbox*.txt
       - uses: actions/upload-artifact@v3
-        name: Upload the snaps as artefacts
+        name: Upload the snaps as artifacts
         with:
           name: series_${{ matrix.type }}${{ matrix.releases }}
           path: checkbox-snap/series_${{ matrix.type }}${{ matrix.releases }}/*.snap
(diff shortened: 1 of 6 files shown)


kissiel (approved):
Looks good!

.github/workflows/checkbox-snap-beta-release.yml
```diff
@@ -49,32 +61,29 @@ jobs:
             /home/runner/.local/state/snapcraft/log/
             checkbox-snap/series_${{ matrix.type }}${{ matrix.releases }}/checkbox*.txt
       - uses: actions/upload-artifact@v3
+        name: Upload the snaps as artefacts
```
mz2, line 64:
```suggestion
        name: Upload the snaps as artifacts
```

.github/workflows/checkbox-snap-beta-release.yml
```diff
@@ -18,7 +18,11 @@ jobs:
     env:
       SERIES: series_${{ matrix.type }}${{ matrix.releases }}
```
mz2, line 19: Nitpick #1: You don't use this env everywhere it could be.

Nitpick #2: Actually `${{ matrix.type }}${{ matrix.releases }}` looks also on its own to be very recurrently used. Not sure it makes sense to create a very dense set of env vars, but `TYPE_RELEASE` perhaps on balance clearer.

mz2 (approved):
Can't spot anything obviously wrong, but this is the thankless kind of PR where probably something goes wrong after N hours on the first attempt :clown_face:

.github/workflows/checkbox-snap-beta-release.yml
Hook25: (woa the #2 PR on this repo is actually the deb workflow)

I don't know if the places where the macro is not used support it (for example path). I would play safe with this and we can change it once we are sure

kissiel (approved):
And now with actual +1

PR #807, opened 2023-10-30
use noninteractive apt in workflows (Infra)

kissiel:
## Description
The spelling workflow failed because it used an interactive invocation of apt.
This patch adds `-y -qq` to the invocation.


p-gentili (approved):
LGTM

PR #808, opened 2023-10-30
add venv installation to workflows so the doc Makefile works (Infra)

kissiel:
## Description
The documentation workflows relied on the Makefile that assumed that the venv module is available, but it is not by default.
This patch adds installation of the python3-venv.
It also adds the `apt update` so we are ensured the cache is fresh.


p-gentili (approved):
LGTM, thanks.

PR #810, opened 2023-11-03
Fixed USB-C OTG cleanup job was unexpectedly executed (BugFix)

rickwu666666:
The USB-C OTG cleanup job was unexpectedly executed. It was executed even if the test job had been skipped because the manifest entry was not supported.

## Description
Add a manifest to the USB-C OTG clean-up jobs to ensure that those jobs will be skipped if USB-C OTG is not supported in the manifest menu.

## Resolved issues
The USB-C OTG cleanup job was unexpectedly executed. It was executed even if the test job had been skipped because the manifest entry was not supported.

## Documentation
The manifest entry for USB-C OTG indicates 'not supported' for side loading."
https://pastebin.canonical.com/p/yWJfdkq4fq/
https://pastebin.canonical.com/p/4Pb5tny86G/

## Tests
1. Side loading this branch
2. Run checkbox-cli and select test plan "com.canonical.certification::usb-c-manual "
3. Select "usb-c-otg" related jobs
4. Select "No" for the manifest entry "Does the platform support USB-C OTG?"
5. Run the test and all select jobs should be skipped due to the manifest entry is no.

### DIFF
--- a/providers/base/units/usb/usb-c.pxu
+++ b/providers/base/units/usb/usb-c.pxu
@@ -251,6 +251,8 @@ _siblings: [
        "_summary": "Cleanup USB OTG serial interface setup after serial device test (after suspend)",
        "after": "after-suspend-usb-c-otg/g_serial"
      }]
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_usbc_otg == 'True'
 
 id: usb-c-otg/g_mass_storage
 _summary: Check DUT can be detected as a mass storage device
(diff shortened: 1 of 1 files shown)


pieqq:
/canonical/self-hosted-runners/run-workflows 03cb9da79201f3e7ef2c82eaabc4f56293ea2a98

pieqq (approved):
Indeed, since the cleanup job is using `after: usb-c-otg/g_serial` instead of `depends: usb-c-otg/g_serial` (because we want to execute it regardless of the outcome of the previous job), it might be unexpectedly triggered. Good catch.

+1

PR #812, opened 2023-11-06
Install the gh command line tool (Infra)

yphus:
## Description

The `gh` tool is not present on self hosted runners and it's required by the stable release workflow to update Github draft releases.


Hook25 (approved):
+1

PR #818, opened 2023-11-08
Select packages published in the last 4 weeks to copy from beta to stable (Infra)

yphus:
## Description

Promoting packages from the testing (beta) ppa to stable is done by copies.
The filter only considers packages published to the archive in the last week.
This is sometime not enough and should be extended to a month.  

## Resolved issues

3.0.0 deb binaries were not copied over to stable ppa because they were published two weeks ago 

## Documentation

Note added to detail the copy criteria (packages published since a month)

## Tests

Tested the patched version to force the 3.0.0 release to stable.


Hook25 (approved):
+1

Issue #816, opened 2023-11-08
Add platform profile testing to certification tests for Lenovo's platforms


binli:
### Bug Description

Original bug: https://bugs.launchpad.net/sutton/+bug/2036435

With Platform Profile (low-power, balanced, performance) support being added to the kernel and GNOME we'd like this to be tested to confirm it is working during HW enablement/certification.
It is expected functionality on these platforms and we need to highlight any issues with it not working to the FW team so it can be fixed before preload signoff.

To test

1) Verify profiles can be changed and stay in the chosen setting.
2) Verify that profiles remain over a suspend and resume cycle. 

### To Reproduce

It should be a feature request, while new a bug.

### Environment

-Hardware: Lenovo's platforms
-OS: 22.04 or later one

Hook25:
We have decided to clean up our issue tracker to systematize our approach to backlog grooming and planning. If this issue is still relevant to you, please reopen it and tag one of the maintainers or reach out on ~Checkbox on Mattermost. Going forward all issues that have received no activity for more than a year will be marked as 'stale' and closed after one week

PR #820, opened 2023-11-08
Switching power mode for Lenovo's platforms (New) (#816)

binli:
## Description

## Resolved issues

https://github.com/canonical/checkbox/issues/816

## Documentation

This testcase could help verify power mode can be changed and stay in the chosen setting.

## Tests

$ checkbox-cli run com.canonical.certification::power-management/switch_power_mode
```
Outcome: job passed
==============[ Running job 3 / 3. Estimated time left: 0:00:10 ]===============
---------------------[ power-management/switch_power_mode ]---------------------
ID: com.canonical.certification::power-management/switch_power_mode
Category: com.canonical.plainbox::power-management
Power mode choices: ['low-power', 'balanced', 'performance']
Switch to power-saver successfully.
Switch to balanced successfully.
Switch to performance successfully.
Outcome: job passed
Finalizing session that hasn't been submitted anywhere: checkbox-run-2023-11-08T13.22.49
 ☑ : Collect information about kernel modules
 ☑ : power-management/switch_power_mode
```


providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+    if not os.path.isdir(sysfs_root):
+        return
+    choices_filename = os.path.join(sysfs_root, "platform_profile_choices")
+    if (not os.path.isfile(choices_filename) or
+            not os.access(choices_filename, os.R_OK)):
+        return
+    profile_filename = os.path.join(sysfs_root, "platform_profile")
+    if (not os.path.isfile(profile_filename) or
+            not os.access(profile_filename, os.R_OK)):
+        return
```
Hook25, line 25: 1) Use `pathlib.Path` and not `os.path` here, it is way more handy
2) As the [documentation of `os.access` says](https://docs.python.org/3/library/os.html#os.access), don't check before accessing, try to open and catch `FileNotFoundError` and `PermissionError`
3) This exits the program with `$?` == 0 (so the test will pass), is this intended?

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+
+    return_value = 0
+
+    # Get the current power mode in quiet mode.
+    old_profile = os.popen("powerprofilesctl get").read().strip().split()[0]
```
Hook25, line 30: Why did you decide to use `os.system` and `os.popen` in this script? If you don't have a reason, please use the `subprocess` module (here you can use `subprocess.check_output`, it will also check the return code for you and raise an exception if something goes wrong!

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Read the power mode from /sys/firmware/acpi/platform_profile_choices
+    with open(choices_filename, "rt", encoding="utf-8") as stream:
+        choices = stream.read().strip().split()
+        if len(choices) < 1:
```
Hook25, line 35: Minor: `if not choices`

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+            print("No power mode to switch.")
+            return
```
Hook25, line 37: This will exit with no error. Possibly not what we need. You can use `raise SystemExit("No power mode to switch.")`, this will print the text and exit with `$? == 1`

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+    print(f'Power mode choices: {choices}')
```
Hook25, line 38: The checkbox codebase has to be compliant with python3.5.1 (sadly), so we can't use f-strings yet. Please change this to `print('Power mode choices: {}'.format(choices))`

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+    # Switch the power mode with powerprofilesctl
+    for choice in choices:
+        # Convert the power mode, e.g. low-power = power-saver,
+        # balanced and performance keep the same.
+        if choice == "low-power":
+            value = "power-saver"
+        else:
+            value = choice
+
+        os.system(f"powerprofilesctl set {value}")
```
Hook25, line 48: Don't use `os.system`, use `subprocess.check_call` instead

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+
+        os.system("sleep 2")
```
Hook25, line 50: To sleep you can use `time.sleep`, also, are you sure you need a fixed sleep here? Can't we poll the active profile somehow (avoiding the static sleep)?

providers/base/bin/switch_power_mode.py
Hook25: I saw there are more, please replace them all with a format call

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
...
+        with open(profile_filename, "rt", encoding="utf-8") as stream:
+            current_profile = stream.read().strip().split()
+        if current_profile[0] == choice:
+            print(f"Switch to {value} successfully.")
+        else:
+            print(f"Failed to switch to {value}.")
+            return_value = 1
+
+    # Switch back to the original power mode
+    os.system(f"powerprofilesctl set {old_profile}")
```
Hook25, line 61: There is another problem with this code. If the program crashes after setting the powerprofile, it will remain changed. Please create a `contextmananger` that restores the power profile.

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# All rights reserved.
+#
+# Written by:
+#   Bin Li <bin.li@canonical.com>
+
+"""Modules providing a function running os or sys commands."""
```
Hook25, line 8: This comment has to be a description of the module I think, this comment is a bit stating the obvious.

Hook25 (changes requested):
Thank you for this contribution. A few things need changing here, the most important one is testing. There is no test here and that is why codecov is complaining. Please do add them. Looking at this contribution reaching 100% coverage should be very easy with a bit of refactoring! 

I'm fixing the title of the PR myself and I've added a few comments here and there to tweak the style of the code. If you need help or guidance, feel free to reach out!

providers/base/bin/switch_power_mode.py
binli: @Hook25 Thanks for your feedback.  I ran pylint to my code and found below error. Anyway, I will use .format to compliant with python3.5.1.

providers/base/bin/switch_power_mode.py:37:10: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)

providers/base/bin/switch_power_mode.py
binli: It's also used to  fix the pylint error, :), I will remove it.

providers/base/bin/switch_power_mode.py:1:0: C0114: Missing module docstring (missing-module-docstring)

providers/base/bin/switch_power_mode.py
binli: @Hook25 
I  agree with you for the first two suggestions.
For the 3rd one, I don't wat the testcase report failure if the platform_profile_choices does not exist, and skip it, any other ideas for this one?

providers/base/bin/switch_power_mode.py
Hook25: I don't think you should. Update it to describe the module (for example, the description of the test you wrote is pretty nice, you can use that)

binli: And after main(), there are also a pylint error, so I did that after main() to describe my test. I might something else to describe my test.

providers/base/bin/switch_power_mode.py:14:0: C0116: Missing function or method docstring (missing-function-docstring)

providers/base/bin/switch_power_mode.py
Hook25: pylint is correct, he can't know that you have to support python3.5

providers/base/bin/switch_power_mode.py
Hook25: What you need is a resource. A test in checkbox can either pass, fail or be skipped. To skip it you have to add a `requires` constraint to the pxu and a resource to require. 

The usual pattern to do this is to add to your script a command line arg that creates a resource output, something like:
```
switch_power_mode.py --resource
```

The code could be something like this:
```python
def print_resource():
    sysfs_root = Path("/sys/firmware/acpi/")
    choices_path = sysfs_root / "platform_profile_choices"
    profile_path = sysfs_root / "platform_profile"
    supported = (sysfs_root.exists()) and profile_path.exists() and choices_path.exists())
    print("acpi_profiles_supported: {}".format(supported)
```

Then you can create a resource job like the following
```yaml
id: switch_power_mode
plugin: resource
_summary:
    Discover if the system supports power modes via acpi
estimated_duration: 1s
command:
    switch_power_mode.py --resource
```

Finally in your pxu you can simply add a constraint like:
```yaml
requires:
    switch_power_mode.acpi_profiles_supported == 'True''
```

If the feature is not supported, your test will be skipped

binli: Done.

providers/base/bin/switch_power_mode.py
binli: Agree with it, I've finished it.

pieqq:
@binli the last automated checks show that `platform_profile_resource.py` fails flake8 tests. Please fix this, and provide unit tests for the two new Python scripts, and then we should be good to go :)

There is a section in the Contributing guide about [writing tests for providers](https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#writing-and-running-unit-tests-for-providers), and [getting a coverage report](https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#coverage) so you can see what is tested and what's not. Let me know if you need help!

kissiel:
After talking with Binli we learned that he's busy ATM, so his PRs are on hold, and because they need work done on them I'm switching the state to "Draft"

providers/base/bin/switch_power_mode.py
binli: Changed to SystemExit, thanks!

providers/base/bin/switch_power_mode.py
binli: Fixed.

kissiel:
This got no love, and the code requires quite a bit of fixing, so this may need a rework from a ground up in the future.

binli:
Sorry for late, I opened another one to trace it. Pleae help review it again, thanks!

https://github.com/canonical/checkbox/pull/1103

Issue #327, opened 2023-02-10
Support pipewire for audio testing for ubuntu release later than 22.10 


kevinyehk:
### Enhancement Proposal

Since pulse audio has been retired in 22.10, so we need to change to use pipewire for audio testing.

pieqq:
As usual, the Arch Wiki contains some good info about pipewire, and how it replaces pulseaudio with its own pulseaudio-compatible deamon: https://wiki.archlinux.org/title/PipeWire

PR #826, opened 2023-11-16
Migrate pulseaudio to pipewire (New)

hanhsuan:
## Description

1. using `check_audio_deamon.sh` to check what environment is under testing to use right function for testing
2. Under pipewire, using `pw-dump` and `wpctl` to replace `pactl` command to collect information and settings.
3. the output of `pw-dump` is json format, therefore `pactl.py` isn't needed a alternative for pipewire.
4. `audio_settings.py` is buggy for selecting right device, change to use `show_default_audio_device.sh` to show the default device and let user to check.
5. using `wpctl status` to replace `audio_settings.py store` command

## Resolved issues

#327
After Ubuntu 22.04, the audio will change from pulseaudio to pipewire. Therefore, the old test case should be migrated from pulseaudio to pipewire.

[JIRA](https://warthogs.atlassian.net/browse/CQT-2094)

## Tests

[Jira Card CQT-2396](https://warthogs.atlassian.net/browse/CQT-2396)

### DIFF
--- a/providers/base/bin/check_audio_deamon.sh
+++ b/providers/base/bin/check_audio_deamon.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+if [ -z "$(command -v pactl)" ]; then
+    if [ -n "$(command -v pw-cli)" ] ;then
+        echo "Daemon is pipewire, use pipewire function"
+        exit 0
+    else
+        # shouldn't be here,
+        # if requires: package.name in ['pulseaudio-utils', 'pipewire']
+        echo "No pipewire or pulseaudio installed!! Stop test"
+        exit 2
+    fi
+elif pactl info | grep -iq pipewire ; then
+    echo "Daemon is pipewire, use pipeiwre function"
+    exit 0
+else
+    echo "Daemon is pulseaudio, use pulseaudio function"
+    exit 1
+fi
--- a/providers/base/bin/pipewire_test.py
+++ b/providers/base/bin/pipewire_test.py
@@ -1,0 +1,608 @@
+#!/usr/bin/env python3
+# This file is part of Checkbox.
+#
+# Copyright 2023 Canonical Ltd.
+# Written by:
+#   Hanhsuan Lee <hanhsuan.lee@canonical.com>
+#
+# Checkbox is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3,
+# as published by the Free Software Foundation.
+#
+# Checkbox is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import division, print_function
+import argparse
+import collections
+import json
+import logging
+import math
+import re
+import subprocess
+import sys
+import time
+try:
+    import gi
+    gi.require_version('GLib', '2.0')
+    gi.require_version('Gst', '1.0')
+    from gi.repository import GObject
+    from gi.repository import Gst
+    from gi.repository import GLib
+    Gst.init(None)  # This has to be done very early so it can find elements
+except ImportError:
+    print("Can't import module: %s. it may not be available for this"
+          "version of Python, which is: " % sys.exc_info()[1], file=sys.stderr)
+    print((sys.version), file=sys.stderr)
+    sys.exit(127)
+
+
+# Frequency bands for FFT
+BINS = 256
+# How often to take a sample and do FFT on it.
+FFT_INTERVAL = 100000000  # In nanoseconds, so this is every 1/10th second
+# Sampling frequency. The effective maximum frequency we can analyze is
+# half of this (see Nyquist's theorem)
+SAMPLING_FREQUENCY = 44100
+# The default test frequency is in the middle of the band that contains 5000Hz
+# This frequency was determined experimentally to be high enough but more
+# reliable than others we tried.
+DEFAULT_TEST_FREQUENCY = 5035
+# only sample a signal when peak level is in this range (in dB attenuation,
+# 0 means no attenuation (and horrible clipping).
+REC_LEVEL_RANGE = (-2.0, -12.0)
+# For our test signal to be considered present, it has to be this much higher
+# than the base level (minimum magnitude). This is in dB.
+MAGNITUDE_THRESHOLD = 2.5
+# Volume for the sample tone (in %)
+PLAY_VOLUME = 70
+
+
+class PIDController(object):
+    """ A Proportional-Integrative-Derivative controller (PID) controls a
+    process's output to try to maintain a desired output value (known as
+    'setpoint', by continually adjusting the process's input.
+
+    It does so by calculating the "error" (difference between output and
+    setpoint) and attempting to minimize it manipulating the input.
+
+    The desired change to the input is calculated based on error and three
+    constants (kp, ki and kd).  These values can be interpreted in terms of
+    time: P depends on the present error, I on the accumulation of past errors,
+    and D is a prediction of future errors, based on current rate of change.
+    The weighted sum of these three actions is used to adjust the process via a
+    control element.
+
+    In practice, kp, ki and kd are process-dependent and usually have to
+    be tweaked by hand, but once reasonable constants are arrived at, they
+    can apply to a particular process without further modification.
+
+    """
+    def __init__(self, kp, ki, kd, setpoint=0):
+        """ Creates a PID controller with given constants and setpoint.
+
+           Arguments:
+           kp, ki, kd: PID constants, see class description.
+           setpoint: desired output value; calls to input_change with
+                     a process output reading will return a desired change
+                     to the input to attempt matching output to this value.
+        """
+        self.setpoint = setpoint
+        self.kp = kp
+        self.ki = ki
+        self.kd = kd
+        self._integral = 0
+        self._previous_error = 0
+        self._change_limit = 0
+
+    def input_change(self, process_feedback, dt):
+        """ Calculates desired input value change.
+
+            Based on process feedback and time interval (dt).
+        """
+        error = self.setpoint - process_feedback
+        self._integral = self._integral + (error * dt)
+        derivative = (error - self._previous_error) / dt
+        self._previous_error = error
+        input_change = (self.kp * error) + \
+                       (self.ki * self._integral) + \
+                       (self.kd * derivative)
+        if self._change_limit and abs(input_change) > abs(self._change_limit):
+            sign = input_change / abs(input_change)
+            input_change = sign * self._change_limit
+        return input_change
+
+    def set_change_limit(self, limit):
+        """Ensures that input value changes are lower than limit.
+
+           Setting limit of zero disables this.
+        """
+        self._change_limit = limit
+
+
+class VolumeController(object):
+    pw_types = {'input': 'SOURCE', 'output': 'SINK'}
+
+    def __init__(self, type, logger=None):
+        """Initializes the volume controller.
+
+           Arguments:
+           type: either input or output
+
+        """
+        self.type = type
+        self._volume = None
+        self.identifier = None
+        self.logger = logger
+
+    def set_volume(self, volume):
+        if not 0 <= volume <= 100:
+            return False
+        command = [
+            "wpctl",
+            "set-volume",
+            "@DEFAULT_AUDIO_%s@" % (self.pw_types[self.type]),
+            str(int(volume)) + "%",
+        ]
+        self._wpctl_output(command)
+        self._volume = volume
+        return True
+
+    def get_volume(self):
+        return self._volume
+
+    def mute(self, mute):
+        mute = str(int(mute))
+        command = [
+            "wpctl",
+            "set-mute",
+            "@DEFAULT_AUDIO_%s@" % (self.pw_types[self.type]),
+            mute,
+        ]
+        self._wpctl_output(command)
+        return True
+
+    def _wpctl_output(self, command):
+        # This method mainly calls wpctl (hence the name). Since wpctl may
+        # return a failure if the audio layer is not yet initialized, we will
+        # try running a few times in case of failure. All our invocations of
+        # wpctl should be "idempotent" so repeating them should not have
+        # any bad effects.
+        for attempt in range(0, 3):
+            try:
+                return subprocess.check_output(command,
+                                               universal_newlines=True)
+            except (subprocess.CalledProcessError):
+                time.sleep(5)
+            except FileNotFoundError:
+                break
+        self.logger.error("Fail to execute: {}".format(command))
+        sys.exit(1)
+
+
+class FileDumper(object):
+    def write_to_file(self, filename, data):
+        try:
+            with open(filename, "w") as f:
+                for i in data:
+                    print(i, file=f)
+            return_value = True
+        except (TypeError, IOError):
+            return_value = False
+        return return_value
+
+
+class SpectrumAnalyzer(object):
+    def __init__(self, points, sampling_frequency=44100,
+                 wanted_samples=50):
+        self.spectrum = [0] * points
+        self.number_of_samples = 0
+        self.wanted_samples = wanted_samples
+        self.sampling_frequency = sampling_frequency
+        # Frequencies should contain *real* frequency which is half of
+        # the sampling frequency
+        self.frequencies = [((sampling_frequency / 2.0) / points) * i
+                            for i in range(points)]
+
+    def _average(self):
+        return sum(self.spectrum) / len(self.spectrum)
+
+    def sample(self, sample):
+        if len(sample) != len(self.spectrum):
+            return
+        self.spectrum = [((old * self.number_of_samples) + new) /
+                         (self.number_of_samples + 1)
+                         for old, new in zip(self.spectrum, sample)]
+        self.number_of_samples += 1
+
+    def frequencies_with_peak_magnitude(self, threshold=1.0):
+        # First establish the base level
+        per_magnitude_bins = collections.defaultdict(int)
+        for magnitude in self.spectrum:
+            per_magnitude_bins[magnitude] += 1
+        base_level = max(per_magnitude_bins,
+                         key=lambda x: per_magnitude_bins[x])
+        # Now return all values that are higher (more positive)
+        # than base_level + threshold
+        peaks = []
+        for i in range(1, len(self.spectrum) - 1):
+            first_index = i - 1
+            last_index = i + 1
+            if self.spectrum[first_index] < self.spectrum[i] and \
+                    self.spectrum[last_index] < self.spectrum[i] and \
+                    self.spectrum[i] > base_level + threshold:
+                peaks.append(i)
+
+        return peaks
+
+    def frequency_band_for(self, frequency):
+        """Convenience function to tell me which band
+           a frequency is contained in
+        """
+        # Note that actual frequencies are half of what the sampling
+        # frequency would tell us. If SF is 44100 then maximum actual
+        # frequency is 22050, and if I have 10 frequency bins each will
+        # contain only 2205 Hz, not 4410 Hz.
+        max_frequency = self.sampling_frequency / 2
+        if frequency > max_frequency or frequency < 0:
+            return None
+        band = float(frequency) / (max_frequency / len(self.spectrum))
+        return int(math.ceil(band)) - 1
+
+    def frequencies_for_band(self, band):
+        """Convenience function to tell me the delimiting frequencies
+           for a band
+        """
+        if band >= len(self.spectrum) or band < 0:
+            return None
+        lower = self.frequencies[band]
+        upper = lower + ((self.sampling_frequency / 2.0) / len(self.spectrum))
+        return (lower, upper)
+
+    def sampling_complete(self):
+        return self.number_of_samples >= self.wanted_samples
+
+
+class GStreamerMessageHandler(object):
+    def __init__(self, rec_level_range, logger, volumecontroller,
+                 pidcontroller, spectrum_analyzer):
+        """Initializes the message handler. It knows how to handle
+           spectrum and level gstreamer messages.
+
+           Arguments:
+           rec_level_range: tuple with acceptable recording level
+                            ranges
+           logger: logging object with debug, info, error methods.
+           volumecontroller: an instance of VolumeController to use
+                             to adjust RECORDING level
+           pidcontroller: a PID controller instance which helps control
+                          volume
+           spectrum_analyzer: instance of SpectrumAnalyzer to collect
+                              data from spectrum messages
+
+        """
+        self.current_level = sys.maxsize
+        self.logger = logger
+        self.pid_controller = pidcontroller
+        self.rec_level_range = rec_level_range
+        self.spectrum_analyzer = spectrum_analyzer
+        self.volume_controller = volumecontroller
+
+    def set_quit_method(self, method):
+        """ Method that will be called when sampling is complete."""
+        self._quit_method = method
+
+    def bus_message_handler(self, bus, message):
+        if message.type == Gst.MessageType.ELEMENT:
+            message_name = message.get_structure().get_name()
+            if message_name == 'spectrum':
+                # TODO: Due to an upstream bug, a structure's get_value method
+                # doesn't work if the value in question is an array (as is the
+                # case with the magnitudes).
+                # https://bugzilla.gnome.org/show_bug.cgi?id=693168
+                # We have to resort to parsing the string representation of the
+                # structure. It's an ugly hack but it works.
+                # Ideally we'd be able to say this to get fft_magnitudes:
+                # message.get_structure.get_value('magnitude').
+                # If an upstream fix ever makes it into gstreamer,
+                # remember to remove this hack and the parse_spectrum method
+                struct_string = message.get_structure().to_string()
+                structure = parse_spectrum_message_structure(struct_string)
+                fft_magnitudes = structure['magnitude']
+                self.spectrum_method(self.spectrum_analyzer, fft_magnitudes)
+
+            if message_name == 'level':
+                # peak_value is our process feedback
+                # It's returned as an array, so I need the first (and only)
+                # element
+                peak_value = message.get_structure().get_value('peak')[0]
+                self.level_method(peak_value, self.pid_controller,
+                                  self.volume_controller)
+
+    # Adjust recording level
+    def level_method(self, level, pid_controller, volume_controller):
+        # If volume controller doesn't return a valid volume,
+        # we can't control it :(
+        current_volume = volume_controller.get_volume()
+        if current_volume is None:
+            self.logger.error("Unable to control recording volume. "
+                              "Test results may be wrong")
+            return
+        self.current_level = level
+        change = pid_controller.input_change(level, 0.10)
+        if self.logger:
+            self.logger.debug(
+                        "Peak level: %(peak_level).2f, "
+                        "volume: %(volume)d%%, Volume change: %(change)f%%" % {
+                            'peak_level': level,
+                            'change': change,
+                            'volume': current_volume})
+        volume_controller.set_volume(current_volume + change)
+
+    # Only sample if level is within the threshold
+    def spectrum_method(self, analyzer, spectrum):
+        if self.rec_level_range[1] <= self.current_level \
+           or self.current_level <= self.rec_level_range[0]:
+            self.logger.debug(
+                "Sampling, recorded %d samples" % analyzer.number_of_samples)
+            analyzer.sample(spectrum)
+        if analyzer.sampling_complete() and self._quit_method:
+            self.logger.info("Sampling complete, ending process")
+            self._quit_method()
+
+
+class GstAudioObject(object):
+    def __init__(self):
+        self.class_name = self.__class__.__name__
+
+    def _set_state(self, state, description):
+        self.pipeline.set_state(state)
+        message = "%s: %s" % (self.class_name, description)
+        if self.logger:
+            self.logger.info(message)
+
+    def start(self):
+        self._set_state(Gst.State.PLAYING, "Starting")
+
+    def stop(self):
+        self._set_state(Gst.State.NULL, "Stopping")
+
+
+class Player(GstAudioObject):
+    def __init__(self, frequency=DEFAULT_TEST_FREQUENCY, logger=None):
+        super(Player, self).__init__()
+        self.pipeline_description = (
+            "audiotestsrc wave=sine freq=%s "
+            "! audioconvert "
+            "! audioresample "
+            "! autoaudiosink" % int(frequency))
+        self.logger = logger
+        if self.logger:
+            self.logger.debug(self.pipeline_description)
+        self.pipeline = Gst.parse_launch(self.pipeline_description)
+
+
+class Recorder(GstAudioObject):
+    def __init__(self, output_file, bins=BINS,
+                 sampling_frequency=SAMPLING_FREQUENCY,
+                 fft_interval=FFT_INTERVAL, logger=None):
+        super(Recorder, self).__init__()
+        pipeline_description = ('''autoaudiosrc
+        ! queue
+        ! level message=true
+        ! audioconvert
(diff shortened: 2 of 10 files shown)


Hook25:
/canonical/self-hosted-runners/run-workflows [94ea095](https://github.com/canonical/checkbox/pull/826/commits/94ea0952a96538512d319153b9e5cd66c7be0c49)

diohe0311:
/canonical/self-hosted-runners/run-workflows a48b4b27e6178f0941bc6604bcffc78bd5423cc1

diohe0311:
> audio_settings.py is buggy for selecting right device

Can you tell more about this issue? in which env?

hanhsuan:
For DP, HDMI and Thunderbolt monitor, the name shows in OS are the same with only index different. Therefore, audio_settings.py will never know which one is correct then force to select one that might not the right one. If the validator isn't noticed this issue, the validator might make the wrong judgement.

diohe0311:
> For DP, HDMI and Thunderbolt monitor, the name shows in OS are the same with only index different. Therefore, audio_settings.py will never know which one is correct then force to select one that might not the right one. If the validator isn't noticed this issue, the validator might make the wrong judgement.

Is this the only issue you have with `audio_settings.py`?
If `audio_settings.py` can provide more information, would you still need `show_default_audio_device.sh`?
It seems that the scripts have overlapping functionalities, perhaps instead of introducing a new script, we could work on improving the existing one?
@Hook25 @kissiel Do you have any suggestions?

hanhsuan:
Other functions depend on pulseaudio that couldn't work on pipewire.

diohe0311:
> Other functions depend on pulseaudio that couldn't work on pipewire.

I'm not familiar with PulseAudio and Pipewire, just start to learning it because of this PR :)
I found Pipewire has a `pipewire-pulse.service`, it supports PulseAudio applications.
https://wiki.archlinux.org/title/PipeWire

We can discuss more once I figure out how to use it.

hanhsuan:
1. `pactl` is not installed by default after lunar
2. The output format of `pactl info` is different between `pulseaudio` and `pipewire` with `pipewire-pulse`, it makes the `pactl.py` parsing failed.

diohe0311 (changes requested):

providers/base/units/audio/jobs.pxu
```diff
@@ -39,13 +39,19 @@ requires:
...
- package.name == 'pulseaudio-utils'
+ package.name in ['pulseaudio-utils', 'pipewire']
 command:
-  audio_settings.py store --file="$PLAINBOX_SESSION_SHARE"/pulseaudio_settings
-  audio_settings.py set --device=pci --volume=50
-  gst_pipeline_test.py -t 2 'audiotestsrc wave=sine freq=512 ! audioconvert ! audioresample ! autoaudiosink'
-  EXIT_CODE=$?
-  audio_settings.py restore --file="$PLAINBOX_SESSION_SHARE"/pulseaudio_settings
+  if check_audio_deamon.sh ; then
+    show_default_audio_device.sh
```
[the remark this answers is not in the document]
hanhsuan: Some jobs that not use `pipewire_utils.py` will use this `show_default_audio_device.sh` such as `id: bluetooth/audio_record_playback`. However, it could be merge into `pipewire_utils.py`, if it is the policy that should be followed.

providers/base/bin/check_audio_deamon.sh
```diff
@@ -0,0 +1,19 @@
...
+        echo "No pipewire or pulseaudio installed!! Stop test"
+        exit 2
+    fi
+elif pactl info | grep -iq pipewire ; then
+    echo "Daemon is pipewire, use pipeiwre function"
+    exit 0
+else
+    echo "Daemon is pulseaudio, use pulseaudio function"
+    exit 1
+fi
```
diohe0311, line 19: The logic here confuse me for 3 reasons:
1. The number of exit code has [meaning](https://pressbooks.senecacollege.ca/uli101/chapter/exit-status-in-shell-scripting/), `1`: general error, `2`: missing attribute. One of your case is calling pulseaudio and exit with `1`, this is not the common way to use it, would leds to misunderstanding.
2. In the upper layer: audio/job.pxu, you wrote as below, which is not right, because if `check_audio_deamon.sh` find no cmd and exit with code: 2, it will return to the `else` section in job.pxu and still trying to call audio_settings.py. This is why we shouldn't write complex sh in job.pxu. Unit test won't be able to catch errors if any.
3. Why need `EXIT_CODE=$?` here? If I understand correct, if any of the script here failed at something will cause execution stop and exit with `1`, and if not, it would be `0` defaultly. Is there a reason to point that out in code?
4. I'm not sure the echo msg is necessary, since the lower layer of script should have enough msg to help user know what is exactly happening(for both pass and error). I might be wrong about this, but current workflow is Jenkins -> TF -> Checkbox -> testplan -> jobs, each stage should only provide the msg relates to itself, otherwise, the log will be tremendous, and not helping us to understand what's happening.
```
  if check_audio_deamon.sh ; then
    show_default_audio_device.sh
    pipewire_utils.py gst -t 2 'audiotestsrc wave=sine freq=512 ! audioconvert ! audioresample ! autoaudiosink'
    EXIT_CODE=$?
  else
    audio_settings.py store --file="$PLAINBOX_SESSION_SHARE"/pulseaudio_settings
    audio_settings.py set --device=pci --volume=50
    gst_pipeline_test.py -t 2 'audiotestsrc wave=sine freq=512 ! audioconvert ! audioresample ! autoaudiosink'
    EXIT_CODE=$?
    audio_settings.py restore --file="$PLAINBOX_SESSION_SHARE"/pulseaudio_settings
  fi
```

Could you please let me know why it can not be as simple as below? Am I missing some thing here?
```
if command -v pw-cli; then
    show_default_audio_device.sh
    pipewire_utils.py gst -t 2 'audiotestsrc wave=sine freq=512 ! audioconvert ! audioresample ! autoaudiosink'

elif command -v pactl; then
    audio_settings.py store --file="$PLAINBOX_SESSION_SHARE"/pulseaudio_settings
    audio_settings.py set --device=pci --volume=50
    gst_pipeline_test.py -t 2 'audiotestsrc wave=sine freq=512 ! audioconvert ! audioresample ! autoaudiosink'
    audio_settings.py restore --file="$PLAINBOX_SESSION_SHARE"/pulseaudio_settings

else
    echo "Test failed: cannot found pipewire or pulseaudio"
    exit 1
```

diohe0311 (changes requested):

providers/base/units/audio/jobs.pxu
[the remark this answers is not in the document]
diohe0311: What is the policy you refering to?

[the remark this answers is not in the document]
hanhsuan: If cert team would like to avoid using shell script in the future.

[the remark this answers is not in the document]
diohe0311: If some jobs only use the functionality of `show_default_audio_device.sh`, I think we can still merge it into `pipewire_utils.py` and made it as one of the parameter, something like `pipewire_utils.py show`.

And Pierre and I have discussed about `pipewire_utils.py`, this case you mentioned just proved that you wrote a high-level utility script that can be used by other jobs. Excelent job! Please move it into `checkbox-support/checkbox_support/scripts`.

providers/base/bin/check_audio_deamon.sh
hanhsuan: For 2. the job has defined `package.name in ['pulseaudio-utils', 'pipewire']`, therefore this job won't execute without one of them.

For 3. I follow the same code style as old one to make it more consistence.

For 4. `pipewire` is not only audio system but also video system. Therefore `pipewire` is used to be video system in jammy alongside with `pulseaudio` audio system. For jammy, `pw-cli` and `pactl` are both installed by default and will always use `pipewire` in your script.

providers/base/units/audio/jobs.pxu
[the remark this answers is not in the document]
hanhsuan: I suggest to change shell script to python script. The reason is
1. It would be easier to know what this script is doing from the script name.
2. If there is the 3rd audio system, it would be easier to extend.

providers/base/bin/pipewire_utils.py
```diff
@@ -0,0 +1,504 @@
...
+
+    :attr PIPELINE_PROCESS_FAIL: gst pipeline process failed
+    :type PIPELINE_PROCESS_FAIL: int
+    """
+    NO_ERROR = 0
+    NOT_DETECTED = -1
+    NO_AVAILABLE_PORT = -2
+    NO_SPECIFIC_DEVICE = -3
+    PIPELINE_PROCESS_FAIL = -4
+    NO_CHANGE_DETECTED = -5
```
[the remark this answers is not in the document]
hanhsuan: `NOT_DETECTED` is for function `detect` and `NO_CHANGE_DETECTED` is for function `monitor`. After refactoring the graphic PR. I have started thinking to refactor this one is necessary or not?

providers/base/bin/check_audio_deamon.sh
hanhsuan: Above for 4 is for 5?

For 4. the echo message in the script is for manual test to notice user to make sure they are testing the right thing.

providers/base/bin/pipewire_test.py
```diff
@@ -0,0 +1,608 @@
...
+            "set-mute",
+            "@DEFAULT_AUDIO_%s@" % (self.pw_types[self.type]),
+            mute,
+        ]
+        self._wpctl_output(command)
+        return True
+
+    def _wpctl_output(self, command):
+        # This method mainly calls wpctl (hence the name). Since wpctl may
+        # return a failure if the audio layer is not yet initialized, we will
```
diohe0311, line 172: Is this audio layer a package?

providers/base/bin/pipewire_test.py
```diff
@@ -0,0 +1,608 @@
...
+class FileDumper(object):
+    def write_to_file(self, filename, data):
+        try:
+            with open(filename, "w") as f:
+                for i in data:
+                    print(i, file=f)
+            return_value = True
+        except (TypeError, IOError):
+            return_value = False
+        return return_value
```
diohe0311, line 197: Might as well provide the error msg here, something like:
except IOError as e:
            print("ERROR: could not find %s: %s" %
                  (filename, repr(e)))

providers/base/bin/pipewire_test.py
```diff
@@ -0,0 +1,608 @@
...
+                # TODO: Due to an upstream bug, a structure's get_value method
+                # doesn't work if the value in question is an array (as is the
+                # case with the magnitudes).
+                # https://bugzilla.gnome.org/show_bug.cgi?id=693168
+                # We have to resort to parsing the string representation of the
+                # structure. It's an ugly hack but it works.
+                # Ideally we'd be able to say this to get fft_magnitudes:
+                # message.get_structure.get_value('magnitude').
+                # If an upstream fix ever makes it into gstreamer,
+                # remember to remove this hack and the parse_spectrum method
```
diohe0311, line 313: Could you please also note this in PR description?

providers/base/bin/pipewire_test.py
```diff
@@ -0,0 +1,608 @@
...
+    # double-quote the identifiers
+    text = re.sub(r"([\w-]+):", r'"\1":', text)
+    # Wrap the whole thing in brackets
+    text = ("{"+text+"}")
+    # Try to parse and return something sensible here, even if
+    # the data was unparsable.
+    try:
+        return json.loads(text)
+    except ValueError:
+        return None
```
diohe0311, line 445: Would be great to have some error msg to indicate what kind of issue we got

providers/base/bin/pipewire_test.py
```diff
@@ -0,0 +1,608 @@
...
+
+    recorder.volumecontroller = VolumeController(type='input',
+                                                 logger=logging)
+    recorder.volumecontroller.set_volume(0)
+    recorder.volumecontroller.mute(False)
+
+    player.volumecontroller = VolumeController(type='output',
+                                               logger=logging)
+    player.volumecontroller.set_volume(PLAY_VOLUME)
+    player.volumecontroller.mute(False)
```
diohe0311, line 530: Why need the second mute?

providers/base/tests/test_pipewire_test.py
```diff
@@ -0,0 +1,364 @@
...
+import unittest
+import sys
+from unittest.mock import MagicMock
+from unittest.mock import Mock, patch, mock_open, call
+try:
+    sys.modules["gi"] = MagicMock()
+    sys.modules["gi.repository"] = MagicMock()
+    from pipewire_test import *
+except ImportError:
+    sys.exit(127)
```
diohe0311, line 26: I thought unit test is to test the function and not affect by the difference of machine, no matter who or what running the test. should come out the same result. Therefore I wonder what's the rationale of `try/except` here

providers/base/tests/test_pipewire_test.py
```diff
@@ -0,0 +1,364 @@
...
+class ParseSpectrumMessageStructureTests(unittest.TestCase):
+    ss = "spectrum, endtime=(guint64)5000000000,"\
+         " timestamp=(guint64)4900000000,"\
+         " stream-time=(guint64)4900000000,"\
+         " running-time=(guint64)4900000000,"\
+         " duration=(guint64)100000000,"\
+         " magnitude=(float){ -60, -60 };"
+    ans = '{"endtime": 5000000000, "timestamp": 4900000000,'\
+          ' "stream-time": 4900000000, "running-time": 4900000000,'\
+          ' "duration": 100000000, "magnitude": [-60, -60]}'
```
diohe0311, line 315: Please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                            "factory.id": 10,\
+                            "clock.quantum-limit": 8192,\
+                            "node.driver": true,\
+                            "object.id": 29,\
+                            "object.serial": 29\
+                          },\
+                          "params": {\
+                          }\
+                        }\
+                      }]'
```
diohe0311, line 185: please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+        wrong_type1 = "videog"
+        wrong_type2 = "xxxx"
+        self.assertEqual("UNKNOWN TYPE",
+                         pt.generate_pw_media_class(wrong_type1,
+                                                    "source"))
+        self.assertEqual("UNKNOWN TYPE",
+                         pt.generate_pw_media_class(wrong_type2,
+                                                    "sources"))
+
+        # example code for testing log output in python 3.10 and above
```
diohe0311, line 122: Do you mean this unit test has different results based on the Python version?

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                               "factory.id": 10,\
+                               "clock.quantum-limit": 8192,\
+                               "node.driver": true,\
+                               "object.id": 29,\
+                               "object.serial": 29\
+                             },\
+                             "params": {\
+                             }\
+                           }\
+                         }]'
```
diohe0311, line 219: please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                            "factory.id": 10,\
+                            "clock.quantum-limit": 8192,\
+                            "node.driver": true,\
+                            "object.id": 29,\
+                            "object.serial": 29\
+                          },\
+                          "params": {\
+                          }\
+                        }\
+                      }]'
```
diohe0311, line 291: please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                   },\
+                   "params": {\
+                       "Route": [{\
+                           "name": "hdmi_demo_output",\
+                           "available": "yes",\
+                           "description": "hdmi demo output"\
+                       }]\
+                   }\
+                 }\
+                }]'
```
diohe0311, line 375: please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                   "params": {\
+                       "Route": [{\
+                           "name": "hdmi_demo_output",\
+                           "available": "yes",\
+                           "direction": "Output",\
+                           "description": "hdmi demo output"\
+                       }]\
+                   }\
+                 }\
+                }]'
```
diohe0311, line 438: please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                   "params": {\
+                       "Route": [{\
+                           "name": "hdmi_after_output",\
+                           "available": "yes",\
+                           "direction": "Output",\
+                           "description": "hdmi after output"\
+                       }]\
+                   }\
+                 }\
+                }]'
```
diohe0311, line 480: please use multiline strings

providers/base/tests/test_pipewire_utils.py
```diff
@@ -0,0 +1,634 @@
...
+                   "params": {\
+                       "EnumRoute": [{\
+                           "name": "hdmi_demo_output",\
+                           "available": "yes",\
+                           "direction": "Output",\
+                           "description": "hdmi demo output"\
+                       }]\
+                   }\
+                 }\
+                }]'
```
diohe0311, line 541: please use multiline strings

diohe0311 (changes requested):

providers/base/bin/pipewire_utils.py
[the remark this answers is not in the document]
hanhsuan: Because kissiel doesn't like this kind of error handling.

providers/base/units/audio/jobs.pxu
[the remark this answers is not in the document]
hanhsuan: I propose to **migrate** `show_default_audio_device.sh` to `show_default_audio_device.py`, not **merge** to `pipewire_utils.py`. And yes, they could move to `checkbox-support/checkbox_support/scripts`.

providers/base/bin/check_audio_deamon.sh
hanhsuan: For 2. I'm fine to remove it. However I prefer to comment it only to make user understand the limitation.

For echo message. If you know the `checkbox audio tests`, `pulseaudio` and `pipewire` well then it could be known from somewhere in the log.

providers/base/bin/pipewire_test.py
```diff
@@ -0,0 +1,608 @@
...
+    gi.require_version('Gst', '1.0')
+    from gi.repository import GObject
+    from gi.repository import Gst
+    from gi.repository import GLib
+    Gst.init(None)  # This has to be done very early so it can find elements
+except ImportError:
+    print("Can't import module: %s. it may not be available for this"
+          "version of Python, which is: " % sys.exc_info()[1], file=sys.stderr)
+    print((sys.version), file=sys.stderr)
+    sys.exit(127)
```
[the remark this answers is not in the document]
hanhsuan: These codes are from the old `audio_test.py`, but you could check the [release history of pyobject](https://pypi.org/project/PyGObject/#history). The oldest one is `2.10.1` on `2006` and `python 3.8` is released on `2019` that means the limitation is for `pyobject 3.36` and above.

providers/base/bin/pipewire_test.py
hanhsuan: `wpctl` is part of `wireplumber`, the session manager of `pipewire`.

providers/base/bin/pipewire_test.py
hanhsuan: I could add these, but this is from the old `audio_test.py`.

providers/base/tests/test_pipewire_utils.py
hanhsuan: `assertLogs` couldn't work well on `python 3.8` and lower as mention in the beginning of the file.

```
# There are some issue with assertLog for python version lower than 3.10,
# https://github.com/python/cpython/commit/6fdfcec5b11f44f27aae3d53ddeb004150ae1f61
# Therefore, please don't add new test cases of assertLog.
```

providers/base/tests/test_pipewire_test.py
hanhsuan: I just copied from the `pipewire_test.py`. Sure let me remove it.

providers/base/bin/pipewire_test.py
hanhsuan: One is for `recorder` and other one is for `player`

providers/base/bin/pipewire_utils.py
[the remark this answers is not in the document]
diohe0311: I can't parse your sentences very well, did Maciej mentioned the reasons? what I don't understand is this is a PR related to audio, why there is a enum member related to monitor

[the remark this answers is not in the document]
hanhsuan: function `monitor` means `def monitor_active_port_change`

providers/base/bin/pipewire_test.py
diohe0311: Then how come it has any chance to return failture if job.pxu already requires packages

hanhsuan: We couldn't expect that image will install everything correctly and that could be a question for all applications: who or where should handle this kind of error ?

providers/base/tests/test_pipewire_test.py
diohe0311: NO, you shouldn't just copy the code from `audio_test.py` and mofidy a few of the content to create another script. Please don't ever do this, coding like this will make the whole checkbox fat and redundant. You can either add your function into it or use `decorator` to create another one if you prefer.

providers/base/units/audio/jobs.pxu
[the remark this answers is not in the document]
diohe0311: I don't think show_default_audio_device needs to be in checkbox-support for 2 reasons:
1. wpctl is not a high-level utility script
2. it already been used in pipewire_utils & pipewire_test

providers/base/tests/test_pipewire_test.py
hanhsuan: My thought is that `audio_test.py` should be retired someday, then your approach will make checkbox fat and redundant.

providers/base/units/audio/jobs.pxu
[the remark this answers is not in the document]
hanhsuan: So put `pipewire_unitls.py` into `checkbox-support` and leave `show_default_audio_device.py` in `checkbox-provider-base` ?

providers/base/bin/check_audio_deamon.sh
diohe0311: Please do remove it, the comment itself doesn't reveal useful information. 
Plainbox will skip the test if requires not meets, so user shouldn't try to debug in test-script layer.

providers/base/tests/test_pipewire_utils.py
diohe0311: From the link you shared, it seems the issue has been resolved.

providers/base/bin/check_audio_deamon.sh
hanhsuan: What if someone add a new job with this script and forget to add package requirement ?

providers/base/tests/test_pipewire_test.py
diohe0311: Are you saying that all the ubuntu version will retire pulseaudio? including uc18, 20, 18.04, 20.04?

providers/base/tests/test_pipewire_utils.py
hanhsuan: the [test result](https://github.com/canonical/checkbox/actions/runs/6888107337/job/18736472019?pr=826) from github action shows it is not for python 3.8.

providers/base/tests/test_pipewire_test.py
hanhsuan: I mean after 22.04 is retired. Reusing code is great, but I do think your approach will create another monster `udevadm.py` in the future for compatibility issue.

providers/base/units/audio/jobs.pxu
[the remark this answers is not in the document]
diohe0311: I'll suggest merge `show_default_audio_device` into `pipewire_unitls`, simply because both of them used `wpctl`, I see it as one of the utility of pipewire. But if you know there is other scenarios might be better to make it individual, that's also fine.

What is unitls stand for?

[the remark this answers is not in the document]
hanhsuan: If I move `pipewire_utils.py` to the checkbox_support, then I have to create other file just like `audio_settings.py` ? or do we have any better way to deal with it ?

providers/base/bin/pipewire_test.py
hanhsuan: This bug is[ fixed](https://bugzilla.gnome.org/show_bug.cgi?id=753754) in 2017, then checkbox has to include `python3-gst-1.0 (gst-python 1.12)` as dependency to solve this. Is it acceptable to add it in to checkbox ?

providers/base/bin/pipewire_test.py
diohe0311: You can use `fail-on-resource` flag (see [doc](https://checkbox.readthedocs.io/en/stable/reference/units/job.html))
Test will fail immediately if the required package isn't exist

providers/base/bin/pipewire_test.py
diohe0311: `gst-python 1.12` only support ubuntu 18+, if we fix this issue properly, need to consider the ubuntu 16 case.

Hook25 (changes requested):
Allow me to endorse this comment: https://github.com/canonical/checkbox/pull/826#issuecomment-1833207607

Looking at the changes from `audio_test.py` -> `pipewire_test.py`, it looks very clear to me that the two should be merged and use inheritance to obtain the result that we need. 

Let me be more explicit, we have a super class here that is `VolumeController`, this class should contain all common functionality and the interface.
```python
def run_retry(command):
    # Since the audio system may return a failure if the audio layer is not 
    # yet initialized, we will try running a few times in case of failure. 
    # Do not use this if your command is not "idempotent", 
    # repeating them should not have any bad effects.
    for attempt in range(0, 3):
        try:
            return subprocess.check_output(
                command,
                universal_newlines=True
            )
        except (subprocess.CalledProcessError):
            time.sleep(5)

class VolumeController:
    def __init__(self, controller_type, logger=None):
        self.controller_type = controller_type
        self.logger = logger
    def set_volume(self, volume): pass
    def get_volume(self): pass
    def set_mute(self, mute): pass
```

Then we have two subclasses, one for pipewire and one for pulseaudio
```python
class PulseVolumeController(VolumeController):
    ... # this is the one from audio_test.py with minor edits

class PipewireVolumeController(VolumeController):
    ... # this is from pipewire_test.py
```

The main function is basically the same for the two modules (but the warning message, that should be moved inside the `PulseVolumeController` class. To select the audio system we want to test we can add a new cli argument with two possible choiches (pipewire, pulseaudio), and we initialize the correct instance in main depending on this argument value. 

This kind of design should also make it trivial to add a new audio system or remove an old one, for example, if we ever deprecate testing pulseaudio systems via checkbox, we can just remove the argument and the class, and everything else will remain the same.

I did read most of this PR but I feel like the comments by @diohe0311 are correct, once those are addressed I can give some additional comments (I don't want them to overlap, lets do one review at the time!)

providers/base/bin/pipewire_test.py
Hook25: If that error is raised here it means that the script being called is wrong, therefore the resource expression that triggered it was wrong (in a way, we were wrong, we must update checkbox!), so lets not handle it. If you want the script to fail when neither are present @diohe0311  is completely correct, use `fail-on-resource`

providers/base/tests/test_pipewire_test.py
diohe0311: @hanhsuan Hi! we have updated the [code review criteria](https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#code-review-criteria--workflow)

Change my formatting related comments to suggestions as it's not the requested changes, please feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

providers/base/tests/test_pipewire_utils.py
diohe0311: Change this to suggestion according to code review criteria, feel free to close it.

diohe0311:
@hanhsuan Hi! we have updated the [code review criteria](https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#code-review-criteria--workflow)

Change my formatting related comments to suggestions as it's not the requested changes, please feel free to close it.

I was try to leave a new comment to overwrite the requested changes, seems github doesn't work like that, I'll close it and suggest to use multiline strings next time

providers/base/tests/test_pipewire_test.py
hanhsuan: This will break the hack parser, therefore it couldn't be modified like this.

Hook25 (approved):
+1, this is probably going to need a followup but its a good base to work on. Thank you

providers/base/bin/pipewire_test.py
Hook25: This is a possible route for a follow up!

providers/base/tests/test_pipewire_test.py
Hook25: Main route for a follow up, a de-duplication of this module, resolving for now

providers/base/tests/test_pipewire_test.py
Hook25: This is probably breaking the hack parser because it contains an extra tab. Probaby a good idea to follow this up with a more robust parser! (or just a `textwrap.dedent`)

providers/base/bin/check_audio_deamon.sh
Hook25: As per our discussion, this is a bit weird, keeping it for now but this is one possible way to follow this up

providers/base/bin/pipewire_utils.py
```diff
@@ -0,0 +1,504 @@
...
+import sys
+import re
+import gi
+gi.require_version('Gst', '1.0')
+gi.require_version('GLib', '2.0')
+from gi.repository import Gst        # noqa: E402
+from gi.repository import GLib       # noqa: E402
+
+
+class PipewireTestError(IntEnum):
```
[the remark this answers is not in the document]
Hook25: True, small and easy follow up!

providers/base/bin/pipewire_utils.py
```diff
@@ -0,0 +1,504 @@
...
+    """
+    A class used to define PipewireTest Error code
+
+    :attr NO_ERROR: process success
+    :type NO_ERROR: int
```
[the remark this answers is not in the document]
Hook25: yup, possible route for a small follow up enum-oriented!

PR #828, opened 2023-11-16
Check for new commits from the main daily workflow (Infra)

yphus:
## Description

Move the check for new commits from the 3 daily workflows to the top / caller workflow.

## Resolved issues

The workflow event type does not seem to be set to workflow_call thus triggering builds every day.
Problem is that ppa uploads are refused if the same version already exists in the ppa. 

Similar issue: https://github.com/actions/runner/discussions/1884

## Documentation

N/A

## Test

The manual workflow dispatch from this branch:
https://github.com/canonical/checkbox/actions/runs/6893400460

### DIFF
--- a/.github/workflows/daily-builds.yml
+++ b/.github/workflows/daily-builds.yml
@@ -6,12 +6,35 @@ on:
   workflow_dispatch:
 
 jobs:
+  check_for_commits:
+    runs-on: [self-hosted, linux, large]
+    name: Check for commits
+    outputs:
+      new_commit_count: ${{ steps.commit_check.outputs.new_commit_count }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check for commits
+        id: commit_check
+        run: |
+          commit_count=$(git log --since="24 hours ago" --oneline -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap | wc -l)
+          echo "new_commit_count=$commit_count" | tee $GITHUB_OUTPUT
+
   checkbox-core-snap-daily:
+    needs: check_for_commits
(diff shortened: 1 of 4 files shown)


.github/workflows/daily-builds.yml
```diff
@@ -6,12 +6,40 @@ on:
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check for new commits
+        id: check_log
+        run: |
+          git rev-list --abbrev-commit --pretty=oneline HEAD --not $(git rev-list -n1 --before="24 hours" --first-parent HEAD) -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap
+          changes=$(git rev-list --abbrev-commit --pretty=oneline HEAD --not $(git rev-list -n1 --before="24 hours" --first-parent HEAD) -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap)
+          if [ -z $changes ]
```
kissiel, line 23: This whole block is a mess. Doing this logic here is unnecessary, all shell logic branches are simply dangerous. This was there before but because this codes gets improved I think it's time to fix it.

Compare all of that big block with:
```
git rev-list --abbrev-commit --pretty=oneline HEAD --not $(git rev-list -n1 --before="24 hours" --first-parent HEAD) -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap >> $GITHUB_OUTPUT
cat The differences: \n $GITHUB_OUTPUT
```



That's it. Linear flow, no branches, less code, same functionality, zero repetition (instead of 3x).

Then, the consumer has only to check if the file is empty.

kissiel (changes requested):
noticed one spot that is growing unnecessarily
see below

kissiel (approved):
Excellent stuff, TYVM for all the simplifications!
The followup to this will be the more reliable way of determining since when to watch for commits.

PR #829, opened 2023-11-20
Strip the executable stack bit on libde265.so.0.1.1, armhf only (BugFix)

yphus:
## Description

On armhf, libde265.so.0.1.1 triggers a review-tools warning about the executable stack (PROT_EXEC).
This patch strips it with 'execstack --clear-execstack'

## Resolved issues

Uploads to the store are blocked for the checkbox22 snap because of the following warning during automated reviews:
`Found files with executable stack. This adds PROT_EXEC to mmap(2) during mediation which may cause security denials. Either adjust your program to not require an executable stack, strip it with 'execstack --clear-execstack ...' or remove the affected file from your snap. Affected files: usr/lib/arm-linux-gnueabihf/libde265.so.0.1.1 functional-snap-v2_execstack`

## Documentation

N/A

## Tests

Tested using `snapcraft remote-build --build-on amd64,arm64,armhf`

Review tools passed:
```
$ review-tools.snap-review ./checkbox22_3.0.1.dev18+gc8878e24a.d20231120_armhf.snap
./checkbox22_3.0.1.dev18+gc8878e24a.d20231120_armhf.snap: pass
```

### DIFF
--- a/checkbox-core-snap/series22/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series22/snap/snapcraft.yaml
@@ -273,6 +273,12 @@ parts:
       python3 manage.py validate
       python3 manage.py build
       python3 manage.py install --layout=relocatable --prefix=/providers/checkbox-provider-base --root="$SNAPCRAFT_PART_INSTALL"
+      # Strip the executable stack bit on libde265.so.0.1.1
+      # Only armhf requires this modification, arm64 and amd64 both pass the
+      # automated store validation (i.e review-tools.snap-review)
+      if [ $SNAPCRAFT_TARGET_ARCH = "armhf" ]; then
+        execstack --clear-execstack $SNAPCRAFT_PART_INSTALL/usr/lib/*/libde265.so.0.1.1
+      fi
     stage-packages:
       - bc
       - bluez-tests
@@ -346,6 +352,7 @@ parts:
       - on arm64:
         - python3-rpi.gpio # only in focal
     build-packages:
+      - execstack
       - libasound2-dev
       - libcap-dev
     organize:


checkbox-core-snap/series22/snap/snapcraft.yaml
```diff
@@ -273,6 +273,9 @@ parts:
       python3 manage.py validate
       python3 manage.py build
       python3 manage.py install --layout=relocatable --prefix=/providers/checkbox-provider-base --root="$SNAPCRAFT_PART_INSTALL"
+      if [ $SNAPCRAFT_TARGET_ARCH = "armhf" ]; then
```
Hook25, line 276: Please add a comment to explain why this is done

Hook25: - Is this happening on any other arch? if not, why? if it is, clean them as well
- What is this library and does it work after it is cleaned of the executable stack?
- Do we need this library? Can we just remove it?

Hook25 (changes requested):
I don't understand the how and the why of this, can you elaborate a bit? (also, write a small comment in the snapcraft.yaml to explain why this is done so the next one who comes by knows what is going on)

checkbox-core-snap/series22/snap/snapcraft.yaml
yphus: This patch is more a hotfix to get back the ability to upload snaps to the store. Do we need libde265-0? probably not, like many other dependencies added because of opencv. 
About the other architectures, they don't have the same problem. I've downloaded and extracted the library file from the [debian packages ](https://launchpad.net/ubuntu/+source/libde265) and indeed only armhf has this bit enabled.

PR #833, opened 2023-11-23
Added version check to camera quality test (Bugfix)

fernando79513:
The camera quality job now includes a requirement for ubuntu version >=22.04

## Description

The camera quality test was created only for 22.04 versions onwards, but it was being run also for previous versions. We have added a check to ensure it is only run for the corresponding versions. 

## Tests

Running the camera quality test in versions earlier than 22.04 should not be executed.

```
==============[ Running job 5 / 5. Estimated time left: 0:00:20 ]===============
------------[ Webcam brisque score for XXX ]-------------
ID: com.canonical.certification::camera/camera-quality_video0
Category: com.canonical.plainbox::camera
 - resource expression "lsb.release >= '22.04'" evaluates to false
```

### DIFF
--- a/providers/base/units/camera/jobs.pxu
+++ b/providers/base/units/camera/jobs.pxu
@@ -117,6 +117,7 @@ flags: also-after-suspend
 _summary: Webcam brisque score for {{ product_slug }}
 estimated_duration: 20s
 depends: camera/detect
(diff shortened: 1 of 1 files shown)


yphus (approved):
LGTM, +1

PR #840, opened 2023-11-24
Fixed missing history in deb builds (infra)

Hook25:
## Description

We need the history to calculate the version. This is the fix from the last beta release


yphus (approved):
+1

PR #842, opened 2023-11-24
Add the app_context module (New)

kissiel:
## Description
The function checking if we're running on Ubuntu Core was previously defined very deep in the unit definition module. This module exposes it for all interested callers. 
Similarly I'm adding a unified way of checking what is the name of the snap we're running from so it's easier to determine so called "app_id" later on (which is used by session assistant to filter out sessions, for instance).

## Tests
The proposed module is mirrored by the test_app_context.py with unit tests.

### DIFF
--- a/checkbox-ng/checkbox_ng/app_context.py
+++ b/checkbox-ng/checkbox_ng/app_context.py
@@ -1,0 +1,56 @@
+# This file is part of Checkbox.
+#
+# Copyright 2023 Canonical Ltd.
+#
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
+1, ty

Hook25 (approved):
+1, strange and esoteric ngl

pieqq (commented):
I think there is a problem with the string being looked for. Please check my inline comments.

PR #843, opened 2023-11-24
Dont require gpg pass when running tests (BugFix)

kissiel:
## Description

Previously when the tests run, if the host had a GPG key in its databse that was protected by a password, the operator was asked to provide the password. This was due to subprocess.call not being mocked properly in the provider_manager tests.

## Tests

The test module is what gets fixed here.


Hook25 (approved):
+1, ty for investigating this!

PR #844, opened 2023-11-25
Fix connecting by hostname (BugFix)

kissiel:
## Description
This fixes a bug where only IP addresses could be used as targets for controller.
The 4df48633 introduced a bug where using hostname as a target for the
control subcommand would yield a traceback. Usually that's not a problem
because we specify agents using IP addresses, but 'localhost' is a great
hostname that should work :)

While fixing this I outlined this check to a function, and ...

## Tests

... wrote tests for the function.


Hook25 (approved):
+1, ty also for adding so many tests

Issue #789, opened 2023-10-25
`tpm2_capabilities.py` will return success while executing `tpm2_getcap` command failed


hanhsuan:
### Bug Description

In the `tpm2_capabilities.py` there is no specific return value from `raise SystemExit`, this makes the test cases `com.canonical.certification::clevis-encrypt-tpm2/detect-rsa-capabilities` and `com.canonical.certification::clevis-encrypt-tpm2/detect-ecc-capabilities` passed with on tpm-tools installed or no TPM chip found.
```python
try:
    algs_caps = subprocess.check_output(['tpm2_getcap', 'algorithms'])
    pcrs_caps = subprocess.check_output(['tpm2_getcap', 'pcrs'])
except subprocess.CalledProcessError:
    raise SystemExit
```

It could be changed to
```python
try:
    algs_caps = subprocess.check_output(['tpm2_getcap', 'algorithms'])
    pcrs_caps = subprocess.check_output(['tpm2_getcap', 'pcrs'])
except subprocess.CalledProcessError:
    raise SystemExit("Please make sure you have installed tpm-tools or tpm chip")
```

If this change is OK to you, I could submit a PR.

### To Reproduce

uninstall tpm-tools or one device without TPM chip and run `checkbox-cli run com.canonical.certification::clevis-encrypt-tpm2/detect-rsa-capabilities`

### Environment

- OS: Ubuntu 22.04
- Checkbox Type: Debain
- Checkbox version: 2.10.1

### Relevant log output

```shell
[submission of device without TPM chip](https://certification.canonical.com/hardware/202310-32222/submission/339052/test/129021/result/37058636/)

Log:

ERROR:tcti:src/tss2-tcti/tcti-device.c:452:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpmrm0: No such file or directory 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-device.so.0 
ERROR:tcti:src/tss2-tcti/tcti-device.c:452:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpm0: No such file or directory 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-device.so.0 
[...]

PR #846, opened 2023-11-26
Fix `tpm2_capabilities.py` always return success while no tpm tool or chip installed (Bugfix)

hanhsuan:
## Description
This PR is simply to add non zero return value to make checkbox could know this test is failed.
If I would like to add unit test for this, the only way I know is to refactor whole script and I think we have arrangement that refractory old codes is not what QA team have to do.

## Resolved issues

#789

### DIFF
--- a/providers/tpm2/bin/tpm2_capabilities.py
+++ b/providers/tpm2/bin/tpm2_capabilities.py
@@ -87,8 +87,9 @@ TPM2_CAP = {
 try:
     algs_caps = subprocess.check_output(['tpm2_getcap', 'algorithms'])
     pcrs_caps = subprocess.check_output(['tpm2_getcap', 'pcrs'])
-except subprocess.CalledProcessError:
-    raise SystemExit
+except (subprocess.CalledProcessError, FileNotFoundError):
+    raise SystemExit("Please make sure you have installed tpm-tools or tpm chip")
+
 
 algs_list = yaml.load(algs_caps, Loader=yaml.FullLoader)
 pcrs_list = yaml.load(pcrs_caps, Loader=yaml.FullLoader)
--- a/providers/tpm2/bin/tpm2_capabilities.py
+++ b/providers/tpm2/bin/tpm2_capabilities.py
@@ -88,7 +88,9 @@ try:
     algs_caps = subprocess.check_output(['tpm2_getcap', 'algorithms'])
     pcrs_caps = subprocess.check_output(['tpm2_getcap', 'pcrs'])
 except (subprocess.CalledProcessError, FileNotFoundError):
-    raise SystemExit("Please make sure you have installed tpm-tools or tpm chip")
+    raise SystemExit("""
+                     Please make sure you have installed tpm-tools and tpm chip
+                     """)
 
 
 algs_list = yaml.load(algs_caps, Loader=yaml.FullLoader)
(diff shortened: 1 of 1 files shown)


kissiel:
The default invocation of tpm2_capabilities.py is not a test, it is a resource job for Checkbox to get TPM2 information. It means it's used to gather information about hardware. If there's no TPM chip, it's a valid information, and the job should not be failed. Overall resource jobs should never fail.

providers/tpm2/bin/tpm2_capabilities.py
```diff
@@ -87,8 +87,9 @@
 try:
     algs_caps = subprocess.check_output(['tpm2_getcap', 'algorithms'])
     pcrs_caps = subprocess.check_output(['tpm2_getcap', 'pcrs'])
-except subprocess.CalledProcessError:
-    raise SystemExit
+except (subprocess.CalledProcessError, FileNotFoundError):
+    raise SystemExit("Please make sure you have installed tpm-tools or tpm chip")
```
kissiel, line 91: Definitely you didn't mean "or" here.

kissiel (changes requested):
If you don't want this job to run at all if there's no tpm binary, you can express that with the `requires` field.

providers/tpm2/bin/tpm2_capabilities.py
hanhsuan: If you have checked  [this submission](https://certification.canonical.com/hardware/202310-32222/submission/339052/test/129021/result/37058636/), you could see [clevis-encrypt-tpm2/detect-ecc-capabilities](https://certification.canonical.com/hardware/202310-32222/submission/339052/test/129020/result/37058635/) is passed while no tpm chip be detected and shouldn't be passed. This bug has been found by ODM too.

hanhsuan: and we use manifest to set TPM is installed or not in the DUT to prevent firmware or hardware issue to skip this test. The submission I provided points to a [real issue](https://bugs.launchpad.net/somerville/+bug/2040205) from the kernel.

providers/tpm2/bin/tpm2_capabilities.py
```diff
@@ -87,8 +87,11 @@
 try:
     algs_caps = subprocess.check_output(['tpm2_getcap', 'algorithms'])
     pcrs_caps = subprocess.check_output(['tpm2_getcap', 'pcrs'])
-except subprocess.CalledProcessError:
-    raise SystemExit
+except (subprocess.CalledProcessError, FileNotFoundError):
+    raise SystemExit("""
+                     Please make sure you have installed tpm-tools and tpm chip
+                     """)
```
kissiel, line 93:
```suggestion
    raise SystemExit(
        "Please make sure you have installed tpm-tools and tpm chip."
    )
```

providers/tpm2/bin/tpm2_capabilities.py
kissiel: So both are needed, right?

kissiel:
/canonical/self-hosted-runners/run-workflows 61071039686a258859ae18270bfa3f72388494a8

kissiel (approved):
Awesome stuff, +1

yphus (approved):
+1

PR #849, opened 2023-11-28
update the search pattern for CAN device (BugFix)

stanley31huang:
BugFix: Fixed the udevadm parser for CAN devices

Modify the search pattern for CAN devices, including new llcecan device, and adding unit tests

## Description
There are two kind of CAN devices on NXP S32G development board, one is generic CAN device named as canxxxx, another one is can devices from Low Latency Communication Engine (LLCE) and named as llcecanxxxx.
The current implementation for CAN devices in udevadm parser is the device name should starts with 'can' string.

So I have modified the search pattern for CAN devices, including new llcecan device, and adding unit tests. 

I also applied this patch to the checkbox to ensure checkbox could identified both generic CAN device and LLCE CAN devices on NXP S32G development board.

## Resolved issues
N/A

## Documentation
N/A

## Tests
stanley@stanley-ThinkPad-T490:~/Desktop/Git_Repos/checkbox-2023$ PYTHONPATH=checkbox-support/:$PYTHONAPATH python3 -m unittest -vv checkbox-support/checkbox_support/parsers/tests/test_udevadm.py 
-k test_NXP_S32G_SOCKETCAN
test_NXP_S32G_SOCKETCAN (checkbox-support.checkbox_support.parsers.tests.test_udevadm.TestUdevadmParser) ... ok

Ran 1 test in 0.079s

OK

C3 submission with side-load checkbox:
https://certification.canonical.com/hardware/202304-31451/submission/349081/test-results/pass/


pieqq (changes requested):
Thanks for this PR!

There are already unit tests for the udevadm parser, they are located in `checkbox-support/checkbox_support/parsers/tests/test_udevadm.py` (with sample data hosted in `checkbox-support/checkbox_support/parsers/tests/udevadm_data`).

Is there a reason why you separated your tests and put them at a different location?

I recommend you add the outcome of your own DUT in the `udevadm_data`, and add your tests to `test_udevadm.py`. You can look at how tests have been implemented there and use similar methods.

For your information, there are already a few sample data files that have a Canbus device: `ZCU104.txt` and `CARA_T_SOCKETCAN.txt`. You can leverage these when you write your tests to make sure your changes don't impact existing data.

stanley31huang:
@pieqq  I just sent a new commits for those suggestion you came up. Please review this PR again, thanks.

checkbox-support/checkbox_support/parsers/udevadm.py
```diff
@@ -256,7 +256,10 @@ def category(self):
                      self._environment["INTERFACE"].startswith('wlan')
                      )):
                     return "WIRELESS"
-                if (self._environment["INTERFACE"].startswith('can')):
+                # generic CAN device will be named as 'can$n'
+                # LLCE CAN device will be named as 'llcecan$n'
+                if (re.search(r"^([a-z]+can|can[a-z]*)[0-9]+$",
```
pieqq, line 261: Is there any reason why you are not very specific with the `llcecan` part?

Also, do we have any examples where the interface name for a socketcan device is something else than `can$n`? If not, how about using `^((llce)*can)[0-9]+$` instead?

pieqq (changes requested):
One question about the regex being used below.

Also, since both `CARA_T_SOCKETCAN.txt` and `ZCU104.txt` have a socketcan device, could you add a check in their respective unit tests? This will help with regression testing.

checkbox-support/checkbox_support/parsers/udevadm.py
stanley31huang: Ok, I have sent a new commit for this issue.

pieqq:
/canonical/self-hosted-runners/run-workflows db8ec4ef90b85bfc38008b45e08f3319560bc4cb

pieqq:
Thanks for the changes! I added a check for socketcan devices in the ZCU104 unit test, so we're good to go!

pieqq (approved):

diohe0311 (approved):
LGTM +1
Just learnt from here that we can run a single unit test, nice!

PR #852, opened 2023-11-29
Replace setuptools_scm version with new get_version script (Infra)

Hook25:
## Description

We have decided to use an alternative way to calculate the version. We have landed the script to do so in 7b2e8d095dcb1419671cd2d6f5a13c83b4819f30. This PR replaces all occurrences of setuptools_scm with this new script with the appropriate calling parameters. 

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-1052

## Documentation

This changes the version of produced artifacts to the one generated by the new script documented [here (link)](https://github.com/canonical/checkbox/blob/2349a73ac8c8aa81b12fa5b92c0247edb3324c7d/tools/release/get_version.py#L20)

The way the artifacts are built isn't changed by this so that will not be documented further.

## Tests

N/A

PR #853, opened 2023-11-29
Fix daily build by removing unsupported syntax (infra)

Hook25:
## Description

[This PR](https://github.com/canonical/checkbox/pull/852) has introduced an unsupported import for python3.10 (that is the standard version of python on jammy. This PR fixes this problem by removing the import for `Self` and making the annotation a string.

## Resolved issues

[Failing daily build due to the import](https://github.com/canonical/checkbox/actions/runs/7036640718)

## Documentation

N/A

## Tests

I have re-run the tests and they still all pass. I have ran the tests on ubuntu jammy and (now) they also pass there.

### DIFF
--- a/tools/release/get_version.py
+++ b/tools/release/get_version.py
@@ -36,7 +36,6 @@ import argparse
 import textwrap
 
 from enum import Enum
-from typing import Self
 from collections import namedtuple
 from subprocess import check_output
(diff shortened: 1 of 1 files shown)


kissiel (approved):
Obvious +1

PR #856, opened 2023-11-30
Fix version format (infra)

Hook25:
## Description

The `get_version` script is generating a version with a `v` at the beginning. This is ok if that version is used to generate the tag but not if it is used to give a version to snaps or debs. This patch introduces the `output-format` option that allows the script to generate a valid version string for tagging and snap/deb building. This also allows us to kill two birds with one stone and remove the string replacement that we did in the deb recipe updater to transform the version it was fed into a valid dpkg version.

## Resolved issues

[Yesterday deb daily builds](https://github.com/canonical/checkbox/actions/runs/7038609106) failed because the version string is invalid. 

## Documentation

The addition has an help prompt and helpful comments near the options.

## Tests

This adds two new tests to verify the deb version and snap version. To run the tests:
```bash
pytest -v .
```


yphus (approved):
Thanks for covering the various version format with tests. +1

PR #854, opened 2023-11-30
Make the promotion script parametric (infra)

Hook25:
## Description

We have decided that we are going to promote from edge to beta and from beta to stable. Currently we are promoting debs from beta to stable but this is done via a script that does not have any input parameter. This PR introduces the input parameters for this script and a couple of tests.

## Resolved issues

Resolved: https://warthogs.atlassian.net/browse/CHECKBOX-1053

## Documentation

This PR also adds an explanation to what the script does

## Tests

This pr adds a new test module. To run it run
```shell
pytest test_lp_copy_packages.py
```

PR #867, opened 2023-12-04
New tool to update ppa recipes (infra)

Hook25:
## Description

As a part of the effort of splitting the build process of debs into more single responsibility steps, this PR creates a new tool that splits the updating of the recipe from the script `lp-recipe-update-build.py`, making it more testable and moving the recipe to the repository (where previously it was only updated but the ground truth was in the PPA).

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-1027

## Documentation

This adds a new section at the top of the script that documents what its purpose is. The help prompt was reworded options were added and documented.

## Tests

This adds a new unit test file that covers all the additions. The utils file is tested in [this PR](https://github.com/canonical/checkbox/pull/854) that introduced the functions. Depending what PR lands first I will rebase and move the tests.

### DIFF
--- a/tools/release/lp_update_recipe.py
+++ b/tools/release/lp_update_recipe.py
@@ -36,7 +36,7 @@ import argparse
 from utils import get_build_recipe
 
 
-def get_build_path(recipe_name: str):
+def get_build_path(recipe_name: str) -> str:
     # recipe name is in the form checkbox-{package}-{risk}
     package = recipe_name.split("-")[:-1]  # remove checkbox and risk
     if "provider" in recipe_name:
@@ -49,7 +49,7 @@ def get_updated_build_recipe(
     recipe_name: str,
     version: str,
     revision: str,
-):
+) -> str:
     target_path = get_build_path(recipe_name)
     new_recipe = textwrap.dedent(
         f"""
@@ -96,10 +96,7 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
 def main(argv):
     args = parse_args(argv)
     update_build_recipe(
-        args.project,
-        args.recipe,
-        args.new_version,
-        args.revision
+        args.project, args.recipe, args.new_version, args.revision
     )
 
 
--- a/tools/release/lp_update_recipe.py
+++ b/tools/release/lp_update_recipe.py
@@ -19,14 +19,10 @@
 # along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Kicks off a recipe build for a branch in Launchpad which has an associated
-recipe.
-Meant to be used as part of a checkbox release to the Hardware Certification
-public PPA and the ~checkbox-dev testing PPA.
+This script updates a recipe on a given ppa with the objective of
+updating the revision to build and the version of the package.
 
-NOTE: The Hardware Certification public PPA has no further use over the
-~checkbox-dev testing PPA. It is kept for historical reasons and should be
-removed at some point.
+Note: This script uses the LP_CREDENTIALS environment variable
 """
 
 import sys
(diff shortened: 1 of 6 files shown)


tools/release/lp_update_recipe.py
```diff
@@ -0,0 +1,107 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Kicks off a recipe build for a branch in Launchpad which has an associated
+recipe.
+Meant to be used as part of a checkbox release to the Hardware Certification
```
kissiel, line 24: heh, I think this is no longer up to date

tools/release/lp_update_recipe.py
```diff
@@ -0,0 +1,107 @@
...
+public PPA and the ~checkbox-dev testing PPA.
+
+NOTE: The Hardware Certification public PPA has no further use over the
```
kissiel, line 27: and this can be deleted too

tools/release/lp_update_recipe.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+
+import sys
+import textwrap
+import argparse
+
+from utils import get_build_recipe
+
+
+def get_build_path(recipe_name: str):
```
kissiel, line 39: `-> str` missing

tools/release/lp_update_recipe.py
```diff
@@ -0,0 +1,107 @@
...
+        component_name = "-".join(package[2:])  # 0-1 are checkbox provider
+        return f"providers/{component_name}"
+    return "-".join(package)
+
+
+def get_updated_build_recipe(
+    recipe_name: str,
+    version: str,
+    revision: str,
+):
```
kissiel, line 52: ditto

tools/release/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    """return Launchpad(
+        None, None, None, service_root="production", version="devel"
+    )"""
+    credentials = os.getenv("LP_CREDENTIALS")
+    if not credentials:
+        raise SystemExit("LP_CREDENTIALS environment variable missing")
+
+    credentials = Credentials.from_string(credentials)
+    return Launchpad(
+        credentials, None, None, service_root="production", version="devel"
```
kissiel, line 19: aren't those `None`s unnecessary?

tools/release/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    build_recipe = (
+        recipe for recipe in project.recipes if recipe.name == recipe_name
+    )
+    try:
+        return next(build_recipe)
+    except StopIteration:
+        # There is no recipe with that name
+        all_recipe_names = ", ".join(recipe.name for recipe in project.recipes)
+        raise SystemExit(
```
kissiel, line 41: raising exceptions while in an exc handling block is discouraged and iffy.

as an alternative you can collect the list comprehension, and run an if over it, like so:
```
build_recipe = [r for r in project.recipes if r.name = recipe_name]
if build_recipe:
    return build_recipe[0]
else:
    raise (...)
```

tools/release/utils.py
kissiel: They are not unnecessary.
Ignore me here :)

tools/release/utils.py
kissiel: Oh, that's smart! And it justifies the complexity. 
Could you copy over the comment you wrote (the second part) to the code? So the next person doesn't have to tilt their head while reading

Hook25: done

Issue #163, opened 2022-11-28
LP1982561: checkbox.shell should make it clear you're inside checkbox


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/checkbox-ng/+bug/1982561

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| New	| 2022-07-22 08:51:34	| 6	| Wishlist	| False  	|

## Description
checkbox.shell opens a runtime shell inside checkbox.

However, unlike other runtime shells (pyenv comes to mind) there is no visible hint that you're in a checkbox shell.

It would be nice if your CLI prompt changed to let you know you're inside checkbox.

Example:

u@u:~$ checkbox.shell
checkbox runtime shell, type 'exit' to quit the session
[checkbox] u@u:~$ exit
exit
u@u:~$ _

## Attachments
No attachments

Tags:
['checkbox']

kissiel:
Good idea.
Let's have a `checkbox-shell` information in the prompt (PS1) within the shell.

kissiel:
in other words, let's get from:
```
~/checkbox/metabox on  metabox-improvements [$!+?] via 🐍 v3.11.6 
❯ checkbox.shell 
checkbox runtime shell, type 'exit' to quit the session
kissiel@venus:~/checkbox/metabox$
```
to
```
~/checkbox/metabox on  metabox-improvements [$!+?] via 🐍 v3.11.6 
❯ checkbox.shell 
checkbox runtime shell, type 'exit' to quit the session
kissiel@venus (checkbox-shell):~/checkbox/metabox$
```

fernando79513:
Python virtual environment is not creating another bash session, so they can modify the prompt (PS1) and works fine.
Since our script is creating a new bash session, it is calling `.bashrc` and PS1 gets reseted.

A way to go over this is adding a new line to your `.bashrc` with the new PS1 update and parse it as an argument to bash:

```
# shell-wrapper
- exec bash
+ exec bash --rcfile <(cat ~/.bashrc; echo 'PS1="(checkbox-shell) $PS1"')
```

fernando79513:
Closed with https://github.com/canonical/checkbox/pull/871

PR #871, opened 2023-12-05
Added a checkbox-shell flag (BugFix)

fernando79513:
When running checkbox shell-wrapper, a flag with
"(checkbox-shell)" is added before the prompt

## Description

A new `(checkbox-support)` flag added when running a shell wrapper.

## Resolved issues

Closes issue  #163 

## Documentation

No documentation update. The shell-wrapper script works the same way as the previous one.

## Tests

To test theses changes, go to `checkbox-snap/common_series_classes/launchers` and run:
```
./shell-wrapper
```
The shell wrapper should output `(checkbox-shell)` before the prompt.

Alternatively, build the checkbox snap, install it and run:
```
checkbox.shell
```

### DIFF
--- a/checkbox-snap/series_uc16/launchers/shell-wrapper
+++ b/checkbox-snap/series_uc16/launchers/shell-wrapper
@@ -4,4 +4,4 @@
 
 echo "$SNAP_NAME runtime shell, type 'exit' to quit the session"
 
-exec bash
+exec bash --rcfile <(cat ~/.bashrc; echo 'PS1="(checkbox-shell) $PS1"')
--- a/checkbox-snap/common_series_classic/launchers/shell-wrapper
+++ b/checkbox-snap/common_series_classic/launchers/shell-wrapper
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 echo "$SNAP_NAME runtime shell, type 'exit' to quit the session"
-exec bash
+exec bash --rcfile <(cat ~/.bashrc; echo 'PS1="(checkbox-shell) $PS1"')
--- a/checkbox-snap/series_uc18/launchers/shell-wrapper
+++ b/checkbox-snap/series_uc18/launchers/shell-wrapper
@@ -4,4 +4,4 @@
 
 echo "$SNAP_NAME runtime shell, type 'exit' to quit the session"
 
-exec bash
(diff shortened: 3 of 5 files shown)


checkbox-snap/series_uc16/launchers/shell-wrapper
```diff
@@ -4,4 +4,4 @@
 
 echo "$SNAP_NAME runtime shell, type 'exit' to quit the session"
 
-exec bash
+exec bash --rcfile <(cat ~/.bashrc; echo 'PS1="(checkbox-shell) $PS1"')
```
kissiel, line 7: The reading of `~/.bashrc ` looks new. It is needed for anything?

kissiel (commented):
one question below

checkbox-snap/series_uc16/launchers/shell-wrapper
fernando79513: As I added in the issue:

Python virtual environment is not creating another bash session, so they can modify the prompt (PS1), and works fine.
Since our script is creating a new bash session, it is calling .bashrc and PS1 gets reset.

A way to go over this is adding a new line to your .bashrc with the new PS1 update and parsing it as an argument to bash:
```
# shell-wrapper
- exec bash
+ exec bash --rcfile <(cat ~/.bashrc; echo 'PS1="(checkbox-shell) $PS1"')
```
So, if I don't have the  `--rcfile` as an argument, your bash rc will execute and overwrite the $PS1 in the new bash session.

fernando79513: Another option that I tested was to directly ignore your `.bashrc` and have a custom prompt for the checkbox bash, but that will override your current PS1

```
# Customize PS1 for session
PS1="(checkbox) \e[0;32m\]\u\e[0;1m\]@\e[0;33m\]\h \e[0;34m\]\w \e[0;1m\]$ \e[m"
export PS1

exec bash --norc
```

kissiel: The more I think about it, the more I see how much off I was :)
Indeed, this looks like the right way forward!
Thanks for the great explanation!

kissiel (approved):
The more I think about it, the more I see how much off I was :)
Indeed, this looks like the right way forward!
Thanks for the great explanation!

PR #874, opened 2023-12-06
Remove reference to bin when importing modules in some unit tests (BugFix)

pieqq:
When running `./manage.py test -u` from the provider, unittest is made aware of the content of the bin/ directory so Python code from this directory can be imported directly.

## Description

When preparing a presentation about unit testing in Checkbox, I found that some unit tests import scripts from bin, which is not required when running unit tests the usual way (`./manage.py test -u`).

## Resolved issues

Running unit tests locally using the `./manage.py test -u` command.

## Tests

- Tests are passing when running the `./manage.py test -u` command from the base provider directory in a venv.
- Single test files can be executed, for example `PYTHONPATH=~/checkbox/providers/base/bin python -m unittest test_camera_quality.py` from the `providers/base/tests/` directory.

### DIFF
--- a/providers/base/tests/test_camera_quality.py
+++ b/providers/base/tests/test_camera_quality.py
@@ -21,13 +21,12 @@
 
 import os
 from pathlib import Path
-import time
 import unittest
 from unittest.mock import patch, MagicMock
 
 import cv2
 
-import bin.camera_quality_test as cqt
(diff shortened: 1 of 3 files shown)


Hook25 (approved):
+1, nice catch

PR #880, opened 2023-12-07
Monitor Launchpad binary builds (Infra)

Hook25:
## Description

The new script only monitors source builds but binary builds do (and infact are more likely) to fail as well. This introduces a new waiting mechanism for binary builds and monitors/retries them the same way source builds are.

## Resolved issues

The current script fails to monitor binary builds.

## Documentation

This changes all names where necessary and documents the workarounds that were needed (namely, there does not seem to exist an api that lists (actually)pending binary builds on a launchpad ppa)

## Tests

This comes with some new tests and updates those that were there before


yphus (dismissed):
`lp_build_monitor_recipe.py` is missing the `+x` execute bit.

Hook25:
I'm marking this as draft as the current implementation doesn't retry for builds fast enough. I have an idea on how to make it better but it needs more testing

tools/release/lp_build_monitor_recipe.py
```diff
@@ -33,7 +33,15 @@ def start_all_build(build_recipe):
             print("An identical build of this recipe is already pending")
 
 
-def wait_every_build_started(build_recipe):
+def are_binary_builds_ongoing(build_recipe):
+    # Note: this is a different definition of pending, this means pending or
```
kissiel, line 37: different compared to?
I'd use a non-referential clarification here, i.e.

>The pending attribute used here means ...

tools/release/lp_build_monitor_recipe.py
```diff
@@ -47,7 +55,30 @@ def wait_every_build_started(build_recipe):
         pending_builds = build_recipe.getPendingBuildInfo()
 
 
-def get_new_builds(build_recipe, started_datetime):
+def recipe_name_to_source_name(name):
+    # name is in the form of source-name-with-dashes-risk
```
kissiel, line 59: are doctests included in the coverage reports? I think they are.
This is a great example how doctest here would be better than this explanation + navigating (searching for) tests.

Also why not a docstring?

tools/release/lp_build_monitor_recipe.py
```diff
@@ -33,7 +33,15 @@ def start_all_build(build_recipe):
             print("An identical build of this recipe is already pending")
 
 
-def wait_every_build_started(build_recipe):
+def are_binary_builds_ongoing(build_recipe):
```
kissiel, line 36: types info missing

tools/release/lp_build_monitor_recipe.py
```diff
@@ -47,7 +55,30 @@ def wait_every_build_started(build_recipe):
...
 
 
-def get_new_builds(build_recipe, started_datetime):
+def recipe_name_to_source_name(name):
+    # name is in the form of source-name-with-dashes-risk
+    # this removes risk
+    return name.rsplit("-", 1)[0]
+
+
+def get_all_binary_builds(build_recipe, started_datetime):
```
kissiel, line 64: types info missing

tools/release/lp_build_monitor_recipe.py
```diff
@@ -47,7 +55,30 @@ def wait_every_build_started(build_recipe):
...
+    """
+    Returns all builds of the current calculated recipe target
+    started after started_date (UTC+0)
+    """
+    recipe_target = recipe_name_to_source_name(build_recipe.name)
+    builds = build_recipe.daily_build_archive.getBuildRecords(
+        source_name=recipe_target
+    )
+    return list(
+        itertools.takewhile(
```
kissiel, line 74: absolutely fantastic code

tools/release/lp_build_monitor_recipe.py
```diff
@@ -63,7 +94,38 @@ def get_new_builds(build_recipe, started_datetime):
     )
 
 
-def monitor_retry_builds(build_recipe, start_date):
+def monitor_retry_binary_builds(source_recipe, start_time) -> list["LPBuild"]:
```
kissiel, line 97: incomplete typing info, also the referenced type is not defined... I know that this "LPBuild" is a super dynamic type, but at least lplib offers some proxy you could use I think

tools/release/lp_build_monitor_recipe.py
```diff
@@ -63,7 +94,38 @@ def get_new_builds(build_recipe, started_datetime):
...
+    builds_unrecoverable = []
+    start_checking_binary = start_time
+    # source builds will trigger new binary build
+    while are_binary_builds_ongoing(source_recipe):
+        # to avoid a race condition, lets get the time at the beginning of the
+        # cycle that we will use to filters binary builds the next cycle
```
kissiel, line 103: cycle == iteration?

kissiel (commented):
Some comments inline, more incoming in ~2h

tools/release/lp_build_monitor_recipe.py
Hook25: Sadly, compared to anywhere else in launchpadlib, included the `buildstatus` in the docstring below. I can change this to your version, it is clearer, I was a bit salty I think

tools/release/lp_build_monitor_recipe.py
Hook25: They are not executed I think, I have written a normal test for this. I will add the docstring.

tools/release/lp_build_monitor_recipe.py
Hook25: I have created classes with the appropriate inheritance that give a little bit of an idea of what is going on here. Launchpadlib sadly has 0 types locally, they are all generated by lazr restful, so there is no way (to the best of my knowledge) to give more informative type hints sadly

tools/release/lp_build_monitor_recipe.py
kissiel: Yeah, I saw the test (tbh, this is how I fully understood what it does :D

tools/release/utils.py
```diff
@@ -8,6 +8,27 @@
...
 
+from lazr.restfulclient.resource import Entry
+
+LPBuild = Entry
+
+class LPSourceBuild(LPBuild):
+    """
+    Documentation: https://api.launchpad.net/devel/#source_package_recipe_build
+    """
+    pass
```
kissiel, line 19: the pass is not required if you have a docstring

tools/release/utils.py
```diff
@@ -8,6 +8,27 @@
...
+
+class LPBinaryBuild(LPBuild):
+    """
+    Documentation: https://api.launchpad.net/devel/#build
+    """
+    pass
```
kissiel, line 25: ditto

kissiel (commented):
The one thing I need more clarity is the typing story...

So previously there was an unknown type of "LPBuild". In the latest change you introduce that type and make it an alias of the LP's Entry. So there wasn't an original LPBuild defined anywhere in the lplib? (even the proxy)

If those functions always returned "Entry" I'd be happy with specifying this type. If you want to do a stricter typing, I guess you'd have to create a proper hierarchy. The classes that are here now are just all different names for the `Entry` class.

What do you think? Do you see value in having those aliases for Entry right now? (I'm not saying there is none :) )

Hook25:
There is no type beside Entry, that is sadly how the launchpadlib api works. 

I want to have a type reference there because it is very hard to keep track of which web api each "entry" actually refers to. Each of those entries have some valid attribute and function calls to them, the only way to know which is to know which web api they are calling or, at runtime, inspecting the `lp_attributes` and `lp_operations` attributes. By creating some alias classes with the documentation link, the next maintainer who comes around can simply refer to the api documentation linked without having to inspect at runtime (or even worse, statically infer from the chain of function calls) to know what those objects are and what is possible with them.

kissiel:
> There is no type beside Entry, that is sadly how the launchpadlib api works.
> 
> I want to have a type reference there because it is very hard to keep track of which web api each "entry" actually refers to. Each of those entries have some valid attribute and function calls to them, the only way to know which is to know which web api they are calling or, at runtime, inspecting the `lp_attributes` and `lp_operations` attributes. By creating some alias classes with the documentation link, the next maintainer who comes around can simply refer to the api documentation linked without having to inspect at runtime (or even worse, statically infer from the chain of function calls) to know what those objects are and what is possible with them.

Perfect justification, thank you!
I wonder if it would be useful to paste the same comment in the code before those classes definition :D

PR #755, opened 2023-10-05
Add documentation for the Checkbox Edge validation process (Infra)

kissiel:
### **The PR enables mermaid diagram rendering in our docs.**

This PR introduces a handful of pages describing how we validate edge build of Checkbox.
See individual added pages for details.

When testing the documentation make sure to run `make clean; make install` first in the docs folder as it needs to add the sphinx-mermaid extensions.

PR #887, opened 2023-12-12
suggest changes to Canary validation docs (infra)

tang-mm:
## Description

Suggested edits on #755 

- moved new docs to `docs/explanation/` instead of `how-to`
- applied CSS to mermaid style so that diagram text is visible under dark mode (accessibility)
- excluded mermaid from spelling check
- In the first diagram for GitHub Workflow, changed node B to use rectangular as it is not a decision point
- other format and wording changes

## Resolved issues

N/A

## Documentation

Preview: https://canonical-checkbox.readthedocs-hosted.com/en/doc-preview/explanation/release_process/canary.html

## Tests

In the docs/ folder, `make install` and then `make run`


docs/explanation/release_process/validation_pipeline_execution.rst
```diff
@@ -101,3 +103,8 @@ There's multiple entities participating in the chain of validating a Checkbox sn
         deactivate TServer
         Jenkins-->>Pipeline: Update Pipeline with Job Outcome
         deactivate Pipeline
+
+.. add code format to link text
+.. |hwcert-jenkins-jobs| replace:: ``hwcert-jenkins-jobs``
+.. _hwcert-jenkins-jobs: https://github.com
```
tang-mm, line 109: `canonical/hwcert-jenkins-jobs` is a private repository. If we decide to include it in the public doc, the link needs to be listed in `linkcheck_ignore` in `conf.py`

kissiel: Understood!
Indeed I'll change that to the full link and add it to the linkcheck_ignore list!

kissiel:
Thank you very much for the help on this!

PR #896, opened 2023-12-14
fix mocks in tests for remote_session_assistant (BugFix)

kissiel:
## Description

This PR fixes 3 unittests that assumed the test will be run on a typical Ubuntu box, which LP builders are clearly not.

## Tests

To test it I've recreated the conditions on LP, meaning I "removed" sudo and made the `guess_normal_user` always raise the exception.


Hook25 (approved):
+1, WOA ty so much for adding so many tests (and for the fix)

PR #897, opened 2023-12-14
Fix pending binary builds (infra)

Hook25:
## Description

[This run of daily builds](https://github.com/canonical/checkbox/actions/runs/7199217221/job/19610485472) failed because the `date_first_dispatched` attribute of a build is filled only once it is actually started, therefore if a build is pending it will be None. This fixes this issue and adds a sanity check on the build status (currently we can not determine how that field is filled for pending builds AND we want to resist/know spurious values in there so that we can proactively fix the script if they come up). 

## Resolved issues

[This build fail](https://github.com/canonical/checkbox/actions/runs/7199217221/job/19610485472)

## Documentation

This explains why the `None` case is included in the selector in a comment both in the function and in the test

## Tests

The new value is now included in the function test.

### DIFF
--- a/tools/release/lp_build_monitor_recipe.py
+++ b/tools/release/lp_build_monitor_recipe.py
@@ -68,6 +68,7 @@ def wait_every_source_build_started(build_recipe: LPSourcePackageRecipe):
                 "  ",
             )
         )
+        # avoid flooding LP with requests
         time.sleep(LP_POLLING_DELAY)
         pending_builds = build_recipe.getPendingBuildInfo()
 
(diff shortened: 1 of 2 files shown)


tools/release/lp_build_monitor_recipe.py
```diff
@@ -197,14 +202,27 @@ def monitor_retry_builds(builds_to_check: list[LPBuild]) -> list[LPBuild]:
             "Gathering build output",
         ]:
             # avoid flooding LP with requests
```
kissiel, line 204: I'd keep the comment just before the sleep

kissiel (approved):
Awesome, +1

kissiel (approved):
still awesome

PR #898, opened 2023-12-14
Disable fail fast in deb daily build (infra)

Hook25:
## Description

Given that every workflow monitors and tries to get as many builds as possible landed in the PPA, killing them all if one fails is not a great idea. Doing so disrupts the monitor->retry loop and removes the information from the daily build log about which ppa(s) failed specifically (namely if checkbox-ng and checkbox-support fail while all the others pass, the daily builds workflow will list 1 failures and all the others cancelled or a few pass, a few cancelled and 1 fail).

## Resolved issues

We can not determine from[ yesterday's build](https://github.com/canonical/checkbox/actions/runs/7199217221/job/19610485472) which PPAs are currently ok and which need some additional attention.

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/.github/workflows/deb-daily-builds.yml
+++ b/.github/workflows/deb-daily-builds.yml
@@ -32,6 +32,7 @@ jobs:
     runs-on: [self-hosted, linux, large]
     needs: ppa_update
     strategy:
+      fail-fast: false
       matrix:
         recipe:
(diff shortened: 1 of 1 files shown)


kissiel (approved):
makes sense, +1

PR #905, opened 2023-12-27
Refactor LED tests (BugFix)

stanley31huang:
Fixed the bugs of led-indicator/sysfs-leds tests (Bugfix)

## Description
This PR trying to solved following issues.
- the led-indicator/sysfs-leds failed for a sysfs LED with heartbeat configuration
- create a new led-indicator/gpio-controller-leds tests for GPIO LEDs with new GPIP_CONTROLLER_LEDS configuration variable, so we could simply defined the gpio chip number and gpio pin in the configuration.
- implemented two python scripts for both LED tests and having unit tests for them.

we also want to retired the shell scripts for led-indicator/sysfs-leds, and deprecated the led-indicator/gpio-leds tests. Once this PR is landed to beta channel, Stanley will send another PR for it.

## Resolved issues

Resolves OEMQA-3588

## Documentation
N/A

## Tests
[C3 submission with following config variable](https://certification.canonical.com/hardware/202309-32095/submission/349863/test-results/pass/)
```
[environment]
GPIP_CONTROLLER_LEDS=dl14:5:1 dl15:5:2 dl16:5:3
SYS_LEDS=SocPower:status

```

### DIFF
--- a/providers/base/tests/test_gpio_control_test.py
+++ b/providers/base/tests/test_gpio_control_test.py
@@ -284,7 +284,8 @@ class TestMainFunction(unittest.TestCase):
         mock_path.return_value = True
         mock_read.return_value = "mock-string"
 
-        dump_gpiochip(None)
+        with redirect_stdout(StringIO()) as stdout:
+            dump_gpiochip(None)
         mock_path.assert_called_once_with()
         mock_read.assert_called_once_with()
 
--- a/providers/base/tests/test_gpio_control_test.py
+++ b/providers/base/tests/test_gpio_control_test.py
@@ -3,6 +3,8 @@ import sys
 import argparse
 from unittest.mock import patch, Mock
 from pathlib import PosixPath, Path
+from io import StringIO
+from contextlib import redirect_stdout
 from gpio_control_test import GPIOController
 from gpio_control_test import blinking_test
 from gpio_control_test import dump_gpiochip
--- a/providers/base/tests/test_gpio_control_test.py
+++ b/providers/base/tests/test_gpio_control_test.py
@@ -303,7 +303,8 @@ class TestMainFunction(unittest.TestCase):
     def test_led_resource(self):
         mock_args = Mock(
             return_value=argparse.Namespace(mapping="DL14:5:1 DL14:5:2"))
-        leds_resource(mock_args())
+        with redirect_stdout(StringIO()) as stdout:
+            leds_resource(mock_args())
 
     def test_led_resource_with_unexpected_format(self):
         mock_args = Mock(
--- a/providers/base/tests/test_gpio_control_test.py
+++ b/providers/base/tests/test_gpio_control_test.py
@@ -6,6 +6,7 @@ from pathlib import PosixPath, Path
 from gpio_control_test import GPIOController
 from gpio_control_test import blinking_test
 from gpio_control_test import dump_gpiochip
+from gpio_control_test import leds_resource
 from gpio_control_test import register_arguments
 
 
@@ -298,6 +299,23 @@ class TestMainFunction(unittest.TestCase):
             str(context.exception), "/sys/kernel/debug/gpio file not exists"
         )
 
+    def test_led_resource(self):
+        mock_args = Mock(
+            return_value=argparse.Namespace(mapping="DL14:5:1 DL14:5:2"))
+        leds_resource(mock_args())
+
+    def test_led_resource_with_unexpected_format(self):
+        mock_args = Mock(
+            return_value=argparse.Namespace(mapping="DL14-5:1"))
+
+        with self.assertRaises(ValueError) as context:
+            leds_resource(mock_args())
+
+        self.assertEqual(
+            str(context.exception),
+            "not enough values to unpack (expected 3, got 2)"
+        )
+
 
 class TestArgumentParser(unittest.TestCase):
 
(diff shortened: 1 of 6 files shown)


kissiel:
What is the meaning of the magic/random number in the title?

stanley31huang:
@kissiel  it's the time I created that branch in my git repo, I was removed it from the title. Thanks.

providers/base/units/led/jobs.pxu
```diff
@@ -347,4 +347,50 @@ estimated_duration: 10
...
+estimated_duration: 3
+environ: GPIP_CONTROLLER_LEDS
+command:
+    awk '{
+        split($0, record, " ")
+        for (i in record) {
+            split(record[i], data, ":")
+            printf "name: %s\nchip_number: %s\nport: %s\n\n", data[1], data[2], data[3]
+        }
+    }' <<< "$GPIP_CONTROLLER_LEDS"
```
pieqq, line 369: This looks a little convoluted; it's a bit hard to read and cannot be tested easily.

How about a tiny Python script that takes `GPIP_CONTROLLER_LEDS` as an argument and does:

```python
def controller_leds_resource(text):
    output = ""
    resource_text = "name: {}\nchip_number: {}\nport: {}\n\n"
    for led in text.split():
        name, chip_number, port = led.split(":")
        output += resource_text.format(name, chip_number, port)
    return output
```

This way you can write a unit test for it.

pieqq (approved):
Sorry for the delay, Stanley! I had started a review of this but never sent it...

It works and there are passing unit tests, so it's all good and can land as is. Good work!

I made a few suggestions in the code, you might want to consider it. Let me know what you think.

providers/base/units/led/jobs.pxu
stanley31huang: Thanks for the feedback, I have updated the gpio_control_test.py scripts and led resource job. Please review the latest changes.

providers/base/tests/test_gpio_control_test.py
```diff
@@ -0,0 +1,351 @@
...
+            gpio_controller.value = "wrong_value"
+
+    @patch("gpio_control_test.GPIOController.get_gpiochip_mapping")
+    @patch("gpio_control_test.GPIOController.off")
+    @patch("gpio_control_test.GPIOController.on")
+    def test_blinking_function(self, mock_on, mock_off, mock_mapping):
+        mock_mapping.return_value = {"1": "32"}
+
+        gpio_controller = GPIOController("1", "1", "out", True)
+        gpio_controller.blinking(1, 0.5)
```
pieqq, line 255:
```suggestion
        gpio_controller.blinking(0.0001, 0.0001)
```

let's keep this as short as possible, otherwise this will slow down every unit test run in the future.

providers/base/tests/test_gpio_control_test.py
```diff
@@ -0,0 +1,351 @@
...
+
+class TestGPIOController(unittest.TestCase):
+
+    @patch("gpio_control_test.GPIOController.get_gpiochip_mapping")
+    @patch("gpio_control_test.GPIOController._unexport")
+    @patch("gpio_control_test.GPIOController._export")
+    @patch("gpio_control_test.GPIOController._write_node")
+    @patch("gpio_control_test.GPIOController._read_node")
+    @patch("gpio_control_test.GPIOController._node_exists")
+    def test_initial_gpio_controller_success(
```
pieqq, line 21:
```suggestion
    @patch("gpio_control_test.time.sleep", new=Mock)
    def test_initial_gpio_controller_success(
```

Prevent the `sleep` function from actually running, which slows down the unit test runs.

providers/base/tests/test_gpio_control_test.py
```diff
@@ -0,0 +1,351 @@
...
+        mock_blinking.assert_called_once_with(
+            mock_args().duration, mock_args().interval)
+
+    @patch("pathlib.Path.read_text")
+    @patch("pathlib.Path.exists")
+    def test_dump_gpiochip_test(self, mock_path, mock_read):
+        mock_path.return_value = True
+        mock_read.return_value = "mock-string"
+
+        dump_gpiochip(None)
```
pieqq, line 287:
```suggestion
        with redirect_stdout(StringIO()) as stdout:
            dump_gpiochip(None)
```

This is to prevent the string `mock-string` from popping up in the unit test results. It uses [contextlib.redirect_stdout()](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout) and `io.StringIO` so you'll have to import these at the top of your test module.

providers/base/tests/test_gpio_control_test.py
```diff
@@ -0,0 +1,351 @@
...
+            dump_gpiochip(None)
+
+        self.assertEqual(
+            str(context.exception), "/sys/kernel/debug/gpio file not exists"
+        )
+
+    def test_led_resource(self):
+        mock_args = Mock(
+            return_value=argparse.Namespace(mapping="DL14:5:1 DL14:5:2"))
+        leds_resource(mock_args())
```
pieqq, line 305:
```suggestion
        with redirect_stdout(StringIO()) as stdout:
            leds_resource(mock_args())
```

Similar to my other comment, this is to prevent a big chunk of data from being printed out in the unit test logs.

pieqq (changes requested):
Thanks a lot for the addition!

I ran the tests locally and realized they took a long time (more than 2 seconds) and they would output some text in the unit test report, so I suggest to make the following changes to avoid this.

pieqq:
/canonical/self-hosted-runners/run-workflows 0537d219415249a1f065699301a27c46be252c7b

pieqq (approved):
+1 Thanks!

PR #913, opened 2024-01-05
Edge to beta validation script (infra)

Hook25:
## Description

Before calling the [edge to beta](https://github.com/canonical/checkbox/blob/main/.github/workflows/checkbox-beta-release.yml) workflow, we need to validate that the operation is what we want to do. This introduces a new script with one automatic check that the operation is actually correct.

The check is: the latest daily build has built/pushed to the store the same hash that we have validated via canary. The rationale is: We are promoting from the store/ppa. If a daily build ran succesfully after we have moved the head of beta_validated, we have overwritten the last beta_validated version (as the workflow uploads to the store/LP). We can not promote the current version as it is untested.

## Resolved issues

This is preliminary work for: CHECKBOX-1056

## Documentation

The script has a documentation header

## Tests

The script comes with unit tests, to run them run:
```
$ pytest test_can_promote_edge.py
```

### DIFF
--- a/tools/release/can_promote_edge.py
+++ b/tools/release/can_promote_edge.py
@@ -14,6 +14,11 @@ import subprocess
 
 
 def get_gh_daily_builds_array() -> list[dict]:
+    """
+    Query github for the latest 20 runs of the daily-builds.yml workflow
+    returning a parsed array of dicts containing their headSha (hash the run
+    was triggered on) and conclusion.
+    """
     return json.loads(
         subprocess.check_output(
             [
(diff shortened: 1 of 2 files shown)


tools/release/can_promote_edge.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python
+
+"""
+This script is ment to be the validation step to gate the promotion of
```
yphus, line 4:
```suggestion
This script is meant to be the validation step to gate the promotion of
```

tools/release/can_promote_edge.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python
```
yphus, line 1:
```suggestion
#!/usr/bin/env python3
```

yphus (changes requested):
LGTM (except the two suggested fixes/typo).

yphus (approved):
+1

PR #914, opened 2024-01-06
Check the installed package instead of from ubuntu-archive (bugfix)

kissiel:
This PR is the rebased version of the following PR: 
https://github.com/canonical/checkbox/pull/640

It might be quicker to land it with this PR than go chasing rebases with the original one.


Hook25 (approved):
ty for doing this! +1

PR #916, opened 2024-01-08
Disable failfast in snap daily builds (infra)

Hook25:
## Description

If any of the snap (frontend) daily builds fail, all of the others are cancelled. This is undesirable because these builds are fairly quick and cancelling them only saves a few minutes but it leads to us having less data to make a deliberation on what the cause on ones failure may be/if there is anything else wrong to fix. Additionally, cancelling the workflow doesn't cancel the build itself, that will run to completion on LP either way.

## Resolved issues

N/A

## Documentation

N/A

## Tests

This was done (for more or less the same reason) for the deb/PPA builds.

### DIFF
--- a/.github/workflows/checkbox-snap-daily-builds.yml
+++ b/.github/workflows/checkbox-snap-daily-builds.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   snap:
     strategy:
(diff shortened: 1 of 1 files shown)


yphus (approved):
+1

PR #920, opened 2024-01-09
Unique validator for packaging metadata (BugFix)

fernando79513:
## Description

This PR is a follow up of #909. We were missing a check during the validation phase to make sure that there are not repeated packages in the packaging metadata files.

For that purpose, we have used the `UniqueValueValidator` for the `Depends`, `Recommends` and `Suggests` fields of the packaging metadata units, so they can't be repeated.

This change is not probable to make any break in previous versions, but it is possible if some package metadata units were using the same package twice.

## Resolved issues

Closes [CHECKBOX-1062](https://warthogs.atlassian.net/browse/CHECKBOX-1062)

## Documentation

Updated packaging-meta-data.rst

## Tests

To run the new packaging tests:
```
cd ~/checkbox/checkbox-ng
python3 -m pytest -k PackagingUnitFieldValidationTests
```

[CHECKBOX-1062]: https://warthogs.atlassian.net/browse/CHECKBOX-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/checkbox-ng/plainbox/impl/unit/packaging.py
+++ b/checkbox-ng/plainbox/impl/unit/packaging.py
@@ -124,6 +124,7 @@ import sys
 from plainbox.i18n import gettext as _
 from plainbox.impl.symbol import SymbolDef
 from plainbox.impl.unit import concrete_validators
+from plainbox.impl.unit.validators import UniqueValueValidator
 from plainbox.impl.unit.unit import Unit
 
 _logger = logging.getLogger("plainbox.unit.packaging")
@@ -153,6 +154,22 @@ class PackagingMetaDataUnit(Unit):
         """Version of the operating system."""
         return self.get_record_value(self.Meta.fields.os_version_id)
 
+    @property
+    def Depends(self):
+        """Package dependencies"""
+        return self.get_record_value(self.Meta.fields.Depends)
+
+    @property
+    def Recommends(self):
+        """Package recommendations"""
+        return self.get_record_value(self.Meta.fields.Recommends)
+
+    @property
+    def Suggests(self):
+        """Package suggestions"""
+        return self.get_record_value(self.Meta.fields.Suggests)
+
+
     class Meta:
 
         name = 'packaging meta-data'
@@ -163,6 +180,9 @@ class PackagingMetaDataUnit(Unit):
 
             os_id = 'os-id'
             os_version_id = 'os-version-id'
+            Depends = 'Depends'
+            Recommends = 'Recommends'
+            Suggests = 'Suggests'
 
         field_validators = {
             fields.os_id: [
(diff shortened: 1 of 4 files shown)


kissiel:
>This change is not probable to make any break in previous versions, but it is possible if some package metadata units were using the same package twice.

So we went from unpredictable behavior, to a predictable one. Having two units like that is/was a problem, so if we're breaking something with your changes, we really are just shedding light on the problem. I wouldn't go as far as calling it a breaking change.

fernando79513:
> > This change is not probable to make any break in previous versions, but it is possible if some package metadata units were using the same package twice.
> 
> So we went from unpredictable behavior, to a predictable one. Having two units like that is/was a problem, so if we're breaking something with your changes, we really are just shedding light on the problem. I wouldn't go as far as calling it a breaking change.

Okey, I will change it to "BugFix" then

checkbox-ng/plainbox/impl/unit/packaging.py
```diff
@@ -153,6 +154,22 @@ def os_version_id(self):
         """Version of the operating system."""
         return self.get_record_value(self.Meta.fields.os_version_id)
 
+    @property
+    def Depends(self):
```
kissiel, line 158: Why is this is capitalized (and the following ones)?

kissiel (changes requested):
If I understand the code correctly, the "Depends" field in the meta were capitalized because they are one-offs (type-like). But the code proposed here adds properties based on that section name (Capitalized), so the field name should be lowercase.

checkbox-ng/plainbox/impl/unit/packaging.py
fernando79513: Good question. I didn't go too deep into this one, I just followed the previous schema to declare properties required for the new validators.
If I changed the name of the field to `depends`, I am getting this error: 
```
  File "/home/fernando/checkbox/checkbox-ng/plainbox/impl/unit/validators.py", line 582, in check_in_context
    value = getattr(unit, field2prop(field))
AttributeError: 'PackagingMetaDataUnit' object has no attribute 'Depends'. Did you mean: 'depends'?
```
Again, this is a bit confusing, but I think since we are getting the attribute based on the field name, they have to be the same. 

I tried changing the ´field2prop´ function so we translate fields starting with caps to lowercase, as we are doing with "-" and "_":
```
  def field2prop(field):
-   return str(field).replace('-', '_')
+   return str(field).replace('-', '_').lower()
```
But doing that this way created a conflict, because there is already a field named "depends" on JobDefinition units

fernando79513: One solution could be to modify the "field2prop" function, translating fields starting with caps to lower, and create a new validator, that only checks the values for the same unit type:

kissiel: Thank you for your explanations.
Indeed it looks like the use of capitalized property names is dictated by the automagic of turning fields from .pxus to props.
I don't like it, but I guess we need to roll with that at least for now.

kissiel (approved):
Thanks for the explanations. This warrants a +1 now.

PR #921, opened 2024-01-09
Add trigger and needs constraint to promotion workflow (infra)

Hook25:
## Description

We have decided that the promotion to the beta channel of the snap is governed by the moving target that is the `beta` branch. Once that branch is moved, the current `edge` version of the snap/deb must be promoted to stable. This introduces the mechanism to the beta promotion workflow along a sanity check to avoid race conditions (namely, the last "green" run of the daily build must have happened on the commit pointed by `beta`, else something went wrong!).

## Resolved issues

Resolves: CHECKBOX-1056

## Documentation

N/A (This is just the workflow work, it works as described before)

## Tests

The mechanism was partially validated offline, it is easier to iterate online on the last details

### DIFF
--- a/.github/workflows/checkbox-beta-release.yml
+++ b/.github/workflows/checkbox-beta-release.yml
@@ -2,10 +2,34 @@ name: Beta version of checkbox
 run-name: Promote edge versions of checkbox to beta
 
 on:
+  push:
+    branches:
+      - beta
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


kissiel (approved):
The double `apt update` is needed indeed. Once for potential curl installation, and the other after we add the PPA.

Good changes, +1

PR #927, opened 2024-01-10
Use action-validator to valdate workflows (infra)

Hook25:
## Description

[This](https://github.com/canonical/checkbox/pull/921/files) has introduced a problem in the `checkbox-beta-release.yml` workflow. We could not tell without because we are missing an automated way to check for trivial regression like these. This adds a new validator that checks yaml file and automatically rejects malformed ones.

## Resolved issues

[Anti-regression for trivial grammar mistakes like here](https://github.com/canonical/checkbox/pull/921/files)

## Documentation

N/A

## Tests

N/A

[This was taken from the official repo here](https://github.com/mpalmer/action-validator/blob/main/.github/workflows/qa.yml#L194C1-L203C56)

### DIFF
--- a/.github/workflows/tox-tools-release.yaml
+++ b/.github/workflows/tox-tools-release.yaml
@@ -4,11 +4,11 @@ on:
   push:
     branches: [ main ]
     paths:
-      - tools/release/**
+      - tools/release/*
   pull_request:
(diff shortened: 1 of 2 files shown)


kissiel (approved):
Goodies! +1

PR #929, opened 2024-01-11
Only stage version.txt like in frontend (infra)

Hook25:
## Description

The version calculator of frontend snaps is currently including in the snap anything in the root of the build directory. 

This is leading to strange consequences, namely right now we are including the build log of a failing when retrying it, but if one or more builds in a build set "pass", they will be staged in the next retry if any failed. This is unnecessary and an unintended consequence that we did fix a while back in the runtime snap.

This fixes the issue by adding the same "stage" constraint to the frontend snap that we have in the backend.

## Resolved issues

Additional content in the snap

## Documentation

N/A

## Tests

N/A

This is the same issue/fix we did for the checkbox core snap


kissiel (approved):
Yes, please!
+1

yphus (approved):
+1

Issue #97, opened 2022-11-28
LP1974209: Unable to detect the insertion of TBT3 storage


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1974209

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Incomplete	| 2022-05-19 15:36:43	| 6	| High	| False  	|

## Description
[Summary]
checkbox-cli cannot detect the insertion of TBT3 storage.
However, system can detect and access it without problem.

screenshot: https://drive.google.com/file/d/1toDC-DPg1ipB62-TpG5-0Uxj5cUB2udM/view?usp=sharing

[Steps to reproduce]
1. Boot into OS
2. Run "checkbox-cli run com.canonical.certification::thunderbolt3/insert"
3. Follow the steps to insert TBT3 storage

[Expected result]
checkbox-cli can detect the TBT3 storage.

[Actual result]
checkbox-cli cannot detect the TBT3 storage.

I/O logs:

INSERT NOW

/tmp/nest-lzry0nu5.6bc01a9b6b0687602c09bf237634540db1ee4c24bf56fb1e9a8066d49c444cda/removable_storage_watcher.py:497: PyGIDeprecationWarning: GObject.timeout_add_seconds is deprecated; use GLib.timeout_add_seconds instead
  GObject.timeout_add_seconds(timeout, self._on_timeout_expired)
You have failed to perform the required manipulation in time

[Failure rate]
3/3

[Additional information]
CID: 202112-29802
SKU: TRBA-DVT2-C4
Image: canonical-oem-somerville-jammy-amd64-20220425-25+jellyfish-tentacool+X17
system-manufacturer: Dell Inc.
system-product-name: XPS 9320
bios-version: 0.2.15
CPU: 12th Gen Intel(R) Core(TM) i7-1280P (20x)
GPU: 0000:00:02.0 VGA compatible controller [0300]: Intel Corporation Device [8086:46a6] (rev 0c)
kernel-version: 5.15.0-9027-generic

[Stage]
Issue reported and logs collected at a later stage

## Attachments
No attachments

Tags:
['jammy-test']

beliaev-maksim:
This thread was migrated from launchpad.net
##### https://launchpad.net/~pieq wrote on 2022-05-24 07:43:10:
Importance set to High because it is still possible to check this manually.

[...]

PR #930, opened 2024-01-11
fix detecting insertion/removal of tb3 drive (BugFix)

kissiel:
## Description

Fix Thunderbolt "insert" and "remove" tests on Classic.

On classic images the "scsi detection" that was employed by the job definition did not work.
Historically we've had two version of insertion/removal detection code: one for Classic and one for Core. The patch that introduced the tests was using this approach, but in fact it was it was not needed, as the version provided by the implementation in checkbox-support (that was used exclusively for UC) works on both.

The patch I'm proposing here makes Checkbox use the working one, regardless of whether we're running on UC or Classic, thus fixing the problem

## Resolved issues
Fixes: #97, CHECKBOX-1123

## Documentation
No changes.

## Tests

Tested on classic image with the tb3 drive.

### DIFF
--- a/providers/base/units/thunderbolt/jobs.pxu
+++ b/providers/base/units/thunderbolt/jobs.pxu
@@ -7,11 +7,7 @@ requires: manifest.has_thunderbolt == 'True'
 estimated_duration: 20.0
 template-engine: jinja2
 command:
-    {%- if __on_ubuntucore__ %}
-        checkbox-support-run_watcher insertion thunderbolt
-    {%- else %}
-        removable_storage_watcher.py insert --timeout 40 scsi
-    {% endif -%}
+ checkbox-support-run_watcher insertion thunderbolt
 _siblings: [
     { "id": "after-suspend-thunderbolt/insert",
       "_summary": "thunderbolt/insert after suspend",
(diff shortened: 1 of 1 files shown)


yphus:
The insert/removal jobs now run the same script on core/classic but the storage test still use `removable_storage_test.py -s 268400000 scsi` for thunderbolt. Have you tested it too?

kissiel:
> The insert/removal jobs now run the same script on core/classic but the storage test still use `removable_storage_test.py -s 268400000 scsi` for thunderbolt. Have you tested it too?

Yes. I got this result:
```
Found the following mounted scsi partitions:
    /dev/nvme1n1p1 : /tmp/tmp7ok46dc5 : None bits/s
--------------------
/dev/nvme1n1p1 (Total Data Size / iteration: 255.9662 MB):
        [Iteration 0] Average Speed: 1085.5509
        Summary:
                Total Data Attempted: 255.9662 MB
                Total Time to write: 0.2358 secs
                Average Write Time: 0.2358 secs
                Average Write Speed: 1085.5509 MB/s
```

yphus (approved):
+1

PR #936, opened 2024-01-15
Avoid breaking the cat pipe by reading from head (infra)

Hook25:
## Description

Currenty the frontend snaps are failing to build. In the logs snapcraft is reporting a non-0 return value on the new workaround part. This error is `141` (broken pipe). This happens because we are `cat`-ing from `/dev/urandom` (an infinite source) and closing the pipe via head, effectively breaking it. This doesn't make some series fail because it may be (not confirmed, but a very likely theory) that snapcraft started to set the `set -o pipefail` only recently-ish.

This PR fixes this problem by making head read the file directly instead of reading it via cat

## Resolved issues

Broken daily builds here: https://github.com/canonical/checkbox/actions/runs/7505763759

## Documentation

N/A 

This doesn't change the effect of the command, it should only make it go through without "breaking" the pipe and making snapcraft mad 

## Tests

The following fails (exits with return code 141) with the same error:
```
!/bin/bash

set -o pipefail
set -e

cat /dev/urandom | head -c 1000 > /dev/null
```

while this doesn't:
```
!/bin/bash

set -o pipefail
set -e

head -c 1000 /dev/urandom > /dev/null
```

### DIFF
--- a/checkbox-snap/series_classic16/snap/snapcraft.yaml
+++ b/checkbox-snap/series_classic16/snap/snapcraft.yaml
@@ -72,4 +72,4 @@ parts:
       echo "This file is a workaround for a bug in the automated snap review tool" > $SNAPCRAFT_PART_INSTALL/size_workaround
       echo "this only contains random bytes to pad the snap to 16kb" >> $SNAPCRAFT_PART_INSTALL/size_workaround
       echo "see: https://bugs.launchpad.net/review-tools/+bug/2049093" >>  $SNAPCRAFT_PART_INSTALL/size_workaround
-      cat /dev/urandom | head -c 16384 >> $SNAPCRAFT_PART_INSTALL/size_workaround
+      head -c 16384 /dev/urandom >> $SNAPCRAFT_PART_INSTALL/size_workaround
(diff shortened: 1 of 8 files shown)


yphus (approved):
Tested with the checkbox classic20 frontend snap:

```
Building workaround-automated-review-issue 
+ echo 'This file is a workaround for a bug in the automated snap review tool'
+ echo 'this only contains random bytes to pad the snap to 16kb'
+ echo 'see: https://bugs.launchpad.net/review-tools/+bug/2049093'
+ head -c 16384 /dev/urandom
Staging config-variables 
+ snapcraftctl stage
Staging launchers 
+ snapcraftctl stage
Staging version-calculator 
+ snapcraftctl stage
Staging workaround-automated-review-issue 
+ snapcraftctl stage
Priming config-variables 
+ snapcraftctl prime
Priming launchers 
+ snapcraftctl prime
Priming version-calculator 
+ snapcraftctl prime
Priming workaround-automated-review-issue 
+ snapcraftctl prime
The 'passthrough' property is being used to propagate experimental properties to snap.yaml that have not been validated.
Snapping |                                                                                        
Snapped checkbox_3.3.0-dev87_amd64.snap
```

PR #938, opened 2024-01-16
Add contrib provider ce-oem GitHub workflow (Infra)

pieqq:
## Description

1. Add workflow for the ce-oem provider (this workflow is adapted from existing workflows used for the generic providers)
2. Update contrib ce-oem tox config:
    - Point to the correct version of Python using {envpython}
    - Remove the "develop installation" step for the ce-oem provider
    - Convert the test step into a test coverage one to generate an XMLreport to be used later by Codecov
3. Update codecov configuration for the ce-oem provider in the contrib area (to make the check non-blocking both as a patch and for the whole project)

## Resolved issues

CHECKBOX-1157

## Tests

Used Codecov validation to check the codecov.yml file was valid:

```
curl -X POST --data-binary @codecov.yml https://codecov.io/validate
```


Hook25 (approved):
+1, ty

PR #939, opened 2024-01-16
Fix namespace for ce-oem jobs (Infra)

pieqq:
## Description

Since the ce-oem provider namespace was switched to `com.canonical.contrib`, some jobs need to be updated, otherwise the provider validation fails.

## Resolved issues

CHECKBOX-1157

## Tests

Locally, in a venv with Checkbox installed:

```
$ ./manage.py develop -d $PROVIDERPATH
$ ./manage.py validate
(...)
```


Hook25 (approved):
+1, ty

Issue #102, opened 2022-11-28
LP1816445: SMART resource should use -x instead of -i for NVME


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-resource/+bug/1816445

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| New	| 2019-02-18 14:59:20	| 8	| Medium	| False  	|

## Description
Currently, we use smartctl -i to determine if a disk supports SMART by looking for the SMART supported and SMART enabled/disabled lines.

problem is that some drives (NVMe specifically and perhaps some SSDs) support SMART functionality but do not display this data.

It was suggested by Western Digital to glean SMART support from -x output rather than -i.

## Attachments
No attachments

Tags:
['hwcert-server']

beliaev-maksim:
This thread was migrated from launchpad.net
##### https://launchpad.net/~rodsmith wrote on 2020-02-26 15:27:46:
Detection of NVMe support in the disk is the least of the problems here. That issue could be worked around fairly easily; however....

I found the following with some information on the status of smartmontools NVMe support:

https://www.smartmontools.org/wiki/NVMe_Support

According to that page, support is "experimental." Notably absent is support for initiating a self-test ("smartctl -t"), which our current test does; however, it is possible to get an overall health assessment ("smartctl -H") and various other pieces of information. I've tried poking around on jehan and on three computers I own with NVMes, but I haven't been able to initiate a test in the way we do on ATA disks. So, do you have pointers to any other information on how to initiate a self-test? If not....

* I could add some NVMe-specific code to use "smartctl -H"
  and/or other calls that produce more detailed information
  to try to assess the NVMe's overall health and use that
  instead of a self-test in the way we do with ATA disks.
* Alternatively, we could shelve this task for a while in
  the hopes that self-test support will eventually be
  added for NVMe devices.
* Another option would be to re-write the test to no
  longer rely on self-tests, even for ATA disks, but to
  instead simply call "smartctl -H" (or whatever) on
  all disks.

I haven't yet had time to think about this in depth, and so I don't currently have a strong preference among these options. The third one would likely be the most time-consuming, though, since we'd need to test the resulting major change on a wide variety of devices (including servers with hardware RAID -- both ones that do and ones that don't support pass-through of SMART data from RAIDed disks).

[...]

PR #940, opened 2024-01-16
Fix disk resource job to fetch SMART status (BugFix)

Hook25:
## Description

Currently the disk_block_device_resource script doesn't correctly fetch the SMART support status of some devices. The reason is that some devices (namely, nvme drives, although more may exist) fail to be reported as supporting SMART by smartctl even though they do. This is due to the fact that this support is not reported in `smartctl -i ` but it is there in `smartctl -x`, although in a slightly different format.

This changes the command that the resource job runs. This also includes some new unit tests for this resource and a small refactoring. 

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/102

## Documentation

N/A

## Tests

I have an nvme drive that was wrongly reported by this resource as not supporting SMART. After these changes it correctly reports the SMART status.

### DIFF
--- a/providers/resource/bin/block_device_resource.py
+++ b/providers/resource/bin/block_device_resource.py
@@ -4,7 +4,8 @@ import os
 import re
 import shlex
 import textwrap
-from glob import glob
+from pathlib import Path
+from itertools import chain
 from subprocess import check_output, CalledProcessError
 
 rootdir_pattern = re.compile("^.*?/devices")
@@ -36,7 +37,7 @@ def usb_support(name, version):
     path = rootdir_pattern.sub("", os.readlink("/sys/block/%s" % name))
 
     # Remove the usb config.interface part of the path
-    m = re.match("((.*usb\d+).*\/)\d-[\d\.:\-]+\/.*", path)  # noqa: W605
+    m = re.match(r"((.*usb\d+).*\/)\d-[\d\.:\-]+\/.*", path)  # noqa: W605
     if m:
         device_path = m.group(1)
         hub_port_path = m.group(2)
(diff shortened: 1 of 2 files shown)


providers/resource/bin/block_device_resource.py
```diff
@@ -116,45 +130,54 @@ def smart_support(name):
...
-rotation: %(rotation)s
-smart: %(smart)s
-""" % {"name": name,
-       "state": state,
-       "usb2": usb2,
-       "usb3": usb3,
-       "rotation": rotation,
-       "smart": smart})
+    command = "smartctl -x /dev/%s" % name
+    diskinfo = check_output(
```
kissiel, line 135: Please add guarding against typical subproc exceptions

kissiel (changes requested):
The version on main doesn't crash on problems with smartctl. With this patch it crashes on my system.

kissiel:
The "crash-y" setup I've got:

```
➜  bin git:(main) ✗ ll /sys/block/*/device            

lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/nvme0n1/device -> ../../nvme0
lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/nvme1n1/device -> ../../nvme1
lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/nvme2c2n1/device -> ../../nvme2
lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/nvme2n1/device -> ../../nvme-subsys2
lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/nvme3n1/device -> ../../nvme3
lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/sda/device -> ../../../5:0:0:0
lrwxrwxrwx 1 root root 0 sty 16 18:36 /sys/block/sdb/device -> ../../../8:0:0:0
lrwxrwxrwx 1 root root 0 sty 16 21:43 /sys/block/sdc/device -> ../../../9:0:0:0
```
The nvme2c2n1 doesn't exist at `/dev/` with the same exact name.

The `c2` part is thrown in for ssd with multipath enabled, but the same thing is not added in `/dev/`
More context is in the `nvme_alloc_ns` function in the `drivers/nvme/host/core.c`

kissiel (approved):
TYVM for the improvements!
Works better now!

PR #944, opened 2024-01-17
Renamed packaging.py to packaging_metadata.py (Bugfix)

fernando79513:
This change was made to avoid naming conflicts when importing the packaging module from packaging.py

## Description

When running unit tests using the `unittest` module we run into an import error:

```
python3 -m unittest test_packaging.py 

ERROR: test_packaging (unittest.loader._FailedTest.test_packaging)
ImportError: Failed to import test module: test_packaging
...
  File "/home/pieq/dev/work/checkbox/checkbox-ng/plainbox/impl/unit/packaging.py", line 118, in <module>
    from packaging import version
ImportError: cannot import name 'version' from partially initialized module 'packaging' (most likely due to a circular import) (/home/pieq/dev/work/checkbox/checkbox-ng/plainbox/impl/unit/packaging.py)
```
This was because the python file had the same name as the module we were importing.

## Documentation

No new Documentation changes

## Tests

Running the unit tests works without any issue:
```
python3 -m unittest test_packaging.py
pytest test_packaging_metadata.py
```

Running `./manage.py packaging` works the same as before. In my case, I had to modify also one of the gitignored files to make it work, `checkbox-ng/checkbox_ng.egg-info/entry_points.txt`  but that should not be necessary if you setup the package again.


kissiel (approved):
And the final +1 :)

PR #949, opened 2024-01-18
Rename ARM Vector Floating Point templates to avoid clash (BugFix)

pieqq:
## Description

These two templates use the same `id` which is a problem when ensuring every template has a unique id field.

## Resolved issues

This will help with CHECKBOX-1074

## Tests

```
$ ./manage.py validate
(...)

$ checkbox-cli list-bootstrapped com.canonical.certification::cpu-automated
$ checkbox-cli list-bootstrapped com.canonical.certification::server-cpu
--> both commands should return the same thing before and after applying the patch
```


kissiel (approved):
Looks good!
+1

Now the question is whether there's a commercial project using the previous naming scheme...

PR #951, opened 2024-01-18
Add uniqueness for Template Units id (BugFix)

pieqq:
## Description

So far, Template Units did not have any unique identifier. The `id` field was used, but it was not unique. This make it difficult to track the origin of an instantiated job.

A few things are addressed:

1. `TemplateUnit` now inherits `UnitWithId` instead of `Unit`, which brings, on top of the usual `id` definition, some validators, including that there are no two templates with the same `id`.
2. A `template-id` field is introduced. It is **optional**: if it's not in the unit definition, it will be computed based on the value of the `id` field (using the same string, with characters like `{`, `}` and ` ` removed). This is mostly done for backward compatibility.
3. When a query is issued to retrieve the value of a Template `id`, it should never be parsed. Before this PR, if the Template unit was using the Jinja2 engine, its `id` would be "rendered", which exposed erroneous information, as explained in commit 81c953598888030eb154954be433a018945e1b85.
4. If there are several Template Units with the same `template-id`, an error message will point to the right Template when running `./manage.py validate` (before, it would point to the unit's `id` field; it now points to the `template-id` field)

## Resolved issues

Required step for CHECKBOX-1074

## Documentation

The Template Unit reference documenation has been updated to mention the `template-id` field.

## Tests

If a provider has 2 different templates defined with the same `id` (as it's the case with the base provider before #949) , its validation will now fail:

```
$ ./manage.py validate
(...)
error: units/cpu/jobs.pxu:179: template 'cpu/arm_vfp_support_platform', field 'id', clashes with 1 other unit, look at: units/cpu/jobs.pxu:173-187, units/cpu/jobs.pxu:189-203
(...)
error: units/cpu/jobs.pxu:179: template 'cpu/arm_vfp_support_platform', field 'id', clashes with 1 other unit, look at: units/cpu/jobs.pxu:173-187, units/cpu/jobs.pxu:189-203
(...)
```

If a provider has 2 different templates that have the same `template-id` (or get the same generated `template-id` from their `id` field), its validation will now fail:

```
$ ./manage.py validate
(...)
error: units/cpu/jobs.pxu:173-187: template 'cpu/arm_vfp_support_platform', field 'template-id', clashes with 1 other unit, look at: units/cpu/jobs.pxu:173-187, units/cpu/jobs.pxu:189-203
(...)
error: units/cpu/jobs.pxu:173-187: template 'cpu/arm_vfp_support_platform', field 'template-id', clashes with 1 other unit, look at: units/cpu/jobs.pxu:173-187, units/cpu/jobs.pxu:189-203
(...)
```
(Note that the `field` mentioned is different from the previous output)

As for exposing the template `id`:

Before:

```
$ checkbox-cli list template
(...)
template 'com.canonical.contrib::ce-oem-optee/xtest--'
template 'com.canonical.contrib::ce-oem-optee/xtest--'
template 'com.canonical.contrib::ce-oem-touchscreen/evdev/single-touch-tap-'
template 'com.canonical.contrib::ce-oem-touchscreen/evdev/maximum-touch-tap-'
template 'com.canonical.certification::optical_drive_{name}'
```

Vs. After:

```
$ checkbox-cli list template
(...)
template 'com.canonical.contrib::ce-oem-optee/xtest-suite-description'
template 'com.canonical.contrib::ce-oem-optee/xtest-suite-description'
template 'com.canonical.contrib::ce-oem-touchscreen/evdev/single-touch-tap-product_slug'
template 'com.canonical.contrib::ce-oem-touchscreen/evdev/maximum-touch-tap-product_slug'
template 'com.canonical.certification::optical_drive_name'
```

Unit tests have been added as well.

### DIFF
--- a/checkbox-ng/plainbox/impl/highlevel.py
+++ b/checkbox-ng/plainbox/impl/highlevel.py
@@ -293,9 +293,11 @@ class Explorer:
 
     def _template_to_obj(self, unit):
         return PlainBoxObject(
-            unit, group=unit.Meta.name, name=unit.id, attrs=OrderedDict((
+            unit, group=unit.Meta.name, name=unit.template_id,
+            attrs=OrderedDict((
                 ('id', unit.id),
                 ('partial_id', unit.partial_id),
+                ('template_id', unit.template_id),
                 ('template_unit', unit.template_unit),
                 ('template_resource', unit.template_resource),
                 ('template_filter', unit.template_filter),
--- a/checkbox-ng/plainbox/impl/test_highlevel.py
+++ b/checkbox-ng/plainbox/impl/test_highlevel.py
@@ -1,0 +1,40 @@
+# This file is part of Checkbox.
+#
+# Copyright 2024 Canonical Ltd.
+# Written by:
+#   Pierre Equoy <pierre.equoy@canonical.com>
+#
+# Checkbox is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3,
+# as published by the Free Software Foundation.
+#
+# Checkbox is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+from unittest import TestCase
+
+from plainbox.impl.highlevel import Explorer
+from plainbox.impl.unit.template import TemplateUnit
+
+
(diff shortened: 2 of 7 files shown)


pieqq:
Discussion with Maciek:

- we could still have `template-id` that would be automatically generated from the `id` field (e.g. `optical_drive_{name}` would become `optical_drive_name`) if not provided manually. This helps having a more presentable identifier if we need to expose it.
- in this case, uniqueness should be tested against `template-id` and not `id`, so `TemplateUnit` should inherit from `Unit` as it was originally
- Validators need to be added to `TemplateUnit`

pieqq:
Still need to

- update PR description
- update documentation to expose the `template-id` field.

Hook25 (approved):
+2, 1 for the feature and 1 for the docs update!

PR #957, opened 2024-01-22
Update contruct to v2.10.57 (bugfix)

Hook25:
## Description

The vendorized version of contruct uses the now removed (as of python3.12) `imp` module. This PR updates the vendorized module to a more up-to-date version of construct with the following changes:
- All imports made relative to vendor 
- Fixed the creation of a module to be compatible with python3.5
- Patched (keeping the old version) py3compact.py to keep the `supportshalffloats` among other functions (used in python3.5 that doesn't support the struct `e` formatter)

## Resolved issues

Partially solves: https://launchpadlibrarian.net/710488943/buildlog_ubuntu-noble-amd64.checkbox-support_3.3.0~dev96~ubuntu24.04.1_BUILDING.txt.gz

## Documentation

This PR documentation and the commit messages should be enough to document what was done

## Tests

N/A

### DIFF
--- a/checkbox-support/checkbox_support/vendor/construct/core.py
+++ b/checkbox-support/checkbox_support/vendor/construct/core.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-import struct, io, binascii, itertools, collections, pickle, sys, os, tempfile, hashlib, importlib, imp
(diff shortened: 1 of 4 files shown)


kissiel (approved):
makes sense; looks good; +1

PR #958, opened 2024-01-22
Fixed unknown method name in checkbox-support test (bugfix)

Hook25:
## Description

When running via the deb runner, this test fails on python 3.12. This should always fail and the reason why it doesn't is still unclear to me at this point. The function name is wrong and it should be **`assert_`**`called_once_with`

This updates every test with the correct function call.

## Resolved issues

Resolves: https://launchpadlibrarian.net/710488943/buildlog_ubuntu-noble-amd64.checkbox-support_3.3.0~dev96~ubuntu24.04.1_BUILDING.txt.gz

## Documentation

N/A

## Tests

N/A


kissiel (approved):
makes sense, +1
(I wonder if toxxing will choke on it or not)

PR #959, opened 2024-01-22
Fixed SyntaxWarning correctly marking regex strings (bugfix)

Hook25:
## Description

Regexes in the pactl parser module are incorrectly put in normal strings.

This marks every string as a raw string, removing the SyntaxWarning that can be seen in the builder and pytest runs.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


kissiel (approved):
Oh yes, please and thank you. +1

PR #961, opened 2024-01-23
Changed variable name to environment from environ (bugfix)

Hook25:
## Description

[This PR](https://github.com/canonical/checkbox/pull/958/files) fixed the test but executing it reveals that it was also wrong. The section is called `environment` not `environ`. This fixes it and updates the function that is used to check to another one because unittest mocks work differently (and the other returns False on unittest and True on pytest)

## Resolved issues
N/A

## Documentation

N/A

## Tests

N/A


kissiel (approved):
+1

PR #962, opened 2024-01-23
Mock module and explain why it was done (Bugfix)

Hook25:
## Description

A few versions ago, urwid splitted the `web.py` file into `web.py`, `_web.js` and `_web.css`. They have also moved the `web` module inside `display`. To be backward compatible they have added a mechanism to redirect calls from the old name `urwid.web_display` to `urwid.display.web`.  For reasons that I don't fully comprehend, `inspect.stack` seems to be triggering this mechanism, leading to the `web` module being imported. This causes an issue on Ubuntu Noble currently that doesn't let us build checkbox-ng. This is because this happens in any test that uses `Origin.get_caller_origin` (many of them) but the Noble package is currently broken [(bug in packaging fixed upstream here)](https://github.com/urwid/urwid/pull/750).

This PR fixes this issue while we wait for the debian package (and subsequently the ubuntu package) to be updated, allowing us to build checkbox-ng on noble again 

## Resolved issues

Fixes daily builds of checkbox-ng

## Tests

Reproducing the issue:
```
$ python -m venv venv
$ . venv/bin/activate
$ pip install urwid
$ rm venv/lib/python*/site-packages/urwid/display/_web.css
$ rm venv/lib/python*/site-packages/urwid/display/_web.js
```

Now you have a package that is the same as the one installable on ubuntu noble. Alternatively, on ubuntu noble, run 
```
$ sudo apt install python3-urwid
```

Verify that the package is broken by running the following:
```
$ python3 -c "import urwid.display.web"
```

Verify that inspect causes the same issue as above by running the following:
```
$ python3.12 -c "import urwid; import inspect; inspect.stack(0)"
```

Now try runnint the following, this is the result of this patch:
```
$ python3.12 -c "import urwid; import inspect; import sys; sys.modules['urwid.display.web']={}; inspect.stack(0)"
```

> Note: Use the latest version of urwid to test this, the mechanism was introduced this bug triggers another bug on an older versions of the library! (RecursionError)

### DIFF
--- a/checkbox-ng/plainbox/impl/secure/origin.py
+++ b/checkbox-ng/plainbox/impl/secure/origin.py
@@ -25,6 +25,7 @@
 import functools
 import inspect
 import os
+import sys
 
 from plainbox.abc import ITextSource
 from plainbox.i18n import gettext as _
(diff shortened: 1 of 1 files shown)


kissiel (approved):
Thank you for the fix and for a great explanation!

Issue #494, opened 2023-05-24
If the ethernet ping test fails, it can cause a traceback


djacobs98:
### Bug Description

com.canonical.certification:ethernet/ping_(interface) is a test that attempts to ping the gateway for the specified Ethernet interface.
This test uses a python script located at: checkbox/providers/base/bin/gateway_ping_test.py

The script uses a dictionary called ping_summary to store the results of the actual ping test. However, if the script is unable to determine the IP to use, this variable remains initialized as None, and this causes a traceback.

lines 276-288 of gateway_ping_test.py
```
    host = get_host_to_ping(
        interface=args.interface, verbose=args.verbose, default=args.host)
    if args.verbose:
        print(_("Checking connectivity to {0}").format(host))
    ping_summary = None
    if host:
        ping_summary = ping(host, args.interface, args.count,
                            args.deadline, args.verbose)
    if ping_summary is None or ping_summary['received'] == 0:
        print(_("No Internet connection"))
        if ping_summary.get('cause'):
            print("Possible cause: {}".format(ping_summary['cause']))
        return 1
```

If get_host_to_ping fails to find a host, it returns None.  
This causes ping_summary to remain None.
The line that checks if ping_summary is None is True.
The script then tries to check the attribute 'cause' which doesn't exist for None, and causes a traceback.

Instead, if get_host_to_ping returns None, the test should fail right there.

### To Reproduce

Run checkbox-iiotg-classic.test-runner and select Automated Tests for 22.04 Server.
Select the Ethernet tests.
Disconnect the Ethernet cable from your device's Ethernet interface.
Run the test.

The test should legitimately fail because its gateway address cannot be determined.
However the test should not fail with a traceback.

### Environment

OS: IOTG Server RT-GA
Checkbox provider: checkbox-iiotg-classic.test-runner
Test plan: Automated for Server.

### Relevant log output

```shell
output from checkbox submission, i/o log:
Adjusting deadline to 3 seconds to fit 2 pings
Desired ip address 10.102.88.1 is not reachable from enp0s30f5, using None instead
Desired ip address 10.102.88.1 is not reachable from enp0s30f5, using None instead
Desired ip address 10.102.88.1 is not reachable from enp0s30f5, using None instead
Desired ip address 10.102.88.1 is not reachable from enp0s30f5, using None instead
Desired ip address 10.102.88.1 is not reachable from enp0s30f5, using None instead
Desired ip address 10.102.88.1 is not reachable from enp0s30f5, using None instead
[...]

PR #969, opened 2024-01-29
Update ping test to not crash on failing pings (BugFix)

Hook25:
## Description

If the ping test fails, it will produce a traceback, this is wrong. This PR changes that leading to a correct failure path

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/494

## Documentation

N/A

## Tests

Tests added to a new file with 100% coverage of relevant code

### DIFF
--- a/providers/base/bin/gateway_ping_test.py
+++ b/providers/base/bin/gateway_ping_test.py
@@ -309,12 +309,18 @@ def main(argv) -> int:
     )
     if args.verbose:
         print(_("Checking connectivity to {0}").format(host))
-    ping_summary = {}
+
     if host:
         ping_summary = ping(
             host, args.interface, args.count, args.deadline, args.verbose
         )
-    if not ping_summary or ping_summary["received"] == 0:
+    else:
+        ping_summary = {
+            "received": 0,
+            "cause": "Unable to find any host to ping",
+        }
+
+    if ping_summary["received"] == 0:
         print(_("No Internet connection"))
         if ping_summary.get("cause"):
             print("Possible cause: {}".format(ping_summary["cause"]))
--- a/providers/base/bin/gateway_ping_test.py
+++ b/providers/base/bin/gateway_ping_test.py
@@ -277,11 +277,11 @@ def main(args):
         interface=args.interface, verbose=args.verbose, default=args.host)
     if args.verbose:
         print(_("Checking connectivity to {0}").format(host))
-    ping_summary = None
+    ping_summary = {}
     if host:
         ping_summary = ping(host, args.interface, args.count,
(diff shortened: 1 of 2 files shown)


providers/base/bin/gateway_ping_test.py
```diff
@@ -219,85 +219,125 @@ def ping(host, interface, count, deadline, verbose=False):
...
+    args = parse_args(argv)
+
     # If given host is not pingable, override with something pingable.
     host = get_host_to_ping(
-        interface=args.interface, verbose=args.verbose, default=args.host)
+        interface=args.interface, verbose=args.verbose, default=args.host
+    )
     if args.verbose:
         print(_("Checking connectivity to {0}").format(host))
-    ping_summary = None
```
fernando79513, line 280: Would't it be better to have something like:
```
    if host:
        ping_summary = ping(
            host, args.interface, args.count, args.deadline, args.verbose
        )
    else:
        ping_summary = {"received": 0, "cause": "Could not find a host"}

    if ping_summary["received"] == 0:
        print(_("No Internet connection"))
        if ping_summary.get("cause"):
            print("Possible cause: {}".format(ping_summary["cause"]))
        return 1
```

providers/base/tests/test_gateway_ping_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @patch("gateway_ping_test.get_host_to_ping")
+    @patch("gateway_ping_test.ping")
+    def test_invalid_arguments_count_deadline(
+        self, mock_ping, mock_get_host_to_ping
+    ):
+        with self.assertRaises(SystemExit):
+            main(["-c", "10", "-d", "8"])
+
+
```
fernando79513, line 82: Flake 8
```suggestion

```

providers/base/tests/test_gateway_ping_test.py
```diff
@@ -0,0 +1,90 @@
+import unittest
+from unittest.mock import patch, MagicMock
```
fernando79513, line 2: Unused
```suggestion
from unittest.mock import patch
```

fernando79513 (commented):
Nice job with the fix and with the tests. Just some small comments.

providers/base/bin/gateway_ping_test.py
Hook25: good idea

providers/base/tests/test_gateway_ping_test.py
Hook25: we use black but sure, fogot to run it

fernando79513 (approved):
LGTM +1

Issue #941, opened 2024-01-16
Traceback on `./manage.py develop` when `/var/tmp/checkbox-providers-develop` doesn't exist


kissiel:
### Bug Description

`develop` assumes that `/var/tmp/checkbox-providers-develop`  exists, and crashes if it doesn't.

### To Reproduce

1. Ensure there's no `/var/tmp/checkbox-providers-develop`  dir
2. run `./manage.py develop`

### Environment

Environment agnostic

### Relevant log output

```shell
(venv-small) ➜  2023.com.kissiel:testing ./manage.py develop 
$PROVIDERPATH is defined, so following provider sources are ignored ['/usr/local/share/plainbox-providers-1', '/usr/share/plainbox-providers-1', '/home/kissiel/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop'] 
CRITICAL plainbox.crashes: Executable 'manage.py' invoked with Namespace(log_level=None, debug_console=False, trace=[], pdb=False, debug_interrupt=False, uninstall=False, force=False, directory='/var/tmp/checkbox-providers-develop', command=<plainbox.provider_manager.DevelopCommand object at 0x7fa124877a10>) has crashed
Traceback (most recent call last):
  File "/home/kissiel/checkbox/checkbox-ng/plainbox/impl/clitools.py", line 560, in dispatch_and_catch_exceptions
    return self.dispatch_command(ns)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kissiel/checkbox/checkbox-ng/plainbox/impl/clitools.py", line 556, in dispatch_command
    return ns.command.invoked(ns)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kissiel/checkbox/checkbox-ng/plainbox/provider_manager.py", line 554, in invoked
    if pp_env and not os.path.samefile(pp_env, ns.directory):
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen genericpath>", line 101, in samefile
FileNotFoundError: [Errno 2] No such file or directory: '/var/tmp/checkbox-providers-develop'
Traceback (most recent call last):
  File "/home/kissiel/checkbox-stuff/providers/2023.com.kissiel:testing/./manage.py", line 16, in <module>
    setup(
  File "/home/kissiel/checkbox/checkbox-ng/plainbox/provider_manager.py", line 1692, in setup
[...]

PR #974, opened 2024-01-31
Fix crash on missing plainbox-provider-develop (BugFix)

Hook25:
## Description

Currently Checkbox crashes when `/var/tmp/checkbox-providers-develop` is missing if one tries to develop a provider (even if `$PROVIDER_PATH` is defined). This fixes this problem.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/941

## Documentation

N/A

## Tests

With `main` Checkbox run:
```
# rm -rf /var/tmp/checkbox-providers-develop
# ./manage.py develop
[! boom !]
```

Re-run on this branch, it should work as expected.

### DIFF
--- a/checkbox-ng/plainbox/provider_manager.py
+++ b/checkbox-ng/plainbox/provider_manager.py
@@ -552,7 +552,9 @@ class DevelopCommand(ManageCommand):
     def invoked(self, ns):
         pp_env = os.getenv("PROVIDERPATH")
         try:
-            samefile = os.path.samefile(pp_env, ns.directory)
+            samefile = pp_env is not None and os.path.samefile(
+                pp_env, ns.directory
+            )
         except FileNotFoundError:
             samefile = False
         if pp_env and not samefile:
--- a/checkbox-ng/plainbox/provider_manager.py
+++ b/checkbox-ng/plainbox/provider_manager.py
@@ -551,7 +551,11 @@ class DevelopCommand(ManageCommand):
 
     def invoked(self, ns):
         pp_env = os.getenv("PROVIDERPATH")
-        if pp_env and not os.path.samefile(pp_env, ns.directory):
+        try:
+            samefile = os.path.samefile(pp_env, ns.directory)
+        except FileNotFoundError:
+            samefile = False
+        if pp_env and not samefile:
             _logger.warning(
                 "$PROVIDERPATH is defined, ignoring -d/--directory"
                 " and developing in: %s", pp_env
--- a/checkbox-ng/plainbox/provider_manager.py
+++ b/checkbox-ng/plainbox/provider_manager.py
@@ -563,10 +563,6 @@ class DevelopCommand(ManageCommand):
                     "$PROVIDERPATH is defined, ignoring -d/--directory"
                     " and developing in: %s", pp_env
                 )
-            _logger.warning(
-                "$PROVIDERPATH is defined, ignoring -d/--directory"
-                " and developing in: %s", pp_env
-            )
             ns.directory = pp_env
         pathname = os.path.join(
             ns.directory, "{}.provider".format(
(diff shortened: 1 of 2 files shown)


checkbox-ng/plainbox/provider_manager.py
```diff
@@ -551,7 +551,13 @@ def register_parser(self, subparsers):
...
     def invoked(self, ns):
         pp_env = os.getenv("PROVIDERPATH")
-        if pp_env and not os.path.samefile(pp_env, ns.directory):
+        try:
+            samefile = pp_env is not None and os.path.samefile(
+                pp_env, ns.directory
+            )
+        except FileNotFoundError:
+            samefile = False
+        if pp_env and not samefile:
```
fernando79513, line 560: The logic is a bit confusing here, I think it would be clearer like this
```suggestion
        # If $PROVIDERPATH is defined, use it instead of the one on the
        # namespace
        if pp_env:
            try:
                samefile = os.path.samefile(pp_env, ns.directory)
            except FileNotFoundError:
                samefile = False
            if not samefile:
                _logger.warning(
                    "$PROVIDERPATH is defined, ignoring -d/--directory"
                    " and developing in: %s", pp_env
                )
```

checkbox-ng/plainbox/test_provider_manager.py
```diff
@@ -288,12 +291,20 @@ def test_develop(self, mock_path_entry):
...
 
+        # PROVIDERPATH defined
+        mock_getenv.return_value = "/tmp/com.example.test.provider"
+        self.tool.main(["develop"])
+        self.assertFileContent(filename, content)
+
     @mock.patch('plainbox.impl.providers.v1.get_universal_PROVIDERPATH_entry')
-    def test_develop__force(self, mock_path_entry):
+    @mock.patch('os.getenv')
+    def test_develop__force(self,mock_getenv, mock_path_entry):
```
fernando79513, line 301:
```suggestion
    def test_develop__force(self, mock_getenv, mock_path_entry):
```

checkbox-ng/plainbox/test_provider_manager.py
```diff
@@ -288,12 +291,20 @@ def test_develop(self, mock_path_entry):
         self.tool.main(["develop"])
         self.assertFileContent(filename, content)
 
+        # PROVIDERPATH defined
+        mock_getenv.return_value = "/tmp/com.example.test.provider"
+        self.tool.main(["develop"])
+        self.assertFileContent(filename, content)
```
fernando79513, line 297: Is this test is wrong? 
You are checking the content of filename but provider path should be prioritized. 
It asserts that the file exists because it has been created in the same tests without a provider path.

checkbox-ng/plainbox/test_provider_manager.py
```diff
@@ -312,11 +323,22 @@ def test_develop__force(self, mock_path_entry):
...
         self.assertFileContent(filename, content)
 
+        # PROVIDERPATH defined
+        mock_getenv.return_value = "/tmp/com.example.test.provider"
+        filename = "/tmp/com.example.test.provider/com.example.test.provider"
+        os.makedirs(os.path.dirname(filename), exist_ok=True)
+        with open(filename, "wt+") as stream:
+            stream.write("should have been overwritten")
+        self.tool.main(["develop", "--force"])
+        self.assertFileContent(filename, content)
```
fernando79513, line 333: Same for this test

fernando79513 (commented):
The fix should be working fine, but I think there might be an error with the tests. 
Also, I added a small comment that could improve the clarity.

checkbox-ng/plainbox/test_provider_manager.py
Hook25: This is ok, note, filename is overwritten here so its the correct name

checkbox-ng/plainbox/test_provider_manager.py
Hook25: Stopped being lazy, splitted the test into its own function

Issue #857, opened 2023-12-01
[Checkbox Control] Unable resume from suspend since operation not permitted problem causes RTC cannot be configured properly


baconYao:
### Problem
When testing via `Checkbox Control` with specific jobs combination, there's high possibility that DUT cannot resume from suspend, then tester can only observer that Host is reconnecting at suspend_advanced_auto job (like screenshot below)
![Screenshot from 2023-12-01 10-55-16](https://github.com/canonical/checkbox/assets/8318678/450805e7-e594-43cf-a2a9-4dd7a7552b7c)

In the beginning, I believed this was a real issue where the Device Under Test (DUT) couldn't enter suspend or hang in suspend. Therefore, I monitored the journal log through SSH or the serial console. However, in this situation, I couldn't reproduce the problem.

Therefore, I eliminated all connections (SSH or serial console) and conducted the test using checkbox locally. I still couldn't reproduce the issue.

Only by using Checkbox Control and without any connections (SSH or serial console), I can consistently reproduce this issue.

Finally, I attempted to manually press the power button when the issue occurred. The result was that the Device Under Test (DUT) could resume normally, and some messages were displayed in the suspend_advanced_auto job (like screenshot below).
![Screenshot from 2023-12-01 10-57-17](https://github.com/canonical/checkbox/assets/8318678/a40fde70-358c-4db2-b8cb-7a35481e99ad)
From the above image, it can be seen that `rtcwake: /dev/rtc0: unable to find device: Operation not permitted` is exposed.
Therefore, I reasonably suspect that during the test execution with Checkbox Control, there might be a chance of encountering `Operation not permitted`, leading to the inability to correctly set the time for rtc.

### Failure Rate

- `Checkbox Control` with `suspend_advanced_auto` job: 0/5
- `Checkbox Control` with `bluetooth/detect-output` + `suspend_advanced_auto` job: 7/10
- `sudo Checkbox Local` with `bluetooth/detect-output` + `suspend_advanced_auto` : 0/5

### Another similar issue of `operation not permitted`

- `rfkill: cannot open /dev/rfkill: Operation not permitted `
  - submission: https://certification.canonical.com/hardware/202308-31931/submission/346071/test/9/result/38010572/
    - `rtcwake: /dev/rtc0: unable to find device: Operation not permitted` can also be observed in same session
      - https://certification.canonical.com/hardware/202308-31931/submission/346071/test/892/result/38010575/

[...]

PR #979, opened 2024-02-01
Improve suspend/suspend_advanced* tests #857 (BugFix)

rmartin013:
## Description
rtwake can sometimes fail to set the alarm. While running the command in background before suspending the system, in such a case the whole test plan will hang.

This patch change the mode to allow the command to return immediately, while still setting the alarm, which allow to serialize its result.

## Resolved issues
https://github.com/canonical/checkbox/issues/857

## Tests
Tested on Riverside jobs using a side loaded provider... Issue is still under observation, but this change isn't preventing the correct execution of the suspend test (at least on this device).


kissiel:
/canonical/self-hosted-runners/run-workflows 2d428f719caa1a819c1a9a846d40679691952916

kissiel (approved):
TYVM! This is exactly what we need.

kissiel:
Codecov is clearly drunk here as it lists loss of coverage for the python files.
Overriding...

Issue #493, opened 2023-05-24
Ping gateway test can fail if 2 or more interfaces are connected to the same subnet.


djacobs98:
### Bug Description

I'm testing a device with 2 Ethernet interfaces.  Both are plugged into the same switch.

Sometimes the Automated test: com.canonical.certification:ethernet/ping_<interface> will fail, even though the command manually passes.
An example of the command the test executes is:
ping 10.102.88.1 -c 2 -w 3  -Ienp0s31f6   
where:
10.102.88.1 is the gateway
(-I as in "aye" not "ell")
enp0s31f6 is the interface to use.  This test has failed with the other interface, enp2s0, as well.

The issue is there are two interfaces on this device and both are plugged into the same switch.  If the system has determined that the other interface is the default one for talking to the subnet, ping packets may be misdirected until the network table is updated.

This can cause some packets to be dropped, which will cause the test to fail.

This is the same underlying issue as with the Manual Ethernet Hotplug test (https://github.com/canonical/checkbox/issues/291)
The work around is also the same: Disconnect the other cable(s) before running this test.  This is inconvenient since this is part of an automated test suite.

Therefore a  better solution would be to alter the test script to down all other interfaces except the interface under test.
Then after the test has completed, restore the network.

### To Reproduce

Have 2 or more Ethernet interfaces plugged into the same Switch and verify they can both ping the gateway.
Run the Ethernet test from the Automated test plan.

Generally the first interface will pass, but the second one to run will fail.

### Environment

Details:
OS: IOTG Server w/RT-GA Kernel.
Checkbox: checkbox-iiotg-classic (--classic --channel=latest/edge)

pieqq:
Is this problem the same as #371 ?

djacobs98:
> Is this problem the same as #371 ?

#371 is for the test in the manual test plan.
This is for a test in the automated test plan.  I think the underlying cause is the same - the network table decides on one interface as the default for the subnet.  Forcing "the other interface" to do the ping will work, eventually. However you will see some dropped packets until the table is updated.

Issue #371, opened 2023-03-16
When running Ethernet Device test using Checkbox remote on a device with multiple Ethernet ports, one port fails to test


pieqq:
### Bug Description

Failure symptom:
Follow the test steps below, and one port will fail

Note:
Use the CDTS and checkbox20 installed online to test, and one of the two ports will fail.
The port failure rate of the specified IP is higher than that of the unspecified IP.

Failure rate:
100%

### To Reproduce

1.Press enter and unplug the ethernet cable from the port enp1s0 of the System.
2.After 15 seconds plug it back in.
3.Checkbox session may be interrupted but it should come back up.
4.Waiting for cable to get disconnected.
5.Cable unplugged!
6.Waiting for the cable to get connected.
7.........Cable detected!

### Environment

- Ubuntu desktop 20.04.5
- Kernel: 5.15.0-46-generic x86_6
- Checkbox remote v2.2

### Relevant log output

```shell
Example from the submission.json from a device with 2 ports:


       {
            "id": "ethernet/hotplug-eno1",
            "full_id": "com.canonical.certification::ethernet/hotplug-eno1",
            "name": "Ensure hotplugging works on port eno1",
            "certification_status": "blocker",
            "category": "Ethernet Device tests",
            "category_id": "com.canonical.plainbox::ethernet",
            "status": "pass",
            "outcome": "pass",
            "comments": null,
[...]

PR #981, opened 2024-02-05
Fix and refactor gateway_ping_test (Bugfix)

Hook25:
## Description

The `gateway_ping_test.py` test regularly fails in many SRU runs. This can be due to the fact that the `get_host_to_ping` function tries to do way more than it should (resorting to regularly scraping the arp table and pinging broadcast  even when the default gateway is reachable). This refactors and hopefully makes more robust the gateway ping test

## Resolved issues

Possibly fixes: https://github.com/canonical/checkbox/issues/493 and https://github.com/canonical/checkbox/issues/371 (TBD)

## Documentation

N/A

## Tests

This also adds tests for every unit and utils functions

### DIFF
--- a/providers/base/bin/gateway_ping_test.py
+++ b/providers/base/bin/gateway_ping_test.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # This file is part of Checkbox.
 #
-# Copyright 2007-2023 Canonical Ltd.
+# Copyright 2007-2024 Canonical Ltd.
 # Written by:
 #   Brendan Donegan <brendan.donegan@canonical.com>
 #   Daniel Manrique <daniel.manrique@canonical.com>
@@ -237,11 +237,11 @@ def is_reachable(ip, interface, verbose=False):
     """
     Ping an ip to see if it is reachable
     """
-    result = ping(ip, interface, 3, 10, verbose)
+    result = ping(ip, interface, 3, 10, verbose=verbose)
     return result["transmitted"] >= result["received"] > 0
 
 
-def get_default_gateway_reachable_on(interface: str, verbose=False) -> str:
+def get_default_gateway_reachable_on(interface: str) -> str:
     """
     Returns the default gateway of an interface if it is reachable
     """
--- a/providers/base/bin/gateway_ping_test.py
+++ b/providers/base/bin/gateway_ping_test.py
@@ -57,6 +57,7 @@ class Route:
 
     def _get_default_gateway_from_ip(self):
         try:
+            # Note: this uses -o instead of -j for xenial/bionic compatibility
             routes = subprocess.check_output(
                 [
                     "ip",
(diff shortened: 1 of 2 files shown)


providers/base/bin/gateway_ping_test.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # This file is part of Checkbox.
 #
-# Copyright 2007-2014 Canonical Ltd.
+# Copyright 2007-2023 Canonical Ltd.
```
baconYao, line 4: Hi @Hook25, is `2023` or `2024` ?

providers/base/bin/gateway_ping_test.py
```diff
@@ -83,151 +108,297 @@ def _get_default_gateway_from_bin_route(self):
...
+
+def is_reachable(ip, interface, verbose=False):
+    """
+    Ping an ip to see if it is reachable
+    """
+    result = ping(ip, interface, 3, 10, verbose)
+    return result["transmitted"] >= result["received"] > 0
+
+
+def get_default_gateway_reachable_on(interface: str, verbose=False) -> str:
```
baconYao, line 244: The `verbose` seems not be used, do we still need it here?

providers/base/bin/gateway_ping_test.py
Hook25: ops you are right, wrong year

providers/base/bin/gateway_ping_test.py
Hook25: no, removed. You also made me spot an error in 3 lines up (verbose is a kwarg, should be passed as such, the order of params has changed)

providers/base/bin/gateway_ping_test.py
```diff
@@ -83,151 +108,297 @@ def _get_default_gateway_from_bin_route(self):
...
-            except subprocess.CalledProcessError:
-                pass
-    # If default host given, ping it as well,
-    # to try to get it into the arp table.
-    # Needed in case it's not responding to broadcasts.
-    if default:
+    def _get_ip_addr_info(self):
+        return subprocess.check_output(
+            ["ip", "-o", "addr", "show"], universal_newlines=True
+        )
```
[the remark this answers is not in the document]
Hook25: I wouldnt add the decorator, there is value to explicitly checking the exception locally (namely, you can not accidentally non-raise a second command that is supposed to fail if you ever add it).

I didn't catch the exception because I don't expect it to ever fail. Do you think I should? I think that if `ip addr show` fails, the script should blow up.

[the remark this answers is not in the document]
fernando79513: Nah, It is okey like that, I just wanted to know your opinion about it

fernando79513 (approved):
LGTM +1

PR #985, opened 2024-02-13
Add template-summary and template-description fields to the Template Unit (New)

pieqq:
## Description

Add two fields to Template Units:

- `template-summary`
- `template-description`

These fields mimick the summary and description fields of the Job Unit to provide a quick explanation and a longer form one about what a template unit does.
    
Both of these fields are translatable, and validation is added for the `template-summary` field to make sure if a value is provided, it is one line only and less than 80 characters.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1239

## Documentation

Template Unit reference page is updated to mention both fields.

## Tests

- Unit tests have been added (you can run them with `pytest checkbox-ng/plainbox/impl/unit/test_template.py`)
- Validators are present, so if you create a template with a `template-summary` field of more than 80 characters, or if the value for this field has more than one line, you will see warnings when running validation for the provider.

For instance, if I add the following `template-summary` to an existing `xxx` template unit in the base provider:

```
template-summary:
 This is a veeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeery long summary!
 and it's more than one line anyway...
```

and then run `./manage.py validate` for the base provider, I will see three warnings:

```
warning: units/.../jobs.pxu:69: template 'xxx', field 'template-summary', field should be marked as translatable
warning: units/.../jobs.pxu:69: template 'xxx', field 'template-summary', please use only one line
warning: units/.../jobs.pxu:69: template 'xxx', field 'template-summary', please stay under 80 characters
```

### DIFF
--- a/checkbox-ng/plainbox/impl/unit/template.py
+++ b/checkbox-ng/plainbox/impl/unit/template.py
@@ -27,6 +27,7 @@ import string
 
 from plainbox.i18n import gettext as _
 from plainbox.i18n import gettext_noop as N_
+from plainbox.impl.decorators import instance_method_lru_cache
 from plainbox.impl.resource import ExpressionFailedError
 from plainbox.impl.resource import Resource
 from plainbox.impl.resource import ResourceProgram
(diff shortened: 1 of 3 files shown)


kissiel (approved):
Good stuff! +1

PR #987, opened 2024-02-14
Implement ExpectNot operator for metabox

Hook25:
Minor: small refactor of expect

## Description

Currently we have no way to assert that something wasn't printed by a metabox scenario if it is interactive. This adds this new functionality by making expect fail when the socket is closed and the input string wasnt found (before it used to always timeout and only return on True)

## Resolved issues

Needed for another PR, I need a way to AssertNotPrinted but for interactive runs

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/metabox/metabox/core/actions.py
+++ b/metabox/metabox/core/actions.py
@@ -49,6 +49,10 @@ class Expect(ActionBase):
     handler = 'expect'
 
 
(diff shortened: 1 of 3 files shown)


kissiel (approved):
Simpler, stronger, nicer, better.
+1

Issue #427, opened 2023-04-14
"save the session and quit" has no effect


pieqq:
### Bug Description

When running Checkbox remote, some jobs (user-interact and user-interact-verify) will first prompt the following to the user:

```
Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]: 
```

`c` and `s` work as expected, but `q` has the same effect as `ENTER`, it launches the job.

### To Reproduce

1. Install Checkbox locally using the contributing guide
2. In a terminal, launch `checkbox-cli service` as root
3. In another one, launch `checkbox-cli remote 127.0.0.1`
4. Select a test plan that has a user-interact or user-interact-verify job, and then select that job
5. When prompted, press `q`

Demonstration with a user-interact job followed by a user-interact-verify job (both have a `command` field set to `true`):

```
-----------------------------[ Running job 2 / 3 ]------------------------------
----------------------------------[ smoke/user-interact ]----------------------------------
ID: com.canonical.certification::smoke/user-interact
Category: Uncategorised
--------------------------------------------------------------------------------
Description:

(...)

Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]: q
--------------------------------------------------------------------------------
Outcome: job passed
-----------------------------[ Running job 3 / 3 ]------------------------------
---------------------------------[ smoke/user-interact-verify ]----------------------------------
ID: com.canonical.certification::smoke/user-interact-verify
Category: Uncategorised
--------------------------------------------------------------------------------
Description:

(...)

Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]: q
Please decide what to do next:
  outcome: job needs verification
  comments: none
Pick an action
  c => add a comment
  p => set outcome to pass
  f => set outcome to fail
  s => set outcome to skip
  r => re-run this job
    => set suggested outcome [job passed]
[cpfsr]: 
```

### Environment

- Checkbox version: 2.4

PR #989, opened 2024-02-14
Fix quit option for manual jobs (bugfix)

Hook25:
## Description

The quit option of a manual job (as described in the linked issue) doesn't work. This fixes the action by doing what a user would expect:
- Finalizes the session on the agent side
- Exits the controller with a message (and an error code, given that the session didn't run to completion)

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/427

## Documentation

N/A (This is the expected behaviour)

## Tests

This adds a new metabox scenario to test this behaviour


kissiel (dismissed):
Good stuff, +1

kissiel (approved):
Still looking good, +1

PR #990, opened 2024-02-15
Fix metabox expect typerror (infra)

Hook25:
## Description

This fixes the expect function to support all use cases and turns the stdout_data buffer into a string instead of always encoding every search pattern, decoding the message data.

## Resolved issues

Broken main metabox

## Documentation

N/A

## Tests

N/A


kissiel (approved):
LGTM, +1

PR #991, opened 2024-02-15
Fix gateway ping (BugFix)

kissiel:
## Description

This PR simplifies the argument handling in `gateway_ping_test` test script.

The `count`, `threshold`, and `deadline` options were never used and they made the logic complicated. The verbosity option was also a mixed bag, and [lately was suffering from a bug](https://github.com/canonical/checkbox/issues/993)

~~This is a partial fix for the [#371](https://github.com/canonical/checkbox/issues/371) bug.~~

Namely it moves code around to make allow for getting the interface name.
This PR doesn't change anything, just adds an command-line option and refactors some functions

## Resolved issues

Fixes: 
https://github.com/canonical/checkbox/issues/993
https://warthogs.atlassian.net/browse/CHECKBOX-1258

https://github.com/canonical/checkbox/issues/992

## Tests

Unit tests were added and the previous ones got adjusted.

### DIFF
--- a/providers/base/bin/gateway_ping_test.py
+++ b/providers/base/bin/gateway_ping_test.py
@@ -38,7 +38,7 @@ import sys
 import time
 
 from contextlib import suppress
-from typing import Dict
+from typing import Dict, List
 
 
 class Route:
@@ -211,11 +211,11 @@ class Route:
         return Route(Route.get_any_interface())
 
 
-def is_reachable(ip, interface, verbose=False):
+def is_reachable(ip, interface):
     """
     Ping an ip to see if it is reachable
     """
-    result = ping(ip, interface, 3, 10, verbose=verbose)
+    result = ping(ip, interface)
     return result["transmitted"] >= result["received"] > 0
 
 
@@ -237,7 +237,7 @@ def get_default_gateway_reachable_on(interface: str) -> str:
     )
 
 
-def get_any_host_reachable_on(interface: str, verbose=False) -> str:
+def get_any_host_reachable_on(interface: str) -> str:
     """
     Returns any host that it can reach from a given interface
     """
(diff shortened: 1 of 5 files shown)


Hook25 (dismissed):
+1, lgtm

kissiel:
After testing it seems that this PR also fixes https://github.com/canonical/checkbox/issues/992

providers/base/bin/gateway_ping_test.py
```diff
@@ -474,52 +400,72 @@ def main(argv) -> int:
...
+
+    :param interface: the interface name to check
+    :return: True if the interface is a cable interface, False otherwise
+
+    Looking at the `man 7 systemd.net-naming-scheme` we can see that
+    even the `eth` matching may be an overkill.
+    """
+    if not isinstance(interface, str) or not interface:
+        return False
+    return interface.startswith("en") or interface.startswith("eth")
```
Hook25, line 467: It is a bit weird, given the `Route` structure, that these are free functions

providers/base/bin/gateway_ping_test.py
```diff
@@ -474,52 +400,72 @@ def main(argv) -> int:
...
 
+def get_default_gateways() -> Dict[str, str]:
+    """
+    Use `route` program to find default gateways for all interfaces.
+
+    returns a dictionary in a form of {interface_name: gateway}
+    """
+    try:
+        routes = subprocess.check_output(
+            ["route", "-n"], universal_newlines=True
```
Hook25, line 442: I didn't double check what it does to be honest, but before we were defining and envvar called `env={"LANGUAGE": "C"},`, is the omission of that intentional?

Hook25 (approved):
Well done, give a thought to the comment below, I would have used the api in Route to instanciate N routes and getting default gateways via that class. This just because the Route class has some robustness to it, trying to fetch the default gateway via multiple routes.

providers/base/bin/gateway_ping_test.py
kissiel: I did not remove it by accident, but I considered that we *never* strip the locale in other places, I'd rather:
* make it consistent
* **if** the need arises it'll be better to introduce like a `locale_less_check_output` that does this to all calls of `check_output`
* make it simpler

providers/base/bin/gateway_ping_test.py
kissiel: I don't know how I feel about using `class`es as namespaces. ATM, for me the less digging, the less indent, the better.

providers/base/bin/gateway_ping_test.py
Hook25: ok I wanted to make sure, we can go back to this if troubles arises

PR #998, opened 2024-02-19
Enable multi interface testing in `gateway_ping_test.py` (New)

kissiel:
## Description

This PR rearranges the code of `gateway_ping_test.py` to enable running the test on more than one interface. This is needed for SRU changes where we want the test to pass if *any* of the cable interfaces can ping the gateway.

## Resolved issues

This partially resolves: [CHECKBOX-1243](https://warthogs.atlassian.net/jira/software/c/projects/CHECKBOX/boards/590?selectedIssue=CHECKBOX-1243)

## Tests

Everything should be covered by the unit tests.
I've also tested it manually with multiple interface configurations


providers/sru/units/sru-ubuntucore.pxu
```diff
@@ -39,7 +39,7 @@ include:
     ethernet/detect
     ethernet/info_automated
     networking/http
-    ethernet/ping_.*
+    ethernet/ping-with-any-cable-interface
```
Hook25, line 42: Ok I can see that this may be very controversial. Are we sure we don't want to test all interfaces? This reduces the coverage of the test suite quite a bit

Hook25: ok, won't we ever have the need of testing all (/some) interfaces? I'm a bit worried we may have a a regression on some interfaces but not others. Is this something that could/may happen? It is a bit strange that we don't exclude some interface from the testing instead of just running only the test on one

kissiel: Until we set up devices in the lab in a way that every port is connected, this is the only way I'm affraid. (after reprovisioning the interface names may change, Pierre reported that).

Hook25 (approved):
+1

PR #1001, opened 2024-02-19
Always escape outdata in metabox output storage (infra)

Hook25:
## Description

Metabox should alway escape all ansi sequences in the output storage else accessing it in any way that leads to it being printed on the screen (printing/debugging/etc.) risks breaking the tty. We also never need them as they could be searched via an assertion but doing so would cause all sorts of problems with no payoff

## Resolved issues

metabox seldomly breaking the tty while testing

## Documentation

N/A

## Tests

N/A


kissiel (approved):
Nice find!
+1

kissiel (approved):
Lol. +1

PR #1002, opened 2024-02-20
Remove $SNAP_DATA refs from checkbox-support UT (BugFix)

kissiel:
## Description

The unit tests' outcome depended on whether they were run from a snap or not (as it's the case for VS Code).
This PR makes the test use mocked env variable
## Tests
Tested by running tests from outside and inside the snap.


Hook25 (approved):
+1, lgtm

Issue #980, opened 2024-02-02
storage_device_dm fails because it is being run on ubuntu-save


kissiel:
### Bug Description

The test yielded storage_device_dm.
```
      Testing disk dm-1
identified as a devmapper device...
/dev/mapper/ubuntu-save-1eb6afa8-7bdf-45bf-ab90-06fc999f871a already mounted at /var/lib/snapd/hostfs/run/mnt/ubuntu-save
6.880859375MB of free space avaialble
Forcing memory setting to 1.72021484375MB
+ bonnie++ -d /var/lib/snapd/hostfs/run/mnt/ubuntu-save -u root -r 1.72021484375
Using uid:0, gid:0.
Writing a byte at a time...done
Writing intelligently...Can't write block.: No space left on device
Can't write block 938.
Traceback (most recent call last):
  File "/tmp/nest-tiwqx3em.be08cee4e0b57534c945fac0784a6efa5b26abe774343e113ac6cb92092baa26/storage_test.py", line 152, in <module>
    main()
  File "/tmp/nest-tiwqx3em.be08cee4e0b57534c945fac0784a6efa5b26abe774343e113ac6cb92092baa26/storage_test.py", line 146, in main
    devmapper_test(udev_name)
  File "/tmp/nest-tiwqx3em.be08cee4e0b57534c945fac0784a6efa5b26abe774343e113ac6cb92092baa26/storage_test.py", line 119, in devmapper_test
    run_bonnie(mount_dir)
  File "/tmp/nest-tiwqx3em.be08cee4e0b57534c945fac0784a6efa5b26abe774343e113ac6cb92092baa26/storage_test.py", line 94, in run_bonnie
    sp.check_call(cmd, shell=True)
  File "/snap/checkbox/4492/checkbox-runtime/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'bonnie++ -d /var/lib/snapd/hostfs/run/mnt/ubuntu-save -u root -r 1.72021484375' returned non-zero exit status 1.
```

Obviously we should skip partitions like that. Possibly skip the tests if we can't find a partition with enough space on any of them.

### To Reproduce

run  `com.canonical/certification__disk_storage_device_dm-1` on `cert-inventec-e200-intel-kernel-20-beta`

### Environment

-

PR #1003, opened 2024-02-20
ignore `dm` devices that serve only ubuntu-* partitions (BugFix)

kissiel:
## Description
On some Core systems there are `dm-*` block devices that udevadm parser fro checkbox-support considered being valid disks, even though they served only as the "disk" to store the recovery/service partitions.
Those "disks" shouldn't be used with storage testing (and the storage tests are generated per disk). Therefore this PR introduces a heuristic where now, disks that host the "service" partitions are not reported as proper disks.

## Resolved issues

Fixes: https://warthogs.atlassian.net/jira/software/c/projects/CHECKBOX/boards/590?selectedIssue=CHECKBOX-1228
Fixes: https://github.com/canonical/checkbox/issues/980

## Tests
I'm providing a test that covers the scenario.
I've also tested it on the [affected device](https://certification.canonical.com/hardware/202104-28914/), by:
* running a test plan without modifying the Checkbox snaps in any way
* observing that two dm-* devices are listed by Checkbox to be tested
* overlaying a filesystem on top of the Checkbox20 snap
* overwriting the udevadm.py file (the one patched with this PR)
* rerunning the same storage test plan and the same storage generator
* observing that only one partition (the valid one) is present in the list of disks to be tested


Hook25 (approved):
Makes sense and ty for the new test/test data

PR #1004, opened 2024-02-20
Add tiny bit of tests to the storage tests (New)

kissiel:
## Description

When working on the #1003 I thought I will have to modify the `storage_test.py`, so when I started reading it, I wrote some tests along the way.
Turns out I didn't have to modify the file, but the tests are still valuable, so let's have them :)

## Tests

are being proposed here...


Hook25 (approved):
+1, ty for adding also the dep to toxing

Issue #700, opened 2023-09-05
Test plan filtering: improvements case-insensitive and filter by ID


pieqq:
When selecting a test plan in the first Checkbox screen, there is the option to filter the list. Although it works well, it would benefit from two improvements:

1. the filter should be case-insensitive: searching for `tutorial` should return a test plan called `Tutorial`.
2. the filter should also search in the test plan IDs. For instance, if a test plan is named `Tutorial Test Plan` and its ID is `tutorial-tp`, searching for `tp` should return it. This should work regardless what's displayed on the screen (user can use the `i` key to toggle between displaying the "human-readable" name, or the ID).

To see the Checkbox filtering option in action:

1. Run Checkbox: `checkbox.checkbox-cli`
2. In the `Select test plan` screen, press `f`, then type a string and press `Enter` to validate

PR #1005, opened 2024-02-21
Support case insensitive searching in names or ids (bugfix)

Hook25:
## Description

Searching is very cumbersome sometimes in checkbox test plans because you either don't know the test plan name (but only the id) or you don't remember the exact case (also, everybody always types lowercase by default). 

This makes the search case insensitive and adds filtering also on ids

## Resolved issues

Resolves: https://github.com/canonical/checkbox/issues/700

## Documentation

N/A

## Tests

N/A


kissiel (approved):
big QoL improvement, thanks
+1

Issue #986, opened 2024-02-13
beacon eddystone test crashes with a traceback


kissiel:
### Bug Description

On some systems, randomly (but not infrequently) we get the following traceback when running the eddystone test:
```
Traceback (most recent call last):
  File "/snap/checkbox-iiotg/46/checkbox-runtime/bin/checkbox-support-eddystone_scanner", line 8, in <module>
    sys.exit(main())
  File "/snap/checkbox-iiotg/46/checkbox-runtime/lib/python3.10/site-packages/checkbox_support/scripts/eddystone_scanner.py", line 70, in main
    init_bluetooth()
  File "/snap/checkbox-iiotg/46/checkbox-runtime/lib/python3.10/site-packages/checkbox_support/scripts/eddystone_scanner.py", line 35, in init_bluetooth
    btctl.writeline('scan on')
  File "/snap/checkbox-iiotg/46/checkbox-runtime/lib/python3.10/site-packages/checkbox_support/interactive_cmd.py", line 84, in writeline
    self._proc.stdin.flush()
BrokenPipeError: [Errno 32] Broken pipe
```

### To Reproduce

`checkbox-cli run com.canonical.certification::bluetooth4/beacon_eddystone_url_hci0.com.canonical.certification::bluetooth4/beacon_eddystone_url_hci0`

### Environment

checkbox snap, uc22 

### Relevant log output

```shell
Traceback (most recent call last):
  File "/snap/checkbox-iiotg/46/checkbox-runtime/bin/checkbox-support-eddystone_scanner", line 8, in <module>
    sys.exit(main())
  File "/snap/checkbox-iiotg/46/checkbox-runtime/lib/python3.10/site-packages/checkbox_support/scripts/eddystone_scanner.py", line 70, in main
    init_bluetooth()
  File "/snap/checkbox-iiotg/46/checkbox-runtime/lib/python3.10/site-packages/checkbox_support/scripts/eddystone_scanner.py", line 35, in init_bluetooth
    btctl.writeline('scan on')
  File "/snap/checkbox-iiotg/46/checkbox-runtime/lib/python3.10/site-packages/checkbox_support/interactive_cmd.py", line 84, in writeline
    self._proc.stdin.flush()
BrokenPipeError: [Errno 32] Broken pipe
```
```


### Additional context

_No response_

yphus:
is it the same issue I reported in lp in 2019? https://bugs.launchpad.net/checkbox-support/+bug/1811675

[...]

PR #1006, opened 2024-02-21
Be verbose when breaking pipes when interacting with bluetoothctl (BugFix)

kissiel:
## Description

We've had a few examples of BT beacon test failures that finished with just an information about broken pipe, with no context. This made it hard to reason about the real cause of failure.

This patch dumps the stdout of the called process (`bluetoothctl`) when the pipe breaks.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/986
Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1249

## Tests

Tested by emulating the called process crashing and by executing bluetoothctl without root privileges. In all cases the warning with more context was attached.


Hook25 (approved):
+1 good idea

Hook25 (dismissed):

PR #839, opened 2023-11-24
Resume screen for checkbox remote (New)

Hook25:
## Description

This PR adds the support for the new urwid resume screen for checkbox remote as well.  

## Resolved issues

Resolves CHECKBOX-937
Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-1168

## Documentation

N/A

## Tests

Change is fully covered with unittests and introduces multiple scenarios (also ran and working for local)

PR #1007, opened 2024-02-21
Resume Screen for Checkbox Remote expose API (Breaking)

Hook25:
## Description

[This PR ](https://github.com/canonical/checkbox/pull/839) modified the remote API but didn't bump the REMOTE_API number.  The api bump is warranted because we need every new api entry:
- delete_session : used by the menu to delete old sessions
- get_resumable_sessions: same, used by the menu to list/manage sessions and get the resumable id
- new kwarg  overwrite_result_dict={}: used to overwrite the result of an older session via the menu, wasn't there before

Additionally (this to do only 1 remote api bump) many parts of the code use private apis of the RemoteAssistant. This is a problem as they are not stable. This also fixes this issue
## Resolved issues

Currently checkbox crashes if launched across revisions because the version wasn't bumped

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -445,15 +445,15 @@ class RemoteController(ReportsStage, MainLoopStage):
         return False
 
     def _resume_session(self, resume_params):
-        metadata = self.sa._sa.resume_session(resume_params.session_id)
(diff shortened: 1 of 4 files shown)


kissiel (approved):
lgtm, +1

kissiel (approved):
+1

PR #1008, opened 2024-02-22
Add tests for resume menu (infra)

Hook25:
## Description

These are all the metabox scenarios for the new resume menu on remote. They were not added to the original resume_remote PR by mistake but given the chance I added a couple more. This should test every scenario the menu must support.

> Note: these are remote only tests because it seems that local is not resuming user-interact jobs correctly (possibly missing the lastest_run_job_name metadata like remote was)  

Driveby fix: this also removes all random newlines in licenses because it seems that every time I init a new metabox scenario I wrongly copy one that has it

## Resolved issues

Missing tests for resume remote

## Documentation

N/A

## Tests

This PR


kissiel (approved):
Goodies, +1

Issue #935, opened 2024-01-15
Auto resume is not working in `edge` channel


LiaoU3:
### Bug Description
In `stable` and `beta`, checkbox could automatically resume back after a reboot.
But in `edge`, checkbox could not automatcally resume back.

Comparison if could auto resume
| channel | remote | local |
|---------|--------|-------|
| stable (3.0.0) | O      | O     |
| beta  (3.3.0-dev19)  | O      | O    |
| edge  (3.3.0-dev93)  | X      | X     |

>Note: The test cases using `pm_test` like [power-management/poweroff](https://github.com/canonical/checkbox/blob/f31e500b5b83e7bd0875719dd9df5fe6e6c4464f/providers/base/units/power-management/jobs.pxu#L54) or [power-management/reboot](https://github.com/canonical/checkbox/blob/f31e500b5b83e7bd0875719dd9df5fe6e6c4464f/providers/base/units/power-management/jobs.pxu#L77) can not resume back in all channels. This is a know issue, so we are not talking about those cases with `pm_test` (By the way the test cases with `pm_test` is removed from testplans in PR #904). In here we only talk about other regular test cases using `reboot`,  `dbus` etc.

#### snap list
We installed `checkbox` and `checkbox22` with `edge` channel.
```
u@u-UPN-EHL01:~$ snap list 
Name                       Version           Rev    Tracking         Publisher            Notes
bare                       1.0               5      latest/stable    canonical✓           base
checkbox                   3.3.0-dev46       5398   latest/edge      ce-certification-qa  classic
checkbox22                 3.3.0-dev87       764    latest/edge      ce-certification-qa  -
core20                     20231123          2105   latest/stable    canonical✓           base
core22                     20231123          1033   latest/stable    canonical✓           base
firefox                    121.0.1-1         3626   latest/stable/…  mozilla✓             -
gnome-3-38-2004            0+git.efb213a     143    latest/stable/…  canonical✓           -
gnome-42-2204              0+git.ff35a85     141    latest/stable    canonical✓           -
gtk-common-themes          0.1-81-g442e511   1535   latest/stable/…  canonical✓           -
snap-store                 41.3-71-g709398e  959    latest/stable/…  canonical✓           -
snapd                      2.61.1            20671  latest/stable    canonical✓           snapd
snapd-desktop-integration  0.9               83     latest/stable/…  canonical✓           -
```
#### checkbox remote
[...]

PR #1016, opened 2024-02-26
Autoresume crashed/norerun sessions on interactive runs for remote (New)

Hook25:
## Description

Checkbox remote used to automatically resume the latest session when it crashed or the machine turned off. This was obtained via an obscure code-path that has since been simplified but the behaviour was useful to QA testing. 

The idea was that a session would be automatically resumed if it was non-interactive, else the user would have to manually trigger the resume. QA's need for autoresume is not adressed by the new resume menu as test session that contain multiple reboots, for example, would be way easier to carry out if checkbox just automatically resumed them.

This introduces a new autoresume behaviour for interactive runs. If the latest resumable session was running a shell job when it was closed unexpectedly, it will be resumed automatically. The status of the latest job follows the usual logic (in order of priority):
1. read from disk when possible
2. pass if the job was noreturn
3. crash for everything else

> Note: the test is remote only as local doesn't seem to autoresume interactive sessions yet. If we want it to do so we can follow this up in the future

> Note2: A possible follow up to this is to give the user some time to interrupt the resume process via the Ctrl+C menu similarly to what is done for automated runs

> Note3: This introduces a small hack in controller.py:490. The reason is that this needs a followup that fixes `get_resumable_sessions` to only return resumable sessions and not incompatible ones

## Resolved issues

Resolves: https://github.com/canonical/checkbox/issues/935

## Documentation

N/A

## Tests

This adds a new metabox scenario where both noreturn and crash are tested in manual interactive runs.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -393,7 +393,6 @@ class RemoteController(ReportsStage, MainLoopStage):
             self.sa.bootstrap()
 
             if not metadata.running_job_name:
-                self.sa.abandon_session()
                 return False
 
             job_state = self.sa.get_job_state(metadata.running_job_name)
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -355,12 +355,19 @@ class RemoteController(ReportsStage, MainLoopStage):
         return self._has_anything_failed
 
     def should_start_via_launcher(self):
+        """
+        Determines if the controller should automatically select a test plan
+        given a launcher. Raises if the launcher tries to skip the test plan
+        selection without providing the test plan that must be automatically
+        selected
+        """
         if self.launcher.get_value("test plan", "forced"):
             tp_unit = self.launcher.get_value("test plan", "unit")
             if not tp_unit:
                 _logger.error(
                     _(
-                        "The test plan selection was forced but no unit was provided"
+                        "The test plan selection was forced but no unit"
+                        " was provided"
                     )
                 )
                 raise SystemExit(1)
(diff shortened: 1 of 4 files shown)


checkbox-ng/checkbox_ng/launcher/controller.py
```diff
@@ -354,7 +355,90 @@ def quitter(msg):
...
+            if not tp_unit:
+                _logger.error(
+                    _(
+                        "The test plan selection was forced but no unit"
+                        " was provided"
+                    )
+                )
+                raise SystemExit(1)
+            return True
+        return False
```
kissiel, line 376:
```suggestion
        tp_forced = self.launcher.get_value("test plan", "forced")
        chosen_tp = self.launcher.get_value("test plan", "unit")
        if tp_forced and not chosen_tp:
            raise SystemExit("The test plan selection was forced but no unit was provided") # split me into lines
        return tp_forced
```

checkbox-ng/checkbox_ng/launcher/controller.py
```diff
@@ -354,7 +355,90 @@ def quitter(msg):
                 break
         return self._has_anything_failed
 
-    def resume_or_start_new_session(self):
+    def should_start_via_launcher(self):
+        """
+        Determines if the controller should automatically select a test plan
+        given a launcher. Raises if the launcher tries to skip the test plan
```
kissiel, line 361:
```suggestion
        if given a launcher. Raises if the launcher tries to skip the test plan
```

checkbox-ng/checkbox_ng/launcher/controller.py
```diff
@@ -354,7 +355,90 @@ def quitter(msg):
...
+                    )
+                )
+                raise SystemExit(1)
+            return True
+        return False
+
+    @contextlib.contextmanager
+    def _resumed_session(self, session_id):
+        """
+        Used to temporarely resume a session to inspect it, abandoning it
```
kissiel, line 381:
```suggestion
        Used to temporarily resume a session to inspect it, abandoning it
```

checkbox-ng/checkbox_ng/launcher/controller.py
```diff
@@ -354,7 +355,90 @@ def quitter(msg):
...
+            self.sa.select_test_plan(app_blob["testplan_id"])
+            self.sa.bootstrap()
+
+            if not metadata.running_job_name:
+                return False
+
+            job_state = self.sa.get_job_state(metadata.running_job_name)
+            if job_state.job.plugin != "shell":
+                return False
+            return True
```
kissiel, line 427: I'm wondering if it would be better to have this set of criteria could be outlined into a more testable and self contained function.

kissiel (changes requested):
A few small fixes proposed here and there, one bigger patch, and one discussion starter :)

checkbox-ng/checkbox_ng/launcher/controller.py
Hook25: I have written this in this format to avoid needless work on non-autoresumable sessions. Given how checkbox work we can not (to the best of my knowledge) inspect the session the way that we need without moving the state of the session via `select_tp`/`bootstrap`. Possibly the last part could be factored out given that I expect more constraints to be put in place on the job state potentially, but as long as it is only one, I would keep this here. I will go back on this either way to port this fix (potentially) to local

kissiel (approved):
lgtm, +1

PR #1020, opened 2024-02-27
Added PR job to test GH actions + testflinger (New)

fernando79513:
## Description

This PR is a draft. We will use it to test the GH actions. It is not intended to be merged but just use it as a test PR to try the connection of GH actions with testflinger


mz2 (changes requested):
Please create a composite action for TF, as requested.

PR #1025, opened 2024-02-28
Fix resuming local manual jobs (bugfix)

Hook25:
## Description

When Checkbox local quits during a manual job, it "forgets" that the job was acutally running, so when restoring the session every action in the resume menu actually acts on the previous job. This is due to the fact that manual jobs do not note the metadata when starting (like automatic jobs do). 

This fixes this behaviour always noting the metadata

Minor: this also adds a pretty text when exiting a session like in remote and changes the return code when doing that.  This should only be positive, as anything catching that return code should know that the session didnt complete
Minor: this also makes the tests that were previously written for remote slightly less lazy by actually checking the status

## Resolved issues

Resolves: [CHECKBOX-1268](https://warthogs.atlassian.net/browse/CHECKBOX-1268), [CHECKBOX-1115](https://warthogs.atlassian.net/browse/CHECKBOX-1115) [CHECKBOX-564](https://warthogs.atlassian.net/browse/CHECKBOX-564) (Title is a bit misleading, from the description this is a duplicate of 1115)

## Documentation

N/A

## Tests

Metabox tests for remote enabled also for local

[CHECKBOX-1268]: https://warthogs.atlassian.net/browse/CHECKBOX-1268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


checkbox-ng/checkbox_ng/launcher/stages.py
```diff
@@ -180,7 +181,7 @@ def _run_single_job_with_ui_loop(self, job, ui):
                                 result_builder.comments = comments
                             break
                     elif cmd == "quit":
-                        raise SystemExit()
+                        raise SystemExit("Session saved")
```
kissiel, line 184:
```suggestion
                        raise SystemExit("Session paused by the operator")
```

metabox/metabox/scenarios/ui/resume_menu.py
```diff
@@ -76,13 +75,12 @@ class ResumeMenuMultipleDelete(Scenario):
         # Now we still have to be able to run test plans
         SelectTestPlan("2021.com.canonical.certification::whoami_as_user_tp "),
         Send(keys.KEY_ENTER),
-        Expect("Results"),
+        Expect(_re("(☑|job passed).*Print who is running the job")),
```
kissiel, line 78: It's testing. We are in control what's printed. 
`_re` is cheating :P

kissiel (changes requested):
`_re` is cheating :P

checkbox-ng/checkbox_ng/launcher/stages.py
Hook25: I wanted to make this uniform across local and remote, do you want me to replace the remote one as well?

kissiel: Yes. And maybe let's not use the word "operator", as it may be confusing to some (too close to "controller"). Let's use "user". I want to put emphasis on the fact that a human did it :)

Hook25: done

kissiel (approved):
solid stuff; +1

PR #1032, opened 2024-03-01
Use pylxd mount disk (infra)

Hook25:
Description
This was introduced a while back as it was unclear to me how to actually mount a disk via pylxd. This PR mounts the disk via the pylxd api instead of using the lxc subprocess

Resolved issues
N/A

Documentation
N/A

Tests
Ran metabox, it works as before.

> Note: recreated as GH was being weird


kissiel (approved):
Let's see :)

PR #1037, opened 2024-03-05
Ensure that jobs that have an after suspend counterpart are run before suspend (New)

pieqq:
## Description

Some jobs should be run before *and* after suspend. To take this into
account, [the `siblings` field](https://checkbox.readthedocs.io/en/stable/reference/units/job.html#job-siblings-field) was developed, along with the
[`also-after-suspend` and `also-after-suspend-manual` flags](https://checkbox.readthedocs.io/en/stable/reference/units/job.html#also-after-suspend-flag).

These would let Checkbox spawn a similar job with an added dependency on
the suspend job (either the manual or the automated version of it).

The problem is that the original job did not have any dependency to
force it to be run *before* the suspend job.

This was not an issue for test plans organized manually, using regular
expressions, because you could have an include section that looks like:

    storage_.*
    com.canonical.certification::suspend/suspend_advanced_auto
    after-suspend-storage.*

However, now that template ids can be added in [test plans](https://checkbox.readthedocs.io/en/latest/reference/units/test-plan.html), this is a
problem.

This patch will make sure the jobs that need to run before suspend are
added as dependencies of their related suspend job, so that regardless
of the order in the test plan, they will be run in the proper order.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1265

## Documentation

The Job Unit reference page has been updated to explain what happens when the `also-after-suspend` flag is used, or when a `sibling` is defined that depends on a suspend job.

## Tests

Given the following units:

```
id: demoresource
plugin: resource
command:
    echo 'name: sda'
    echo 'path: /dev/sda'
    echo
    echo 'name: sdb'
    echo 'path: /dev/sdb'

unit: template
template-resource: demoresource
template-id: demo-template
id: demo-test-storage-{name}
plugin: shell
command: echo "This is the path: {path}"
flags: also-after-suspend

unit: test plan
id: demo-test-plan
bootstrap_include:
  demoresource
include:
  demo-template     # template id generated from job id
```

Before applying the patch:

```
$ checkbox-cli list-bootstrapped com.canonical.certification::demo-test-plan

com.canonical.certification::demo-test-storage-sda
com.canonical.certification::sleep
com.canonical.certification::rtc
com.canonical.certification::suspend/suspend_advanced_auto
com.canonical.certification::after-suspend-demo-test-storage-sda
com.canonical.certification::demo-test-storage-sdb
com.canonical.certification::after-suspend-demo-test-storage-sdb
```

The `demo-test-storage-sdb` job should be run **before** `suspend/suspend_advanced_auto`, but is run after.

After applying the patch:

```
$ checkbox-cli list-bootstrapped com.canonical.certification::demo-test-plan

com.canonical.certification::demo-test-storage-sda
com.canonical.certification::rtc
com.canonical.certification::sleep
com.canonical.certification::demo-test-storage-sdb
com.canonical.certification::suspend/suspend_advanced_auto
com.canonical.certification::after-suspend-demo-test-storage-sda
com.canonical.certification::after-suspend-demo-test-storage-sdb
```

I also ran the test plan to make sure it executed properly and in the right order.

### DIFF
--- a/checkbox-ng/plainbox/impl/ctrl.py
+++ b/checkbox-ng/plainbox/impl/ctrl.py
@@ -53,6 +53,7 @@ from subprocess import check_output, CalledProcessError, STDOUT
 
 from plainbox.abc import IJobResult
 from plainbox.abc import ISessionStateController
+from plainbox.configuration import Suspend
 from plainbox.i18n import gettext as _
 from plainbox.impl import get_plainbox_dir
 from plainbox.impl.depmgr import DependencyDuplicateError
@@ -102,12 +103,14 @@ class CheckBoxSessionStateController(ISessionStateController):
           of resource definitions.
     """
 
-    def get_dependency_set(self, job):
+    def get_dependency_set(self, job, job_list=None):
         """
         Get the set of direct dependencies of a particular job.
 
         :param job:
             A IJobDefinition instance that is to be visited
+        :param job_list:
+            List of jobs to check dependencies from
         :returns:
             set of pairs (dep_type, job_id)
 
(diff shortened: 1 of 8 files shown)


checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -125,15 +128,64 @@ def get_dependency_set(self, job):
...
+            return suspend_deps
+        for dep_job in job_list:
+            if dep_job.flags and expected_flag in dep_job.flags:
+                suspend_deps.add(dep_job.id)
+            if dep_job.siblings:
+                for sibling_data in json.loads(dep_job.tr_siblings()):
+                    if (
+                        sibling_data.get("depends")
+                        and suspend_job_id in sibling_data["depends"]
+                    ):
```
kissiel, line 185:
```suggestion
                    if suspend_job_id in sibling_data.get("depends", []):
```

checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -125,15 +128,64 @@ def get_dependency_set(self, job):
...
+        """
+        suspend_deps = set()
+        expected_flag = ""
+        if suspend_job_id == Suspend.AUTO_JOB_ID:
+            expected_flag = Suspend.AUTO_FLAG
+        elif suspend_job_id == Suspend.MANUAL_JOB_ID:
+            expected_flag = Suspend.MANUAL_FLAG
+        if not expected_flag:
+            return suspend_deps
+        for dep_job in job_list:
```
kissiel, line 177: If you create a function that takes one job (as in one item in the dep_job collection), and do same check in that function, you can then run return set(filter(the_predicate_fn, job_list)).
Testability and test clarity being the big benefit with that.

checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -236,8 +288,63 @@ def get_inhibitor_list(self, session_state, job):
...
                 )
                 inhibitors.append(inhibitor)
+        if job.id in [Suspend.AUTO_JOB_ID, Suspend.MANUAL_JOB_ID]:
+            for inhibitor in self._get_suspend_inhibitor_list(
+                session_state, job
+            ):
+                inhibitors.append(inhibitor)
         return inhibitors
 
+    def _get_suspend_inhibitor_list(self, session_state, suspend_job):
```
kissiel, line 298: The code in this function looks very similar to the one above it. Same things apply + outlining of predicates will help emphasize the differences between functions.

kissiel (changes requested):
Great stuff; landable as is. 
Some tips below, tho - it can be made less complicated (especially tests, which have to cover a lot of branching). If you think it's a lot of work, ping me, but I'm kindly asking for a few tweaks.

checkbox-ng/plainbox/impl/ctrl.py
kissiel: I meant similar to `_get_before_suspend_dependency_set`

checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -125,15 +128,64 @@ def get_dependency_set(self, job):
...
+            List of jobs to search dependencies on.
+        :returns:
+            A set of job ids that need to be run before the suspend job
+        """
+        suspend_deps = set()
+        expected_flag = ""
+        if suspend_job_id == Suspend.AUTO_JOB_ID:
+            expected_flag = Suspend.AUTO_FLAG
+        elif suspend_job_id == Suspend.MANUAL_JOB_ID:
+            expected_flag = Suspend.MANUAL_FLAG
```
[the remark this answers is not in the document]
pieqq: Done

checkbox-ng/plainbox/impl/ctrl.py
pieqq: This twisted my brain for a few hours today, but I think I managed to get something to work along these lines. Thanks for the feedback, TIL!

kissiel (approved):
Great stuff, +1

PR #1038, opened 2024-03-06
[checkbox-ce-oem] Modify dbus warm cold boot job (BugFix)

rickwu666666:
- Modify the command called in the job
  Since we change the command name in the test-strict-confinement snap.
- Fix the wrong job name in test-plan
- Change job structure
  Make the job easier to read

## Description
We modify the command in the test-strict-confinement snap. Therefore, we modify the command that is called in the job.
Also, fix the wrong job ID that is included in the test plan.

## Resolved issues
N/A

## Documentation
N/A

## Tests
Warm-boot 
https://certification.canonical.com/hardware/202304-31485/submission/359313/

Cold-boot
https://certification.canonical.com/hardware/202304-31485/submission/359311/


contrib/checkbox-provider-ce-oem/units/strict-confinement/powermanagement-strict-confinement.pxu
```diff
@@ -18,7 +14,10 @@ imports:
 requires:
   lsb.distributor_id == 'Ubuntu Core'
   snap.name == 'test-strict-confinement'
-depends: com.canonical.certification::init-boot-loop-data
+command:
+  set -e
+  rtcwake -d "${RTC_DEVICE_FILE:-rtc0}" -v -m on -s "${STRESS_BOOT_WAKEUP_DELAY:-120}" &
```
kissiel, line 19: This line can lead to just shutting down the whole machine.
The ampersand at the end doesn't guarantee that the spawned process will be scheduled on the CPU before the next line runs, rendering this a race condition.

kissiel (changes requested):
Race condition with the background process scheduling found.

contrib/checkbox-provider-ce-oem/units/strict-confinement/powermanagement-strict-confinement.pxu
rickwu666666: Hi @kissiel,
This command using argument `-m on` means it will keep the system on and read the RTC device until an alarm time appears. It won't shut down the system. However, I guess I should use `-m no` instead.

https://man7.org/linux/man-pages/man8/rtcwake.8.html

LiaoU3 (dismissed):
+1

baconYao:
> Race condition with the background process scheduling found.

@rickwu666666 , I think you can reference this PR
https://github.com/canonical/checkbox/pull/979

rickwu666666:
I've pushed a fix for the race condition.

stanley31huang (changes requested):
In test plan, we are going to compare the logs between the reboot test, so we need to add a  job to collect logs before reboot test.

rickwu666666:
@stanley31huang , Yes you are correct. I've pushed the fix. Thanks for the review.

contrib/checkbox-provider-ce-oem/units/strict-confinement/powermanagement-strict-confinement.pxu
```diff
@@ -42,18 +41,18 @@ _summary: Perform warm reboot via dbus
...
+depends: com.canonical.certification::init-boot-loop-data
 imports:
   from com.canonical.certification import snap
   from com.canonical.certification import lsb
 requires:
   lsb.distributor_id == 'Ubuntu Core'
   snap.name == 'test-strict-confinement'
-depends: com.canonical.certification::init-boot-loop-data
+command:
+  test-strict-confinement.warm-boot
```
kissiel, line 55: Why is going through additional snap needed here?
As far as I understand it, this just reboots the machine, why introduce extra steps, that involve more ways the job can go wrong (calling a snap from a snap is problematic, also debugging becomes more difficult).

contrib/checkbox-provider-ce-oem/units/strict-confinement/powermanagement-strict-confinement.pxu
```diff
@@ -5,20 +5,19 @@ _description: This test will check your system shutdown/booting cycle via dbus c
...
 imports:
   from com.canonical.certification import snap
   from com.canonical.certification import lsb
 requires:
   lsb.distributor_id == 'Ubuntu Core'
   snap.name == 'test-strict-confinement'
-depends: com.canonical.certification::init-boot-loop-data
+command:
+  set -e
+  rtcwake -d "${RTC_DEVICE_FILE:-rtc0}" -v -m no -s "${STRESS_BOOT_WAKEUP_DELAY:-120}" && test-strict-confinement.cold-boot || exit 1
```
kissiel, line 20: There are two problems with this line:

- This command creates a race against checkbox. As soon as the dbus call is made to Poweroff, Checkbox proceeds to the next job (and writes progress to the disk, which leads to other IO problems, btw.), and if Checkbox starts the next job before the actual poweroff happens, the next job will be interrupted by the system. 

`|| exit 1` is redundant, if the first command failed then the error code will be preserved (also if the second one fails).

kissiel (changes requested):
Going through an extra snap seems unnecessary. It:
* obfuscates what is being done
* reduces clarity and makes the whole thing less readable and debuggable
* creates unnecessary dependencies

Also there's a serious race condition with Checkbox being created here that may lead to breakage of **other** jobs that may follow the one modified here.

contrib/checkbox-provider-ce-oem/units/strict-confinement/powermanagement-strict-confinement.pxu
rickwu666666: Hi @kissiel,
Introducing extra snap is because we want to test those commands in a strict confinement environment. This requirement is from the project [Shiner](https://warthogs.atlassian.net/browse/OEMQA-2989) and it is due to a [snap issue](https://bugs.launchpad.net/shiner/+bug/2034257). The Shiner team requires us to run some test cases that they are concerned about in a strict confinement environment, and that means a snap that is not installed in devmode and also connected necessary interfaces.
And this is kind of a POC and also a short-term solution for the shiner team.

contrib/checkbox-provider-ce-oem/units/strict-confinement/powermanagement-strict-confinement.pxu
rickwu666666: Hi @kissiel,
Does this race condition happen because the dbus command is called by another SNAP which is outside of the checkbox instead of calling dbus itself? If so, do you have any recommendations for ways to avoid this kind of issue?

LiaoU3 (approved):
+1

rickwu666666:
@stanley31huang,
I've pushed the fix based on our discussed yesterday.

LiaoU3 (approved):
LGTM

stanley31huang:
I think we should have some extra steps to reboot the system in case we failed to power control the system via `test-strict-confinement` snap.
Like what we did in Baoshan project https://git.launchpad.net/~baoshan-team/baoshan/+git/checkbox-provider-baoshan/tree/units/thermal/jobs.pxu

rickwu666666:
https://certification.canonical.com/hardware/202403-33886/submission/386662/test-results/fail/
Test result of if reboot via test-strict-confinement snap failed.

stanley31huang (approved):
LGTM

PR #1039, opened 2024-03-06
Fix dkms build validation in mantic

fernando79513:
## Description

The previous bash implementation contained a bug that was only fixed for ubuntu 22.04. I am not sure why this issue was not caught in lunar lobster, but it was appearing in mantic.

The test has been fixed to work with the new syntax of dkms_status

It has been refactored in python and includes tests, so it would be easier to change it again in the future.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1252

## Documentation

There are no changes on the documentation

## Tests

To run the dkms validation tests:
```
cd ~/checkbox/providers/sru
python3 -m pytest -k TestDKMSValidation
```


providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+            command,
+            stderr=subprocess.STDOUT,
+            shell=True,
+            universal_newlines=True,
+        )
+        return result.strip()
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
+            "Command '{0}' failed with error:\n{1}".format(command, e.output)
+        )
```
kissiel, line 45: This is not getting caught anywhere, so it'll cause a backtrace when this, predicted, problem happens. So I'd recommend raising SystemExit with a nice info.

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+    return kernel_info
+
+
+def parse_version(ver: str) -> version.Version:
+    """Parse the version string and return a version object"""
+    match = re.match(r"(\d+\.\d+\.\d+(-\d+)?)", ver)
+    if match:
+        parsed_version = version.parse(match.group(1))
+    else:
+        raise ValueError("Invalid version string: {0}".format(ver))
```
kissiel, line 67: Same applies here with raising of SystemExit

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+        )
+        logger.warning(msg)
+        logger.warning("=== DKMS status ===\n{0}".format(dkms_status))
+
+    # Scan the APT log for errors during system update
+    if has_dkms_build_errors(kernel_ver_current):
+        return 1
+
+    return 0
```
kissiel, line 148:
```suggestion
    return has_dkms_build_errors(kernel_ver_current)
```

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+        return result.strip()
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
+            "Command '{0}' failed with error:\n{1}".format(command, e.output)
+        )
+
+
+def parse_dkms_status(dkms_status: str, ubuntu_release: str) -> List[Dict]:
+    """Parse the output of 'dkms status' and return a list of dictionaries"""
```
kissiel, line 49: "return a list of dictionaries" introduces no info, all was said with the typehint.
More useful information would be what it returns, and not of what type (and even the type is incomplete).

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+    )
+    with open(log_path, "r") as f:
+        for line in f.readlines():
+            if err_msg in line:
+                logger.error(
+                    "Found dkms build error messages in {}".format(log_path)
+                )
+                logger.error("\n=== build log ===")
+                result = run_command(
+                    "grep '{}' {} -C 5".format(err_msg, log_path)
```
kissiel, line 84: why do we need to run grep here?

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+                )
+                logger.error(result)
+                return True
+    return False
+
+
+def main():
+    # Get the kernel version and DKMS status
+    ubuntu_release = run_command("lsb_release -r | cut -d ':' -f 2 | xargs")
```
kissiel, line 93: do we need to use multiple pipes to effectively run a simple string operation?

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+logger = logging.getLogger("dkms_build_validation")
+
+
+def run_command(command: str) -> str:
+    """Run a shell command and return its output"""
+    try:
+        result = subprocess.check_output(
+            command,
+            stderr=subprocess.STDOUT,
+            shell=True,
```
kissiel, line 38: after reading all of the program, the infamous, discouraged `shell=True` is not needed.

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+            universal_newlines=True,
+        )
+        return result.strip()
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
+            "Command '{0}' failed with error:\n{1}".format(command, e.output)
```
kissiel, line 44: `e.output` because the out and err pipes are mixed is not "error".

providers/sru/bin/dkms_build_validation.py
```diff
@@ -0,0 +1,152 @@
...
+                logger.error("\n=== build log ===")
+                result = run_command(
+                    "grep '{}' {} -C 5".format(err_msg, log_path)
+                )
+                logger.error(result)
+                return True
+    return False
+
+
+def main():
```
kissiel, line 91: decomposition of all the branches that happen in main would lead to a much cleaner set of tests

kissiel (changes requested):
No correctness problems found, but a few design problems. See below.

providers/sru/bin/dkms_build_validation.py
fernando79513: Tbh, I just left the grep there so I don't have to implement a functionality to get the context around each line. I will remove the grep.

fernando79513: It should be possible to handle the context lines (the lines arround the error match). I saw that there is a grep-like implementation here that handles the context, but I don't think it is worth the effort to implement it here just for this small test. Maybe it is better to have the "/var/log/apt/term.log" as an attatchment.
If we think it is required, I could try to add it as a separate function.

fernando79513 (commented):
I have solved most of the issues here. I will handle the decomposition in another commit.

providers/sru/bin/dkms_build_validation.py
fernando79513: Done!

kissiel (approved):
Great work! +1
Thank you for all the last-straight tweaks!

PR #1040, opened 2024-03-06
Allow listing instances from resume state (bugfix)

Hook25:
## Description

It in unclear why get_resumable_sessions is called after resume_session currently but given that the cache that get_resumable builds is re-invalidated/built if resume_session doesn't find the session is looking for in it after it is called, then there is no danger on this new edge.

This adds the possibility of listing resumable sessions after the resume_session function was called. 

## Resolved issues

Solves: https://github.com/canonical/checkbox/issues/935#issuecomment-1980005512

(although it is not clear how this was triggered at all, I am unable to recreate this)

## Documentation

N/A

## Tests

N/A


kissiel (approved):
+1

PR #1042, opened 2024-03-06
Remote resistant no launcher (Bugfix)

Hook25:
## Description

When trying to resume a session that was started without any launcher Checkbox remote will crash complaining about the missing launcher. This makes the agent use an empty configuration (as it should do) when the launcher is not provided by the controller

## Resolved issues

N/A

## Documentation

N/A

## Tests

Modified a test so that the launcher is not provided but the flow continues as expected


kissiel (approved):
LGTM, +1

PR #1043, opened 2024-03-06
Remove snappy-revert sub-tp from sru (BugFix)

kissiel:
## Description

This patch removes the snappy-revert sub test plan from the SRU test plan. At the current state of things it's main suspect for causing instability of other tests (due to lack of reboots).

The followup story is [CHECKBOX-1292](https://warthogs.atlassian.net/browse/CHECKBOX-1292)


pieqq (approved):
+1, let's fix this before landing it back!

Issue #1056, opened 2024-03-12
False positives on `suspend/network_after_suspend_auto`


kissiel:
### Bug Description

The aforementioned job fails with the enigmatic output, see below.

### To Reproduce

Running virtually any TP that has the problematic job and its base sibling included reproduces it.

### Environment

https://test-observer.canonical.com/#/debs/26330

### Relevant log output

```shell
connected
WARNING: systemd-networkd is not running, output will be incomplete.

Failed to query link bit rates: Unit dbus-org.freedesktop.network1.service not found.
Failed to query link DHCP leases: Unit dbus-org.freedesktop.network1.service not found.
2,3c2,3
< 64 bytes from 10.102.160.1: icmp_seq=1 ttl=64 time=0.337 ms
< 64 bytes from 10.102.160.1: icmp_seq=2 ttl=64 time=0.177 ms
---
> 64 bytes from 10.102.160.1: icmp_seq=1 ttl=64 time=0.174 ms
> 64 bytes from 10.102.160.1: icmp_seq=2 ttl=64 time=0.155 ms
6,7c6,7
< 2 packets transmitted, 2 received, 0% packet loss, time 1005ms
< rtt min/avg/max/mdev = 0.177/0.257/0.337/0.080 ms
---
> 2 packets transmitted, 2 received, 0% packet loss, time 1029ms
> rtt min/avg/max/mdev = 0.155/0.164/0.174/0.009 ms
11,12c11,12
< 64 bytes from 10.102.160.1: icmp_seq=1 ttl=64 time=0.130 ms
< 64 bytes from 10.102.160.1: icmp_seq=2 ttl=64 time=0.164 ms
---
> 64 bytes from 10.102.160.1: icmp_seq=1 ttl=64 time=0.136 ms
> 64 bytes from 10.102.160.1: icmp_seq=2 ttl=64 time=0.159 ms
16c16
< rtt min/avg/max/mdev = 0.130/0.147/0.164/0.017 ms
---
> rtt min/avg/max/mdev = 0.136/0.147/0.159/0.011 ms
```
```


### Additional context

_No response_

PR #1057, opened 2024-03-12
Don't diff results of the network test (BugFix)

kissiel:
## Description

This patch removes the reliance on diff for the after-suspend network test.

Previously it compared the output of the before and after suspend jobs diffing them. The job now is more reliable and produces precise information about the pings rendering the diffing useless.
This patch removes that.

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CHECKBOX-1304
Fixes: https://github.com/canonical/checkbox/issues/1056

## Tests

Tested with a stripped down test plan that included the jobs.


Hook25 (approved):
Makes sense, the test itself is gateway_ping_test so idk what the diff was trying to get here

PR #1059, opened 2024-03-13
Increase step timeout to 10h for snaps and 13h for debs build (infra)

Hook25:
## Description

This increases the maxium timeout of a job step from 6h to 10+10h for snaps and 13h for debs. This is hopefully enough.

Minor: this moves all of these jobs to small workers, we don't want/need to hog large ones for such inactive tasks

## Resolved issues

Failing daily builds due to timeouts (ty armhf)

## Documentation

N/A (comments to translate minutes into hours)

## Tests

N/A


kissiel (approved):
sad +1

PR #1061, opened 2024-03-13
Replace `create` with `up --no-start` in docker tests (bugfix)

Hook25:
## Description

`docker-compose create` is now fully deprecated. This replaces the call to an equivalent `up` call.

> Note: Here --no-start was used because removing --no-start and the subsequent start seems to hang the test 

## Resolved issues

Fixes: CHECKBOX-1254

## Documentation

N/A

## Tests

Tested locally via `checkbox-cli` and independent script (resolved the template and dumped it into a file)

```
#!/bin/bash
export -n PYTHONHOME PYTHONUSERBASE PYTHONPATH
set -e
#read content from stdin
yml=$(cat <<EOF
version: '2'
services:
  test:
    image: ubuntu
    command: bash
    tty: true
EOF
)
echo "$yml" | docker-compose -f - pull
echo "$yml" | docker-compose -f - up --no-start
echo "$yml" | docker-compose -f - start
ID=$(echo "$yml" | docker-compose -f - ps -q test)
echo $ID
[ `docker inspect -f {{.State.Status}} $ID` != running ] && exit 1
echo "$yml" | docker-compose -f - stop
[ `docker inspect -f {{.State.Status}} $ID` != exited ] && exit 1
echo "$yml" | docker-compose -f - up -d --scale test=3
docker ps | grep 'test' | grep -c 'Up'
ret=$(docker ps -a | grep 'test' | grep -c 'Up')
[ $ret != 3 ] && exit 1
echo "$yml" | docker-compose -f - up -d --scale test=1
ret=$(docker ps -a | grep 'test' | grep -c 'Up')
[ $ret != 1 ] && exit 1
echo "$yml" | docker-compose -f - down
docker ps -a | grep 'test' || exit 0
echo "The container is still available"
exit 1
```

### DIFF
--- a/providers/docker/units/docker.pxu
+++ b/providers/docker/units/docker.pxu
@@ -285,7 +285,7 @@ command:
  EOF
  )
  echo "$yml" | {compose_command} -f - pull
- echo "$yml" | {compose_command} -f - create
+ echo "$yml" | {compose_command} -f - up --no-start
  echo "$yml" | {compose_command} -f - start
  ID=$(echo "$yml" | {compose_command} -f - ps -q test)
  [ `docker inspect -f {status} $ID` != running ] && exit 1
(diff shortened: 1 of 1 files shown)


kissiel (approved):
TYVM! +1

PR #1067, opened 2024-03-14
Fix lp copy packages arbitrary lookback (infra)

Hook25:
## Description

The `lp_copy_packages.py` script failed to update our stable PPA because it assumed that any package that had to be copied over was less than 4 weeks old. 

This is false for two reasons:
1. A package has to be copied over regardless of how old it is if it is still the most recent version we have (namely, it took us way more than 1month to release stable this time around)
2. A package doesn't have to be copied over if it is `Superseeded` (namely, we have a more current version in the PPA of that same package)

> Note: there is a bit more insight on the filter I have left in the comment, but this is the top level view.
> Note1: the old comment stating that LP will choke on the request if no date is provided is only true if the iterator that the request returns is fully unwrapped. For this reason this PR only unwraps the relevant part.
## Resolved issues

Last stable release failed to update stable channel, see log: https://github.com/canonical/checkbox/actions/runs/7987143879/job/21809020771

(No copying over message -> no package copied over)

## Documentation

This PR adds documentation for what the function does and also adds a note about the more "counterintuitive" side of it (the fact that non-superseeded packages may still be non-copy-wrothy)

## Tests

This updates the test with the new two cases at hand

### DIFF
--- a/tools/release/lp_copy_packages.py
+++ b/tools/release/lp_copy_packages.py
@@ -28,6 +28,7 @@ Note: This script uses the LP_CREDENTIALS environment variable
 import sys
 import datetime
 import argparse
+import itertools
 
 from utils import get_launchpad_client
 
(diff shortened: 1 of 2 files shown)


fernando79513 (commented):
LGTM +1

fernando79513 (approved):
LGTM +1

Issue #1064, opened 2024-03-14
`stress/memory_stress_ng` will be shown as crash without any reason


hanhsuan:
### Bug Description

While using checkbox to test memory_stress_ng remotely, it will be shown as crash without any reason. The log in the [submission](https://certification.canonical.com/hardware/202312-33252/submission/360061/test/207367/result/39789642/) shows as below
```
Automatically passed after resuming execution
```

### To Reproduce

using checkbox remote to test stress/memory_stress_ng test case.

### Environment

- OS: Jammy
- checkbox version: 4.0.0.dev139
- checkbox type: debian

PR #1070, opened 2024-03-15
Clearer autoresume text (bugfix)

Hook25:
## Description

Currently the autoresume comment is not clear:
- For marking as passed, it doesn't state why it is doing so
- For marking as crashed, the text is inaccurate as it is using the same as passing

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/1064

## Documentation

This counts as documentation, it clarifies the behaviour with new text

## Tests

Also updated the text in the tests


kissiel (approved):
TYVM! +1

PR #1072, opened 2024-03-15
New timeout helper module (New)

Hook25:
## Description

This is a recurring need in many providers. We need a way to run a function for up to X seconds or bail out. Instead of re-implementing this functionality multiple times, lets create a helper function. 

## Resolved issues

This is partially needed for this: https://warthogs.atlassian.net/browse/CHECKBOX-1250

## Documentation

This has a new docstring that explain what it does

## Tests

This adds unittests for every situation


kissiel (approved):
This is a fantastic culmination of an awesome idea.
Big +1, can't wait to see it used in the wilds.

PR #1075, opened 2024-03-18
[Checkbox-ce-oem] Modify job and script to support use x-test tool(Bugfix)

rickwu666666:
## Description

Modify the job and script to support running the [x-test](https://github.com/canonical/x-test-snap) snap. The x-test snap is a snap that executes the OPTEE test. We intend to allow users to test OPTEE either via the xtest included in the gadget snap or the xtest included in the x-test snap.

## Resolved issues
N/A

## Documentation
N/A

## Tests
Sideload results with the xtest that is included in gadget snap:
https://pastebin.canonical.com/p/HPbYZtVyrs/

Sideload results with the xtest that is included in x-test snap:
https://certification.canonical.com/hardware/202401-33430/submission/360530/

Sideload results with no xtest found in the system:
https://certification.canonical.com/hardware/202401-33430/submission/360527/

### DIFF
--- a/contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
+++ b/contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
@@ -2,32 +2,69 @@
 
 import json
 import argparse
+import subprocess as sp
+import shlex
+import re
+from checkbox_support.snap_utils.snapd import Snapd
+from checkbox_support.snap_utils.system import get_gadget_snap
 
 
-def parse_json_file(filename, filter=False):
-    with open(filename, 'r') as file:
+def parse_json_file(filepath, filter=False, xtest=None):
+    with open(filepath, 'r') as file:
         data = json.load(file)
     for test in data:
         if (filter and test['suite'] == 'pkcs11') or \
            (not filter and test['suite'] not in ['pkcs11']):
-            print_test_info(test)
+            print_test_info(test, xtest)
 
 
-def print_test_info(test):
+def print_test_info(test, xtest):
     print("suite: {}".format(test['suite']))
     print("test_id: {}".format(test['test_id']))
-    print("description: {}\n".format(test['test_description']))
+    print("description: {}".format(test['test_description']))
+    print("tool: {}\n".format(xtest))
+
+
+def look_up_xtest():
+    xtest = look_up_gadget()
+    if xtest is not False:
+        pass
+    elif Snapd().list("x-test"):
+        xtest = "x-test.xtest"
+    else:
+        xtest = None
+    return xtest
+
+
+def look_up_gadget():
+    gadget = get_gadget_snap()
+    cmd = "snap info {}".format(gadget)
+    gadget_info = sp.run(shlex.split(cmd),
+                         stdout=sp.PIPE,
+                         stderr=sp.PIPE,
+                         text=True,
+                         timeout=10)
+    xtest = "{}.xtest".format(gadget)
+    if re.findall(xtest, gadget_info.stdout):
+        return xtest
+    else:
+        return False
 
 
 def main():
     parser = argparse.ArgumentParser(description='Parse a JSON file.')
-    parser.add_argument('filename', help='The name of the file to parse.')
+    parser.add_argument('-f', '--filepath',
+                        help='The path of the file to parse.')
     parser.add_argument('-p', '--pkcs11',
-                        help='To filter out PKCS11 for the suite'
+                        help='To filter out PKCS11 for the suite.'
                         'field in JSON.',
                         action='store_true')
     args = parser.parse_args()
-    parse_json_file(args.filename, args.pkcs11)
+    xtest = look_up_xtest()
+    if args.filepath is None:
+        print(xtest)
+    else:
+        parse_json_file(args.filepath, args.pkcs11, xtest)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
```diff
@@ -2,32 +2,69 @@
...
+
+def look_up_xtest():
+    xtest = look_up_gadget()
+    if xtest is not False:
+        pass
+    elif Snapd().list("x-test"):
+        xtest = "x-test.xtest"
+    else:
+        xtest = None
+    return xtest
```
baconYao, line 36: Adjust the order, if `x-test` snap exists, use it, else fallback to `gadget`. What do you think?

baconYao (commented):
See my inline comment

contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
rickwu666666: @baconYao I've updated the logic to prioritize the x-test snap. Thanks for the review.

contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
```diff
@@ -2,32 +2,68 @@
...
+    gadget_info = sp.run(shlex.split(cmd),
+                         stdout=sp.PIPE,
+                         stderr=sp.PIPE,
+                         text=True,
+                         timeout=10)
+    xtest = "{}.xtest".format(gadget)
+    if re.findall(xtest, gadget_info.stdout):
+        return xtest
+    else:
+        return False
```
stanley31huang, line 50:
```suggestion
    xtest = False
    snap = Snapd().list(gadget)
    if snap["apps"]:
        for app in snap["apps"]:
            if app["name"] == "xtest":
                xtest = ".".join([app["snap"], app["name"])
        xtest = ["{}.{}".format(name] for app in snap["apps"]]
    return xtest
```

rickwu666666: @stanley31huang I've pushed the fix. Thanks for the review.

contrib/checkbox-provider-ce-oem/units/optee/jobs.pxu
```diff
@@ -25,22 +25,13 @@ _summary: Check if xtest is in the gedget snap.
...
-            echo -e "\nInfo: Gadget snap support xtest command!"
-            echo -e "$xtest"
-            exit 0
-        else
-            echo -e "\nError: Gadget snap not support xtest command!"
-            echo -e "\nError: Please identify the tool name if needed!"
-            exit 1
-        fi
+    tool=$(parse_optee_test.py)
+    if [[ "$tool" != "None" ]]; then
```
fernando79513, line 30: This test is a bit weird. It only checks if the string "None" is printed, but if the test fails, since whatever is printed in stderr is not equal to "None", the test will pass and output `echo "Info: Found xtest runnable $tool"`

Ideally, I would prefer avoiding bash scripting as much as possible, and handle everything in a python script, but at least add here a way to catch the errors on the script:
```suggestion
    exit_status=$?
    if [[ "$exit_status" -eq 0 ]]; then
        echo "Info: Found xtest runnable $tool"
```

contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
```diff
@@ -2,32 +2,62 @@
...
+    parser.add_argument('-f', '--filepath',
+                        help='The path of the file to parse.')
     parser.add_argument('-p', '--pkcs11',
-                        help='To filter out PKCS11 for the suite'
+                        help='To filter out PKCS11 for the suite.'
                         'field in JSON.',
                         action='store_true')
     args = parser.parse_args()
-    parse_json_file(args.filename, args.pkcs11)
+    xtest = look_up_xtest()
```
fernando79513, line 56: I don't get the need to make the python script work for two different things.
I would rather have one script to parse the contents of the json file (`parse_optee_test.py`) and another one to look for the name of xtest (`look_up_xtest.py`)

fernando79513 (commented):
Good job fixing the bug!
I have some suggestions to the PR. 
The main one is handling errors on `parse_optee_test.py`. If I am not wrong, as it is now, if the script fails, it will not print "None" and the `ce-oem-optee/xtest-check` test will pas

contrib/checkbox-provider-ce-oem/bin/parse_optee_test.py
```diff
@@ -2,32 +2,62 @@
...
+    else:
+        xtest = None
+    return xtest
+
+
+def look_up_gadget():
+    gadget = get_gadget_snap()
+    xtest = False
+    snap = Snapd().list(gadget)
+    for app in snap["apps"]:
```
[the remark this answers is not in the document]
rickwu666666: @fernando79513 
`Snapd().list(gadget)` will return a dictionary if given a snap name. And I think this behavior is handled by snapd API.

rickwu666666:
@fernando79513 Thanks for the review. I've pushed the fix.

Sideload result with xtest is in the gadget snap:
https://pastebin.canonical.com/p/876jgc7fks/

The result with xtest is in the x-test snap:
https://pastebin.canonical.com/p/bpfcTHy6Jv/

The result with no xtest found in the system:
https://pastebin.canonical.com/p/C8CMGFw6gC/

fernando79513 (approved):
LGTM +1

PR #1079, opened 2024-03-18
Take inline overrides into account in mandatory_include and bootstrap_include sections of a test plan (BugFix)

pieqq:
## Description

Although inline overrides were taking into account in the `include` section of a test plan, it was not the case for the `mandatory_include` and `bootstrap_include` sections.

So a test plan like this:

```
unit: test plan
id: test-plan
name: My test plan
bootstrap_include:
    bootstrap-cert-blocker-job  certification_status=blocker
mandatory_include:
    mandatory-cert-blocker-job  certification_status=blocker
include:
    regular-cert-blocker-job    certification_status=blocker
```

would only make "regular-cert-blocker-job" a cert-blocker, not "mandatory-cert-blocker-job".

## Tests

Tested by running a test plan similar as the one described above and making sure the HTML report includes the job in `mandatory_include` and `bootstrap_include` sections as a cert-blocker.

### DIFF
--- a/checkbox-ng/plainbox/impl/unit/testplan.py
+++ b/checkbox-ng/plainbox/impl/unit/testplan.py
@@ -843,6 +843,8 @@ PatternMatcher('^job-[x-z]$'), inclusive=False)])
                     override_list.append((pattern, field_value_list))
 
             V().visit(IncludeStmtList.parse(testplan.include))
+            if testplan.mandatory_include:
+                V().visit(IncludeStmtList.parse(testplan.mandatory_include))
         for tp_unit in testplan.get_nested_part():
             override_list.extend(self._get_inline_overrides(tp_unit))
         return override_list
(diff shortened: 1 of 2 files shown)


checkbox-ng/plainbox/impl/unit/testplan.py
```diff
@@ -843,6 +843,8 @@ def visit_IncludeStmt_node(self, node: IncludeStmt):
                     override_list.append((pattern, field_value_list))
 
             V().visit(IncludeStmtList.parse(testplan.include))
+            if testplan.mandatory_include:
+                V().visit(IncludeStmtList.parse(testplan.mandatory_include))
```
Hook25, line 847: this is below `if testplan.include is not None`. Is this intended? Can a test have no include field and mandatory include? Also, what about `boostrap_include`?

Hook25 (commented):
Small comment below

Hook25 (approved):
+1, ty

PR #1089, opened 2024-03-20
Add fwupdmgr attachment job (New)

stanley31huang:
## Description
This is a follow up PR for the [PR965](https://github.com/canonical/checkbox/pull/965), I have applied the workaround (provided by Pierre). And I have verified the tests on EGW 3200.
https://certification.canonical.com/hardware/202312-33237/submission/360846/

This attachment job is requested from the OEM SWE team. They'd like to parse the `fwupdmgr` device information from the submission.
This PR added a `firmware/fwupdmgr_get_devices` attachment job and the related test plan to be nested in the client-cert-desktop test plan.

The `firmware/fwupdmgr_get_devices` only does the following command:
```
fwupdmgr get-devices --json
```

#### Requires
```
requires:
    executable.name == "fwupdmgr"
    environment.SNAP == ""  # only execute on Debian checkbox
```
The tests will only be executed when `fwupdmgr` executable is available.
Also, we found a limitation:

Our purpose is to run the `fwupdmgr` command from the default deb package to get the needed information. However, while Checkbox is running in the snap environment, the `fwupdmgr` will try to find the snap services `snap.fwupd.fwupd.service`, which leads to the following daemon and client mismatch error,
```
root@u-Inspiron-15-5508:/home/u# fwupdmgr get-devices
Mismatched daemon and client, use fwupd.fwupdmgr instead
```
This environment check is written in https://github.com/fwupd/fwupd/blob/main/src/fu-util-common.c#L42

``` C
#define SYSTEMD_FWUPD_UNIT	"fwupd.service"
#define SYSTEMD_SNAP_FWUPD_UNIT "snap.fwupd.fwupd.service"

fu_util_get_systemd_unit(void)
{
	if (g_getenv("SNAP") != NULL)
		return SYSTEMD_SNAP_FWUPD_UNIT;
	return SYSTEMD_FWUPD_UNIT;
}
```
Using the Debian version checkbox will not encounter this problem. Currently, I don’t have any good idea to solve this problem. So I intend to add a job for Debian Checkbox for now. And the `environment.SNAP == ""` is to ensure the job will only be executed with Debian Checkbox.

## Resolved issues
Request from SWE team
https://warthogs.atlassian.net/browse/OEMQA-3797

## Tests
Sideload result on an Ubuntu Desktop 22.04 laptop with Debian Checkbox
https://pastebin.canonical.com/p/4Jf5gXDKYV/
Sideload result on an Ubuntu Desktop 22.04 laptop with Snap Checkbox -> job skipped
https://pastebin.canonical.com/p/8wnCchBnWv/

### DIFF
--- a/providers/base/bin/get_firmware_info_fwupd.py
+++ b/providers/base/bin/get_firmware_info_fwupd.py
@@ -18,44 +18,91 @@
 # along with Checkbox. If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import sys
 import json
 import shlex
+import logging
 import subprocess
 from checkbox_support.snap_utils.snapd import Snapd
 
 
-def get_firmware_info_fwupd():
+def get_fwupdmgr_services_versions():
+    """Show fwupd client and daemon versions
 
-    fwupd_snap = Snapd().list("fwupd")
-    if fwupd_snap:
+    Returns:
+        list: fwupd client and daemon versions
+    """
+    fwupd_vers = subprocess.run(
+        shlex.split("fwupdmgr --version --json"),
+        capture_output=True)
+    fwupd_vers = json.loads(fwupd_vers.stdout).get("Versions", [])
+
+    return fwupd_vers
+
+
+def get_fwupd_runtime_version():
+    """Get fwupd runtime version
+
+    Returns:
+        tuple: fwupd runtime version
+    """
+    runtime_ver = ()
+
+    for ver in get_fwupdmgr_services_versions():
+        if (ver.get("Type") == "runtime" and
+                ver.get("AppstreamId") == "org.freedesktop.fwupd"):
+            runtime_ver = tuple(map(int, ver.get("Version").split(".")))
+
+    return runtime_ver
+
+
+def get_firmware_info_fwupd():
+    """
+    Dump firmware information for all devices detected by fwupd
+    """
+    if Snapd().list("fwupd"):
         # Dump firmware info by fwupd snap
         subprocess.run(shlex.split("fwupd.fwupdmgr get-devices --json"))
     else:
         # Dump firmware info by fwupd debian package
-        fwupd_vers = subprocess.run(
-            shlex.split("fwupdmgr --version --json"),
-            capture_output=True)
-        fwupd_vers = json.loads(fwupd_vers.stdout)
-
-        runtime_ver = ()
-        for ver in fwupd_vers.get("Versions", []):
-            if (ver.get("Type") == "runtime" and
-                    ver.get("AppstreamId") == "org.freedesktop.fwupd"):
-                runtime_ver = tuple(map(int, ver.get("Version").split(".")))
+        runtime_ver = get_fwupd_runtime_version()
         # Apply workaround to unset the SNAP for the fwupd issue
         # See details from following PR
         # https://github.com/canonical/checkbox/pull/1089
 
         # SNAP environ is avaialble, so it's running on checkbox snap
         # Unset the environ variable if debian fwupd lower than 1.9.14
-        if os.environ["SNAP"] and runtime_ver < (1, 9, 14):
+        if os.environ.get("SNAP") and runtime_ver < (1, 9, 14):
             del os.environ["SNAP"]
 
         subprocess.run(shlex.split("fwupdmgr get-devices --json"))
 
 
 if __name__ == "__main__":
+
+    root_logger = logging.getLogger()
+    root_logger.setLevel(logging.INFO)
+    logger_format = "%(asctime)s %(levelname)-8s %(message)s"
+    date_format = "%Y-%m-%d %H:%M:%S"
+
+    # Log DEBUG and INFO to stdout, others to stderr
+    stdout_handler = logging.StreamHandler(sys.stdout)
+    stdout_handler.setFormatter(logging.Formatter(logger_format, date_format))
+
+    stderr_handler = logging.StreamHandler(sys.stderr)
+    stderr_handler.setFormatter(logging.Formatter(logger_format, date_format))
+
+    stdout_handler.setLevel(logging.DEBUG)
+    stderr_handler.setLevel(logging.WARNING)
+
+    # Add a filter to the stdout handler to limit log records to
+    # INFO level and below
+    stdout_handler.addFilter(lambda record: record.levelno <= logging.INFO)
+
+    root_logger.addHandler(stderr_handler)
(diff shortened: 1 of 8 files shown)


providers/base/units/firmware/jobs.pxu
```diff
@@ -105,3 +105,15 @@ plugin: attachment
...
+plugin: attachment
+category_id: com.canonical.plainbox::firmware
+_summary: Collect the device firmware update information
+requires:
+    executable.name == "fwupdmgr"
+command:
+    snap_var="$SNAP"
+    unset SNAP
+    fwupdmgr get-devices --json
+    export SNAP="$snap_var"
```
superm1, line 119: I'm not sure the specific purpose of the test, but if it's to check end to end updates or supportability you probably should also be running `fwupdmgr refresh` as well to ensure the supported flag is accurately set.  Just make sure you catch the return code of 2 (nothing to do) in case it's up to date already.  something like:

```
fwupdmgr refresh --json || $? == 2
```

If it's used in pre-production there should be a step to add an embargo remote too if it doesn't exist.

stanley31huang: Mario,

the testing purpose of this job is to capture the all devices and current firmware version detected by fwupd. And I think the main reason to log these information, is we would like to compare the firmware version if any issues happens on a specific system.

kissiel, line 119: The saving and restoring of the $SNAP is not needed, as any changes to the environment is local to the  command being run.

providers/base/units/firmware/jobs.pxu
```diff
@@ -105,3 +105,15 @@ plugin: attachment
 depends: firmware/fwts_dump
 command:
   [ -f "$PLAINBOX_SESSION_SHARE/acpidump.log" ] && gzip -c "$PLAINBOX_SESSION_SHARE/acpidump.log"
+
+id: firmware/fwupdmgr_get_devices
+plugin: attachment
+category_id: com.canonical.plainbox::firmware
+_summary: Collect the device firmware update information
```
kissiel, line 112:
```suggestion
_summary: Collect the device firmware update information
_purpose: Attach information about the devices, as reported by the fwupdmgr
```

kissiel (changes requested):
Minor addition proposed in line, and one possible simplification.

providers/base/units/firmware/jobs.pxu
stanley31huang: Thanks for the comments, I have removed the step to saving and restoring the SANP env var.

providers/base/units/firmware/jobs.pxu
```diff
@@ -105,3 +105,14 @@ plugin: attachment
...
+_purpose: Attach information about the devices, as reported by the fwupdmgr
+requires:
+    executable.name == "fwupdmgr"
+command:
+    unset SNAP
```
superm1, line 117: If you upgrade fwupd to 1.9.14/15 in 22.04 you won't need to do this.

stanley31huang: @superm1  
thanks for the feedback.
But on the PC enablement project, system will include the fwupd Debian package instead of fwupd snap package.
And considering the fwupd Debian package is installed by default, so we should use it for the testing.
So we still need to apply the workaround for now, until we get any fixed for fwupd Debian package.

superm1: Right; I just mean to say you can do backport/SRU and it can be dropped or you can have different test case for different releases.

stanley31huang: I see your concerns now, but this test case would be included across different Ubuntu LTS version.
So I have implemented a small scripts to detect the fwupd package to allow check could retrieve firmware information  with Debian fwupd and Snap fwupd now.

stanley31huang: @superm1  let me know if you have any concerns with current implementation.

superm1: No it looks fine to me. The only suggestion I would offer you is that you can totally avoid a lot of this if you just use gobject introspection instead. Then you don't need to distinguish any versions snap or deb. You would just use the libfwupd gi instead.

You can look at simple_client.py upstream to see what I'm talking about.

But if you choose not to do it what you did looks functional.

superm1: I think you must be missing a slot or connection to properly communicate with confined snap. But yeah that's fine.

providers/base/bin/get_firmware_info_fwupd.py
```diff
@@ -0,0 +1,61 @@
...
+            del os.environ["SNAP"]
+
+        subprocess.run(shlex.split("fwupdmgr get-devices --json"))
+
+
+if __name__ == "__main__":
+    try:
+        get_firmware_info_fwupd()
+    except Exception as err:
+        print(err)
```
kissiel, line 61: This means that if things go catastrophically bad, the job will still be considered OK.
Stdout shouldn't be used for error messaging.

providers/base/bin/get_firmware_info_fwupd.py
```diff
@@ -0,0 +1,61 @@
...
+        fwupd_vers = subprocess.run(
+            shlex.split("fwupdmgr --version --json"),
+            capture_output=True)
+        fwupd_vers = json.loads(fwupd_vers.stdout)
+
+        runtime_ver = ()
+        for ver in fwupd_vers.get("Versions", []):
+            if (ver.get("Type") == "runtime" and
+                    ver.get("AppstreamId") == "org.freedesktop.fwupd"):
+                runtime_ver = tuple(map(int, ver.get("Version").split(".")))
```
kissiel, line 44: This logic is not tested anywhere, nor is it described what it does. It qualifies as a separate function with a well defined input and output, and obviously tests.

kissiel (changes requested):
The tests provided here look like they are here just to have test coverage instead of testing the logic (and the most difficult piece of logic is untested).

providers/base/bin/get_firmware_info_fwupd.py
stanley31huang: Fixed it, print out the error message by logging.error function.

providers/base/bin/get_firmware_info_fwupd.py
stanley31huang: Fixed, I have implemented two separate function for this part. Thanks

kissiel (approved):
Awesome! Thanks for all the amendments!
+1

PR #1091, opened 2024-03-20
Fix missing lsb-release build-depends in core16 snapcraft.yaml (infra)

Hook25:
## Description

One of the reasons why core snaps are failing to build currently is that lsb-release is now missing and needed by a (python) dependency of checkbox-ng. This adds it to build-depends. 

## Resolved issues

Fixes: CHECKBOX-1316

## Documentation

N/A

## Tests

N/A


kissiel (approved):
Small patch, big fix. +1

PR #1092, opened 2024-03-20
[checkbox-ce-oem] Fix serial tests (Bugfix)

LiaoU3:
## Description
1. Fix the field `requires`: True -> 'True'
2. Fix undecodable byte from `os.urand`: Use ASCII letters, numbers, punctuation instead
3. Correct `_template-summary` by not using jinja variable
4. Increase the testing byte from 128 to 1024
5. Also increase the timeout value from 1 to 3

## Resolved issues
N/A

## Documentation
N/A

## Tests
submission: https://certification.canonical.com/hardware/202401-33430/submission/361465/

@rickwu666666 I also observed that your machine can not receive message with /dev/ttyMAX1 with baudrate 115200

### DIFF
--- a/contrib/checkbox-provider-ce-oem/bin/serial_test.py
+++ b/contrib/checkbox-provider-ce-oem/bin/serial_test.py
@@ -19,6 +19,8 @@ import serial.rs485
 import time
 import logging
 import os
+import random
+import string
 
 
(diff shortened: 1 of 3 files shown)


rickwu666666 (approved):
LGTM

Issue #1080, opened 2024-03-18
Some issues with test case power-management/lid_close_suspend_open


eugene-yujinwu:
### Bug Description

There are some issues related to the test case [power-management/lid_close_suspend_open](https://github.com/canonical/checkbox/blob/44498aa07112594d4634a94af9be4134b9977128/providers/base/units/power-management/jobs.pxu#L158):

1. A typo that make the test steps jump from 3 to 5.

2. There is a [HoldoffTimeoutSec](https://www.freedesktop.org/software/systemd/man/latest/logind.conf.html#HoldoffTimeoutSec=) which will "Specifies a period of time after system startup or system resume in which systemd will hold off on reacting to lid events....Defaults to 30s. " So if we run this test just after a resume from suspend, it will failed.  Please see LP bugs: [LP:2042406](https://bugs.launchpad.net/somerville/+bug/2042406),[ LP:2047136](https://bugs.launchpad.net/sutton/+bug/2047136). Maybe we need add some notes to make the tester aware of this information.

3. This test case seems to be redundant with another test case [power-management/lid](https://github.com/canonical/checkbox/blob/44498aa07112594d4634a94af9be4134b9977128/providers/base/units/power-management/jobs.pxu#L99C5-L99C25). Each of them listed in the test plan [after-suspend-graphics-integrated-gpu-cert-manual ](https://github.com/canonical/checkbox/blob/a12b050c58e8440285f2eb6ab4409ebfeea46d03/providers/base/units/graphics/test-plan.pxu#L122), so it's recursively included in the test plan client-cert-desktop-22-04-manual. Maybe we can remove one of them.

4. A improvement suggestion from Kent: Maybe we could move this test case away from the suspend_advanced_auto test case, to avoid the situation that run this test case just after the resume from suspend. 

### To Reproduce

checbox-cli run com.canonical.certification::power-management/lid_close_suspend_open

### Environment

checkbox: 4.0.0.dev139. Lenovo Desktop OEM image 1016-oem kernel

zongminl:
This has been fixed by #1093

PR #1093, opened 2024-03-20
Fix some issues related with the test case power-management/lid_close_suspend_open (Bugfix)

eugene-yujinwu:
Fix some issues related with the test case power-management/lid_close_suspend_open (bugfix) (#1080)

- Include a note to ensure the test is executed at the appropriate time
- Fix the typo in the test order number
- Remove functionally similar test cases power-management/lid from the test plan
- Adjust the place of test case to avoid running it just after the suspend case

### DIFF
--- a/providers/base/bin/lid_close_suspend_open.sh
+++ b/providers/base/bin/lid_close_suspend_open.sh
@@ -1,4 +1,16 @@
-#! /usr/bin/bash
+#!/usr/bin/bash
+
+holdoff_timeout_usec=$(gdbus introspect --system --dest org.freedesktop.login1 --object-path /org/freedesktop/login1 -p | grep HoldoffTimeoutUSec | awk '{print $5}' | awk -F\; '{print $1}')
+holdoff_timeout_sec=$(echo "scale=0; $holdoff_timeout_usec / 1000000" | bc)
+
+if [ "$holdoff_timeout_sec" -ne 0 ]; then
+    while (journalctl --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit" >/dev/null 2>&1)
+    do
+        echo "The system just resume from suspend, please wait for $holdoff_timeout_sec seconds and continue the test"
+        sleep 1
+    done
+fi
+
 prev_suspend_number=$(cat /sys/power/suspend_stats/success)
 echo "Number of successful suspends until now: $prev_suspend_number"
 echo "Please close the lid and wait for 5 sec to make it suspend~"
--- a/providers/base/bin/lid_close_suspend_open.sh
+++ b/providers/base/bin/lid_close_suspend_open.sh
@@ -1,15 +1,15 @@
 #!/usr/bin/bash
 
 holdoff_timeout_usec=$(gdbus introspect --system --dest org.freedesktop.login1 --object-path /org/freedesktop/login1 -p | grep HoldoffTimeoutUSec | awk '{print $5}' | awk -F\; '{print $1}')
-holdoff_timeout_sec=$(echo "scale=0; $holdoff_timeout_usec / 1000000" | bc)
+holdoff_timeout_sec=$((holdoff_timeout_usec / 1000000))
 
 previous_sleep_log=$(journalctl --output=short-unix --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit")
 if [[ "$holdoff_timeout_sec" != 0 && "$previous_sleep_log" != "" ]]; then
     # set the previous_sleep_time
     previous_sleep_time=$(echo "$previous_sleep_log" | awk -F'.' '{ print $1 }')
-    # sleep a period of time
-    sleep_time=$(( $(date +"%s") - "$previous_sleep_time" ))
-    echo "sleep for $sleep_time seconds ..."
+    # sleep a period of time. wait_time = holdoff_time - wakeup_time_from_last_suspend
+    sleep_time=$(("$holdoff_timeout_sec" - ($(date +"%s") - "$previous_sleep_time")))
+    echo "DUT was resumed less than ${holdoff_timeout_sec} seconds ago. Waiting for ${sleep_time} seconds before running the test..."
     sleep "$sleep_time"
 fi
 echo "System is ready for suspend test"
--- a/providers/base/bin/lid_close_suspend_open.sh
+++ b/providers/base/bin/lid_close_suspend_open.sh
@@ -6,7 +6,7 @@ holdoff_timeout_sec=$(echo "scale=0; $holdoff_timeout_usec / 1000000" | bc)
 previous_sleep_log=$(journalctl --output=short-unix --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit")
 if [[ "$holdoff_timeout_sec" != 0 && "$previous_sleep_log" != "" ]]; then
     # set the previous_sleep_time
-    previous_sleep_time=$(echo $previous_sleep_log | awk -F'.' '{ print $1 }')
+    previous_sleep_time=$(echo "$previous_sleep_log" | awk -F'.' '{ print $1 }')
     # sleep a period of time
     sleep_time=$(( $(date +"%s") - "$previous_sleep_time" ))
     echo "sleep for $sleep_time seconds ..."
--- a/providers/base/bin/lid_close_suspend_open.sh
+++ b/providers/base/bin/lid_close_suspend_open.sh
@@ -3,18 +3,16 @@
 holdoff_timeout_usec=$(gdbus introspect --system --dest org.freedesktop.login1 --object-path /org/freedesktop/login1 -p | grep HoldoffTimeoutUSec | awk '{print $5}' | awk -F\; '{print $1}')
 holdoff_timeout_sec=$(echo "scale=0; $holdoff_timeout_usec / 1000000" | bc)
 
-if [ "$holdoff_timeout_sec" -ne 0 ]; then
-    if (journalctl --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit" >/dev/null 2>&1); then
-        echo "The system just resume from suspend, the lid event will be hold off on in $holdoff_timeout_sec seconds"
-        echo "Please wait for the new prompt before starting the test."
-    fi
-    while (journalctl --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit" >/dev/null 2>&1)
-    do
-        echo "waiting... "
-        sleep 3
-    done
-    echo ""
+previous_sleep_log=$(journalctl --output=short-unix --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit")
+if [[ "$holdoff_timeout_sec" != 0 && "$previous_sleep_log" != "" ]]; then
+    # set the previous_sleep_time
+    previous_sleep_time=$(echo $previous_sleep_log | awk -F'.' '{ print $1 }')
+    # sleep a period of time
+    sleep_time=$(( $(date +"%s") - "$previous_sleep_time" ))
+    echo "sleep for $sleep_time seconds ..."
+    sleep "$sleep_time"
 fi
+echo "System is ready for suspend test"
 
 prev_suspend_number=$(cat /sys/power/suspend_stats/success)
 echo "Number of successful suspends until now: $prev_suspend_number"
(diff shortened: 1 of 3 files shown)


providers/base/units/power-management/jobs.pxu
```diff
@@ -161,10 +161,12 @@ requires: device.product == 'Lid Switch'
 _purpose:
     This test will check your lid sensor can detect lid close/open, and DUT will suspend when the lid closed
 _steps:
-    1. Press "Enter" to start the test
-    2. Close the lid (Please close the lid within 10 sec)
+    (Please wait at least 30 seconds after resuming from sleep before running this test. In most cases,
+     systemd will hold off on reacting to lid events in a period of time after system startup or resume.)
+    1. Press "Enter" to start the test.
+    2. Close the lid (Please close the lid within 10 sec).
     3. Wait 5 seconds with the lid closed.
```
stanley31huang, line 168: it's a bit confuse to me that the waiting time is not match in step 3 and the above notes.

stanley31huang:
@eugene-yujinwu  please remove the latest white space character from the title for the validate_title check.

eugene-yujinwu:
> @eugene-yujinwu please remove the latest white space character from the title for the validate_title check.

Done. Thanks!

providers/base/units/power-management/jobs.pxu
eugene-yujinwu: @stanley31huang Thanks for comment. They are different time, the time mentioned in the notes is the time should wait to run this test if system just resume from suspend. The time in step 3 is the time to close the lid. It's maybe really a bit confusing, I'm not sure how to make it a little more clear.

clairlin53:
@eugene-yujinwu 
I recommend keeping the second times with lid test, there were some issues observed with lid close after second times and there is an issue when closing the lid over 10 seconds, please refer to the following bugs for more details.
https://bugs.launchpad.net/stella/+bug/1974120
https://bugs.launchpad.net/stella/+bug/1979038
https://bugs.launchpad.net/stella/+bug/1962826

eugene-yujinwu:
> @eugene-yujinwu I recommend keeping the second times with lid test, there were some issues observed with lid close after second times and there is an issue when closing the lid over 10 seconds, please refer to the following bugs for more details. https://bugs.launchpad.net/stella/+bug/1974120 https://bugs.launchpad.net/stella/+bug/1979038 https://bugs.launchpad.net/stella/+bug/1962826

eugene-yujinwu:
Add the power-management/lid test back per Clair's comments.

eugene-yujinwu:
Reopen it. Thank @Vincent!

providers/base/units/power-management/jobs.pxu
stanley31huang: It seems to me that we could simply write a scripts to check if current time is too close to previous suspend process.
If previous suspend process within 30 seconds, Ask user to wait a period of time and then close the lid.

eugene-yujinwu: I think this is good idea. But I don't know how to get the how long the last suspend time to now. Do you have any idea about that? Thanks!

hanhsuan:
@eugene-yujinwu  I am thinking about that the 30 seconds is the default value defined in the `logind.conf` by `HoldoffTimeoutSec`. Is that possible to show the real value while this test case at beginning ? Then the user don't have to guess how long they have to wait for next test.

providers/base/bin/lid_close_suspend_open.sh
```diff
@@ -1,4 +1,16 @@
-#! /usr/bin/bash
+#!/usr/bin/bash
+
+holdoff_timeout_usec=$(gdbus introspect --system --dest org.freedesktop.login1 --object-path /org/freedesktop/login1 -p | grep HoldoffTimeoutUSec | awk '{print $5}' | awk -F\; '{print $1}')
+holdoff_timeout_sec=$(echo "scale=0; $holdoff_timeout_usec / 1000000" | bc)
+
+if [ "$holdoff_timeout_sec" -ne 0 ]; then
+    while (journalctl --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit" >/dev/null 2>&1)
+    do
+        echo "The system just resume from suspend, please wait for $holdoff_timeout_sec seconds and continue the test"
```
hanhsuan, line 9: The `$holdoff_timeout_sec` is never changed during the loop, it would be better to echo just once at beginning.

eugene-yujinwu: @hanhsuan A new commit was pushed, please kindly help review it. Many thanks!

providers/base/bin/lid_close_suspend_open.sh
```diff
@@ -1,4 +1,21 @@
...
+if [ "$holdoff_timeout_sec" -ne 0 ]; then
+    if (journalctl --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit" >/dev/null 2>&1); then
+        echo "The system just resume from suspend, the lid event will be hold off on in $holdoff_timeout_sec seconds"
+        echo "Please wait for the new prompt before starting the test."
+    fi
+    while (journalctl --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit" >/dev/null 2>&1)
+    do
+        echo "waiting... "
+        sleep 3
+    done
```
stanley31huang, line 15: instead of having a while loop to polling the journal logs, I would suggest we simply sleep within a period of time if needed, likes
```
previous_sleep_log=$(journalctl --output=short-unix --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit")
if [ "$holdoff_timeout_sec" -ne 0 and "$previous_sleep_log" -ne "" ]; then
    # set the previous_sleep_time
    previous_sleep_time=$(echo $previous_sleep_log | awk -F'.' '{ print $1 }')
    # sleep a period of time
    sleep_time=$(( $(date +"%s") - "$previous_sleep_time" ))
    echo "sleep for $sleep_time seconds ..."
    sleep "$sleep_time"
fi
echo "System is ready for suspend test"
```

providers/base/units/power-management/jobs.pxu
stanley31huang: If the scripts will make sure the system is ready for suspend test, we need to modify the steps here.

providers/base/bin/lid_close_suspend_open.sh
eugene-yujinwu: @stanley31huang a new commit was pushed per this advise. Just a little change. Please have a look when you have time. thanks a lot.

stanley31huang:
@eugene-yujinwu  please change the status of this PR to `Ready for review` when you think it's ready. And please make sure all unit tests passed with your changes.

eugene-yujinwu:
@stanley31huang just rebase the commit and reopen the review. But I cannot figure out why some of the unit tests failed. Do you have any idea or hints? Thanks!

eugene-yujinwu:
@stanley31huang , finally I make the unit tests passed:). could you kindly review this, thanks!

providers/base/units/power-management/jobs.pxu
```diff
@@ -189,6 +189,7 @@ _steps:
     4. Open the lid
 _verification:
     Did the system suspend when the lid was closed, and resume back when the lid was opened?
+    (Note: Systemd will hold off on reacting to lid events in a period of time after system startup or resume. So please make sure the system doesn't just resume from a suspend before run the test.)
```
pieqq, line 192:
```suggestion
    Note: Systemd will not react to lid events if the DUT was just started or resumed. Please make sure the DUT has been running for long enough before running this test.
```

providers/base/units/graphics/test-plan.pxu
```diff
@@ -220,6 +218,8 @@ include:
     after-suspend-graphics/1_video_.*                        certification-status=blocker
     after-suspend-graphics/1_cycle_resolution_.*             certification-status=non-blocker
     suspend/1_xrandr_screens_after_suspend.tar.gz_auto
+    power-management/lid_close_suspend_open                  certification-status=blocker
+    power-management/lid                                     certification-status=blocker
```
pieqq, line 222: Any specific reason why the jobs order has been modified?

providers/base/bin/lid_close_suspend_open.sh
```diff
@@ -1,4 +1,19 @@
-#! /usr/bin/bash
+#!/usr/bin/bash
+
+holdoff_timeout_usec=$(gdbus introspect --system --dest org.freedesktop.login1 --object-path /org/freedesktop/login1 -p | grep HoldoffTimeoutUSec | awk '{print $5}' | awk -F\; '{print $1}')
+holdoff_timeout_sec=$(echo "scale=0; $holdoff_timeout_usec / 1000000" | bc)
```
pieqq, line 4: Bash comes with [arithmetic expansion feature](https://www.gnu.org/software/bash/manual/html_node/Arithmetic-Expansion.html) to work with numbers, so I think you should be able to use this instead:

```suggestion
holdoff_timeout_sec=$(($holdoff_timeout_usec / 1000000))
```

providers/base/bin/lid_close_suspend_open.sh
```diff
@@ -1,4 +1,19 @@
...
+
+previous_sleep_log=$(journalctl --output=short-unix --since "$holdoff_timeout_sec seconds ago" -b 0 -r | grep "suspend exit")
+if [[ "$holdoff_timeout_sec" != 0 && "$previous_sleep_log" != "" ]]; then
+    # set the previous_sleep_time
+    previous_sleep_time=$(echo "$previous_sleep_log" | awk -F'.' '{ print $1 }')
+    # sleep a period of time
+    sleep_time=$(( $(date +"%s") - "$previous_sleep_time" ))
+    echo "sleep for $sleep_time seconds ..."
```
pieqq, line 12:
```suggestion
    echo "DUT was resumed less than 30 seconds ago. Waiting for ${sleep_time seconds} before running the test..."
```

pieqq (changes requested):
Nice quality of life improvement on this job!

I feel like the Shell script is becoming really big and hard to maintain, and it is not tested... I think it would be worthwhile to convert it into a Python script with unit tests.

That said, I've made a few comments and added some suggestions.

providers/base/bin/lid_close_suspend_open.sh
eugene-yujinwu: This works. But an error will be reported:
[Line 4:](javascript:setPosition(4, 24))
holdoff_timeout_sec=$(($holdoff_timeout_usec / 1000000))
                       ^-- [SC2004](https://www.shellcheck.net/wiki/SC2004) (style): $/${} is unnecessary on arithmetic variables.
So remove the $ before it.

providers/base/bin/lid_close_suspend_open.sh
eugene-yujinwu: clearer expression. But 30 seconds should be ${holdoff_timeout_sec} seconds.

providers/base/units/graphics/test-plan.pxu
eugene-yujinwu: The second close lid test will wait at least ${holdoff_timeout_sec} to sleep since the DUT just resume from last lid test.So it will take more time to finished this test than put it at other place which not just after a suspend. However, put this test to other place are not very appropriate too. And put it at here will test the hold off time in some way, so just put it back here.

providers/base/units/power-management/jobs.pxu
eugene-yujinwu: Clearer expression. ACK.

eugene-yujinwu:
@pieqq Thanks for review. A new commit was pushed per your suggestion. Could you kindly review it. Thanks!

eugene-yujinwu:
> Nice quality of life improvement on this job!
> 
> I feel like the Shell script is becoming really big and hard to maintain, and it is not tested... I think it would be worthwhile to convert it into a Python script with unit tests.
> 
> That said, I've made a few comments and added some suggestions.

@pieqq Do you think I should drop this PR, and rewrite it with python and submit a new PR?

pieqq (approved):
+1, thanks!

pieqq:
> > Nice quality of life improvement on this job!
> > I feel like the Shell script is becoming really big and hard to maintain, and it is not tested... I think it would be worthwhile to convert it into a Python script with unit tests.
> > That said, I've made a few comments and added some suggestions.
> 
> @pieqq Do you think I should drop this PR, and rewrite it with python and submit a new PR?

Let'd land this. You can keep a note to improve it later (switch to Python + add unit tests).

eugene-yujinwu:
> > > Nice quality of life improvement on this job!
> > > I feel like the Shell script is becoming really big and hard to maintain, and it is not tested... I think it would be worthwhile to convert it into a Python script with unit tests.
> > > That said, I've made a few comments and added some suggestions.
> > 
> > 
> > @pieqq Do you think I should drop this PR, and rewrite it with python and submit a new PR?
> 
> Let'd land this. You can keep a note to improve it later (switch to Python + add unit tests).

Thank you @pieqq . OK, let's do it later.

PR #1094, opened 2024-03-20
Fix the rest of job defintions (BugFix)

kissiel:
Another here's the final portion of changes done by the pxu-fixer.
Changes proposed were added in bulk with `git add -u`.
They should include only cosmetic changes.

### DIFF
--- a/providers/base/units/gpio/jobs.pxu
+++ b/providers/base/units/gpio/jobs.pxu
@@ -18,7 +18,6 @@ flags: preserve-locale also-after-suspend
 imports: from com.canonical.plainbox import manifest
 requires: manifest.gpio_loopback == 'True'
 
-```
 # Filtering this job by the same devices as above as uses the same pin 
 # definition file and uses the RPi python module
 unit: template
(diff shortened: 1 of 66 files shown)


providers/base/units/gpio/jobs.pxu
```diff
@@ -17,7 +17,7 @@ flags: preserve-locale also-after-suspend
 imports: from com.canonical.plainbox import manifest
 requires: manifest.gpio_loopback == 'True'
 
-
+```
```
tang-mm, line 20: Is this a valid syntax in PXU?

tang-mm (changes requested):
The updates are mostly about splitting fields with minor wording changes. They look good to me in general, except for the syntax that seems to be markdown.

providers/base/units/gpio/jobs.pxu
kissiel: Nope, that's an LLM hallucination :)
OMW to fix that.

tang-mm (dismissed):

tang-mm (approved):
LGTM +1 Thanks for improving them!

Issue #1090, opened 2024-03-20
Rerun the failed job then reopen the session will make the job status is changed to crash


hanhsuan:
### Bug Description

If using UI to run the test cases and there are some test cases failed, then we will try to rerun them. However, some test cases might be tested before suspend/reboot then we will close the session manually to make sure the result is saved. After this, the latest rerun job status will always be changed to crash. The behavior of remote and local test are the same.

The steps is similar:
**some test cases failed**
![Screenshot from 2024-03-20 14-44-41](https://github.com/canonical/checkbox/assets/32028620/e5fd72e9-a87d-4068-858d-f2a82251ad6a)
**rerun and passed**
![job_pssed](https://github.com/canonical/checkbox/assets/32028620/88814aa4-f3e1-4880-bc11-cb221d5b0f79)
![Screenshot from 2024-03-20 14-45-10](https://github.com/canonical/checkbox/assets/32028620/8cacd046-582f-4660-aedc-828d3db670a5)
**close session**
![close_session](https://github.com/canonical/checkbox/assets/32028620/47268508-4fc8-42f1-8011-0a35e266cc23)
**reopen the session**
![Screenshot from 2024-03-20 14-45-58](https://github.com/canonical/checkbox/assets/32028620/0113e09f-77ba-47e2-8be8-e9e52765db7b)

### To Reproduce

As above

### Environment

- OS: jammy
- checkbox type: debian
- checkbox version: v4.0.0.dev139

PR #1095, opened 2024-03-20
Fix resume forgetting previous job outcome (bugfix)

Hook25:
## Description

When resuming a session Checkbox doesn not take into consideration the previous outcome that a job may have had. This causes issues when a session is terminated after an outcome is assigned but before moving on (for example, when the interrupt menu is brought up right after the outcome is assigned). This makes Checkbox use the infromation if it is present.

## Resolved issues

Resolves: CHECKBOX-1315
Resolves: https://github.com/canonical/checkbox/issues/1090 

## Documentation

Updated the docstrings

## Tests

Added unit tests and metabox scenarios


kissiel (approved):
wow, indeed the previous logic flow was bad...
This is fantastic.

PR #1097, opened 2024-03-21
Add example to the Template unit template-summary field (Infra)

pieqq:
## Description

Following a discussion with QA, add an example of a `template-summary` field in the Template Unit reference page.

### DIFF
--- a/docs/reference/units/template.rst
+++ b/docs/reference/units/template.rst
@@ -48,6 +48,14 @@ Template-Specific Fields
     translation into other languages. It must be one line long, ideally it
     should be short (50-70 characters max).
 
+    Example: if a template has a ``summary`` field is set to:
+
+        *Test serial connection for port {serial_port}*
+
+    then its ``template-summary`` could be:
+
+        *Test serial connection for every serial port on the system*
+
     This field is optional (Checkbox will only advise you to provide one when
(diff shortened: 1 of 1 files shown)


LiaoU3 (approved):
It is awesome with examples!

docs/reference/units/template.rst
```diff
@@ -48,6 +48,14 @@ Template-Specific Fields
     translation into other languages. It must be one line long, ideally it
     should be short (50-70 characters max).
 
+    Example: if a template has a ``summary`` field is set to:
+
+        *Test serial connection for port {serial_port}*
+
```
tang-mm, line 54:
```suggestion

Each instantiated job will include the same ``summary`` field, generated with the actual value of the serial port.

```

tang-mm (approved):
The example looks clear to me. I added a one-line suggestion to further distinguish the two fields, but it is not blocking as it's already explained in other parts of the doc

docs/reference/units/template.rst
pieqq: As discussed elsewhere, this is a good idea to explain what happens to these generated jobs, but it might be better to put it in the examples sections at the bottom of the document. I will do that next.

PR #1102, opened 2024-03-22
Moved genio from contrib to providers (New)

fernando79513:
## Description

Since we have met the coverage requirements for the genio provider, we have decided to move it from the "contrib" area to the providers. The following changes needed to include the last changes on the [Launchpad branch](https://git.launchpad.net/~baoshan-team/baoshan/+git/checkbox-provider-baoshan/) will be added via PR to checkbox and should adhere to the coverage requirements.

## Resolved issues

N/A

## Documentation

N/A

## Tests

The pipelines should be executed and passed for the genio provider.


kissiel (changes requested):
I think the workflow filename could use a rename too

fernando79513:
> I think the workflow filename could use a rename too

Nice catch, I have already renamed the workflow file

PR #1104, opened 2024-03-22
Move lsb_release install up to stage and build of checkbox-support (bugfix)

Hook25:
## Description

The previous patch was incomplete (but I didnt notice because of snapcraft caching). We need lsb_release both in build and in stage. Also, we need it for checkbox-support as well. This moves the install in the first part that needs it.

## Resolved issues

Failing daily builds

## Documentation

N/A

## Tests

Built via LXD (as LP remote-builds are still broken)


kissiel (approved):
Makes sense, +1

PR #1105, opened 2024-03-25
Add sysfs led indicator test into automated test coverage(New)

rickwu666666:
## Description
Add led indicator test auto-test job and plan. Also, add it into ce-oem auto test coverage due to we have a [regression issue](https://bugs.launchpad.net/shiner/+bug/2057719) found in the post-gm stage.

## Resolved issues
N/A

## Tests
Sideload results with make up some non-existent led paths
https://certification.canonical.com/hardware/202205-30270/submission/361748/test-results/

### DIFF
--- a/providers/base/units/led/jobs.pxu
+++ b/providers/base/units/led/jobs.pxu
@@ -352,6 +352,7 @@ command:
 unit: template
 template-resource: led-indicator/sysfs-leds
 template-unit: job
+template-id: led-indicator/sysfs-leds-auto
 id: led-indicator/sysfs-leds-{name}-auto
 category_id: led
 _summary: Automated check control of {name} LED.
--- a/providers/base/units/led/jobs.pxu
+++ b/providers/base/units/led/jobs.pxu
@@ -331,7 +331,7 @@ command:
 unit: template
 template-resource: led-indicator/sysfs-leds
 template-unit: job
-id: led-indicator/sysfs-leds-{name}
+id: led-indicator/sysfs-leds-{name}-manual
 category_id: led
 _summary: Check control of {name} LED.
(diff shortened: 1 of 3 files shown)


providers/base/units/led/jobs.pxu
```diff
@@ -349,6 +349,23 @@ requires: manifest.has_led_indicator == 'True'
...
+  Automated check that {name} LED turns on and off.
+plugin: shell
+user: root
+flags: also-after-suspend
+estimated_duration: 10
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_led_indicator == 'True'
+command:
+    led_control_test.py -n {path} -d 10
+
```
kissiel, line 368: The job doesn't check if LED turns on and off.

kissiel (changes requested):
The job doesn't check if LED turns on and off.

rickwu666666:
> The job doesn't check if LED turns on and off.

Hi @kissiel,
Yes, you are correct, this job is not able to tell if the LED actually is on or off. However, the script checks whether the value is expected or not after writing the value to the node.  Wondering if that is good enough.
https://github.com/canonical/checkbox/blob/main/providers/base/bin/led_control_test.py#L61

providers/base/units/led/jobs.pxu
rickwu666666: Hi @kissiel,
Yes, you are correct, this job is not able to tell if the LED actually is on or off. However, the script checks whether the value is expected or not after writing the value to the node. Wondering if that is good enough.
https://github.com/canonical/checkbox/blob/main/providers/base/bin/led_control_test.py#L61

PR #1106, opened 2024-03-25
migrate `keys/microphone-mute` from pulseaudio to pipewire (BugFix)

hanhsuan:
This test case is missed to migrate in the #826

## Description

This PR is going to make `keys/microphone-mute` could be executed on both environment.

## Resolved issues

`keys/microphone-mute` could be only executed on `pulseaudio` environment only.

## Tests

The command is same with `audio/alsa_record_playback_internal`


pieqq (approved):

pieqq:
LGTM :)

PR #1111, opened 2024-03-25
Add submision for camera quality test (New)

fernando79513:
## Description

The test camera/camera-quality_.* fails from time to time. Since this test uses an algorithm to determine the output, it can be difficult to assess whether there is really a problem with the webcam or not, especially when reviewing submissions that do not have a sample of the image used for this.

We have added an attachment job to include the image used to compute the quality score so we can check it later.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1281

## Tests

Run the camera automated tests:
```
com.canonical.certification::camera-cert-automated
```
Check that the image has been created in the `PLAINBOX_SESSION_SHARE` directory

### DIFF
--- a/providers/base/bin/camera_quality_test.py
+++ b/providers/base/bin/camera_quality_test.py
@@ -30,6 +30,7 @@ import cv2
 import numpy as np
 
 from checkbox_support.vendor.brisque.brisque import BRISQUE
+from checkbox_support.helpers.timeout import timeout
 from tempfile import NamedTemporaryFile
 
 
@@ -157,6 +158,7 @@ def evaluate_score(score: float) -> int:
     return 0
 
 
+@timeout(120)
 def main(argv: list) -> int:
     parser = argparse.ArgumentParser(description="Run the image quality test")
     parser.add_argument(
--- a/providers/base/bin/camera_quality_test.py
+++ b/providers/base/bin/camera_quality_test.py
@@ -31,7 +31,6 @@ import numpy as np
 
 from checkbox_support.vendor.brisque.brisque import BRISQUE
 from checkbox_support.helpers.timeout import timeout
-from tempfile import NamedTemporaryFile
 
 
 THRESHOLD = 60
--- a/providers/base/bin/camera_quality_test.py
+++ b/providers/base/bin/camera_quality_test.py
@@ -22,6 +22,7 @@
 import argparse
 from collections import deque
 import logging
+import os
 import time
 import sys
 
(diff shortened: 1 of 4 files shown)


providers/base/units/camera/jobs.pxu
```diff
@@ -119,13 +119,31 @@ estimated_duration: 20s
 depends: camera/detect
 requires: lsb.release >= '22.04'
 command:
-  timeout 120 camera_quality_test.py -d {{ name }} ||
-  (>&2 echo "Timeout computing score"; false)
+  camera_quality_test.py -d {{ name }}
```
pieqq, line 122: In cases where we want to save a file/document somewhere, we usually add an argument to specify where to save this. You could add an `-o/--output` argument for instance:

```suggestion
  camera_quality_test.py -d {{ name }} -o "${PLAINBOX_SESSION_SHARE}"
```

providers/base/bin/camera_quality_test.py
```diff
@@ -22,22 +22,51 @@
...
 
 from checkbox_support.vendor.brisque.brisque import BRISQUE
+from checkbox_support.helpers.timeout import timeout
 from tempfile import NamedTemporaryFile
 
+
 THRESHOLD = 60
 TIMEOUT = 10
 MIN_INTERVAL = 0.5
+PLAINBOX_SESSION_SHARE = os.environ.get("PLAINBOX_SESSION_SHARE", "")
```
pieqq, line 40: Instead of hardcoding the output location, use an argument (for instance `--output/-o`) when calling the script, so that it's easier to try out the script without calling Checkbox.

providers/base/bin/camera_quality_test.py
```diff
@@ -99,15 +126,7 @@ def get_score_from_device(device: str, save: bool = False) -> float:
...
-    # Save the image if requested
-    if save:
-        with NamedTemporaryFile(
-            prefix="camera_test_brisque_{}".format(device),
-            suffix=".jpg",
-            delete=False,
-        ) as f:
-            cv2.imwrite(f.name, image)
-            print("Saved image to {}".format(f.name))
+    save_image(image, device)
```
pieqq, line 129: If you use a `--output/-o` argument, you could just call `save_image()` if it's set.

providers/base/bin/camera_quality_test.py
```diff
@@ -22,22 +22,51 @@
...
+        cv2.imwrite(filepath, image)
+        print("Saved image to {}".format(filepath))
+    else:
+        with NamedTemporaryFile(
+            prefix="quality_image_{}".format(device),
+            suffix=".jpg",
+            delete=False,
+        ) as f:
+            cv2.imwrite(f.name, image)
+            print("Saved image to {}".format(f.name))
```
pieqq, line 67: If you use a `--output/-o` argument, I don't think the `else` logic is very useful. Moreover, the logic to check for the output folder could be done outside of the `save_image()` function.

pieqq (changes requested):
Thanks for tackling this, it's gonna be super helpful!

I have a few suggestions inline below. Moreover, I think this PR includes changes that are not related to the problem you're fixing. Maybe rebase this PR against `main`?

providers/base/units/camera/jobs.pxu
```diff
@@ -119,13 +119,31 @@ estimated_duration: 20s
...
   Takes a picture and computes the quality based on a No-Reference image
   quality assessment algorithm called BRISQUE. This test will timeout and fail
   if the quality has not been compute within 120 seconds.
 
+unit: template
+template-resource: device
+template-filter: device.category == 'CAPTURE' and device.name != ''
+template-unit: job
+plugin: attachment
+template-engine: jinja2
```
pieqq, line 133: Any reason for using `jinja2` templating instead of the default?

pieqq (changes requested):

providers/base/units/camera/jobs.pxu
fernando79513: I was just using the same as the one we are using with the camera test. I am not sure about the differences between both.

pieqq (approved):
All good, thanks!

PR #1113, opened 2024-03-26
Disk resource gated jobs (bugfix)

Hook25:
## Description

The test `disk/hdd-parking` is currently being ran on every device that has a disk. This leads to other tests failing  because the script enables a service that instantly goes in the degraded state, making other tests (that check that there is no service enabled and degraded) fail. This also pollutes submissions as the test can never pass on devices that do not have an hard disk.

This introduces a new resource job that lists various useful information about block devices, including if the block device is "rotational". This information is integrated in the requires constraint of the hdd test.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1136

## Documentation

N/A

## Tests

The new resource job is unit tested.

### DIFF
--- a/providers/base/bin/disk_resource.py
+++ b/providers/base/bin/disk_resource.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+#
+# This file is part of Checkbox.
+#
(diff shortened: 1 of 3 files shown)


fernando79513 (approved):
Really nice and clean, GJ!
LGTM +1

PR #1116, opened 2024-03-27
Switching power mode for Lenovo's platforms (New)

binli:
## Description

## Resolved issues

https://github.com/canonical/checkbox/issues/816

## Documentation

This testcase could help verify power mode can be changed and stay in the chosen setting.

## Tests

$ checkbox-cli run com.canonical.certification::power-management/switch_power_mode

---------------------[ power-management/switch_power_mode ]---------------------
ID: com.canonical.certification::power-management/switch_power_mode
Category: com.canonical.plainbox::power-management
Power mode choices: ['low-power', 'balanced', 'performance']
Switch to low-power successfully.
Switch to balanced successfully.
Switch to performance successfully.
Outcome: job passed
Finalizing session that hasn't been submitted anywhere: checkbox-run-2024-03-22T16.31.38
 ☑ : Collect information about kernel modules
 ☑ : Discover if the system supports power modes via acpi
 ☑ : power-management/switch_power_mode

### DIFF
--- a/providers/base/bin/switch_power_mode.py
+++ b/providers/base/bin/switch_power_mode.py
@@ -57,6 +57,20 @@ def set_power_profile(profile):
         ) from e
 
 
+@contextlib.contextmanager
+def preserve_power_profile():
+    """
+    Rolls back the power profile to the original before calling
+    """
+    sysfs_root = Path("/sys/firmware/acpi/")
+    profile_path = sysfs_root / "platform_profile"
+    old_profile = get_sysfs_content(profile_path)
+    try:
+        yield
+    finally:
+        set_power_profile(old_profile)
+
+
 def main():
     """main function to switch the power mode."""
     sysfs_root = Path("/sys/firmware/acpi/")
@@ -64,10 +78,7 @@ def main():
     profile_path = sysfs_root / "platform_profile"
 
     # use a context manager to ensure the original power mode is restored
-    with contextlib.ExitStack() as stack:
-        # Read the current power mode from /sys/firmware/acpi/platform_profile
-        old_profile = get_sysfs_content(profile_path)
-        stack.callback(set_power_profile, old_profile)
+    with preserve_power_profile():
 
         choices = get_sysfs_content(choices_path).split()
 
--- a/providers/base/bin/switch_power_mode.py
+++ b/providers/base/bin/switch_power_mode.py
@@ -9,6 +9,7 @@
 import os
 from pathlib import Path
 import sys
+import subprocess
 
 
 def main():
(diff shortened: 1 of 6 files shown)


providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,86 @@
...
+    """main function to switch the power mode."""
+    sysfs_root = Path("/sys/firmware/acpi/")
+    choices_path = sysfs_root / "platform_profile_choices"
+    profile_path = sysfs_root / "platform_profile"
+
+    # use a context manager to ensure the original power mode is restored
+    with contextlib.ExitStack() as stack:
+        # Read the current power mode from /sys/firmware/acpi/platform_profile
+        old_profile = get_sysfs_content(profile_path)
+        stack.callback(set_power_profile, old_profile)
```
Hook25, line 70: I would go with something like this instead of `ExitStack` as it is a bit more complicated and the code is less descriptive imo

```suggestion
@contextlib.contextmanager
def preserve_power_profile():
    """
    Rolls back the power profile to the original  before calling
    """
    sysfs_root = Path("/sys/firmware/acpi/")
    profile_path = sysfs_root / "platform_profile"
    original_profile = get_sysfs_content(profile_path)
    try:
        yield
    finally:
        set_power_profile(original_profile)
        

def main():
    """main function to switch the power mode."""
    sysfs_root = Path("/sys/firmware/acpi/")
    choices_path = sysfs_root / "platform_profile_choices"
    profile_path = sysfs_root / "platform_profile"

    # use a context manager to ensure the original power mode is restored
    with preserve_power_profile():
```

> Note: if you use this, double check it runs, I'm editing on GH

providers/base/bin/switch_power_mode.py
```diff
@@ -0,0 +1,86 @@
...
+
+def set_power_profile(profile):
+    """
+    Sets the power profile to the specified value.
+    Args:
+        profile (str): The power profile to set (e.g., "power-saver").
+    Raises:
+        SystemExit: If the power profile could not be set.
+    """
+    profile = "power-saver" if profile == "low-power" else profile
```
Hook25, line 51: Where does this come from?

Hook25 (approved):
Please consider changing that ExitStack call, beside that everything here works for me.
Can you write why that replacement of profile for low power? I don't really get it

**NOTE:** Pull before committing or pushing anything to your local branch, I have rebased this PR because Codecov was drunk!

providers/base/bin/switch_power_mode.py
binli: Thank you, I've fixed this part.

providers/base/bin/switch_power_mode.py
binli: The value in /sys and powerprofile is different, so I need convert it before set mode with powerprofilectl.

```
$ cat /sys/firmware/acpi/platform_profile_choices 
low-power balanced performance

$ powerprofilesctl list 
  performance:
  balanced:
* power-saver:
 ```

Comment (binli) [reply providers/base/bin/switch_power_mode.py:53]: I added a comment above this code.
```
diff --git a/providers/base/bin/switch_power_mode.py b/providers/base/bin/switch_power_mode.py
index 81811fa31..e17073885 100755
--- a/providers/base/bin/switch_power_mode.py
+++ b/providers/base/bin/switch_power_mode.py
@@ -48,6 +48,8 @@ def set_power_profile(profile):
     Raises:
         SystemExit: If the power profile could not be set.
     """
+    # In sys file the modes are low-power, balanced, or performance
+    # but powerprofilesctl only accepts power-saver, balanced or performance
     profile = "power-saver" if profile == "low-power" else profile
     try:
         subprocess.check_call(["powerprofilesctl", "set", profile])
```

Hook25 (approved):

Issue #1114, opened 2024-03-27
`bluetooth4/beacon_eddystone_url_*` failed


hanhsuan:
### Bug Description

While executing `bluetooth4/beacon_eddystone_url_*`, the test will fail due to `TypeError: main() missing 1 required positional argument: 'argv'`. This should relate to #1086.

[submission](https://certification.canonical.com/hardware/202401-33397/submission/362062/test/76077/result/40043054/)

### To Reproduce

1. execute `bluetooth4/beacon_eddystone_url_*` test case

### Environment

- OS: jammy
- Checkbox type: debian
- checkbox version: 4.0.0.dev194

### Relevant log output

```shell
Traceback (most recent call last):
  File "/usr/bin/checkbox-support-eddystone_scanner", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 69, in _f
    return run_with_timeout(f, timeout_s, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 56, in run_with_timeout
    raise exception_queue.get()
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 43, in _f
    result_queue.put(f(*args, **kwargs))
TypeError: main() missing 1 required positional argument: 'argv'
```

Hook25:
thank you for the report, I know what is the cause and I will fix it as soon as possible

PR #1118, opened 2024-03-27
Set default argv (bugfix)

Hook25:
## Description

When a console script is installed by python, it is automatically given a wrapper main that looks something like this:
```python
#!/home/h25/prj/canonical/venv/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from checkbox_support.scripts.eddystone_scanner import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

This is a problem when main expects argv as a parameter (or expects to be called by the statement under `if __name__ == "__main__"` for that matter. 

This fixes our eddystone-beacon scriptlet by setting the default value of argv to what main sets it to.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/1114

## Documentation

N/A

## Tests

Tested locally I was able to reproduce the issue without the patch and see the fix after the patch

### DIFF
--- a/checkbox-support/checkbox_support/scripts/eddystone_scanner.py
+++ b/checkbox-support/checkbox_support/scripts/eddystone_scanner.py
@@ -71,7 +71,9 @@ def beacon_scan(hci_device):
 
 
 @timeout(60 * 10)  # 10 minutes timeout
-def main(argv=sys.argv[1:]):
+def main(argv=None):
+    if argv is None:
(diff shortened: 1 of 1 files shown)


kissiel (dismissed):
TY! +1

kissiel (approved):
Still +1

Issue #1115, opened 2024-03-27
Cannot get config variables in checkbox.conf with local run


LiaoU3:
### Bug Description

When local run, checkbox is not able to get the config variable. This is critical issue because we are not able to test the DUT which can not connect to the network (ethernet, wireless, etc.)
![Screenshot from 2024-03-27 02-06-51](https://github.com/canonical/checkbox/assets/58060146/d6e553b0-39f7-46d4-83cf-43666e96b638)

### To Reproduce

1. Install checkbox
   1. sudo snap install checkbox22 --beta
   2. sudo snap install checkbox --beta --classic
2. List checkbox.conf make sure it is there
   1. sudo checkbox.configure -l
3. Make sure there is an incompleted session
   1. checkbox.checkbox-cli
   2. Select [client-cert-desktop-22-04-stress](https://github.com/canonical/checkbox/blob/11a42fadbae0518c192f375ae526ec2dc13202e0/providers/certification-client/units/client-cert-desktop-22-04.pxu#L161)
   3. Check if boot count for coldboot and warmboot is 100
   4. ctr + c to interrupt
5. Local run checkbox again
   1. checkbox.checkbox-cli
   2. Select [client-cert-desktop-22-04-stress](https://github.com/canonical/checkbox/blob/11a42fadbae0518c192f375ae526ec2dc13202e0/providers/certification-client/units/client-cert-desktop-22-04.pxu#L161)
   3.. You will see that now the boot count is 2 (which is the default value when "$STRESS_BOOT_ITERATIONS" not assigned)

### Environment

 - OS: Desktop 22.04
 - Type: Snap
 - Version: 4.0.0-dev194

### Relevant log output

```shell
ubuntu@kria:~$ sudo checkbox.configure -l
ALSADEVICE=hw:CARD=bytrt5660,DEV=0
MODEL_GRADE=signed
NET_DEVICE_INFO=ven_rsi_sdio ven_rsi_91x
OBEX_TARGET=00:02:72:C5:F9:1F
OPEN_AC_SSID=ubuntu-cert-ac-open
OPEN_AX_SSID=ubuntu-cert-ax-open
OPEN_BG_SSID=ubuntu-cert-bg-open
OPEN_N_SSID=ubuntu-cert-n-open
STRESS_BOOT_ITERATIONS=100  <------------- Here is 100, but in local run is 2 (default)
STRESS_BOOT_WAIT_DELAY=120
STRESS_BOOT_WAKEUP_DELAY=60
[...]

PR #1119, opened 2024-03-27
Fix config loading after resume (bugfix)

Hook25:
## Description

When resuming a session, checkbox local doesn't currently load any config. This is an issue for any job that relies of a config (for example, the environment), so it must be fixed.

This introduces basically the same mechanism that remote uses (storing the launcher in the app blob, loading it and reading the configs on the machine after resume, using the alternative config).

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/1115
Fixes: CHECKBOX-1340

## Documentation

New function is documented

## Tests

unit tests and metabox tests added


kissiel (approved):
Looks great! +1

PR #1123, opened 2024-03-29
Add checking RAID mode test (New)

stanley31huang:
## Description
The purpose for this PR is validate the RAID mode on the system by checking /proc/mdstat and dump the detail information for each RAID device.

## Resolved issues
N/A

## Documentation
N/A

## Tests
Tested on EGW5200 (an RAID1 is created on BIOS): https://certification.canonical.com/hardware/202307-31837/submission/362442/


providers/base/units/disk/manifest.pxu
```diff
@@ -0,0 +1,4 @@
+unit: manifest entry
+id: has_md_raid
+_name: The Multiple devices (Software RAID)
```
pieqq, line 3: "Multiple devices" is too generic. Software RAID stands on its own.

```suggestion
_name: Software RAID
```

providers/base/bin/check_software_raid.py
```diff
@@ -0,0 +1,143 @@
...
+
+def dump_raid_info(nodes: list) -> None:
+    """
+    Display detail information for all of RAID devices
+
+    Args:
+        nodes (list): the name of MD devices
+    """
+    for node in nodes:
+        subprocess.run(shlex.split("mdadm -D /dev/{}".format(node)))
```
pieqq, line 69: If you want to test this function, you can mock `subprocess`, call `dump_raid_info` with a list of 2 nodes, and check that `subprocess.run` was called twice, for instance.

pieqq (changes requested):
Nice stuff!

You've added `mdadm` as a requirement for the snaps, but not for the Debian packages. Please add a packaging meta data unit to install mdadm when using the Debian version (see [example](https://github.com/canonical/checkbox/blob/main/providers/base/units/camera/packaging.pxu#L23-L26))

For the test data, I would rename it as `*.txt` files instead of `*.log` files, as they are not really logs.

Finally, a few comments inline.

providers/base/bin/check_software_raid.py
stanley31huang: added a unit test for dump_raid_info function, thanks.

pieqq (approved):
+1. Thanks!

Issue #1029, opened 2024-02-29
usb/insert job get error when no `udisks2`


LiaoU3:
### Bug Description

usb/insert related job will get error when there's no package `udisks2`.
I have 2 solutions to this issue.
1. Package `udisks2` into checkbox snap.
2. Add require `package.name == 'udisks2'` and add flag `fail-on-resource`

### To Reproduce

1. Prepare a DUT with image which does not have  `udisks2`.
2. Install checkbox
   1. sudo snap install checkbox22 --beta
   2. sudo snap install checkbox --classic --beta
3. checkbox.checkbox-cli run com.canonical.certification::usb/insert

### Environment

- OS: Ubuntu Desktop
- Checkbox Type: Snap
- Checkbox Version: 3.3.0-dev19
- Hardware: KD240 ([202307-31906](https://certification.canonical.com/hardware/202307-31906/))

### Relevant log output

```shell
ubuntu@kria:~$ checkbox.checkbox-cli run com.canonical.certification::usb/insert
$PROVIDERPATH is defined, so following provider sources are ignored ['/home/ubuntu/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop']
===========================[ Running Selected Jobs ]============================
==============[ Running job 1 / 1. Estimated time left: 0:02:00 ]===============
-----------------[ USB 2.0 storage device insertion detected ]------------------
ID: com.canonical.certification::usb/insert
Category: com.canonical.plainbox::usb
Purpose:

Check system can detect USB 2.0 storage when inserted.
NOTE: Make sure the USB storage device has a partition before starting
the test.

Steps:

1. Press continue
2. Wait until the message "INSERT NOW" is printed on the screen
3. Connect USB 2.0 storage device

Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]:
... 8< -------------------------------------------------------------------------


INSERT NOW


Traceback (most recent call last):
  File "/snap/checkbox22/current/usr/lib/python3/dist-packages/dbus/bus.py", line 177, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/snap/checkbox22/current/usr/lib/python3/dist-packages/dbus/bus.py", line 361, in get_name_owner
    return self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
[...]

PR #1130, opened 2024-04-01
Added udisks2 requirement (BugFix)

fernando79513:
## Description

Following the second suggestion of #1029, I have add a requirement for the systems that are running the `usb/insert`, so the `udisks2` package has to be installed for the test to be executed.  

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1279

## Documentation

There are no changes on the documentation.

## Tests

1. Prepare a DUT with image which does not have udisks2.
2. Install checkbox
```
 sudo snap install checkbox22 --beta
 sudo snap install checkbox --classic --beta
```
3. Run the test
```
 checkbox.checkbox-cli run com.canonical.certification::usb/insert
```

The test should fail with an error message:
```
==============[ Running job 2 / 2. Estimated time left: 0:02:00 ]===============
-----------------[ USB 2.0 storage device insertion detected ]------------------
ID: com.canonical.certification::usb/insert
Category: com.canonical.plainbox::usb
 - resource expression "package.name == 'udisks2'" evaluates to false
Outcome: job failed
package.name == 'udisks2'
False
```

### DIFF
--- a/providers/base/units/usb/usb.pxu
+++ b/providers/base/units/usb/usb.pxu
@@ -54,7 +54,7 @@ id: usb/insert
 flags: also-after-suspend fail-on-resource
 template-engine: jinja2
 requires:
- package.name == 'udisks2'
+ package.name == 'udisks2' or snap.name == 'udisks2'
 _summary: USB 2.0 storage device insertion detected
 _purpose:
  Check system can detect USB 2.0 storage when inserted.


Hook25 (approved):
+1, as per discussion, it makes sense for now to go this way, as bundling udisk2 in the snap opens the gate for quite a few more problems. We can go back to this if it causes any issue

Hook25 (approved):

PR #1133, opened 2024-04-02
Correct desktop 2404 suspend cycles plan (BugFix)

seankingyang:
## Description
Correct the desktop 24.04 suspend cycles plan.
It should use the new suspend cycles test plan.


scotthu27 (approved):
LGTM~

Issue #1054, opened 2024-03-12
sleepgraph suspend without notifying systemd


KaiChuan-Hsieh:
### Bug Description

The sleepgraph utilize $ echo "mem" > /sys/power/state to trigger suspend, it will impact Nvidia suspend/resume function, since the NV driver has system services nvidia-suspend.service and nvidia-resume.serivce to handle suspend/resume for the NV card, the test job should be adjusted for not running on system with NV cards.

### To Reproduce

it is a better approach to simulate user setup.

### Environment

OS: ubuntu 22.04.4
Checkbox Type: Debian
Checkbox Version: 4.0.0
Hardware being tested: Hardware with NV

PR #1136, opened 2024-04-02
Limit the platform to perform the sleepgraph test (BugFix)

KaiChuan-Hsieh:
Limit the platform to perform the sleepgraph test (BugFix) (#1054)

* Filter cpuinfo to GenuineIntel only

* Align the sleep delay time with Windows, https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/modern-standby-stress-and-long-duration-validation#test-operation

## Description

- sleepgraph trigger suspend by utilizing sysfs node without notifying systemd (an [upstream bug has been opened for this](https://github.com/intel/pm-graph/issues/28))
- However, Nvidia uses a userspace daemon that relies on systemd status, so triggering suspend using sysfs directly may lead to issues on devices with Nvidia graphics cards
- Align the sleep delay with Windows

## Resolved issues

https://github.com/canonical/checkbox/issues/1054

## Documentation

https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/modern-standby-stress-and-long-duration-validation#test-operation

## Tests

Verify the test by running the plan

$ checkbox-cli run com.canonical.certification::stress-pm-graph

### DIFF
--- a/providers/base/units/stress/jobs.pxu
+++ b/providers/base/units/stress/jobs.pxu
@@ -607,51 +607,67 @@ plugin:shell
 category_id: com.canonical.plainbox::stress
 id: stress/s2idle_pm-graph_30
 estimated_duration: 10m
-requires: 
+requires:
+ cpuinfo.type == 'GenuineIntel'
  executable.name == 'sleepgraph'
  sleep.mem_sleep == 's2idle'
 user: root
 _summary: Resume from idle by using Intel pm-graph
 command:
- sleepgraph -m freeze -rtcwake 10 -sync -gzip -multi 30 0 -skiphtml -o "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph/s2idle-"$(date -d today +%Y-%m-%d-%H%M)"
+ if [ ! -f /proc/driver/nvidia/suspend ]; then
+     sleepgraph -m freeze -rtcwake 60 -sync -gzip -multi 30 30 -skiphtml -o "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph/s2idle-"$(date -d today +%Y-%m-%d-%H%M)"
+ else
+     echo "The platform has Nvidia driver loaded, which is not capable of running sleepgraph. Exiting..."
+ fi
 
 plugin: attachment
 category_id: com.canonical.plainbox::stress
 id: stress/s2idle_pm-graph_30.tar.xz
 estimated_duration: 1
-requires: 
+requires:
+ cpuinfo.type == 'GenuineIntel'
  sleep.mem_sleep == 's2idle'
 after:
  stress/s2idle_pm-graph_30
 user: root
 _summary: Attach pm-graph logs (s2idle)
 command:
- tar Jcf "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph_30.tar.xz "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph && cat "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph_30.tar.xz
+ if [ -d "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph ]; then
+     tar Jcf "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph_30.tar.xz "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph && cat "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph_30.tar.xz
+ fi
 
 plugin:shell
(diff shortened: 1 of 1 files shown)


providers/base/units/stress/jobs.pxu
```diff
@@ -603,51 +603,77 @@ _description:
...
      Verify the system is not frozen and the wifi and bluetooth applets are still visible and functional
 
+unit: template
+template-resource: cpuinfo
+template-filter: cpuinfo.type == 'GenuineIntel'
+template-unit: job
 plugin:shell
 category_id: com.canonical.plainbox::stress
-id: stress/s2idle_pm-graph_30
+id: stress/s2idle_pm-graph_30_{model_number}_{model_version}_{model_revision}
```
pieqq, line 612: What's the purpose for using a template for this job?

pieqq (commented):
Question inline.

I found the [internal issue](https://bugs.launchpad.net/somerville/+bug/1999593) where this was discussed, which brought some clarification. I'll update the bug description to match this, and also I think when this lands, the commit message should include this info.

providers/base/units/stress/jobs.pxu
KaiChuan-Hsieh: The purpose is to make the job be seen on Intel platform only.

pieqq: Why not using

```
requires:
  cpuinfo.type == 'GenuineIntel'
```

instead? This way, you don't need a template and the job will be skipped on every platform except Intel ones.

KaiChuan-Hsieh:
@pieqq 

I intend to make the platform other than Intel can't see the job, is it not allowed? It is why I use template, I don't know which is better.

pieqq:
@KaiChuan-Hsieh it's one way to do it, indeed, but we usually just go with the `requires` section instead, since we don't mind having skipped tests in our submissions. This is why you will see a lot of `requires` sections with a condition on `cpuinfo.*` in [the CPU jobs](https://github.com/canonical/checkbox/blob/main/providers/base/units/cpu/jobs.pxu), for instance.

KaiChuan-Hsieh:
@pieqq 

Hello,
I got my PR updated, would you please check it again?

Thanks,

providers/base/units/stress/jobs.pxu
```diff
@@ -607,19 +607,26 @@ plugin:shell
...
  sleep.mem_sleep == 's2idle'
 user: root
 _summary: Resume from idle by using Intel pm-graph
 command:
- sleepgraph -m freeze -rtcwake 10 -sync -gzip -multi 30 0 -skiphtml -o "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph/s2idle-"$(date -d today +%Y-%m-%d-%H%M)"
+ if [ ! -f /proc/driver/nvidia/suspend ]; then
+     sleepgraph -m freeze -rtcwake 60 -sync -gzip -multi 30 30 -skiphtml -o "$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph/s2idle-"$(date -d today +%Y-%m-%d-%H%M)"
+ else
+     echo "The platform has Nvidia driver loaded, which is not capable of running sleepgraph"
+     exit 1
```
pieqq, line 621:
```suggestion
     echo "The platform has Nvidia driver loaded, which is not capable of running sleepgraph. Exiting..."
```

If you `exit 1`, the job will be marked as failed on all the devices with a Nvidia GPU, which is probably not what you want. Just `echo` something clear for the reviewer and you're good to go.

providers/base/units/stress/jobs.pxu
```diff
@@ -607,19 +607,26 @@ plugin:shell
...
+ fi
 
 plugin: attachment
```
pieqq, line 624: Have you tried running this job with the modification you made to `stress/s2idle_pm-graph_30` on a device with a Nvidia GPU? `stress/s2idle_pm-graph_30` will not run sleepgraph, so nothing will be generated and the attachment job will complain it doesn't find the directory you're talking about.

You probably need to modify the `command` to check for the existence of `"$PLAINBOX_SESSION_SHARE"/s2idle_pm-graph` first.

pieqq (changes requested):
I wanted to find a better way to skip these jobs if a Nvidia GPU is detected, but the resource expressions in Checkbox are very tricky, and I couldn't find something simple.

That said, I made a comment inline and a suggestion. You will need to apply them to both `stress/s3_pm-graph_30` and `stress/s2idle_pm-graph_30` jobs and their related attachment jobs.

pieqq (approved):
Thanks!

PR #1140, opened 2024-04-02
Fix config loading local when resumable session available (Bugfix)

Hook25:
## Description

When Checkbox has a manually resumable session or an incompatible automatically resumable session on the machine, it is unable to use any configuration. This is because any loaded configuration is "used" by the session assistant created when Checkbox is launched that is then discarded when probing to see if a session is automatically resumable. This fixes the issue by postponing the `use_alternate_configuration` call.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/1115
Relates to: https://github.com/canonical/checkbox/pull/1119

## Documentation

N/A

## Tests

This adds a new metabox scenario that fails in main and passes after the patch.


kissiel (approved):
Awesome, thank you!

kissiel (approved):
round 2...

PR #1141, opened 2024-04-03
Improve the 24.04 IoT Cert Test Plans (New)

baconYao:
## Documentation
  - [OQ012 - IoT Checkbox Certification Test Plans for 24.04](https://docs.google.com/document/d/19sXrqIR6PkGtR6TAV_avf2Jt8YqE3Pi4DRWZtMb5Jnw/edit?usp=sharing)
  - [OEMQA-4157 Improve the scope and update the IoT Certify Test Plans](https://warthogs.atlassian.net/browse/OEMQA-4157)

### DIFF
--- a/providers/certification-client/units/client-cert-iot-ubuntucore-24.pxu
+++ b/providers/certification-client/units/client-cert-iot-ubuntucore-24.pxu
@@ -35,25 +35,25 @@ bootstrap_include:
   interface
   net_if_management
 nested_part:
-  submission-cert-full
-  self-manual
-  ubuntucore-manual
   audio-manual
-  bluetooth-manual
+  bluetooth-cert-manual
   camera-manual
   cpu-manual
   disk-manual
   ethernet-manual
+  led-indicator-manual
   led-manual
   mediacard-manual
   memory-manual
-  monitor-manual
   networking-manual
   rtc-manual
+  self-manual
   serial-manual
   snappy-snap-manual
   socketcan-manual-remote
   socketcan-manual-local
+  submission-cert-full
+  ubuntucore-manual
   usb-manual
   usb-c-manual
   usb3-manual
@@ -63,9 +63,9 @@ nested_part:
   wireless-wifi-master-mode-manual
   wwan-manual
   after-suspend-audio-manual
-  after-suspend-bluetooth-manual
+  after-suspend-bluetooth-cert-manual
   after-suspend-ethernet-manual
-  after-suspend-monitor-manual
+  after-suspend-led-indicator-manual
   after-suspend-networking-manual
   after-suspend-serial-manual
   after-suspend-socketcan-manual-remote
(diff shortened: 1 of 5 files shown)


providers/certification-client/units/client-cert-iot-ubuntucore-24.pxu
```diff
@@ -35,25 +35,25 @@ bootstrap_include:
...
+  bluetooth-cert-manual
   camera-manual
   cpu-manual
   disk-manual
   ethernet-manual
+  led-indicator-manual
   led-manual
   mediacard-manual
   memory-manual
-  monitor-manual
```
rickwu666666, line 50: Why we don't need monitor manual test for Ubuntu Core? Some of the IoT devices do have display support such as HDMI and DP.

providers/certification-client/units/client-cert-iot-ubuntucore-24.pxu
```diff
@@ -77,6 +77,9 @@ nested_part:
   after-suspend-wireless-manual
   after-suspend-wireless-wifi-master-mode-manual
   after-suspend-wwan-manual
+exclude:
+  bluetooth4/HOGP-mouse
+  after-suspend-bluetooth4/HOGP-mouse
```
rickwu666666, line 82: Why we don't need to test the BT mouse? I suggest that if we are nesting the same plan as PC nesting, we could retain the same coverage.

rickwu666666 (commented):
Hi @baconYao,
Please see my in-line comments. Thanks.

providers/certification-client/units/client-cert-iot-ubuntucore-24.pxu
baconYao: Core and Server images don't have GUI by default, it means you cannot see the cursor even the BT mouse be connected.

baconYao:
Hi @yphusm, could you merge this PR if  you don't have any concern?

linjuiche:
@yphus @jocave 
We expect this could also replace the test plan for the ODM program as well.

pieqq (dismissed):
Thanks for working on this!

Let's land it as is, we can always investigate the use of template ids in the nested parts, and their impact regarding the need for `after-suspend-*` counterparts.

pieqq:
After a second look, I have a question:

Is there any reason to rewrite almost the same test plans twice?

For instance, when I compare  `client-cert-iot-ubuntucore-24-manual` with `client-cert-iot-server-24-04-manual`, and `client-cert-iot-ubuntucore-24-automated` with `and client-cert-iot-server-24-04-automated`, they look very similar.

Maybe one should still be based on the other (as a nested part), with the tests you don't want excluded?

(this also applies to the new Desktop IoT test plan: maybe we should base everything – UC and Desktop – on Server and exclude/include)

PR #1142, opened 2024-04-03
Fix the audio_after_suspend test failing on mantic (BugFix)

hanhsuan:
## Description

The device id in the output of `wpctl status` will be changed after suspend since kernel 6.5.0.26 with dGPU installed  .
Therefore, changing shell script to python to make the parsing easier and more robust.  

Test cases change will be addressed in other PR.

## Resolved issues

as described in the [card](https://warthogs.atlassian.net/browse/CHECKBOX-1282)

## Tests

Unit test

### DIFF
--- a/providers/base/bin/pipewire_utils.py
+++ b/providers/base/bin/pipewire_utils.py
@@ -76,7 +76,7 @@ class PipewireTest:
         :param media_class: sink(s) or source(s)
         :type media_class: str
 
-        :returns: "Ouput", "Input" or "UNKNOWN CLASS"
+        :returns: "Output", "Input" or "UNKNOWN CLASS"
         :"rtype": str
         """
         if media_class.lower() in ["sink", "sinks"]:
@@ -183,7 +183,7 @@ class PipewireTest:
         :type media_class: str
 
         :param device: device type, such as hdmi, usb and bluez etc.
-        :type devide: str
+        :type device: str
         """
         mclass = self.generate_pw_media_class(media_type, media_class)
         if mclass in ["UNKNOWN CLASS", "UNKNOWN TYPE"]:
@@ -255,7 +255,7 @@ class PipewireTest:
                             self.logger.info(
                                     "[ Audio sink ]".center(80, '='))
                             self.logger.info(
-                                    "Device: [{}] availavle: [{}]"
+                                    "Device: [{}] available: [{}]"
                                     .format(route["description"],
                                             available))
                             return True
@@ -276,7 +276,7 @@ class PipewireTest:
         :type timeout: int
 
         :param device: device type, such as hdmi etc.
-        :type devide: str
+        :type device: str
         """
         if device:
             if not self._check_state(device):
@@ -409,7 +409,7 @@ class PipewireTest:
         """
         Get node description from the output of wpctl inspect
 
-        :param properties: output of wpctl inxpect
+        :param properties: output of wpctl inspect
         :type properties: str
 
         :returns: the node description
--- a/providers/base/bin/pipewire_utils.py
+++ b/providers/base/bin/pipewire_utils.py
@@ -465,21 +465,21 @@ class PipewireTest:
                    " ├─ Sink endpoints:", " ├─ Sources:",
                    " ├─ Source endpoints:", " └─ Streams:",
                    " └─ Default Configured Node Names:"]
-        storted_lines = []
+        sorted_lines = []
         sub_items = []
         for line in lines:
             clean_line = re.sub(", pid:.*| [0-9]+\\. ", "", line)
             # only sort needed items
             if clean_line in catalog:
                 # found next catalog, append sorted sub_items
-                [storted_lines.append(item) for item in sorted(sub_items)]
+                [sorted_lines.append(item) for item in sorted(sub_items)]
                 # clean up after append
                 sub_items = []
-                storted_lines.append(clean_line)
+                sorted_lines.append(clean_line)
             else:
                 sub_items.append(clean_line)
-        [storted_lines.append(item) for item in sorted(sub_items)]
-        return storted_lines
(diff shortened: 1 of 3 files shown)


providers/base/bin/pipewire_utils.py
```diff
@@ -533,6 +603,11 @@ def _args_parsing(self, args=sys.argv[1:]):
                 "-t", "--type", type=str, required=True,
                 help="VIDEO or AUDIO")
 
+        # Add parser for show current status function
+        parser_show = subparsers.add_parser(
+                'status',
+                help='show pipewire current status from wpctl status')
+
```
fernando79513, line 610: Instead of having a function that prints the modified output of wpctl status, I think it would be easier to just compare the output of two raw statuses, that way, you won't have to modify much of the checkbox job and you could get rid of the diff from the CLI:
```
        # Add parser to compare to statuses
        parser_status = subparsers.add_parser(
                'compare_status',
                help='compare the output from two wpctl status')
        parser_status.add_argument(
                "-s1", "--status1", type=str,
                help="path to the second output of wpctl status")
        parser_status.add_argument(
                "-s2", "--status1", type=str,
                help="path to the second output of wpctl status")
```

In the job you would have something like this:

```
plugin: shell
category_id: com.canonical.plainbox::suspend
id: suspend/audio_before_suspend
...
command:
  if check_audio_daemon.sh ; then
    wpctl status > "$PLAINBOX_SESSION_SHARE"/audio_settings_before_suspend
```

```
plugin: shell
category_id: com.canonical.plainbox::suspend
id: suspend/audio_after_suspend
...
command:
  if check_audio_daemon.sh ; then
      wpctl status > "$PLAINBOX_SESSION_SHARE"/audio_settings_after_suspend
      pipewire_utils.py compare_status -s1 "$PLAINBOX_SESSION_SHARE"/audio_settings_after_suspend -s2 "$PLAINBOX_SESSION_SHARE"/audio_settings_after_suspend
  else
      audio_settings.py store --file="$PLAINBOX_SESSION_SHARE"/audio_settings_after_suspend
      diff "$PLAINBOX_SESSION_SHARE"/audio_settings_before_suspend "$PLAINBOX_SESSION_SHARE"/audio_settings_after_suspend
  fi
```

providers/base/tests/test_pipewire_utils.py
```diff
@@ -20,11 +20,44 @@
...
 sys.modules["gi"] = MagicMock()
 sys.modules["gi.repository"] = MagicMock()
-from unittest.mock import patch
 from pipewire_utils import *
+from io import StringIO
+
+
+class RunCommandTests(unittest.TestCase):
+
+    @patch("subprocess.check_output")
```
fernando79513, line 32: This won't be necessary if we just add the compare_status function to pipewire_utils.py

providers/base/bin/pipewire_utils.py
```diff
@@ -68,6 +69,25 @@ class PipewireTest:
...
+        """
+        try:
+            return subprocess.check_output(cmd,
+                                           shell=shell,
+                                           universal_newlines=True)
+
+        except (subprocess.CalledProcessError, FileNotFoundError) as e:
+            raise SystemExit("Running command:{} failed due to {}"
+                             .format(cmd, repr(e)))
+
```
fernando79513, line 90: This won't be necessary if we just add the compare_status function to pipewire_utils.py

providers/base/bin/pipewire_utils.py
```diff
@@ -452,6 +472,56 @@ def show_default_device(self, device_type):
...
+        :returns: is sorted and printed or not
+        """
+        if len(lines) > 0:
+            sorted_lines = sorted(lines)
+            for e in sorted_lines:
+                # logger couldn't pipe to file
+                print("{}".format(e))
+            return True
+        else:
+            return False
```
fernando79513, line 490: If we just do the comparison, we can get rid of the print and simplify this part.

fernando79513 (changes requested):
Thanks for taking care of this fix, good work here!
I have some comments, and I think we should also have to modify the job in `suspend.pxu` so it uses the new changes.

fernando79513:
It is not directly related to this PR, but when running the tests I noticed that `test_pipewire_test.py` creates a new file each time you run it. I think it will be fixed if you change the [`test_with_spectrum`](https://github.com/canonical/checkbox/blob/main/providers/base/tests/test_pipewire_test.py#L360) test
```
    @patch("pipewire_test.FileDumper.write_to_file")
    def test_with_spectrum(self, mock_wpctl, mock_fd):
        mock_fd.return_value = True
        self.assertEqual(1,
                         run_test(process_arguments(["-u", "u", "--quiet"])))
```

providers/base/bin/pipewire_utils.py
```diff
@@ -68,6 +69,25 @@ class PipewireTest:
...
 
+    def _run_command(self, cmd: dict, shell=False) -> str:
+        """
+        use subprocess.check_output to execute command
+
+        :param cmd: the command will be executed
+
+        :param shell: enbale shell or not
+
+        :returns: Ouput of command
```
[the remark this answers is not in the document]
fernando79513: There are some other spell issues on the test, maybe it is worth using a spell checker to fix them.

providers/base/bin/pipewire_utils.py
hanhsuan: This will be in another PR.

fernando79513 (changes requested):
Thank you for the changes!
I added one suggestion and pleas fix the typos in pipewire_utils.py:
```
79          :returns: "Output", "Input" or "UNKNOWN CLASS"
186         :type device: str
258                                     "Device: [{}] available: [{}]"
279        :type devide: str
412        :param properties: output of wpctl inspect

fernando79513 (approved):
LGTM +1

PR #1144, opened 2024-04-03
Move jobs back to paid runners (infra)

Hook25:
## Description

Now that `small` self hosted workers are gone, we were told to move every small workflow back to github workers. This PR does that.

We are not moving back daily builds because we need the 10h timeout and it is not possible to increase it above 6 for paid runners.

Minor: this removes outdated "beta release" jobs (old ones where we re-built for beta on `v` tag push 

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


kissiel (approved):
Let's be prepared to roll that back as soon as we notice any lags.

PR #1145, opened 2024-04-03
Check power mode before and after suspend (New)

binli:
## Description

## Resolved issues

## Documentation

This testcase could help verify power mode can stay in the chosen setting after suspend.

## Tests

```
==============[ Running job 8 / 8. Estimated time left: 0:00:01 ]===============
----------------------[ suspend/power_mode_after_suspend ]----------------------                                                        ID: com.canonical.certification::suspend/power_mode_after_suspend
Category: com.canonical.plainbox::suspend                                                                                               
1c1                                                                 
< performance                                                       
---                                                                 
> low-power                                                         
Outcome: job failed                                                                                                                     
Finalizing session that hasn't been submitted anywhere: checkbox-run-2024-04-03T16.03.02
 ☑ : Collect information about kernel modules                                                                                            ☑ : Discover if the system supports power modes via acpi
 ☑ : Dumps power_mode info to a file for comparison after suspend
 ☑ : Create resource info for supported sleep states
 ☑ : Creates resource info for RTC                                  
 ☑ : Automated test of suspend function 
 ☒ : suspend/power_mode_after_suspend
```


providers/base/units/suspend/suspend.pxu
Hook25, line 1: This works but it doesn't do what you think it will. The issue here is that while the after_suspend job runs after suspend (due to the dependency), the before suspend may itself run after suspend, making the job pointless. Recently we have made this change https://github.com/canonical/checkbox/pull/1037, so you can use the `also-after-suspend` flag to get the result you want (I see that this is a specific suspend job, so idk if it is the same for you as the normal suspend job).

Hook25 (changes requested):
The suspend story is sadly still to be defined, so this doesn't do what you think it will do. Please consider using one job that has an also-after-suspend flag and:
1. if the power mode file doesn't exist, it creates it
2. if the power mode file exists, it diffs it

This doesn't 100% work (as re-runs are a thing), but is more correct than what this PR would yield

PR #1164, opened 2024-04-09
Fix timeout children (bugfix)

Hook25:
## Description

When a process spawns childrens and it is killed, if the childrens were non-deamons they are re-parented by the super reaper to the closest relative, falling back on init if none are found. This behaviour is not what we want for our timeout decorator, namely, when a function times out, all processes spawned by the decorated functions must be killed. This updates the decorator so that this happens.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1250

## Documentation

Updated docstring with the new limitations and features

## Tests

All tests are still passing

### DIFF
--- a/checkbox-support/checkbox_support/helpers/timeout.py
+++ b/checkbox-support/checkbox_support/helpers/timeout.py
@@ -75,14 +75,3 @@ def timeout(timeout_s):
         return _f
 
     return timeout_timeout_s
-
-
-@timeout(1)
-def main():
-    import subprocess
-
-    subprocess.check_call(["sleep", "10000"])
-
-
-if __name__ == "__main__":
-    main()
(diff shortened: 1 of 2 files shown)


checkbox-support/checkbox_support/helpers/timeout.py
```diff
@@ -71,3 +75,14 @@ def _f(*args, **kwargs):
         return _f
 
     return timeout_timeout_s
+
+
+@timeout(1)
+def main():
+    import subprocess
+
+    subprocess.check_call(["sleep", "10000"])
```
kissiel, line 84: leftovers?

kissiel (changes requested):
some leftovers left over :)

checkbox-support/checkbox_support/helpers/timeout.py
Hook25: fixed

kissiel (approved):
Nice battle.
Happy to see us conquering this problem.

PR #1165, opened 2024-04-09
Fetch last passing daily build before rebuilding (infra)

Hook25:
## Description

The daily build workflow always rebuilds if there were changes in the last 24h, this implies that:
- If a daily build fail, the next one doesn't take this fact into consideration, leading to some daily builds being skipped while it would be beneficial if they actually did run
- If a daily build is triggered manually, the next one may be reduntant and lead to failures (for example, duplicated deb source uploads for the same version get automatically rejected)

With this change we will now fetch the last time the "daily-builds.yaml" workflow passed, and use that as the since barrier to see if we should re-build or not.

> Note: this does not 100% solve the duplicated source uploads for debs. If the deb workflow fails after uploading, this will lead to the same rejection but this is not a problem of the trigger (that will now be correct) rather the deb workflow, that should check if the source is in the ppa and put an `~rc` postfix (or something to that effect) to the source version

## Resolved issues

N/A

## Documentation

N/A

## Tests

Tested this offline, this  currently yields true (as would the 24 hours check) but now with the correct number of commits 6 instead of 3. This is because yesterday's daily build failed, so today we must take that into consideration

### DIFF
--- a/.github/workflows/daily-builds.yml
+++ b/.github/workflows/daily-builds.yml
@@ -17,8 +17,11 @@ jobs:
           fetch-depth: 0
       - name: Check for commits
         id: commit_check
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


kissiel (approved):
TYVM! +1

PR #1166, opened 2024-04-09
Add timeout-minutes to second job in the workflow (infra)

Hook25:
## Description

When adding global `timeout-minutes` I forgot one job. If this is not set the global 6h timeout is enforced instead.

## Resolved issues

Daily build deb build killed after 6h

## Documentation

N/A

## Tests

N/A


kissiel (approved):
hurtful +1

Issue #571, opened 2023-06-20
Checkbox remote will keep start over the whole test plan when running memory stress tests


seankingyang:
### Bug Description

 Using the checkbox remote /testflinger to run the "com.canonical.certification::client-cert-desktop-22-04-stress", when running to the stressor "bigheap" or others, the checkbox master will disconnect with salve. 
 After reconnecting the salve, the master will start over the test plan (not keep going).

### To Reproduce

1. Using checkbox remote to run com.canonical.certification::client-cert-desktop-22-04-stress 

### Environment

- OS: ubuntu 22.04

### Additional context
[Jenkins_log.log](https://github.com/canonical/checkbox/files/11796295/Jenkins_log.log):
```
----------------------------[ Running job 35 / 74 ]-----------------------------
------------------------[ Stress test of system memory ]------------------------
ID: com.canonical.certification::memory/memory_stress_ng
Category: Memory tests
--------------------------------------------------------------------------------
INFO:master:remote: Waiting for job to finish.
stress-ng: info:  [3988613] setting to a 300 second (5 mins, 0.00 secs) run per stressor
stress-ng: info:  [3988613] dispatching hogs: 20 bsearch
stress-ng: info:  [3988613] successful run completed in 300.01s (5 mins, 0.01 secs)
stress-ng: info:  [3988671] setting to a 300 second (5 mins, 0.00 secs) run per stressor
stress-ng: info:  [3988671] dispatching hogs: 20 context
stress-ng: info:  [3988671] successful run completed in 300.01s (5 mins, 0.01 secs)
stress-ng: info:  [3988696] setting to a 300 second (5 mins, 0.00 secs) run per stressor
stress-ng: info:  [3988696] dispatching hogs: 20 hsearch
stress-ng: info:  [3988696] successful run completed in 300.01s (5 mins, 0.01 secs)
stress-ng: info:  [3988763] setting to a 300 second (5 mins, 0.00 secs) run per stressor
stress-ng: info:  [3988763] dispatching hogs: 20 lsearch
stress-ng: info:  [3988763] successful run completed in 300.01s (5 mins, 0.01 secs)
stress-ng: info:  [3988830] setting to a 300 second (5 mins, 0.00 secs) run per stressor
stress-ng: info:  [3988830] dispatching hogs: 20 matrix
stress-ng: info:  [3988830] successful run completed in 300.11s (5 mins, 0.11 secs)
[...]

PR #1167, opened 2024-04-10
Fix checkbox remote running memory stress test (Bugfix)

pseudocc:
## Description

Once a sub process of checkbox-ng service got killed by OOM killer, the service will be stopped, because of `DefaultOOMPolicy=stop`, and the memory stress test will be marked as crashed, or the whole test plan will start over after the service restarts.

This PR explicitly set `OOMPolicy=continue` so the service will be able to remain alive, and checkbox remote could finish the memory stress test.

## Resolved issues

Fixes: #571

> systemd[1]: checkbox-ng.service: A process of this unit has been killed by the OOM killer.
> kernel: oom_reaper: reaped process 908499 (stress-ng-stack), now anon-rss:15720kB, file-rss:704kB, shmem-rss:0kB
> systemd[1]: checkbox-ng.service: State 'final-sigterm' timed out. Killing.
> systemd[1]: checkbox-ng.service: Killing process 908487 (stress-ng) with signal SIGKILL.
> systemd[1]: checkbox-ng.service: Killing process 908488 (stress-ng-stack) with signal SIGKILL.
> systemd[1]: checkbox-ng.service: Killing process 908489 (stress-ng-stack) with signal SIGKILL.

## References

> systemd[1]: checkbox-ng.service: A process of this unit has been killed by the OOM killer.

Logged from [systemd source](https://github.com/systemd/systemd/blob/v249.11/src/core/cgroup.c#L2998).

OOMPolicy part of `man systemd.service`:
```
 OOMPolicy=
     Configure the out-of-memory (OOM) killing policy for the kernel and the
     userspace OOM killer systemd‐oomd.service(8). On Linux, when memory becomes
     scarce to the point that the kernel has trouble allocating memory for
     itself, it might decide to kill a running process in order to free up memory
     and reduce memory pressure. Note that systemd-oomd.service is a more
     flexible solution that aims to prevent out-of-memory situations for the
     userspace too, not just the kernel, by attempting to terminate services
     earlier, before the kernel would have to act.

     This setting takes one of continue, stop or kill. If set to continue and a
     process in the unit is killed by the OOM killer, this is logged but the unit
     continues running. If set to stop the event is logged but the unit is
     terminated cleanly by the service manager. If set to kill and one of the
     unit's processes is killed by the OOM killer the kernel is instructed to
     kill all remaining processes of the unit too, by setting the
     memory.oom.group attribute to 1; also see kernel documentation[2].

     Defaults to the setting DefaultOOMPolicy= in systemd‐system.conf(5) is set
     to, except for units where Delegate= is turned on, where it defaults to
     continue.

     Use the OOMScoreAdjust= setting to configure whether processes of the unit
     shall be considered preferred or less preferred candidates for process
     termination by the Linux OOM killer logic. See systemd.exec(5) for details.

     This setting also applies to systemd‐oomd.service(8). Similarly to the
     kernel OOM kills performed by the kernel, this setting determines the state
     of the unit after systemd-oomd kills a cgroup associated with it.
```

## Tests

Control group: https://certification.canonical.com/hardware/202303-31332/submission/363424/

Experiment group: https://certification.canonical.com/hardware/202303-31332/submission/363564/

The `stress-ng` stressors got killed but `checkbox-ng.service` remain alive. Kindly check [dmesg-after-memory-stress.txt](https://github.com/canonical/checkbox/files/14926707/after-memory-stress.txt) for more.

### DIFF
--- a/checkbox-ng/debian/checkbox-ng.service
+++ b/checkbox-ng/debian/checkbox-ng.service
@@ -10,6 +10,7 @@ Restart=always
 RestartSec=1
 TimeoutStopSec=30
 Type=simple
+OOMPolicy=continue
 
 [Install]
 WantedBy=multi-user.target


LiaoU3:
Wow! This change seems could fix the problem that bothers us for a long time.
However, I have a question which is it looks like this change is for the debian checkbox, can we also do the same thing on snap checkbox?

pieqq:
This looks promising for Debian-packaged version of Checkbox!

I would like to find a way to replicate this for the snap version too. Sadly, I'm not sure how to do. I've [asked on the Snapcraft forum](https://forum.snapcraft.io/t/setting-systemd-service-specific-option-in-a-snap-service-daemon/39692), we'll see!

Hook25:
I will merge this as it fixes at least half of the issue. Tyvm for the contribution, we will try to fix it on snaps as well

Hook25 (approved):

PR #1169, opened 2024-04-10
New mock_timeout (bugfix)

Hook25:
## Description

The timeout decorator spawns a new process undoing every mock that was applied to the module. This makes every test that uses it fail. This introduces a new decorator for tests called `mock_timeout` that mocks the timeout effect so that every test is still run in the same subprocess. Note that this doesn't make the timeout function timeout anymore, but given that the purpose of this decorator is to hard fail a procedure, that is not supposed to be a tested behaviour of a unit (it is always preferable to gently fail within the unit if possible rather than using this timeout, this is more of a last resort option)

## Resolved issues

Failing camera unit tests

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/checkbox-support/checkbox_support/helpers/timeout.py
+++ b/checkbox-support/checkbox_support/helpers/timeout.py
@@ -26,7 +26,9 @@ import os
 import subprocess
 import multiprocessing
 
+from functools import partial
 from contextlib import wraps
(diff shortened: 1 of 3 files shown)


fernando79513 (approved):
LGTM +1
Thanks again for the silly test!

PR #1170, opened 2024-04-10
Adding rt test plan (New)

fernando79513:
## Description

To test the real-time kernel, we have to include RT tests into the generic providers and create a SRU specific to RT devices.

The rt-performance tests are already in canonical/checkbox under the base provider and the TSN and TGPIO tests will be included in the future.

At the moment, the `sru-rt` will only include the rt-performance tests, but the TSN and TGPIO will be added in the future. 

Since the `rt-performance-tests-stress` takes 24 hours to complete, the `sru-rt` only includes `rt-performance-tests-automated` with a shorter version of the cyclic test for latency. Once we have tested the `sru-rt` on some devices and we ensure it is stable, we should probably add this test back, but taking into consideration that test runs using this test are going to take more than one day.

## Documentation

There are no changes to the documentation

## Tests

To list the RT jobs currently present on checkbox:
```
checkbox-cli expand com.canonical.certification::rt-performance-tests-full
```
To list the jobs currently in `sru-rt`:
```
checkbox-cli expand com.canonical.certification::sru-rt
...
Job 'com.canonical.certification::recovery_info_attachment'
Job 'com.canonical.certification::rt-performance-tests/cyclictest_rt'
Job 'com.canonical.certification::stress/cpu_stress_ng_test'
...
```

### DIFF
--- a/providers/base/units/rt-performance-tests/jobs.pxu
+++ b/providers/base/units/rt-performance-tests/jobs.pxu
@@ -1,7 +1,13 @@
 id: rt-performance-tests/cyclictest_rt_latency
 category_id: rt-performance-tests
+estimated_duration: 24h
 _summary:
     Run cyclictest to test scheduling latency. Default run time is 24 hours.
+_purpose:
+    This test executes the cyclictest command for a long period of time to 
+    evaluate the system's ability to maintain scheduling latency. It checks
+    that the maximum latency did not exceed the threshold and that there are no
+    overflows (scheduling misses) for each thread.
 plugin: shell
 requires:
     package.name == 'rt-tests'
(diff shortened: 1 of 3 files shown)


bladernr:
Is the plan here to require the 24h test as part of a certification baseline?

kissiel:
> Is the plan here to require the 24h test as part of a certification baseline?

In the RT testing that is being rolled out we intend to run this only for 60s.
Testing devices how they run the RT  kernel is not [yet?] part of the certification AFAIK.

kissiel (approved):
Fantastic stuff, TYVM!

fernando79513:
> Is the plan here to require the 24h test as part of a certification baseline?

Most surely we are not going to require runnug the test for 24 hours. We still have to set a definite value for this specific test. The 24 hour limit was set for the Intel IoTG devices on the iiotg provider. For the Real-time SRU, we will set a smaller duration in the future.

PR #1172, opened 2024-04-11
Modify bluetooth/bluetooth_obex_send job able to run on both classic and core (BugFix)

rickwu666666:
Modify bluetooth/bluetooth_obex_send job able to run on both classic and core

Replace l2ping command with hcitool inq.

Because l2ping tool is not a part of bluez snap

## Description
OBEX is able to test on Ubuntu core, but the current job can only run on classic images.
And replace `l2ping` command with `hcitool inq` which is able to inquire the Bluetooth devices. Since `l2ping` does not include in Bluez snap. And the reason that we don't use `hcitool scan` is because the `inq` is faster than `scan`.

## Tests
Test with desktop image
pass: https://certification.canonical.com/hardware/202311-32337/submission/363603/test-results/
fail: https://certification.canonical.com/hardware/202311-32337/submission/363616/test-results/

Test with core image
pass: https://certification.canonical.com/hardware/202401-33430/submission/363604/test-results/
fail: https://certification.canonical.com/hardware/202401-33430/submission/363617/test-results/

### DIFF
--- a/providers/base/units/bluetooth/jobs.pxu
+++ b/providers/base/units/bluetooth/jobs.pxu
@@ -423,9 +423,10 @@ category_id: com.canonical.plainbox::bluetooth
 id: bluetooth/bluetooth_obex_send
 depends: bluetooth/detect-output
 estimated_duration: 10.0
+environ: BTDEVADDR PLAINBOX_PROVIDER_DATA
 requires:
- package.name == 'bluez'
- executable.name == 'obexftp'
+ package.name == 'bluez' or snap.name == 'bluez'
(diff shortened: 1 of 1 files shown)


pieqq (approved):
Thanks for providing sample submissions on both classic and core! Makes the review much easier.

I'll approve it because it modifies an existing job and brings clear benefits (being able to run on UC), but I have a few improvements in mind:

1. At the very least, all these lines of Shell could be put into a script in `bin/`
2. Even better, that script could be written in Python, so it could be unit-tested :)

PR #1124, opened 2024-03-29
Make sure the device reboots after snap refresh/revert tests (BugFix)

pieqq:
## Description

The `snap_update_test.py` script is currently running asynchronous snapd requests and simply waiting 90 seconds before rebooting the device, without checking the status of the snapd command. This leads to very flaky results.

This PR updates the script (and parts of  `checkbox_support.snapd_utils.snapd`) to

- wait for the snapd request to finish
- poll the snapd change to provide more information to the tester
- keep all of this information in a log that is then attached to the session (because Checkbox currently does not keep output for `noreturn` jobs)
- initiate a reboot after each snap refresh/revert command instead of in a separate job. This is to avoid a situation we have seen in production where non-related jobs would be run between the refresh/revert job and the reboot job, making the results for these jobs unreliable

## Resolved issues

CHECKBOX-1264

## Documentation

Changes are meticulously documented in the git commits. Please read them for more information.

## Tests

- Unit tests
- Tested on the Xinyi device by running the `snap-refresh-revert` test plan on it after making sure the snapd, gadget and pc kernels were all using their respective beta channels. [Submission available here](https://certification.canonical.com/hardware/202207-30448/submission/362450/). The skipped tests are because the targeted revision is the same as the currently installed revision, which is expected for the gadget snap as the base version, stable version and beta version are all the same revision. Please note that the submission archive has an `attachment_files` directory containing the output of the `noreturn` jobs, such as:

```
[content of /attachment_files/com.canonical.certification__snapd_log-attach-after-snap-refresh-snapd-snapd-to-base-rev]

Refreshing snap snapd from revision 21465 to 19267
(info) (Doing) Ensure prerequisites for "snapd" are available
(info) (Doing) Monitoring snap "snapd" to determine whether extra refresh steps are required
(info) (Doing) Fetch and check assertions for snap "snapd" (19267)
(info) (Doing) Monitoring snap "snapd" to determine whether extra refresh steps are required
(info) (Doing) Setup snap "snapd" (19267) security profiles
(info) (Doing) Monitoring snap "snapd" to determine whether extra refresh steps are required
(info) (Doing) Setup snap "snapd" (19267) security profiles
(info) (Doing) Monitoring snap "snapd" to determine whether extra refresh steps are required
(info) (Doing) Setup snap "snapd" (19267) security profiles
(info) (Doing) Monitoring snap "snapd" to determine whether extra refresh steps are required
(info) (Doing) Automatically connect eligible plugs and slots of snap "snapd"
[...]

PR #1176, opened 2024-04-11
Bring back kernel/snapd/gadget snaps revert/refresh jobs to the SRU test plan (BugFix)

pieqq:
## Description

The snap-refresh-revert nested part had been removed from the SRU test plan in cd17d9a51b6565b8736f5fddf3a09102197c3efd following some discoveries that, when executed, the jobs executed would change the system state in the wrong order (e.g. updating the kernel snap without reverting it while running other jobs afterwards).

The jobs were [modified to prevent this kind of scenario](https://github.com/canonical/checkbox/pull/1124), and the use of [manifest entries](https://github.com/canonical/checkbox/tree/main/providers/base/units/snapd#manifest-entries) prevents the jobs from being run on all the devices yet (currently, the only project that enables them is the Xinyi project).

## Resolved issues

Fix LM-1522

## Tests

[Tested on Xinyi](https://github.com/canonical/checkbox/pull/1124#issuecomment-2048809738) with a local build of Checkbox.

### DIFF
--- a/providers/sru/units/sru.pxu
+++ b/providers/sru/units/sru.pxu
@@ -138,6 +138,7 @@ nested_part:
     # Now we ask to switch to the integrated graphics card.
     after-suspend-graphics-integrated-gpu-cert-automated
     # after-suspend-monitor-integrated-gpu-cert-automated # not defined
+    com.canonical.certification::snap-refresh-revert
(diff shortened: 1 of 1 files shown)


kissiel (approved):
Let's go!
+1

PR #1177, opened 2024-04-11
Add .git-blame-ignore-revs and update contrib guide (infra)

pieqq:
## Description

- .git-blame-ignore-revs prevents git blame from showing the black formatting commit
- add suggestion to use black from PyPI instead of version from the Ubuntu repo because the PyPI version is the one used in our CI/CD and repo versions are usually older


Hook25 (approved):
hash matches, +1

PR #1179, opened 2024-04-12
Add disconnect function to snapd class (New)

rickwu666666:
Add disconnect function to snapd class 

## Description
We intend to test the slots for SNAP, and the disconnect slots will be a part of our test scenario.

## Tests
We tried to call this function and that works fine.
#### Test script #####
```
#!/usr/bin/env python3
from snapd import Snapd
import os
import requests

def disconnect_interface(gpio_slot, gadget_name):
    """
    Connect GPIO plugs of checkbox to GPIO slots of gadget snap.

    Args:
        gpio_slot: A GPIO slot information.
        gadget_name: The name of the gadget snap.

    Raises:
        SystemExit: If failed to connect any GPIO.
    """

    # Get the snap name of checkbox
    timeout = int(os.environ.get('SNAPD_TASK_TIMEOUT', 60))

    print("Attempting disconnect gpio to {}:{}".format(gadget_name, gpio_slot))
    try:
        Snapd(task_timeout=timeout).disconnect(
            gadget_name,
            gpio_slot,
            "checkbox",
            "gpio"
            )
        print("Success")
    except requests.HTTPError:
        print("Failed to connect {}".format(gpio_slot))
        raise SystemExit(1)

def main():
    disconnect_interface('gpio-510', 'gadget')

if __name__ == "__main__":
    main()
```

#### Test result #####
```
root@ubuntu:/home/ceqa/snap_utils# snap connections checkbox|grep gpio-510
gpio                                    checkbox:gpio                          gadget:gpio-510      manual
root@ubuntu:/home/ceqa/snap_utils# python3 test.py 
Attempting disconnect gpio to gadget:gpio-510
Success
root@ubuntu:/home/ceqa/snap_utils# snap connections gadget|grep gpio-510
gpio                     -                                                    gadget:gpio-510    
```

### DIFF
--- a/checkbox-support/checkbox_support/snap_utils/snapd.py
+++ b/checkbox-support/checkbox_support/snap_utils/snapd.py
@@ -215,9 +215,7 @@ class Snapd:
             self._poll_change(r["change"])
 
     def connect(self, slot_snap, slot_slot, plug_snap, plug_plug):
-        self.connect_or_disconnect(
-            slot_snap, slot_slot, plug_snap, plug_plug
-        )
+        self.connect_or_disconnect(slot_snap, slot_slot, plug_snap, plug_plug)
 
     def disconnect(self, slot_snap, slot_slot, plug_snap, plug_plug):
         self.connect_or_disconnect(
--- a/checkbox-support/checkbox_support/snap_utils/snapd.py
+++ b/checkbox-support/checkbox_support/snap_utils/snapd.py
@@ -13,7 +13,6 @@ import requests_unixsocket
 
 
 class AsyncException(Exception):
-
     def __init__(self, message, abort_message=""):
         self.message = message
         self.abort_message = abort_message
@@ -203,9 +202,11 @@ class Snapd:
     def interfaces(self):
         return self._get(self._interfaces)["result"]
 
-    def connect(self, slot_snap, slot_slot, plug_snap, plug_plug):
+    def connect_or_disconnect(
+        self, slot_snap, slot_slot, plug_snap, plug_plug, action="connect"
+    ):
         data = {
-            "action": "connect",
+            "action": action,
             "slots": [{"snap": slot_snap, "slot": slot_slot}],
             "plugs": [{"snap": plug_snap, "plug": plug_plug}],
         }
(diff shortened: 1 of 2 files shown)


rickwu666666:
Depend on this [PR](https://github.com/canonical/checkbox/pull/1180)

checkbox-support/checkbox_support/snap_utils/snapd.py
```diff
@@ -213,6 +213,16 @@ def connect(self, slot_snap, slot_slot, plug_snap, plug_plug):
...
+    def disconnect(self, slot_snap, slot_slot, plug_snap, plug_plug):
+        data = {
+            "action": "disconnect",
+            "slots": [{"snap": slot_snap, "slot": slot_slot}],
+            "plugs": [{"snap": plug_snap, "plug": plug_plug}],
+        }
+        r = self._post(self._interfaces, json.dumps(data))
+        if r["type"] == "async" and r["status"] == "Accepted":
+            self._poll_change(r["change"])
+
```
pieqq, line 225: Since this function is exactly the same as `connect`, except its `action` is set to `disconnect` instead of `connect`, I recommend you refactor this.

For example, create a `connect_or_disconnect(self, action, slot_snap, slot_slot, plug_snap, plug_plug)` function that takes `"connect"` or `"disconnect"` as `action`, and does exactly the same as the `connect` function.

Then, you can call `connect_or_disconnect` from `connect` and `disconnect`, turning these functions into one-liners.

You can then revamp a little bit the unit tests (what you wrote is good, just turn this into tests for the new `connect_or_disonnect` function instead, and for the `connect` and `disconnect` functions, just test that they call `connect_or_disonnect` with the proper `action` or something like this)

pieqq (dismissed):
It looks good, but I think you should refactor it to prevent code redundancy, since `connect` and `disconnect` functions are very similar. See my comment below.

rickwu666666:
> It looks good, but I think you should refactor it to prevent code redundancy, since `connect` and `disconnect` functions are very similar. See my comment below.

Hi Pierre,
I've pushed the fix. Thanks for the review.

Hook25 (approved):

Issue #1022, opened 2024-02-27
Virtualization Tests Fail to Run on Noble


mreed8855:
### Bug Description

The virtualization tests are skipped when attempting to run them on Noble 6.8 kernel.  

job cannot be started: resource expression "package.name == 'lxd' or snap.name == 'lxd'" evaluates to false, resource expression "package.name == 'lxd' or snap.name == 'lxd'" evaluates to false

This happens for both verify_lxd and verify_lxd_vm.

	
Kernel: 6.8.0-11-generic x86_64
Distro: Ubuntu 24.04 (Noble Numbat)

### To Reproduce

1.  run test-virtualization

### Environment

Kernel: 6.8.0-11-generic x86_64
Distro: Ubuntu 24.04 (Noble Numbat)

### Relevant log output

```shell
virtualization/verify_lxd	

not supported	blocker		

job cannot be started: resource expression "package.name == 'lxd' or snap.name == 'lxd'" evaluates to false, resource expression "package.name == 'lxd' or snap.name == 'lxd'" evaluates to false
```

mreed8855:
The current workaround for this is to install lxd prior to running the test suite.

sudo snap install lxd

PR #1181, opened 2024-04-12
Require one of the lxd or lxd-installer debs or the lxd snap for the virtualization tests. (BugFix)

bladernr:
## Description
Noble does not provide the lxd snap by default now in the server images, instead it provides a wrapper called lxd-installer that triggers the snap install the first time a user tries using either the lxc or lxd commands.  

The jobs were not looking for the lxd-installer package, instead they were looking for the lxd snap or the old lxd package.  This changes the constraints so that the jobs will run if only lxd-installer is present (on noble server) . In this case, the 'lxd init' commands that virtualization.py runs will trigger the snap install on server, and then tests proceed as expected.

Added a check to run the job if lxd-installer is on the system, since on Noble, the lxd snap and deb will not be there at first.

## Resolved issues

#1022 
CHECKBOX-1275

## Documentation

NO docs changes needed.

## Tests

Manually  modified jobs.pxu on a fresh noble install to add the package lxd-installer to the requires section for the virtualization jobs.

Ran test-virtualization to call both lxd and lxdvm tests.

Before changes, the tests are skipped because the lxd snap is not installed.
After the changes, the tests are executed.

### DIFF
--- a/providers/base/units/virtualization/jobs.pxu
+++ b/providers/base/units/virtualization/jobs.pxu
@@ -5,7 +5,7 @@ environ: LXD_TEMPLATE KVM_IMAGE
 estimated_duration: 60.0
 requires:
  executable.name == 'lxc'
- snap.name == 'lxd'
+ package.name == 'lxd-installer' or snap.name == 'lxd'
 command: virtualization.py --debug lxdvm
 _description:
  Verifies that an LXD Virtual Machine can be created and launched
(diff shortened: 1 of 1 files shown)


bladernr:
Why is this failing on "black formatting"? I pulled a updated copy of checkbox, edited a branch and pushed it, and the black formatting thing seems to be complaining about things I didn't even touch.

bladernr:
Saw @pieqq 's update and rebased to fix the formatting issues.

kissiel (approved):
Makes sense! +1

PR #1182, opened 2024-04-15
[checkbox-ce-oem] checkbox-ce-oem change to daily build (Infra)

LiaoU3:
## Description
Make checkbox-ce-oem from building when commit to building daily.

## Resolved issues
N/A

## Documentation
N/A

## Tests
https://github.com/canonical/checkbox/actions/runs/8682803386

### DIFF
--- a/.github/workflows/daily-builds.yml
+++ b/.github/workflows/daily-builds.yml
@@ -29,17 +29,17 @@ jobs:
           commit_count=$(git log --since="$LAST_PASS_DAILY_BUILD" --oneline -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap | wc -l)
           echo "new_commit_count=$commit_count" | tee $GITHUB_OUTPUT
 
-  # checkbox-core-snap-daily:
-  #   needs: check_for_commits
-  #   if: ${{ needs.check_for_commits.outputs.new_commit_count > 0 || github.event_name == 'workflow_dispatch' }}
-  #   uses: ./.github/workflows/checkbox-core-snap-daily-builds.yml
-  #   secrets: inherit
+  checkbox-core-snap-daily:
+    needs: check_for_commits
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-builds.yml
```diff
@@ -41,6 +41,12 @@ jobs:
     uses: ./.github/workflows/checkbox-snap-daily-builds.yml
     secrets: inherit
 
+  checkbox-ce-oem-snap-daily:
+    needs: check_for_commits
+    if: ${{ needs.check_for_commits.outputs.new_commit_count > 0 || github.event_name == 'workflow_dispatch' }}
```
Hook25, line 46: you need to edit the line 29 of this file adding the path to the provider! else your commits to the provider are not enough to trigger a rebuild. (I would suggest change but github doesn't let me comment on unedited lines)

Basically go from 
```
commit_count=$(git log --since="$LAST_PASS_DAILY_BUILD" --oneline -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap | wc -l)
```
to
```
commit_count=$(git log --since="$LAST_PASS_DAILY_BUILD" --oneline -- checkbox-ng checkbox-support providers checkbox-core-snap checkbox-snap contrib/checkbox-ce-oem | wc -l)
```

Hook25 (changes requested):
I'm not opposed to adding this to daily build but we use the status of daily build also to decide if we can promote a beta version of checkbox so I have to know. How reliable is that snap build? Does it build all the times or does it seldomly fail?

Also, small tweek below, needed to actually trigger if there are new commits to contrib.

Hook25:
The black formatting ierror should be now fixed in main, rebase the PR, it should be ok

.github/workflows/daily-builds.yml
LiaoU3: Thank you! Fixed!

PR #1187, opened 2024-04-16
Mention allowed characters in template ids in the reference page (infra)

pieqq:
Template ids, like job ids, should only include characters in [a-z0-9/-].


tang-mm (approved):
LGTM +1

PR #1200, opened 2024-04-19
Created launchers for 24.04 Release (New)

bladernr:
## Description

Add new launchers for 24.04 release for Server Cert. New launchers and test plans are now version agnostic.  The older 22.04 specific launchers and testplans will remain for a while and will eventually be deprecated in the future.

Going forward users will use commands like `certify-ubuntu-server` rather than `certify-24.04`

## Resolved issues
SERVCERT-1538

## Documentation
Will need minor updats to self-test guide which are already underway (CDOC-42)

## Tests

Simply diffed the 22.04 launchers and testplans that these were copied from and verified the new testplan IDs matched. No further tests needed


kissiel (approved):
LGTM, +1

PR #1201, opened 2024-04-22
Checkbox snap for Core24 builds (infra)

Hook25:
## Description

This introduces the core24 snapcraft.yaml. The snaps for arm64 and amd64 are already in the store and were built from this recipe. To aid the review of this huge PR, just know that most of the content here is copy-pasted from core22 so for an easier review I can reccomend
```
diff --color -r series22 series24
```

Here is a full description of what changed:

0. Name changed :)
1. grade/build-base: devel, temporary as core24 is not stable yet, this is needed for now
2. Every `SNAPCRAFT...` environment variable was renamed into `CRAFT...` as per the new snapcraft guideline
4. Every `snapcraft pull/build/stage` etc. explicit action  (in overrides) replaced with `craft default` as per new snapcraft guideline
5. fwts installed from universe (as the build was broken and its unclear why, this can be reverted once all dependencies are avialable)
6. stress-ng from universe (same reason as fwts)
7. acpi-tools from universe (same reason as fwts)
8. python3-opencv from universe (same reason as fwts)
9. /s/python3.10/python3.12 (new base, new python)
10. perl explicitly staged (no longer in base as of core24)
11. pytotherside metapackage is no longer in the archive, replaced with the actual packages

The snap contains a small hack (documented above it) to clear the exestack bit on one library `libamdhip64.so.5.2` for amd64. This is necessary as the snap will require a manual review on every build if it is not done. The change doesn't seem to have any impact and should be removable once the library is updated.

## Resolved issues

N/A

## Documentation

From a documentation point of view the only change from series22 is the exestack bit clearing, that is documented.

## Tests

Snaps were built and are currently in the store.


kissiel (approved):
Great stuff!
+1

PR #1202, opened 2024-04-22
Workflow to build core24 snaps (infra)

Hook25:
## Description

This is a workflow to build and publish Core24 to the edge channel in the store. This is only includes the runtime for now and it is here to experiment (as github doesn't seem to add a workflow to the list of "runnables" till it is landed in main at least once).

## Resolved issues

Make core24 buildable automatically like the others.

## Documentation

N/A

## Tests

N/A


kissiel (approved):
Let's go! :-)

PR #1204, opened 2024-04-23
Update possible certification status values for Job units and modify its default (new)

pieqq:
## Description

Until now, a Job unit could have 4 possible values for certification status (`unspecified`, `not-part-of-certification`, `blocker`, `non-blocker`). This PR reduces the number to 2 (`blocker` and `non-blocker`), and set the default to `non-blocker` (instead of `unspecified` in the past).

## Resolved issues

CER-2586

## Documentation

- Job unit reference page is updated
- Examples making use of `unspecified` certification values are updated throughout the docs
- Checkbox submission JSON schema updated

## Tests

The `disk-cert-automated` nested part has a mix of default and cert-blockers.

Running `checkbox-cli list-bootstrapped -f "{id} ({certification_status})\n" "com.canonical.certification::disk-cert-automated"` allows to see the difference before and after the patch.

Before:

```
executable (unspecified)
disk/detect (blocker)
disk/stats_dm-1 (unspecified)
disk/read_performance_dm-1 (blocker)
disk/storage_device_dm-1 (blocker)
benchmarks/disk/hdparm-read_dm-1 (unspecified)
benchmarks/disk/hdparm-cache-read_dm-1 (unspecified)
disk/apste_support_on_nvme0 (unspecified)
disk/apste_support_on_nvme1 (unspecified)
manifest (unspecified)
disk/check-software-raid (unspecified)
```

After:

```
executable (non-blocker)
disk/detect (blocker)
disk/stats_dm-1 (non-blocker)
disk/read_performance_dm-1 (blocker)
disk/storage_device_dm-1 (blocker)
benchmarks/disk/hdparm-read_dm-1 (non-blocker)
benchmarks/disk/hdparm-cache-read_dm-1 (non-blocker)
disk/apste_support_on_nvme0 (non-blocker)
disk/apste_support_on_nvme1 (non-blocker)
manifest (non-blocker)
disk/check-software-raid (non-blocker)
```

I also ran this test plan before and after the patch and uploaded the results to C3.

[Before](https://certification.canonical.com/hardware/202204-30172/submission/365427/test-results/):

image

[After](https://certification.canonical.com/hardware/202204-30172/submission/365429/test-results/):

image


checkbox-ng/checkbox_ng/launcher/subcommands.py
Hook25, line 1: One thing that worries me is that this field is overridden for local but not for remote (or at least, I can't puzzle together how it is for remote). Will this cause any issue possibly?

checkbox-ng/plainbox/impl/exporter/test_html.py
Hook25, line 1: The html end to end testing is pretty weird and not a unit test for sure, but it is not 100% pointless. Why did you remove them instead of updating them?

checkbox-ng/plainbox/test-data/html-exporter/with_both_certification_status.html
Hook25, line 1: related to removal above, why?

checkbox-ng/plainbox/test-data/html-exporter/with_certification_non_blocker.html
Hook25, line 1: related to removal above, why?

Hook25 (approved):
+1 on the change, less random statusses for jobs is better! A few questions, the most important one is: this doesn't change anything in the remote codebase (as far as I can see). Is this change propagated there somehow I can't understand right now? If so, how?

Minor other questions below.

checkbox-ng/plainbox/impl/exporter/test_html.py
pieqq: Because we went from 4 kinds of cert status to 2, so the test matrix is much smaller and some tests were redundant, so I removed them, which included removing the associated generated sample HTML files.

checkbox-ng/checkbox_ng/launcher/subcommands.py
pieqq: As discussed on MM, the change you're talking about is not a local Vs. remote, it's just to modify the way the `expand` subcommand returns the cert status.

checkbox-ng/plainbox/impl/exporter/test_html.py
Hook25: ok I was kind of asking why not re-sampling the "golden" result but I guess that if the coverage is not impacted by the removal fire on!

PR #1212, opened 2024-04-26
removed requirement of udisks2 on core (BugFix)

fernando79513:
## Description

Acording to #1029, the test was failing on some ubuntu core devices, because they required having the udisks2 snap, but the udisks2 snap was not needed for devices that run ubuntu core, since they were using another script to test the USB.
In summary:
 - The removable_storage_watcher.py wouldn't work on core even if udisks2 snap is installed
 - The run_watcher.py wouldn't work on some desktop devices
 - On core devices, since they are using run_watcher.py, we do not require udisks2 snaps

## Resolved issues
#1029, https://warthogs.atlassian.net/browse/CHECKBOX-1279

## Documentation

There are no documentation changes.

## Tests

When running:
```
checkbox.checkbox-cli run com.canonical.certification::usb/insert
```
Ubuntu Desktop requires the `udisks2` package or the `udisks2` snap, and fails if they don't exist
Ubuntu Core does not require this check.

### DIFF
--- a/providers/base/units/usb/usb.pxu
+++ b/providers/base/units/usb/usb.pxu
@@ -54,7 +54,9 @@ id: usb/insert
 flags: also-after-suspend fail-on-resource
 template-engine: jinja2
 requires:
+ {%- if not __on_ubuntucore__ %}
  package.name == 'udisks2' or snap.name == 'udisks2'
+ {% endif -%}
 _summary: USB 2.0 storage device insertion detected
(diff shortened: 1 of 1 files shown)


Hook25 (commented):
Does this work? is an empty `requires` constraint valid?

Hook25 (approved):
yessir it does, +1

you learn something new every day out here :)

PR #1216, opened 2024-04-29
Fix base provider's unit tests (BugFix)

kissiel:
## Description

Over the course of last 2 months, yet again, tests that calling `sleep` or other unmocked time-related functions had been committed to the repository making the iteration time unnecessary long.
One of the test also made the whole thing unable to be run from within editors or IDEs that are running from a snap, because it actually asserted we're not running from snap (and of course the CI let it through, as it runs the tests not within a snap.

# Overall this PR makes the tests run  **20x** faster.

See individual commits for details.

### DIFF
--- a/providers/base/bin/prime_offload_tester.py
+++ b/providers/base/bin/prime_offload_tester.py
@@ -122,17 +122,13 @@ class PrimeOffloader:
             )
 
     def check_offload(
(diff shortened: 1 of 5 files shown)


Hook25 (approved):

pieqq (approved):
LGTM! Thanks :)

PR #1222, opened 2024-05-02
Disable failfast strategy for core snap (infra)

Hook25:
## Description

Similarly to what was done for the other jobs, avoid aborting all other builds if one fails. This is pretty wasteful, as it doesn't cancel the snapcraft build on launchpad, but also potentially limits the amount of snaps we get to push in one day with no benefit. 

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


kissiel (approved):
LGTM

PR #1227, opened 2024-05-03
Core24 SH build for frontend (infra)

Hook25:
## Description

This mirrors what was done for the backend snap to the frontend one. Namely, indroduce an automated way to build and publish the snaps for all configurations/arch we support.

## Resolved issues

Frontend snap for classic 22.04 on armhf was actually a core snap. I have ran this issue and confirmed that it fixes the problem. 

## Documentation

N/A

## Tests

Last run: https://github.com/canonical/checkbox/actions/runs/8936471962

### DIFF
--- a/.github/workflows/core24-builds.yml
+++ b/.github/workflows/core24-builds.yml
@@ -5,7 +5,7 @@ on:
   workflow_call:
 
 jobs:
-  snap:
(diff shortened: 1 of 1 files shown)


fernando79513 (approved):
LGTM +1

PR #1230, opened 2024-05-06
describe the black workflow better (infra)

kissiel:
The previously used naming was misleading, suggesting we're formatting with black, and that the formatting action failed.
And even when interpreted properly (as a check) there was the mention of the line length even when the code was actually fitting in 79 chars.
Also `black` as an action is a bit too terse.


p-gentili (approved):
LGTM!

PR #1231, opened 2024-05-06
Fix core22 not building issue (infra)

Hook25:
## Description

The core22 build is broken because now snapcraft doesn't use the bodge that we created for armhf anymore nor does it set a default rust channel for rustup. This fixes it by removing the bodge (that is no longer needed) and sets the rust channel

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CER-2608

## Documentation

N/A

## Tests

I have built the snap on armhf and the build surpassed the part


p-gentili (approved):
LGTM, thanks snapcraft team :)

PR #1238, opened 2024-05-09
Get the host keyboard device path dinamically (BugFix)

p-gentili:
## Description

This PR adapts the `input/keyboard` to the latest changes in host mainboard FW (especially https://warthogs.atlassian.net/browse/ZAP-717). It's now possible to reconfigure the HID devices provided by host which means the device ID of host keyboard can change depending on the device order.

The path to the keyboard device is now obtained via pattern matching, making it compatible with different interface number.

## Resolved issues

N/A

## Documentation

N/A

## Tests

- Added unit tests for the new function.
- Tested on HW in two scenarios:
    -  With keyboard as first HID device, as before
    - With keyboard as second HID device

### DIFF
--- a/providers/base/bin/zapper_keyboard_test.py
+++ b/providers/base/bin/zapper_keyboard_test.py
@@ -10,6 +10,7 @@ import struct
 import sys
 import threading
 from enum import Enum
+from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
Makes sense +1

PR #1240, opened 2024-05-10
Allow bypassing the check via manual trigger (infra)

Hook25:
## Description

Sometimes the beta release is created via other means, for instance the last one, due to a snapcraft regression, was partially built locally. When this happens, the last successful daily build is not "the one that originated the beta currently in the store/ppa". This adds the possibility to "force" the release of beta by manually triggering the workflow.

## Resolved issues

Sometimes we need to bypass this check when the build was put together manually

## Documentation
N/A

## Tests

This is the workaround I always use to release when we don't have a "clean" daily build


p-gentili (approved):
Makes sense.

PR #1243, opened 2024-05-10
Bump snapcraft version core24 (infra)

Hook25:
## Description

Snapcraft 8 seems to choke on the legacy command. This replaces it with the new one


fernando79513 (approved):
LGTM +1

Issue #1122, opened 2024-03-28
Checkbox couldn't recognize job failed


hanhsuan:
### Bug Description

The condition of #1117 shows that checkbox couldn't recognize this job should be marked as failed. This behavior could be found in [graphics/1_gl_support](https://certification.canonical.com/hardware/202401-33397/submission/362354/test/200240/result/40073727/) that worked [before](https://certification.canonical.com/hardware/202401-33399/submission/354087/test/206378/result/39028089/).

### To Reproduce

1. modify the `suspend/audio_before_suspend` to naming miss match
2. run `suspend/audio_before_suspend` test

### Environment

- OS: jammy
- checkbox type: debian
- checkbox version: 4.0.0.dev194

zongminl:
Fixed by #1247 and #1268

PR #1247, opened 2024-05-15
Fix `id: graphics/{index}_gl_support_{product_slug}` alway return ture (New)

hanhsuan:
## Description

This PR is going to create a python script to fix `id: graphics/{index}_gl_support_{product_slug}` issue as describing in
https://github.com/canonical/checkbox/issues/1122.

The job and test plan will be addressed in another PR.

## Resolved issues

The command used in `id: graphics/{index}_gl_support_{product_slug}`
```shell
"$CHECKBOX_RUNTIME"/usr/lib/nux/unity_support_test -p 2>&1 | sed -e "s/\x1b\[[0-9;]*m//g"
```
will always return true to make this test always pass.

## Tests

Unit test

If you would like to make the renderer to be changed to llvmpipe while testing, you could use following command to test.
```shell
LIBGL_ALWAYS_SOFTWARE=1 python3 gl_support.py
```


providers/base/tests/test_gl_support.py
```diff
@@ -0,0 +1,105 @@
...
+
+class RemoveColorCode(unittest.TestCase):
+    """
+    This function should remove color code
+    """
+
+    unchanged = r"""
+OpenGL vendor string:   Intel
+OpenGL renderer string: Mesa Intel(R) Arc(tm) Graphics (MTL)
+OpenGL version string:4.6 (Compatibility Profile)Mesa 23.2.1-1ubuntu3.1~22.04.2
```
stanley31huang, line 31: It seems to me that the data format is a bit different with the output by unity_support_test utility, and I believe that's because you did not changes to fit black format check.
So I would suggest we place the original data in test_data directory, and use it in the unit test.

e.g. https://github.com/canonical/checkbox/blob/main/providers/base/tests/test_data/mdstat_intel_rst.txt

hanhsuan: This only thing changed in this line is removing spaces and I don't think this will impact the unit test correctness for removing color code.

Hook25 (approved):
+1

PR #1249, opened 2024-05-16
Add testptp for TGPIO testing (Infra)

LiaoU3:
## Description
Add testptp for TGPIO testing #1239 

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A

### DIFF
--- a/providers/base/src/EXECUTABLES
+++ b/providers/base/src/EXECUTABLES
@@ -2,3 +2,4 @@ alsa_test
 clocktest
 threaded_memtest
 ptrace_test
+testptp
--- a/providers/base/src/Makefile
+++ b/providers/base/src/Makefile
@@ -1,6 +1,10 @@
 .PHONY:
 all: alsa_test clocktest threaded_memtest ptrace_test testptp
 
+ifeq ($(shell uname -m), x86_64)
+all: testptp
+endif
+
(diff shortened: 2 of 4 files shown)


Hook25 (commented):
So this is interesting, you initially vendorized the files from the intel project, but then you rolled back the commit and added the rule in the snapcraft.yaml. I've tested this, it snaps, it is fine but what about all the other platforms (we support source, debs and 4 other snaps beside 22)? Wi will never support tgpio on these platforms? Can you explain me why you decided to roll back the vendorization and add this as a part in the yaml instead?

LiaoU3:
> So this is interesting, you initially vendorized the files from the intel project, but then you rolled back the commit and added the rule in the snapcraft.yaml. I've tested this, it snaps, it is fine but what about all the other platforms (we support source, debs and 4 other snaps beside 22)? Wi will never support tgpio on these platforms? Can you explain me why you decided to roll back the vendorization and add this as a part in the yaml instead?

You are right. My PR cannot fit other platforms. Do you have any idea how I could fix this?
The reason why I rollback the commit is due to [LP#2065260](https://bugs.launchpad.net/lookout-canyon/+bug/2065260)

Hook25 (approved):

PR #1252, opened 2024-05-22
Fix example (Infra)

waveform80:
The parsed example is incorrect (or, I suppose, the example is incorrect, but either way one of them needs fixing!)


Hook25 (approved):
Good catch, thanks

+1

PR #1257, opened 2024-05-27
Fix: Make checkbox-ce-oem uc20 as a runnable checkbox (BugFix)

stanley31huang:
## Description
make checkbox-ce-oem/uc20 as a runnable checkbox, so we could use it as our checkbox runtime if needed.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A


LiaoU3 (approved):
+1

PR #1259, opened 2024-05-29
Checkbox24 for armhf via remote builds (infra)

Hook25:
## Description

TLDR: This introduces a pipeline to build checkbox24 on armhf. While the other checkbox24 we can build natively, this we have to dispatch to launchpad so that it is built on armhf.

This build runtime and frontend using remote-build because we have no armhf shr. The new remote-build works differently, namely it requires a clean branch to push on launchpad. This is an issue for us because all the prepare.sh scripts copy around files to create the build directory. This introduces a new trio of prepare scripts that do the same thing but copying "additional files" (namely: the common files and the recipe) to the root of the repository. Those changes are then automatically committed by the pipeline so that the branch is "clean".

## Resolved issues

N/A

## Documentation

N/A

## Tests

The version currently in the store was built with this pipeline (although due to an issue it was pushed manually)


fernando79513 (approved):
LGTM +1!

fernando79513 (approved):
LGTM +1!

PR #1264, opened 2024-05-31
Stable core24 snapcraft yaml (infra)

Hook25:
## Description

These changes are to prepare the checkbox24 snap to be propoted into stable. 
This changes:
- move the build-base to stable (no need anymore as core24 is now released)
- move back the grade to stable (this is needed to allow the snap to be actually promoted to stable)
- extend the exestack to any library (there are a lot that are violating this policy on armhf, instead of tracking them lets just clean the bits for now, else we fail to publish to the store)
- updated the snapcraft8_prepare_[classic|uc] to also include hooks in the build
- updated 1 dependency because it changes name on newer versions of python (this will probably need to be extended, but for now lets check what happens)

## Resolved issues

N/A

## Documentation

Added a comment to explain why we ignore the return value of the execstack cleaning 

## Tests

Snaps currently in the store were built on this branch: https://github.com/canonical/checkbox/actions/runs/9304327974

### DIFF
--- a/checkbox-core-snap/series24/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series24/snap/snapcraft.yaml
@@ -223,7 +223,8 @@ parts:
       # a manual review
       for libamd_path in $(find .. -name "*.so.*"); do
         echo "Clearing execstack on $libamd_path";
-        execstack --clear-execstack "$libamd_path" || exit 1
+        # the || true is because there are som  *.so*.py in gdblib
+        execstack --clear-execstack "$libamd_path" || true
       done;
       craftctl default
     stage-packages:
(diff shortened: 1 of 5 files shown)


checkbox-core-snap/series24/snap/snapcraft.yaml
```diff
@@ -222,9 +221,10 @@ parts:
       # cleaning this bit should not cause any issue but if it does do not only remove this line
       # contact also the store/update the package because else this will always trigger
       # a manual review
-      for libamd_path in $(find .. -name "libamdhip64.so.5*"); do
+      for libamd_path in $(find .. -name "*.so.*"); do
         echo "Clearing execstack on $libamd_path";
-        execstack --clear-execstack "$libamd_path" || exit 1
+        # the || true is because there are som  *.so*.py in gdblib
```
fernando79513, line 226:
```suggestion
        # the || true is because there are some  "*.so*.py" in gdblib
```

fernando79513 (commented):
Good job! Just a small typo.

fernando79513 (approved):
LGTM +1!

PR #1269, opened 2024-06-03
Unplug the test monitor at the end of the EDID test (BugFix)

p-gentili:
## Description

The EDID job leaves the "monitor" connected at the end of the test, making tests like "resolution_after_suspend" fail. This PR introduces a simple context manager to clear the environment at the end of the test.

## Resolved issues

Resolves [CHECKBOX-1436](https://warthogs.atlassian.net/browse/CHECKBOX-1436)

## Documentation

Tests are included and functions documented.

## Tests

Locally, side-loading the provider and running in sequence
```
DISPLAY=:0 xrandr -q | grep "[*]" | awk '{{print $1}}'
DISPLAY=:0 checkbox.checkbox-cli run com.canonical.certification::monitor/edid
DISPLAY=:0 xrandr -q | grep "[*]" | awk '{{print $1}}'
```

[CHECKBOX-1436]: https://warthogs.atlassian.net/browse/CHECKBOX-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/base/bin/edid_cycle.py
+++ b/providers/base/bin/edid_cycle.py
@@ -11,12 +11,22 @@ import os
 import re
 import subprocess
 import time
+from contextlib import contextmanager
 
 from checkbox_support.scripts.zapper_proxy import zapper_run  # noqa: E402
 
 EDID_FILES = list(
     pathlib.Path(os.getenv("PLAINBOX_PROVIDER_DATA", ".")).glob("edids/*.edid")
 )
+
+
+@contextmanager
+def zapper_monitor(zapper_host: str):
+    """Unplug the Zapper monitor at the end of the test."""
+    try:
+        yield
+    finally:
+        _clear_edid(zapper_host)
 
 
 def discover_video_output_device(zapper_host):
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
LGTM not merging as I have a question: is this the only reason why the "cycle" is not a cycle? Can't it happen that we unintentionally change the resolution and never put it back to the original value? That is what the diff was reporting in my imagination

p-gentili:
> LGTM not merging as I have a question: is this the only reason why the "cycle" is not a cycle? Can't it happen that we unintentionally change the resolution and never put it back to the original value? That is what the diff was reporting in my imagination

Three things to keep in mind:
- the idle state of test monitor is disconnected
- the other jobs, like *resolution_after_suspend* are not xxx-aware. It means that the xxx jobs should clean the environment when they exit (similarly the USB hot-plugging job VS usb-detect job)
- this job is not modifying the resolution of the internal monitor (for now at least)

So the cycle is OFF -> resolution 1 -> res. 2 -> res. 3 and, with this, -> OFF. 

For instance, the device linked in the Jira card was reporting before suspend only a FHD monitor and, after suspend, two FHD monitors.

Hook25:
ty for the explanation

Issue #1278, opened 2024-06-07
What's plainbox?


waveform80:
### Enhancement Proposal

As someone diving into checkbox for the first time, and attempting to define some new tests, I keep seeing references to "plainbox" (e.g. in the [file format spec](https://checkbox.readthedocs.io/en/stable/reference/units/rfc822.html#plainbox-rfc822-specification)). Unfortunately, nowhere has actually defined *what plainbox is*. After a bit of digging I did discover [the Launchpad page](https://launchpad.net/plainbox) and was pleased to read it has a "well ... documented core". Unfortunately, I haven't manage to find those docs. I did find [the PyPI page](https://pypi.org/project/plainbox/) for the project which has several, unfortunately dead, [documentation links](http://plainbox.readthedocs.org/en/latest/usage.html).

If the answer is basically "it's the core of checkbox" and if plainbox isn't really a separate project at all any more (I see its last release on PyPI was 7 years ago, at the time of writing) perhaps it'd be better to just search'n'replace all those references in the docs and change them to "checkbox"?

I'm happy to file a PR for this, if this is desirable.

Hook25:
Hello, you are right, there is no plainbox anymore, as in: our documentation should not reference it all over the place. 
If you are willing to, feel free to file a pr where you find/replace all occurrences with Checkbox. Do so only in the `docs/` subdirectory, as in the code some plainbox still ramains separate.

As for what plainbox was: it was basically a base on top of which one could build a testing frameword. One such framework is Checkbox (and that we know of, the only one) . Most of this separation is gone nowadays because it caused (and still causes sometimes) quite a few headaches and complications without yielding any benefit.

PR #1285, opened 2024-06-12
Replace plainbox references (Infra)

waveform80:
## Description

Replaces references to plainbox with checkbox or, where sensible, simply removes them. This only affects files under `docs/` but even then there are places where plainbox references remain, such as in category and job identifiers. However, I'm reasonably confident that all the changes proposed are within prose only.

There are some references I've simply removed, because they reference specific plainbox versions and attempting to map these to checkbox versions is difficult and probably pointless (as they'll apply to all checkbox versions in current usage). There are also some "other" fixups, like a couple of obvious typos, a corrected path, and some fixed links.

The changes are organized into separate commits with the trivial obvious stuff first (where I'm confident that replacement of plainbox with checkbox is all that's needed), and the later commits covering stuff I'm a little less sure about which should probably be double-checked.

Each commit has additional commentary in the commit message. Where long paragraphs have been changed, I've re-wrapped things. If this makes things too hard to review, I'm happy to separate the re-wrap into a separate commit or just avoid the re-wrap entirely (and to rebase or re-work generally).

## Resolved issues

Fixes: #1278

## Tests

Tested build with sphinx locally; changes look reasonable and well formatted.

### DIFF
--- a/docs/how-to/custom-app.rst
+++ b/docs/how-to/custom-app.rst
@@ -74,9 +74,9 @@ in another provider it is best to reference that provider in the snap, and
 include the test, or whole test plans from that provider in your new testing
 project.
 
-Let's reuse disk tests from the "plainbox-provider-snappy" provider that we
-can use from the checkbox generic snap. All we need to do is add chosen tests
-to the ``include`` field of the test plan.
+Let's reuse disk tests from the "base" provider that we can use from the
+checkbox generic snap. All we need to do is add chosen tests to the ``include``
+field of the test plan.
 
 .. code-block:: none
     :caption: units/test-plan.pxu
@@ -95,7 +95,7 @@ to the ``include`` field of the test plan.
     (...)
 
 You can also include the whole *external* test plan. Let's reuse the CPU
-testing suite from plainbox-provider-snappy.
+testing suite from the base provider.
 
 .. code-block:: none
     :caption: unit/test-plan.pxu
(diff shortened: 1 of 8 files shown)


docs/how-to/custom-app.rst
```diff
@@ -28,7 +28,7 @@ Let's create a Checkbox Snap for project called "myproject".
 
 .. code-block:: bash
 
-    $ ./configure --with-provider-included project=myproject base=18 && ./cleanup.sh
+    $ ./configure --with-provider-included project=myproject base=18
```
Hook25, line 31: Why are you removing the cleanup command? also, probably unrelated to this

Hook25:
/canonical/self-hosted-runners/run-workflows 9c8760f60ae5bff5f8ebeedc2facf24260f19138

Hook25 (changes requested):
I have to ask you just one change on this, please make all the `checkbox` -> `Checkbox` (I would do it myself but I can't push to your fork!)

Ty for this btw, really helps us out

Hook25:
/canonical/self-hosted-runners/run-workflows 3cb4e459788c6d5cbe534f09cfa3b5f41dd624fb

Hook25 (approved):
Thank you very much for taking this up!

+1

Hook25:
In order to land this we need signed commits, can you please sign your commits?

https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#pull-requests

waveform80:
Should all be signed now (and verified after realizing the GPG key in GitHub hadn't been updated with new emails in forever)

Hook25:
/canonical/self-hosted-runners/run-workflows fc197c3ce4ec5b9b7358ab9a5f259c94fe8f577f

PR #1229, opened 2024-05-06
Improvements in EDID cycling (BugFix)

kissiel:
## Description

TL;DR: A more reliable way to do EDID cycling.

When debugging jobs that tested the changes in resolutions of the connected monitor, I noticed that the regexes failed to detect the right port on which the monitor was connected. I also discovered there's plenty of corner cases which would be very hard to cover with just a regex extraction. Thus, I wrote a fully fledged parser for `xrandr`'s and `gnome-randr`'s output.  Both parsers are used in the new function I'm proposing here to land in `checkbox-support` where it would return the information about display modes available on given video output.

This in turn, when plugged into the `edid_cycle` test script started yielding true results from the job.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/ZAP-677
Fixes: https://warthogs.atlassian.net/browse/ZAP-678

## Tests
I tested the code by running it on multiple devices running both X11 and Wayland.
I also propose unit tests with the code.

PR #1286, opened 2024-06-12
Set extend mode after plugging a new monitor (BugFix)

p-gentili:
## Description

The EDID cycling job is failing on laptops because Mirror mode is making the external monitor match the internal display resolution. Like this:

```
Testing EDID cycling on HDMI-1
switching EDID to 1920x1080
PASS
switching EDID to 1280x1024
FAIL, got 1920x1080 but 1280x1024 expected
switching EDID to 2560x1440
FAIL, got 1920x1080 but 2560x1440 expected
```

`gnome-randr` can't be used for applying monitor configurations, and in general having two utilities (xrandr and gnome-randr) for x11 and wayland doesn't help with maintainability of this test.

This PR relies on DBus and Mutter to retrieve the current monitor configuration and apply extended mode every time we switch EDID.

## Resolved issues

Resolves [ZAP-677](https://warthogs.atlassian.net/browse/ZAP-677)

## Documentation

Up to date

## Tests

Tested on [202302-31240](https://certification.canonical.com/hardware/202302-31240) ~~side-loading the provider~~ running from source. Previously failing like above ([ref](https://certification.canonical.com/hardware/202302-31240/submission/346523/test/176564/result/38078692/)).

[ZAP-677]: https://warthogs.atlassian.net/browse/ZAP-677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


checkbox-support/checkbox_support/dbus/monitor_config.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+            )
+            position_x += max_resolution[1]
+
+        self._apply_monitors_config(state[0], extended_logical_monitors)
+
+    def _get_current_state(self) -> GLib.Variant:
+        """
+        Run the GetCurrentState DBus request and assert the return
+        format is correct.
```
Hook25, line 87: > assert the return  format is correct.

Does it?

checkbox-support/checkbox_support/dbus/monitor_config.py
```diff
@@ -0,0 +1,132 @@
...
+            object_path=self.OBJECT_PATH,
+            interface_name=self.INTERFACE,
+            cancellable=None,
+        )
+
+    def get_current_resolutions(self) -> Dict[str, str]:
+        """Get current active resolutions for each monitor."""
+
+        state = self._get_current_state()
+        return {
```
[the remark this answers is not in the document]
Hook25: Applies to the whole file, namely, when you use an item it would be useful to do so by name. I would use a namedtuple and "waste" one operation to caste the state arr to it (`MyType(*state)`) else this is very hard to read imo.

Alternatively you can use named variables/enums as indeces, but it is not as good/easy to read imo

checkbox-support/checkbox_support/dbus/monitor_config.py
```diff
@@ -0,0 +1,132 @@
...
+            return max_resolution_mode, max_resolution[0], max_resolution[1]
+
+        return {monitor[0][0]: get_max(monitor[1]) for monitor in state[1]}
+
+    def _apply_monitors_config(self, serial: str, logical_monitors: List):
+        """Apply the given monitor configuration."""
+        self._proxy.call_sync(
+            method_name="ApplyMonitorsConfig",
+            parameters=GLib.Variant(
+                "(uua(iiduba(ssa{sv}))a{sv})",
```
Hook25, line 121: Please inline an explanation for what this magic is

Hook25 (changes requested):
A few changes below, but I've tried both this and the "original" script on my machine and it doesn't seem to work with i3 (unsurprisingly, as it doesn't use Mutter). My question is, how limiting is this? We do use Checkbox on non-mutter platforms as well (like Mir for example). Could this be a problem? Is it possible to somehow calculate the `OBJECT_PATH`, `NAME` and `INTERFACE` to "fix" this?

p-gentili:
> A few changes below, but I've tried both this and the "original" script on my machine and it doesn't seem to work with i3 (unsurprisingly, as it doesn't use Mutter). My question is, how limiting is this? We do use Checkbox on non-mutter platforms as well (like Mir for example). Could this be a problem? Is it possible to somehow calculate the `OBJECT_PATH`, `NAME` and `INTERFACE` to "fix" this?

With the current implementation, relying on xrandr and gnome-randr, we are covering everything X11 + Gnome Wayland. With the latest few commits I re-included support for X11 via `xrandr` so that the coverage is the same. I still prefer to use Mutter over xrandr on Gnome because I feel like it should handle corner cases better (let me know if you're against this choice). We'll need to add support for other Wayland compositors, such as Mir or wlroots-based.

Got part of the `xrandr` parser from https://github.com/canonical/checkbox/pull/1229 and added support for setting a new display configuration, in extended mode.

Both handlers inherit from the same abstract class so that the interface is the same and the EDID test itself doesn't care about the actual system.

Tested on:
- http://certification.canonical.com/hardware/202203-30003 (20.04, X11, GNOME and not GNOME)
- https://certification.canonical.com/hardware/202302-31240 (22.04, wayland, GNOME)

Hook25 (approved):
+1, ty for having updated this

PR #1291, opened 2024-06-13
Usb tests change base script (BugFix)

fernando79513:
## Description

The insert/remove/storage tests were failing on devices without `udisks2` since that was the main tool we used for these tests (removable_storage_watcher.py). Since the addition of core devices, we have a secondary script for this kind of tests (run_watcher.py) that just looks at journalctl messages instead of requiring `udisks2` to asses the insertion/removal of a USB device.

The main requirement for this PR was to use `run_watcher.py` only for devices without `udisks2`, but I extended the usage of this script for all the devices, but only for the jobs that were already using `run_watcher.py` on ubuntu core.

We did the same for the "usb_read_write.py" script.

I've also modified both of the scripts to fix some bugs and increase the clarity.

The `removable_storage_test.py` is still used for some automated tests, this PR focuses on removing the `udisks2` from tests that were already run on ubuntu_core.

I also removed the `usb/disk_detect` job, since it seemed to be a copy of usb/detect and it was not included in any other test-plan
 
 ## Resolved issues

https://warthogs.atlassian.net/browse/CER-2624
https://github.com/canonical/checkbox/issues/1029

## Tests

A selection of the tests have been tested on core and on classic.
The thunderbolt tests have not been tested, since I don't have access to a thunderbolt drive.
To run the modified tests:
```
checkbox-cli run com.canonical.certification::usb/insert
checkbox-cli run com.canonical.certification::usb/remove
checkbox-cli run com.canonical.certification::usb/storage-automated
```
Also available for:
 - usb3
 - usbc
 - mediacard
 - thunderbolt

### DIFF
--- a/checkbox-support/checkbox_support/scripts/run_watcher.py
+++ b/checkbox-support/checkbox_support/scripts/run_watcher.py
@@ -67,8 +67,6 @@ class StorageWatcher(StorageInterface):
     """
     StorageWatcher watches the journal message and triggers the callback
     function to detect the insertion and removal of storage.
-
-
     """
 
     def __init__(self, testcase, storage_type, zapper_usb_address):
@@ -274,10 +272,17 @@ class MediacardStorage(StorageWatcher):
         super().__init__(*args)
         self.mounted_partition = None
         self.action = None
+        self.device = None
+        self.address = None
 
     def _validate_insertion(self):
-        if self.mounted_partition:
+        if self.mounted_partition and self.action == "insertion":
             logger.info("usable partition: {}".format(self.mounted_partition))
+            logger.info(
+                "{} card inserted. Address: {}".format(
+                    self.device, self.address
+                )
+            )
             logger.info("Mediacard insertion test passed.")
 
             # backup the storage info
@@ -306,8 +311,18 @@ class MediacardStorage(StorageWatcher):
                 match.group("dev_num"), match.group("part_name")
             )
 
+        # Look for insertion action
+        insertion_re = re.compile(
+            "new (?P<device>.*) card at address (?P<address>[0-9a-fA-F]+)"
+        )
+        insertion_match = re.search(insertion_re, line_str)
+        if re.search(insertion_re, line_str):
+            self.action = "insertion"
+            self.device = insertion_match.group("device")
+            self.address = insertion_match.group("address")
+
         # Look for removal action
-        removal_re = re.compile("card [0-9a-fA-F]+ removed")
+        removal_re = re.compile("card ([0-9a-fA-F]+) removed")
         if re.search(removal_re, line_str):
             self.action = "removal"
 
@@ -327,6 +342,7 @@ class ThunderboltStorage(StorageWatcher):
         # The insertion will be valid if the insertion action is detected and
         # the mounted partition is found.
         if self.action == "insertion" and self.mounted_partition:
+            logger.info("usable partition: {}".format(self.mounted_partition))
             logger.info("Thunderbolt insertion test passed.")
 
             # backup the storage info
@@ -343,9 +359,6 @@ class ThunderboltStorage(StorageWatcher):
 
     def _parse_journal_line(self, line_str):
 
-        # Prefix of the thunderbolt device for regex matching
-        RE_PREFIX = "thunderbolt \d+-\d+:"
-
         # Extract the partition name. Looking for string like "nvme0n1: p1"
         part_re = re.compile("(?P<dev_num>nvme\w+): (?P<part_name>p\d+)")
         match = re.search(part_re, line_str)
@@ -354,9 +367,11 @@ class ThunderboltStorage(StorageWatcher):
                 match.group("dev_num"), match.group("part_name")
             )
 
+        # Prefix of the thunderbolt device for regex matching
+        RE_PREFIX = "thunderbolt \d+-\d+:"
+
         insertion_re = re.compile("{} new device found".format(RE_PREFIX))
         if re.search(insertion_re, line_str):
-            logger.debug("Found new thunderbolt device string in journal")
             self.action = "insertion"
 
         removal_re = re.compile("{} device disconnected".format(RE_PREFIX))
--- a/checkbox-support/checkbox_support/scripts/run_watcher.py
+++ b/checkbox-support/checkbox_support/scripts/run_watcher.py
@@ -39,27 +39,27 @@ class StorageInterface(ABC):
     """
 
     @abstractmethod
-    def _check_logs_for_insertion(self, line_str):
+    def _parse_journal_line(self, line_str):
         """
-        callback that looks for the expected log lines in the journal during
-        the insertion test.
-        """
-        pass
+        Parse the journal line and update the attributes based on the line
+        content.
 
-    @abstractmethod
-    def _check_logs_for_removal(self, line_str):
-        """
-        callback that looks for the expected log lines in the journal during
-        the removal test.
+        :param line_str: str of the scanned log lines.
         """
         pass
 
     @abstractmethod
     def _validate_insertion(self):
+        """
+        Check if the that the storage was inserted correctly.
+        """
         pass
 
     @abstractmethod
     def _validate_removal(self):
+        """
+        Check if the that the storage was removed correctly.
+        """
         pass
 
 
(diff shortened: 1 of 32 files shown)


checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
-    def _callback(self, lines):
+    def _process_lines(self, lines):
         for line in lines:
             line_str = str(line)
             logger.debug(line_str)
-            self._storage_strategy.do_callback(line_str)
+            self.callback(line_str)
 
-    def _no_storage_timeout(self, signum, frame):
+    def _storage_info_helper(self, reserve, mounted_partition=""):
```
Hook25, line 107: ok this is very, very hard to understand. My best understanding is that this function basically dumps some infos in a file so that subsequent jobs can read them and try to use them, so the idea is that this is called, all jobs that test that specific partition are called and then this is called again to remove the infos from the file. 

If this is the correct interpretation consider doing the following:
- split this into two, the `reserve` arg only toggles between storing and removing, you can have two functions, one called `_store_storage_info` and one `_delete_stored_storage_info` 
- update the docstring, I tried several time to read it and it is not clear 
- also update the docstring as it documents a non-existing parameter that is used everywhere in this file (`storage_type`)

Hook25: also, storage type is gone and this is used to store more than one storage type. All of them are called `usb_insert_info`. I've read the TODO but is that something we can just TODO and leave as is?

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
+                os.path.join(plainbox_session_share, file_name), "w"
+            )
+            file_to_share.write(mounted_partition + "\n")
+            file_to_share.close()
+
+        # remove the back info
+        if not reserve:
+            file_to_share = os.path.join(plainbox_session_share, file_name)
+            with contextlib.suppress(FileNotFoundError):
+                os.remove(file_to_share)
```
Hook25, line 154: This is troubles waiting to happen, how did this job run to here without failing if the file was never actually created? I assume this would happen only if you crash while trying to insert but the dependency would not let us run right?

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
+                    file_name, plainbox_session_share
+                )
             )
-        )
-        sys.exit(1)
+            file_to_share = open(
+                os.path.join(plainbox_session_share, file_name), "w"
+            )
+            file_to_share.write(mounted_partition + "\n")
+            file_to_share.close()
```
Hook25, line 148: Minor:
```suggestion
            with open(
                os.path.join(plainbox_session_share, file_name), "w"
            ) as file_to_share:
                file_to_share.write(mounted_partition + "\n")
```

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
         part_re = re.compile("sd\w+:.*(?P<part_name>sd\w+)")
         match = re.search(part_re, line_str)
         if match:
-            self.MOUNTED_PARTITION = match.group("part_name")
+            self.mounted_partition = match.group("part_name")
 
     def _refresh_detection(self, line_str):
         """
-        refresh values of the dictionary FLAG_DETECTION.
+        refresh values with the lines from journal.
```
Hook25, line 241: Please update this docstring, my best understanding is that tries, incrementally, line by line, to see if there is something relevant to this class in the journal. If there is, `self` is updated with the new info

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
-        "device": {
-            "new high-speed USB device number": False,
-            "new SuperSpeed USB device number": False,
-            "new SuperSpeed Gen 1 USB device number": False,
-        },
-        "driver": {"using ehci_hcd": False, "using xhci_hcd": False},
-        "insertion": {"USB Mass Storage device detected": False},
-        "removal": {"USB disconnect, device number": False},
-    }
+    class Device(Enum):
```
Hook25, line 162: Please add a docstring to this and consider changing the name to... `DeviceName` or something similar

Minor: you can use `class DeviceName(str, Enum)` so that below you can simply say
```
for device_name in DeviceName:
  if device_name in line: ...
```

Hook25: ditto for the other lines, in general it is not clear what these are when you use them in the code

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
+        self.driver = None
+        self.action = None
 
-    def do_callback(self, line_str):
+    def callback(self, line_str):
         self._refresh_detection(line_str)
         self._get_partition_info(line_str)
         self._report_detection()
 
     def report_insertion(self):
```
Hook25, line 188: Why is this called `report_insertion`, this tests insertion when all needed infos are fetched from the journal. Also, add a docstr here as well

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -93,124 +90,171 @@ def run(self):
...
     def __init__(self, args):
         self.args = args
+        self.mounted_partition = None
+        self.device = None
+        self.number = None
+        self.driver = None
+        self.action = None
 
-    def do_callback(self, line_str):
+    def callback(self, line_str):
```
Hook25, line 183: Please add a docstring that explains how this works, this basically tries to get the detection infos, tries to get the partition info and then tries to see if it got all the info needed to run the test, and if it did, it does. Find some form that is not confused like I'm being here, and write it to a docstring

checkbox-support/checkbox_support/scripts/run_watcher.py
Hook25: Maybe I understood why this happens, if you have more than 1 job then only the first "remove" job is going to work. This is debeatably wrong though, as if if we have an ordering like the following, the process if broken: 
1. insert a
2. use info a
3. insert b
4. remove a
5. use info b <- BOOM

> Note: I don't know if this is what is going on here, just a theory, correct me if I'm wrong

checkbox-support/checkbox_support/scripts/run_watcher.py
```diff
@@ -252,9 +293,11 @@ def report_removal(self, line_str):
         if match:
             logger.info("Mediacard removal test passed.")
 
-        # remove the storage info
-        storage_info_helper(reserve=False, storage_type=self.args.storage_type)
-        sys.exit()
+            # Storage removal info
+            self._storage_info_helper(
```
Hook25, line 297: ditto for below, how does this work? it can't right? the function is not defined anywhere and we don't inherit from the usb class

checkbox-support/checkbox_support/scripts/run_watcher.py
Hook25, line 1: One additional note I didnt know where to put. Please don't pass args to the subclasses, rather pop out the individual parameters for each class. Each method reads directly from `self.args.[...]` making this a de-facto obscure (and hard to refactor) lump of arguments. I would rather see them individualy added to only the classes that need them

Hook25 (changes requested):
A few comments below, only major is that I think this actually breaks non-usb tests, please check that the moving of that function inside the usb class didnt break the others

fernando79513 (commented):
I tried to follow most of the suggestions and make the code a bit more clear.

LiaoU3:
Verified pass for 22.04 for thunderbolt ssd testing

https://certification.canonical.com/hardware/202012-28495/submission/380910/

checkbox-support/checkbox_support/scripts/run_watcher.py
fernando79513: I have created an jira ticket to follow up on this issue https://warthogs.atlassian.net/browse/CHECKBOX-1497

Hook25 (approved):
+1, tyvm for working through this huge thing

PR #1292, opened 2024-06-14
Fix the c binary install path in pc-sanity (Bugfix)

xanthein:
## Description
Fix the compiled c binary file install path


binli (approved):
LGTM

PR #1297, opened 2024-06-18
Replaced open(Path, ...) with Path.open(...) (BugFix)

p-gentili:
## Description

Built-in `open()` function does not support a `pathlib.Path`  as input prior to  [Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#pep-519-adding-a-file-system-path-protocol). This PR fixes the problem using the `Path.open()` method instead.

## Resolved issues

N/A

## Documentation

N/A

## Tests

- Updated unit tests
- Tested locally running `bluetooth/detect`


Hook25 (approved):
+1, ty python3.5

PR #1299, opened 2024-06-19
plainbox-provider-pc-sanity (noble): depends openscap-scanner as well (BugFix)

cyruslien:
## Description
In Noble, the oscap command comes from package openscap-scanner, depends on it as well.


binli (approved):
LGTM.

PR #1302, opened 2024-06-20
Return error on timeout (bugfix)

Hook25:
## Description

If checkbox times out connecting to the agent, it prints an error but exists with return code 0. This is an issue because we can't rely on `$?` in this situation (like our canary pipeline does) and it clearly a bug. This fixes said bug.

## Resolved issues

Fixes: CHECKBOX-1468

## Documentation

N/A

## Tests

This adds a test for the "new" situation


p-gentili (approved):
LGTM

PR #1304, opened 2024-06-21
Add missing checkbox dependency to pc-sanity (BugFix)

xanthein:
## Description
Add checkbox-provider-resource and checkbox-ng to pc-sanity depends


binli (approved):
LGTM

PR #1301, opened 2024-06-20
Support both networkd and NetworkManger in netplan (New)

seankingyang:
## Description
Netplan serves as a network configuration abstraction renderer and was officially released in the 24.04 LTS version.
Netplan functions as a configuration tool for back-end daemons managing network interfaces, such as NetworkManager or systemd-networkd. 
In essence, Netplan acts as a configuration tool for back-end daemons managing network interfaces, independent of specific technologies.

In this PR, modify the `wifi_client_test_netplan.py` script to make it support both networkd and NetworkManger.
- Introduce an input parameter in this script to determine whether the renderer should be networkd or NetworkManager.
- Include a new input parameter for the func: `generate_test_config` to create the configuration with the specified renderer.
- Enhance the func: `wait_for_routable` to support both networkd and NetworkManager.
- Implement the func: `_get_nmcli_state` to retrieve the connection state for NetworkManager.
- Integrate the func: `check_routable_state` to determine the appropriate `_get_{nmcli | networkctl}_state` based on the renderer parameter.
- Update the func: `perform_ping_test` to accommodate both networkd and NetworkManager. (Obtain the Gateway method)
-  Separate the gateway retrieval into a new function `get_gateway`, which will select the appropriate `_get_{networkctl | nmcli}_gateway` based on the renderer parameter.
- Introduce the func: `_get_networkctl_gateway` extracted from the func: perform_ping_test to support this functionality.
- Implement the func: `_get_nmcli_gateway` to obtain the gateway IP address from NetworkManager.
- Update the func: `print_journal_entries` to cater to both network and NetworkManager scenarios.

> When utilizing this script in a test job, the renderer cannot be retrieved as an input parameter.
> There is a limitation regarding checkbox usage: **the outcome of the bootstrap resource job cannot be utilized in the command field.**
> 
> An alternative approach is required until the following issues are resolved:
> - The resource job: net_if_management needs to be split into two parts, one for wireless and the other for ethernet.
> - The script: net_if_management.py should only output either wireless or ethernet information.
> - Revise all jobs that utilize the output from: net_if_management
> 
> WIFI jobs (nm and np) should avoid using device as the template-resource, instead opt for net_if_management_wireless (currently unavailable).

According to the above issue, it need to add the following action:

-  Add an input parameter: renderer for AutoDetect
    - Auto-detection involves checking all .yaml files to identify any specific renderer (networkd, NetworkManger) definitions.
[...]

PR #1306, opened 2024-06-21
netplan bugfix declare auth for non wpa3 (bugfix)

Hook25:
## Description

This is a bugfix ported from #1301

The issue is that by not declaring the auth kind, the netbplan is not correctly parsed.

## Resolved issues

Fixes: CHECKBOX-1471

## Documentation

N/A

## Tests

@p-gentili is testing this to see if it works in production, unit tests were updated


p-gentili (approved):
Tested on https://certification.canonical.com/hardware/201805-26252, works fine.

PR #1310, opened 2024-06-24
Remove poitless job from canary (bugfix)

Hook25:
## Description

This job fails on many machines, including some that are certified. I think we can remove it as it makes finding canary running candidates harder

## Resolved issues

Canary failing on target queue

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
LGTM+1!

PR #1312, opened 2024-06-24
Add readme file for cases and environ variable for ce-oem-provider (New)

stanley31huang:
## Description
Add readme file for cases and environ variable for ce-oem-provider
We would like to have a document to describe which environ variable has been defined in the checkbox-ce-oem provider

**AI**
The README files are currently generated by a script, and we plan to introduce another PR to facilitate their updates in the future. The idea is to run a script to generate the README files and compare the differences in the GitHub workflow when a PR is created (similar to the tox check). If any differences are found, the script will add a comment to the PR to notify the reviewer.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/cases_and_environ.md
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/cases_and_environ.md
@@ -1,0 +1,272 @@
+
+## <a id='top'>environ keys for crypto test</a>
+- HWRNG
+	- Affected Test Cases:
+		- [ce-oem-crypto/hwrng-current](#ce-oem-crypto/hwrng-current)
+
+## Detailed test cases
+### <a id='ce-oem-crypto/hwrng-current'>ce-oem-crypto/hwrng-current</a>
+- **environ :**  HWRNG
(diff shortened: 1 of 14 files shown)


pieqq (changes requested):
This must have required a lot of work, unless you had some kind of script to generate the content!

A few comments:

- In the "Affected Test Cases" sections, I would point directly to the job in the pxu file. You can use relative links, and GitHub will understand them. Doing this will avoid a lot of duplicated data (`environ, summary, description, command`) that will probably go out of date quickly
- How do you plan on maintaining this? Are other contributors to ce-oem-provider aware they have to update the associated `md` files if they introduce a new environment key? Maybe you can add this to the [README file](https://github.com/canonical/checkbox/blob/main/contrib/checkbox-ce-oem/README.md).

zongminl (approved):
As the first step of capturing environ variables into README files, this PR looks good enough.
Some environ variables still missing detailed descriptions, this will be something to be fixed in the future.

stanley31huang:
@pieqq  I have updated the description to have an followup action to implement a script to keep these document up to date. Could you please review this PR again?

pieqq (approved):
This PR fell out of my radar. But as discussed before, it's good to land as is, and we'll see where this takes us.

Issue #125, opened 2022-11-28
LP1910516: disk/hdd-parking failed on non-hdd system and cause the system enters "degraded"


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1910516

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| Confirmed	| 2021-01-07 09:42:37	| 10	| Medium	| False  	|

## Description
Here are two problems.

1. disk/hdd-parking needs to check the system has HDD(s) as extra "requires:"
2. the hdapds should not be installed in non-HDD system OR the hdapds.service should be disabled if there is no HDD.

* plainbox-provider-certification-client installed hdapsd as Recommends by default.[1]
* hdapds will create a systemd service "hdapds.service" and if there is no hdd in the system then the system will enter "degraded" mode. [2]
* disk/hdd-parking use the command "command: hdd_parking.py" and it use 
```
    try:
        hdapsd = Popen(
            ['/usr/sbin/hdapsd'], stdout=PIPE, stderr=PIPE,
            universal_newlines=True)
    except OSError as err:
```
without checking the hdd, it cause the test case failed[3].
* The DUT is only has nvme[4]

As least, the "disk/hdd-parking" needs to be executed only if the HDD presents.
I suggest to disable "hdapds.service" when there is no HDD in the system if the "plainbox-provider-certification-client" dependencies could not be checked in runtime.

[1]
$ apt show plainbox-provider-certification-client
Package: plainbox-provider-certification-client
...
Recommends: ..., hdapsd, ...

[2]
● hdapsd.service - hdapsd hard drive active protection system daemon
     Loaded: loaded (/lib/systemd/system/hdapsd.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Wed 2021-01-06 18:32:25 CST; 22h ago
       Docs: man:hdapsd(8)
    Process: 760 ExecStart=/usr/sbin/hdapsd (code=exited, status=1/FAILURE)
   Main PID: 760 (code=exited, status=1/FAILURE)

-- Logs begin at Wed 2020-12-30 20:24:13 CST, end at Thu 2021-01-07 17:35:58 CST. --
systemd[1]: Started hdapsd hard drive active protection system daemon.
hdapsd[760]: Wed Jan  6 18:32:25 2021: Could not detect any devices.
[...]

PR #1315, opened 2024-06-26
Remove hdapsd packaging dependency from base provider (Bugfix)

pieqq:
## Description

The HDD parking test requires HDAPSD (Hard Drive Active Protection
System Daemon) to be installed on the system. However, this only makes
sense on portable devices (laptops) using mechanical, rotational hard
drives.

Nowadays, most laptops are shipped with SSD or NVMe drives that have no
mechanical components.

By force-installing hdapsd, the associated service starts and
immediately fails on devices without any mechanical hard drives. This is
an issue for other tests, notably the poweroff and reboot tests, which
both check for failed services. This has a big impact on automated test
reviews in our lab.

This PR removes the packaging dependency. By doing so, the `disk/hdd-parking` job will be skipped by default since the `hdapsd` daemon won't be running. It can be manually installed on devices that require it to test it.

## Resolved issues

Fix https://github.com/canonical/checkbox/issues/125


Hook25 (approved):
ty for fixing this

PR #1316, opened 2024-06-27
[checkbox-ce-oem] Remove manifest from resource job (BugFix)

rickwu666666:
To prevent the issue of get empty manifest in very first run of checkbox test

## Description
A resource job that require manifest will cause menifest becom empty in the very first run of checkbox test. 
e.g. The checkbox test run right after system installed.


stanley31huang (approved):
LGTM

PR #1317, opened 2024-06-28
Checkbox24 promotion (infra)

Hook25:
## Description

We need to promote via the action checkbox24 as well (checkbox24 + uc24 + 24.04)


pieqq (approved):
+1, thanks!

PR #1323, opened 2024-07-05
Convert wan/3gpp-scan test as a automated test (New)

stanley31huang:
## Description
Convert wan/3gpp-scan test as a automated test (New)

## Resolved issues
[OEMQA-4494](https://warthogs.atlassian.net/browse/OEMQA-4494)

## Documentation
N/A

## Tests
Please refer following submission for the test results (w/ side loaded provider)
https://certification.canonical.com/hardware/202302-31253/submission/379312/ (Updated on 2024/07/08)

[OEMQA-4494]: https://warthogs.atlassian.net/browse/OEMQA-4494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/base/bin/wwan_tests.py
+++ b/providers/base/bin/wwan_tests.py
@@ -362,6 +362,45 @@ class ThreeGppConnection:
         sys.exit(ret_code)
 
 
+class ThreeGppScanTest:
+
+    def register_argument(self):
+
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            "hw_id", type=str, help="The hardware ID of the modem"
+        )
+        parser.add_argument(
+            "--timeout",
+            type=int,
+            default=300,
+            help="timeout for 3gpp-scan",
+        )
+        return parser.parse_args(sys.argv[2:])
+
+    def invoked(self):
+
+        args = self.register_argument()
+        ret_code = 1
+        try:
+            mm = MMCLI()
+            mm_id = mm.equipment_id_to_mm_id(args.hw_id)
+            _wwan_radio_on()
+            cmd = [
+                "mmcli",
+                "-m",
+                str(mm_id),
+                "--3gpp-scan",
+                "--timeout",
+                str(args.timeout),
+            ]
+            ret_code = subprocess.check_call(cmd)
+        except subprocess.CalledProcessError:
+            pass
+        _wwan_radio_off()
+        sys.exit(ret_code)
+
+
 class CountModems:
 
     def invoked(self):
@@ -462,6 +501,7 @@ class WWANTests:
             "count": CountModems,
             "resources": Resources,
             "3gpp-connection": ThreeGppConnection,
+            "3gpp-scan": ThreeGppScanTest,
             "sim-present": SimPresent,
             "sim-info": SimInfo,
         }
--- a/providers/base/bin/wwan_tests.py
+++ b/providers/base/bin/wwan_tests.py
@@ -427,11 +427,7 @@ class ThreeGppScanTest:
                     cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
                 )
                 print(ret_sp.stdout.decode("utf-8"))
-                if (
-                    ret_sp.returncode == 0
-                    and b"error" not in ret_sp.stdout.lower()
-                ):
-                    ret_code = 0
+                ret_code = ret_sp.returncode
         except subprocess.CalledProcessError:
             pass
 
--- a/providers/base/bin/wwan_tests.py
+++ b/providers/base/bin/wwan_tests.py
@@ -297,6 +297,16 @@ def _wwan_radio_off():
     print()
 
 
+def _wwan_radio_status():
+    print_head("Get radio status")
+    cmd = ["nmcli", "r", "wwan"]
+    ret_sp = subprocess.run(
+        cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, check=True
+    )
+    print(ret_sp.stdout.decode("utf-8"))
+    return ret_sp.stdout.decode("utf-8").strip()
+
+
 def _destroy_3gpp_connection():
     print_head("Destroy 3GPP Connection")
     cmd = ["nmcli", "c", "delete", GSM_CON_ID]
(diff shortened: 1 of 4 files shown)


providers/base/units/wwan/test-plan.pxu
```diff
@@ -16,6 +16,7 @@ _description: Automated wwan tests for Snappy Ubuntu Core devices
 include:
     # Note these tests require snap calling snap support
     wwan/detect
+    wwan/3gpp-scan-.*-auto
```
pieqq, line 19: Use the `template-id` instead of regex.
```suggestion
    wwan/3gpp-scan-manufacturer-model-hw_id-auto
```

providers/base/units/wwan/test-plan.pxu
```diff
@@ -36,6 +37,8 @@ _name: Automated wwan tests (after suspend)
 _description: Automated wwan tests for Snappy Ubuntu Core devices
 include:
     after-suspend-wwan/detect
+    after-suspend-wwan/check-sim-present-.*-auto
+    after-suspend-wwan/3gpp-scan-.*-auto
```
pieqq, line 41: Same here:

```suggestion
    after-suspend-wwan/check-sim-present-manufacturer-model-hw_id-auto
    after-suspend-wwan/3gpp-scan-manufacturer-model-hw_id-auto
```

providers/base/bin/wwan_tests.py
```diff
@@ -362,6 +362,45 @@ def invoked(self):
...
+            _wwan_radio_on()
+            cmd = [
+                "mmcli",
+                "-m",
+                str(mm_id),
+                "--3gpp-scan",
+                "--timeout",
+                str(args.timeout),
+            ]
+            ret_code = subprocess.check_call(cmd)
```
pieqq, line 397: Using `check_call`, you will only get back the return code of the `mmcli` command. It's fine if there is no useful feedback we could get from the output, but otherwise it might be interesting to capture the output in case the command fails. (I forgot what output `mmcli --3gpp-scan` provides, so I'm not sure if my comment makes sense :))

providers/base/tests/test_wwan_tests.py
```diff
@@ -109,3 +111,98 @@ def test_invoked_with_mmdbus(self, mock_mmdbus):
...
+            obj_3gppscan.invoked()
+
+        self.assertEqual(context.exception.code, 1)
+        self.assertTrue(mock_mmcli.called)
+        self.assertTrue(mmcli_instance.equipment_id_to_mm_id.called)
+        self.assertTrue(mock_r_on.called)
+        self.assertTrue(mock_r_off.called)
+        mock_subprocess.assert_called_with(
+            ["mmcli", "-m", "0", "--3gpp-scan", "--timeout", "200"]
+        )
```
pieqq, line 180: Given that you also test when a `CalledProcessError` is raised, I don't think this test is required, as it covers the same parts of your code.

(even if you remove it, the coverage for `wwan_tests.py` will still be 100%)

pieqq (changes requested):
A few comments below, and a change request to use `template-id`s instead of regex in the test plans.

providers/base/tests/test_wwan_tests.py
stanley31huang: no, the coverage will down to 89%. So that's why I added this unit test.

providers/base/bin/wwan_tests.py
stanley31huang: Here is some errors we might seen during the scanning network test, so we probably could check is `error` shown in the log.
```
# the modem is not available
admin@C0310A2043-00843:~$ sudo mmcli -m 0 --3gpp-scan --timeout 200; echo $?
error: couldn't find modem
1

# the modem is not enabled
admin@C0310A2043-00843:~$ sudo mmcli -m 1 --3gpp-scan --timeout 200; echo $?
error: modem not enabled yet
1

# Unable to scan available network within a period of time
admin@C0310A2043-00843:~$ sudo mmcli -m 1 --3gpp-scan --timeout 30; echo $?
error: couldn't scan networks in the modem: 'Timeout was reached'
1
```

providers/base/tests/test_wwan_tests.py
stanley31huang: here are some error symptom while scanning 3GPP network. So we could simply check if the `error` message exists or not.
```
# the modem is not enabled
admin@C0310A2043-00843:~$ sudo mmcli -m 1 --3gpp-scan --timeout 200; echo $?
error: modem not enabled yet
1

# Failed to scan available network within 100s
admin@C0310A2043-00843:~$ sudo mmcli -m 0 --3gpp-scan --timeout 100; echo $?
error: couldn't scan networks in the modem: 'Timeout was reached'
1
```

providers/base/bin/wwan_tests.py
```diff
@@ -383,21 +412,29 @@ def invoked(self):
...
+                ]
+                print_head("Scanning 3GPP available network")
+                print("+ {}".format(" ".join(cmd)))
+                ret_sp = subprocess.run(
+                    cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+                )
+                print(ret_sp.stdout.decode("utf-8"))
+                if (
+                    ret_sp.returncode == 0
+                    and b"error" not in ret_sp.stdout.lower()
```
pieqq, line 432: Why is this needed? Are there cases where the return code from the 3gpp scanning is `0` but there is still an error?

stanley31huang: just in case that the `3gpp-scan` will always return `0` no matter the connection is detected or not.

BTW, here is the output while the system can't scan any available connection. 
```
admin@C0310A2043-00843:~$ sudo mmcli -m 0 --3gpp-scan --timeout 200; echo $?
error: couldn't scan networks in the modem: 'Timeout was reached'
1
```

providers/base/bin/wwan_tests.py
```diff
@@ -362,6 +372,72 @@ def invoked(self):
...
+                ):
+                    ret_code = 0
```
stanley31huang, line 434:
```suggestion
                ret_code = ret_sp.returncode
```

providers/base/bin/wwan_tests.py
stanley31huang: Thanks @pieqq  for the suggestion, I have sent a fixed.

pieqq (approved):
Looks good to me, thanks for applying all the changes we discussed here and on Mattermost!

PR #1334, opened 2024-07-12
Add missing references to Ubuntu 24 test plans (Bugfix)

RadxaYuntian:
## Description

Radxa is an ODM for single board computers. We are currently looking to get our product Ubuntu certified, ideally with the latest version.  However, with the latest `checkbox24` snap installed, we are unable to see the 24.04 test plans listed in `checkbox.odm-certification` per our training material. We are able to see them listed in `checkbox.checkbox-cli` following [the official documentation](https://checkbox.readthedocs.io/en/stable/tutorial/using-checkbox/installing-checkbox.html).

Checking the code, it appears to be simply missing the correct reference when the folder structure was bootstrapped in #1201. This PR is an attempt to fix the existing configuration files based on the similar implementation of the older versions.

## Resolved issues

N/A

### DIFF
--- a/checkbox-snap/series_classic18/launchers/odm-certification
+++ b/checkbox-snap/series_classic18/launchers/odm-certification
@@ -5,12 +5,15 @@ stock_reports = text, submission_files
 
 [test plan]
 unit = com.canonical.certification::client-cert-odm-desktop-18-04-manual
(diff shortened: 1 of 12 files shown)


fernando79513 (approved):
LGTM +1!

PR #1340, opened 2024-07-17
[CE-OEM] Add rpmsg auto test into main ce-oem test plan (New)

rickwu666666:
## Description
Add rpmsg related into main ce-oem test plan.

## Tests
List bootstrapped.
https://pastebin.canonical.com/p/8523NKWwm9/


rickwu666666:
After discussing with Stanley, we concluded that RPMSG-related tasks are not suitable for inclusion in the CE-OEM test plan, as the M4 firmware could exhibit different behaviors. Therefore, I'll close this PR and apply the modifications in the Checkbox project.

Issue #233, opened 2022-11-28
LP1978265: iperf3 test will fail if CAN interface can't set to up


beliaev-maksim:
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1978265

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| New	| 2022-06-10 05:26:33	| 6	| Undecided	| False  	|

## Description
[Summary]
ethernet/iperf3 test result will be fail if DUT has CAN interface. And seems the iperf performance is normal, but it fail while trying to bring up CAN interface.

INFO:root:Testing eno1 against 10.102.88.25
INFO:root:Have successfully pinged 10.102.88.25 on eno1
INFO:root:-------------------- Test Run Number 1 --------------------
INFO:root:Using 1 thread.
INFO:root:Connecting to port 5201 on server....
INFO:root:Avg Transfer speed: 938.6555555555556 Mb/s
INFO:root:93.87% of theoretical max 1000 Mb/s
INFO:root:Average CPU utilization: 10.4%
INFO:root:
INFO:root:-------------------- Test Run Number 2 --------------------
INFO:root:Using 1 thread.
INFO:root:Connecting to port 5201 on server....
INFO:root:Avg Transfer speed: 938.8333333333334 Mb/s
INFO:root:93.88% of theoretical max 1000 Mb/s
INFO:root:Average CPU utilization: 10.0%
INFO:root:
INFO:root:-------------------- Test Run Number 3 --------------------
INFO:root:Using 1 thread.
INFO:root:Connecting to port 5201 on server....
INFO:root:Avg Transfer speed: 938.6444444444444 Mb/s
INFO:root:93.86% of theoretical max 1000 Mb/s
INFO:root:Average CPU utilization: 9.6%
INFO:root:
INFO:root:-------------------- Test Run Number 4 --------------------
INFO:root:Using 1 thread.
INFO:root:Connecting to port 5201 on server....
INFO:root:Avg Transfer speed: 938.8333333333334 Mb/s
INFO:root:93.88% of theoretical max 1000 Mb/s
INFO:root:Average CPU utilization: 9.5%
INFO:root:
RTNETLINK answers: Invalid argument
ERROR:root:Failed to restore can0:Command '['ip', 'link', 'set', 'dev', 'can0', 'up']' returned non-zero exit status 2.

Please refer to submission[1] for more detail.
[...]

Issue #401, opened 2023-03-31
Iperf3 tests are incorrectly failing if there are other Ethernet interfaces that cannot be up'd.


djacobs98:
### Bug Description

On some of the HW platforms I am testing, I am seeing additional Ethernet interfaces that cannot be used.  For example, one platform has a Fiber Ethernet card, but we don't have the media converters.    

So these interfaces appear to Ubuntu, but you can't use '$ sudo ip link set dev <interface> up' to bring them up.  
You'll get an error.

When the iperf3 tests under the Stress test plan run, the test brings down all other Ethernet interfaces except for the interface being tested.  After iperf3 finishes, the test tries to bring all interfaces back up.  But because these unusable interfaces cause an error when you try to bring them up, this error causes the entire iperf3 test to fail.

The iperf3 test should make a note of the interfaces that are actually up at the beginning of the test, and only bring those down and back up.  The rest should be ignored so they don't cause false failures.

### To Reproduce

1. Take a system with 2 or more interfaces.
2. Only connect a cable to 1 interface (e.g. enp1s0)
3. Run the iperf3 test on that interface (enp1s0)
4. iperf3 test will fail when it tries to restore the connection on the other, disconnected, interface.

### Environment

Checkbox snap on either Classic or Core.
System must have 2 or more Ethernet interfaces.

### Relevant log output

```shell
Launchpad : https://bugs.launchpad.net/lookout-canyon/+bug/2000283

[Checkbox job `com.canonical.certification::ethernet/iperf3_enp1s0` output]

stderr
------
WARNING:root:Removing iperf server 10.102.182.100 (10.102.182.100) from
WARNING:root:test list since it's not within 10.102.88.0/23.
WARNING:root:Removing iperf server 10.102.182.137 (10.102.182.137) from
WARNING:root:test list since it's not within 10.102.88.0/23.
WARNING:root:Removing iperf server 10.102.182.101 (10.102.182.101) from
WARNING:root:test list since it's not within 10.102.88.0/23.
INFO:root:Testing enp1s0 against 10.102.89.198
INFO:root:Have successfully pinged 10.102.89.198 on enp1s0
INFO:root:-------------------- Test Run Number 1 --------------------
INFO:root:Using 1 thread.
INFO:root:Connecting to port 5201 on server....
INFO:root:Avg Transfer speed: 939.1 Mb/s
INFO:root:93.91% of theoretical max 1000 Mb/s
INFO:root:Average CPU utilization: 3.0%
[...]

PR #1342, opened 2024-07-18
restructure the setup and teardown functions for network.py scripts (Bugfix)

stanley31huang:
## Description
There are some improvement in this PR as following:
1. restructure the setup and teardown functions with context manager library. 
2. not bring up the network interfaces if it was down before testing.
3. not shutdown the conduit network interface (master network interface) while testing with an user network interface (slave network interface) (for Linux DSA network)

## Resolved issues
https://github.com/canonical/checkbox/issues/233
https://github.com/canonical/checkbox/issues/401

## Documentation
Reference:
- https://dri.freedesktop.org/docs/drm/networking/dsa/dsa.html
- https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net

### Tested on system with multiple physical network
- all passed results: 
https://certification.canonical.com/hardware/202405-34042/submission/391153/
- eno3 network interface failed as expected (no ethernet cable connected), it's to simulate the test failed and make sure this scripts could restore the network configuration back. 
https://certification.canonical.com/hardware/202405-34042/submission/381308/

### Tested on system with DSA network
- the lan2 network interface failed as expected, it's to simulate the iperf server is not stable and make sure this scripts could restore the network configuration back.
https://certification.canonical.com/hardware/202405-34042/submission/381472/

### DIFF
--- a/providers/base/bin/network.py
+++ b/providers/base/bin/network.py
@@ -862,11 +862,15 @@ def setup_network_ifaces(
     if target_if_attrs["status"] == "down" and not turn_up_network(
         target_network, timeout
     ):
-        raise SystemExit("Failed to bring up %s interface", target_network)
+        raise SystemExit(
+            "Failed to bring up {} interface".format(target_network)
+        )
 
     if not underspeed_ok and check_underspeed(target_network):
         raise SystemExit(
-            "the network speed is incorrect for %s interface", target_network
+            "the network speed is incorrect for {} interface".format(
+                target_network
+            )
         )
 
     if (
@@ -891,7 +895,9 @@ def setup_network_ifaces(
             conduit_net, timeout
         ):
             raise SystemExit(
-                "Failed to bring up %s conduit interface", conduit_net[0]
+                "Failed to bring up {} conduit interface".format(
+                    conduit_net[0]
+                )
             )
 
     # Shutdown other network interfaces
@@ -901,7 +907,7 @@ def setup_network_ifaces(
             and toggle_status
             and not turn_down_network(iface)
         ):
-            raise SystemExit("Failed to shutdown %s interface", iface)
+            raise SystemExit("Failed to shutdown {} interface".format(iface))
 
 
 def restore_network_ifaces(cur_network_info, back_network_info, timeout):
--- a/providers/base/bin/network.py
+++ b/providers/base/bin/network.py
@@ -770,7 +770,7 @@ def wait_for_iface_up(iface, timeout):
     deadline = time.time() + timeout
 
     net_if = Interface(iface)
-    while (time.time() < deadline):
+    while time.time() < deadline:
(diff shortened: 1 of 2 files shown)


providers/base/bin/network.py
```diff
@@ -753,24 +767,208 @@ def make_target_list(iface, test_targets, log_warnings):
...
+        cur_network_info = get_network_ifaces()
+        recover_success = restore_network_ifaces(
+            cur_network_info, network_info, recover_timeout
+        )
+
+        # Restore routing table to original state
+        logging.debug("Restore routing table")
+        try:
+            # Harmless "RTNETLINK answers: File exists" messages on stderr
+            with open(os.devnull, "wb") as DEVNULL:
```
Hook25, line 958: Please use: https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL

providers/base/bin/network.py
```diff
@@ -753,24 +767,208 @@ def make_target_list(iface, test_targets, log_warnings):
...
+
+    finally:
+        cur_network_info = get_network_ifaces()
+        recover_success = restore_network_ifaces(
+            cur_network_info, network_info, recover_timeout
+        )
+
+        # Restore routing table to original state
+        logging.debug("Restore routing table")
+        try:
```
Hook25, line 956: Like the comment! ty, I would use `with suppress(CalledProcessError)` or alternatively... `run` istead of `check_call` there is no point to using `check_call` if you ignore the check!

Hook25 (changes requested):
For me it is very hard to understand if this is ok actually, I have a few nitpicks about the code but nothing major. The submissions you provide actually have 2 failed tests, so I'm not sure what is going on there. Could this just randomly fail on some machines? Did you test on just two? Can you please test this on more? This is not ran during sru right?

providers/base/bin/network.py
stanley31huang: thanks for the suggestion, I have applied it into the latest commit

providers/base/bin/network.py
stanley31huang: fixed

stanley31huang:
> For me it is very hard to understand if this is ok actually, I have a few nitpicks about the code but nothing major. The submissions you provide actually have 2 failed tests, so I'm not sure what is going on there. 

> Could this just randomly fail on some machines? Did you test on just two? Can you please test this on more? 

I am sorry for the confusion on those failed test cases, it's trying to simulate some test failed scenario and make sure the network configuration could be restore. I have added more description for those submissions with failed tests, and attached a new submission without failed tests. 

> This is not ran during sru right?

I think so, it's not include in the SRU test plan as this is only include in the stress test plan.

> 
> (Oh and fantastic job with the context manager, I really like that!)

Hook25 (approved):

PR #1343, opened 2024-07-18
Implement desktop environment detector (New)

tomli380576:
## Description

This PR adds a quick utility function to check if the machine has a desktop environment. It works by checking `dpkg`'s existence and looking for the `ubuntu-desktop` and `ubuntu-desktop-minial` packages.

### DIFF
--- a/checkbox-support/checkbox_support/scripts/image_checker.py
+++ b/checkbox-support/checkbox_support/scripts/image_checker.py
@@ -3,12 +3,12 @@
 #
 # Written by:
 #   Patrick Chang <patrick.chang@canonical.com>
+#   Zhongning Li <zhongning.li@canonical.com>
 
 import argparse
 from os.path import exists
 from checkbox_support.snap_utils.system import on_ubuntucore
-import shutil
-from subprocess import PIPE, run
+from subprocess import DEVNULL, run
 
 
 def get_type() -> str:
@@ -18,26 +18,24 @@ def get_type() -> str:
     return "core" if on_ubuntucore() else "classic"
 
 
-def has_desktop_environment() -> bool:
+def has_desktop_environment(
+    desktop_packages=["ubuntu-desktop", "ubuntu-desktop-minimal"]
+) -> bool:
     """
     Returns whether there's a desktop environment
     """
-    if not shutil.which("dpkg") or on_ubuntucore():
-        # core doesn't have dpkg
+    if on_ubuntucore():
+        # ubuntu core doesn't have a desktop env by default
         return False
 
-    # if we found any of these packages, we are on desktop
-    if (
-        run(
-            ["dpkg", "-l", "ubuntu-desktop"], stdout=PIPE, stderr=PIPE
-        ).returncode
-        == 0
-        or run(
-            ["dpkg", "-l", "ubuntu-desktop-minimal"], stdout=PIPE, stderr=PIPE
-        ).returncode
-        == 0
-    ):
-        return True
+    for package in desktop_packages:
+        if (
+            run(
+                ["dpkg", "-l", package], stdout=DEVNULL, stderr=DEVNULL
+            ).returncode
+            == 0
+        ):
+            return True
 
     return False
 
--- a/checkbox-support/checkbox_support/scripts/image_checker.py
+++ b/checkbox-support/checkbox_support/scripts/image_checker.py
@@ -7,16 +7,41 @@
 import argparse
 from os.path import exists
 from checkbox_support.snap_utils.system import on_ubuntucore
+import shutil
+import subprocess
+from subprocess import PIPE
+from typing import Literal
 
 
-def get_type():
+def get_type() -> Literal["core", "classic"]:
     """
     Return the type of image.
     """
     return "core" if on_ubuntucore() else "classic"
 
 
-def get_source():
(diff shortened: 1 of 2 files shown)


checkbox-support/checkbox_support/scripts/image_checker.py
```diff
@@ -7,16 +7,42 @@
...
 
 
-def get_source():
+def has_desktop_environment() -> bool:
+    """
+    Returns whether there's a desktop environment
+    """
+    if not shutil.which("dpkg") or on_ubuntucore():
+        # core doesn't have dpkg
+        return False
```
baconYao, line 27:
```suggestion
    if on_ubuntucore():
        # Ubuntu Core doesn't have a desktop environment by default. 
        return False
```

I think `dpkg` is the default command installed in Ubuntu. Therefore, I think we don't have to check it.

checkbox-support/checkbox_support/scripts/image_checker.py
```diff
@@ -7,16 +7,42 @@
...
+        ).returncode
+        == 0
+        or run(
+            ["dpkg", "-l", "ubuntu-desktop-minimal"], stdout=PIPE, stderr=PIPE
+        ).returncode
+        == 0
+    ):
+        return True
+
+    return False
```
baconYao, line 42:
```suggestion
    packages = ["ubuntu-desktop", "ubuntu-desktop-minimal"]
    for package in packages:
        result = run(["dpkg", "-l", package], stdout=DEVNULL, stderr=DEVNULL)
        if result.returncode == 0:
            return True
    return False
```

Here are some suggestions that you can take
- Use `DEVNULL` instead of `PIPE` since we don't care ant output.
- Return function once we get the positive outcome

checkbox-support/checkbox_support/scripts/image_checker.py
```diff
@@ -7,16 +7,42 @@
 import argparse
 from os.path import exists
 from checkbox_support.snap_utils.system import on_ubuntucore
+import shutil
+from subprocess import PIPE, run
```
baconYao, line 11:
```suggestion
from subprocess import DEVNULL, run
```

baconYao (commented):
Please see my suggestions.

checkbox-support/checkbox_support/scripts/image_checker.py
```diff
@@ -3,20 +3,44 @@
...
+        return False
+
+    for package in desktop_packages:
+        if (
+            run(
+                ["dpkg", "-l", package], stdout=DEVNULL, stderr=DEVNULL
+            ).returncode
+            == 0
+        ):
+            return True
```
Hook25, line 38: Minor: I think this is easier to read (If you take this you need to import `contextlib.suppress`, `subprocess.check_call` and `subprocess.CalledProcessError`)
```suggestion
        with suppress(CalledProcessError):
            check_call(["dpkg", "-l", package])
            return True
```

checkbox-support/checkbox_support/scripts/image_checker.py
```diff
@@ -42,12 +66,15 @@ def main():
...
     parser.add_argument("-s", "--source", action="store_true")
+    parser.add_argument("-d", "--detect_desktop", action="store_true")
     args = parser.parse_args()
 
     if args.type:
         print("type: {}".format(get_type()))
     if args.source:
         print("source: {}".format(get_source()))
+    if args.detect_desktop:
+        print("Has desktop environment? {}".format(has_desktop_environment()))
```
Hook25, line 77: The previous lines seem to be a resource-like output, should this be as well?

Hook25 (changes requested):
See below, major is the resource one, is this intended?

checkbox-support/checkbox_support/scripts/image_checker.py
tomli380576: By "resource-like" did you mean replacing L77 with:
```
has_desktop: true
```
so that it looks like a key-value pair, or did you mean changing the return value of `has_desktop_environment` to a string to match the other functions?

checkbox-support/checkbox_support/scripts/image_checker.py
tomli380576: Ohh this looks like a cooler
```
try:
    ...
except:
    passs
```
Thanks for the suggestion! :D

checkbox-support/checkbox_support/scripts/image_checker.py
Hook25: first option, it seems to me that the output is a (key value pair) that can be used (I don't know if it actually is used) to build a resource. With the output you added, that purpose would no longer work, and it would also be pretty strange to have such a discrepancy.

checkbox-support/checkbox_support/scripts/image_checker.py
Hook25: it is :)

checkbox-support/checkbox_support/scripts/image_checker.py
tomli380576: Makes sense, I can add a comment as well to let people know that we should print key-value pairs :+1:

Hook25 (approved):
+1, ty. well done!

PR #1355, opened 2024-07-22
Use the upstream v1.3.0 version of action build (infra)

Hook25:
## Description

We were using a custom fork of the action build in order to count the results of remote-build and fail the action if too few (or too many?) snaps were created. This is no longer needed as snapcraft 8 fails if any of the remote build fails.

I'm creating this PR because I suspect that the fact that the action-build is so out of date (was forked last year) is a contributing factor in the failure we are currently seeing in the self-hosted builds.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
LGTM +1!

PR #1359, opened 2024-07-23
Add additional gpu tests (New)

pedro-avalos:
## Description

- Update gpu-setup script to work with arm64
- Update gpu-setup script to build tests from cuda-samples
- Add 4 new gpgpu tests from cuda-samples
- Separate gpgpu automated tests from stress tests

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-967

## Documentation

N/A

## Tests

To run tests: install nvidia drivers; then run tools/gpu-setup bash script to install cuda-toolkit and make the cuda-sample tests.

Run `checkbox-cli run com.canonical.certification::gpgpu-automated`.

### DIFF
--- a/providers/gpgpu/tools/gpu-setup
+++ b/providers/gpgpu/tools/gpu-setup
@@ -43,7 +43,7 @@ echo "*  Adding nVidia package repository"
 # SAUCE: https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&target_distro=Ubuntu&target_version=1804&target_type=debnetwork
 
 OSRELEASE=ubuntu`lsb_release -r | cut -f2 |sed -e 's/\.//'`
-ARCH=`uname -i`
+ARCH=`uname -m`
 if [[ "$ARCH" =~ "aarch64" ]]; then
     ARCH="arm64"
 fi
--- a/providers/gpgpu/tools/gpu-setup
+++ b/providers/gpgpu/tools/gpu-setup
@@ -44,13 +44,29 @@ echo "*  Adding nVidia package repository"
 
 OSRELEASE=ubuntu`lsb_release -r | cut -f2 |sed -e 's/\.//'`
 ARCH=`uname -m`
-if [[ "$ARCH" =~ "aarch64" ]]; then
-    ARCH="arm64"
-fi
+case $ARCH in
+    arm)
+        ;&
+    aarch64_be)
+        ;&
+    aarch64)
+        ARCH="arm64"
+        ;;
+    x86_64)
+        ;;
+    *)
+        echo "ERROR: Unsupported architecture $ARCH"
+        exit 1
+        ;;
+esac
 REPO_URL="https://developer.download.nvidia.com/compute/cuda/repos/$OSRELEASE/$ARCH"
 
 # Import and verify cuda gpg key
 wget -O cuda-archive-keyring.gpg "$REPO_URL/3bf863cc.pub"
+if [[ $? -eq 8 ]] ; then
+    echo "ERROR: wget failed. $OSRELEASE/$ARCH not supported?"
+    exit 1
+fi
 gpg --no-default-keyring --keyring ./temp-keyring.gpg \
     --import cuda-archive-keyring.gpg
 gpg --no-default-keyring --keyring ./temp-keyring.gpg \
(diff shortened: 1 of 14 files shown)


providers/gpgpu/tools/gpu-setup
```diff
@@ -65,6 +66,17 @@ apt install -y --no-install-recommends cuda-toolkit
 CUDA_PATH=$(find /usr/local -maxdepth 1 -type d -iname "cuda*")/bin
 export PATH=$PATH:$CUDA_PATH
 
+# clone cuda-samples repo
+echo "*  Cloning cuda-samples repo"
+CUDA_SAMPLES_DIR=/opt/cuda-samples
```
Hook25, line 71: My best understanding is that these are executables of some sort that run a test. If that is the case, the correct location to put them is under `providers/gpgpu/bin` after you build them. This makes it so you can call them directly "by name" from a Checkbox test (as the bin directory is automagically added to your PATH). If you do it this way, they will never work outside of a debian package/source run. If you need some data along with the binaries you should store it in `providers/gpgpu/data` (you can get the path of this directory from a test via the `PLAINBOX_PROVIDER_DATA` variable)

Also note that we support more than arm64 and amd64 (we build for a lot of other archs like armhf but also riscv64 or ppc, what does the script do under those circumstances?)

Hook25 (changes requested):
Please consider the comments below. In general we have to consider distribution when adding dependencies to providers and do so carefully, there are a lot of moving parts. 

Consider also rebasing this PR, you will get a new pipeline that builds the packages and checks that everything is allright.

providers/gpgpu/tools/gpu-setup
```diff
@@ -43,16 +43,17 @@ echo "*  Adding nVidia package repository"
...
+    ARCH="arm64"
 fi
+REPO_URL="https://developer.download.nvidia.com/compute/cuda/repos/$OSRELEASE/$ARCH"
+PINFILE="cuda-$OSRELEASE.pin"
 wget -O /etc/apt/preferences.d/cuda-repository-pin-600 "$REPO_URL/$PINFILE"
-wget -O /etc/apt/trusted.gpg.d/$KEYFILE "$REPO_URL/$KEYFILE"
-add-apt-repository -y "deb http://developer.download.nvidia.com/compute/cuda/repos/$OSRELEASE/x86_64/ /"
+curl -s -L "$REPO_URL/7fa2af80.pub" | tee /etc/apt/trusted.gpg.d/cuda-af80.asc
+# Had to add the next line because they now have a new key in the repos
+curl -s -L "$REPO_URL/3bf863cc.pub" | tee /etc/apt/trusted.gpg.d/cuda-63cc.asc
```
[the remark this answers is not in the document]
pedro-avalos: This should now be resolved in commit 9fe8e20

providers/gpgpu/tools/gpu-setup
pedro-avalos: The script is now building into `providers/gpgpu/bin`. I've added wrappers to address some limitations of nvidia's cuda-samples programs. See aca24be.

As to the architectures, the architecture used to be hardcoded to x86_64 before. They support arm64 in *some* cases, but not in others (eg Ubuntu 24.04); since it technically is supported in some cases, I added the option to use arm64, but I'm not sure if that's the best option. Let me know what you think (ie if we should limit this provider to x86_64 for now). See a854201.

Hook25 (changes requested):
Ok the changes are taking shape. Well done!

I've tested the snap builds on your branch and they don't work anymore. This is the "tail" of the log that leads to a failure
```
Configuring system for GPU Testing
**********************************
*
*  Testing network connectivity
../../tools/gpu-setup: line 10: ping: command not found
ERROR: This script requires internet access to function correctly
make: *** [gpu-setup] Error 1
../../src/Makefile:5: recipe for target 'gpu-setup' failed
Failed to run 'override-build': Exit code was 2.
Build failed
```

To reproduce this you can do the following
```
cd checkbox-core-snap; ./prepare.sh series16 
cd series16
snapcraft remote-build
```

This should fail with the same error I'm giving you. It is crucial that you use remote-build as the image LP uses is slightly different to the one you have in lxd

As to how to fix this, try to add ping to the build-depends of the part. If that works please do also test series18, 20, 22 and 24. Note that for 24 the build process is slightly different (you have to use snapcraft8_prepare.sh and commit the changes, you can then reset HEAD~1 once you are done).

Hook25 (approved):
As per our agreement, this is way too (legacy) broken to fix it all without a follow up and we need to decide what is the best way to do that. For now we land this excluded from the provider building system.

Please do create the cards in jira to highlight the issue at hand and we will discuss them and schedule them in a future pulse

PR #1361, opened 2024-07-24
removed python3-systemd and using libsystemd instead (BugFix)

fernando79513:
## Description

The systemd-python package, that was required to use the systemd library in python, required libsystemd-dev to be installed, but we were not installing that library during the build of the snaps.
libsystemd-dev has been added to the build-packages and libsystemd0 to the stage-packages.

The systemd-python version has been updated from 232 to 233 to fix an issue when running the 16.04 snap 

## Tests

All the snaps have been built remotely successfully.
Tested that the job can be executed in all the snaps.


Hook25 (approved):
+1, tyvm for taking this on

PR #1363, opened 2024-07-24
Fix systemd debian package (BugFix)

fernando79513:
## Description
Hopefully the last PR to fix the debian package dependencies related to the USB tests change


Hook25 (approved):
python3-systemd is becoming worse than setuptools-scm :(

PR #1364, opened 2024-07-24
Move all supported build to snapcraft8 (infra)

Hook25:
Note: 16 and 18 remain there as they are not supported by snapcraft8

## Description

We have to transition to snapcraft8 eventually and given that we are experiencing significant delays due to an internal issue, this seems like the best time to experiment. This transitions 20 and 22 to snapcraft 8 builds, making arm64 and amd64 self hosted. 18 and 16 were not moved as they are no longer supported by snapcraft.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
LGTM +1!

PR #1365, opened 2024-07-25
Add multi head test for 24.04 (BugFix)

hanhsuan:
## Description
The monitor multi-head tests are missed in the 24.04 test-plan and needed to add back.
At the same time, we only focus on testing output of one dGPU and the NVlink isn't supported to output on two dGPUs at the same time. Create a new job without the complexity of template will be better choice.

## Resolved issues

No

## Tests

Have sideloaded the provider to check the multi-head test cases are exist or not on Desktop(exist) and laptop(not exist).


Hook25 (approved):
+1, this is a manual test so it wont interact with SRU. LGTM

PR #1369, opened 2024-07-25
New debian validation pipeline (infra)

Hook25:
## Description

This new workflow builds the debian packages when relevant files are changed.

## Resolved issues

We have broken the debian packages twice in the last 2 days

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/.github/workflows/deb_validator.yaml
+++ b/.github/workflows/deb_validator.yaml
@@ -25,7 +25,6 @@ on:
       - providers/certification-client/**
       - providers/certification-server/**
       - providers/docker/**
-      - providers/genio/**
       - providers/gpgpu/**
       - providers/iiotg/**
       - providers/resource/**
@@ -52,7 +51,6 @@ jobs:
           - providers/certification-client
           - providers/certification-server
           - providers/docker
-          - providers/genio
           - providers/gpgpu
           - providers/iiotg
           - providers/resource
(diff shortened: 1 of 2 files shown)


.github/workflows/deb_validator.yaml
```diff
@@ -0,0 +1,68 @@
+name: Verify debian packages
+
+on:
+  push:
```
p-gentili, line 4: Any situation where we would push w/o PR?

.github/workflows/deb_validator.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - providers/tpm2/**
+  workflow_dispatch:
+
+jobs:
+  deb_validation:
+    strategy:
+      fail-fast: false
+      matrix:
+        # Version of Ubuntu
+        # note: no 16, we dont have 16 debs anymore
```
p-gentili, line 37:
```suggestion
        # note: no 16, we don't support debs for 16 anymore
```

p-gentili (approved):
It would be nice to filter out configurations from the matrix that are not involved. 

Since it only takes 4 minutes to run instead of hours debugging problems with dependencies, we can address this later if it's not trivial.

.github/workflows/deb_validator.yaml
Hook25: It is a nice to have to push the branch and work without a pr sometimes

Hook25:
I agree that this is subottimal as is, but it does give a nice barrier to trivial mistakes. I will search for a clean solution to only run what is necessary but the metabox one can't be easily adapted to this so I will do it in a follow up

p-gentili (approved):
+1

PR #1372, opened 2024-07-26
Disable verbosity for remote-build action (infra)

Hook25:
## Description

When used with core20, `--verbosity` seems to fail. The reason is presumably that the option is forwarded to the runner but, given that it auto-detects the snapcraft version, fails because it tries to use it with snapcraft7 (where the option was simply `--verbose`).

## Resolved issues

Fixes: https://github.com/canonical/checkbox/actions/runs/10099135074/job/27927719163#step:9:322

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
LGTM +1!

PR #1379, opened 2024-07-30
Move all possible builds to self hosted (infra)

Hook25:
## Description

Currently, the core20 builds are broken due to snapcraft8 not supporting core20 remote builds anymore. Additionally, many of the things we build via remote build could actually be partially built via self hosted runners. 
This PR does the following: 
- Move core18 frontend uc+classic and runtime to self hosted for arm64 and amd64
- Move core20 armhf builds back to the old workflow
- Minor: replace exclude with include (as it is more readable and correct)
- Minor: explicitly expand the SNAPCRAFT_BUILDER_ID variable as the action doesn't seem to expand envvars anymore

> Note: this doesn't move any core16 build to the self-hosted as the new version of the snapcore action doesn't seem to support it, see: https://github.com/canonical/checkbox/actions/runs/10143907516/job/28046289302

## Resolved issues

Resolves: https://warthogs.atlassian.net/browse/CER-2693

## Documentation

Inclusions documented in the comment above

## Tests

Full run (without submit) here: https://github.com/canonical/checkbox/actions/runs/10160203825/job/28097642975

Note: remote builds for 22 and 24 fixed in 3aac73e053d8649b5e8843bb77a40fa1d9c97f19
Note: duplicated builds of frontend amd64 18/20 fixed in 767777e824188fafd47a2c8db25334dc4e2f6675

### DIFF
--- a/.github/workflows/checkbox-core-snap-daily-builds.yml
+++ b/.github/workflows/checkbox-core-snap-daily-builds.yml
@@ -16,11 +16,11 @@ jobs:
           - releases: 20
             arch: arm64
           - releases: 20
-            arch: arm64
+            arch: amd64
           - releases: 18
             arch: arm64
           - releases: 18
-            arch: arm64
+            arch: amd64
     runs-on:
       group: "Canonical self-hosted runners"
       labels:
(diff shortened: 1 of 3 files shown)


pieqq (approved):
+1. Thanks for working on this!

PR #1380, opened 2024-07-30
Gate SRU with clear warning in the description (Infra)

Hook25:


.github/workflows/pr_validation.yaml
```diff
@@ -26,3 +24,77 @@ jobs:
...
+          checkbox-cli expand -f json com.canonical.certification::sru >> sru_testplan_main.json
+          # https://stackoverflow.com/a/57795761
+          # compare two jsons without comparing the order arrays but just
+          # the content of objects
+
+          echo
+          echo
+          echo "Difference of com.canonical.certification::sru-server (diff this_branch main)"
+          diff \
+            <(jq -S 'def post_recurse(f): def r: (f | select(. != null) | r), .; r; def post_recurse: post_recurse(.[]?); (. | (post_recurse | arrays) |= sort)' "sru_server_testplan_this_branch.json") \
```
pieqq, line 68: Can you just add a comment somewhere to explain what these few lines of madness are actually doing?

pieqq (changes requested):
Just one request, but otherwise it looks good to me!

.github/workflows/pr_validation.yaml
Hook25: yup, of course this needed an explanation! fixed

.github/workflows/pr_validation.yaml
```diff
@@ -64,11 +64,19 @@ jobs:
           echo
           echo
           echo "Difference of com.canonical.certification::sru-server (diff this_branch main)"
+          # the following line sorts the array of jobs/templates obtained from the pre/post checkout
+          # this is done because we want compare the list/content of jobs but not necessarely the order
+          # that for most jobs is not deterministic either way. The array keys are sorted as well
+          # as what counts is the content of each key (id, environ, command) and not the order
+          # the exporter prints them out
```
pieqq, line 71:
```suggestion
          # The following line sorts the array of jobs/templates obtained from the pre/post checkout.
          # This is done because we want compare the list/content of jobs but not necessarily the order
          # that for most jobs is not deterministic either way. The array keys are sorted as well
          # as what counts is the content of each key (id, environ, command) and not the order
          # the exporter prints them out.
```

pieqq (changes requested):

pieqq (approved):
Thanks!

PR #1386, opened 2024-08-05
Re-introduce default provider (infra)

Hook25:
## Description

The default provider constraint is used to automatically plug interfaces and install a sound default dependency for a frontend snap. We had to remove this as it doesn't allow us to build a snap if the default-provider doesn't have a "stable" version. Now we do have one, so we can re-introduce it

## Resolved issues

N/A

## Documentation

N/A

## Tests

Built and installed the snap. It now works as intended


pieqq (approved):
+1. Thanks for remembering this! :)

PR #1387, opened 2024-08-05
Add manifest entries for special key jobs (BugFix)

pieqq:
## Description

Instead of relying on the generic `has_special_keys` manifest entry, create a manifest entry for each of the dedicated keys test present in Checkbox (for instance volume up/down, audio mute, microphone mute, video out, etc.)

Use this manifest entries in the related `keys/*` jobs.

Remove the unused `has_special_keys` manifest entry.

Delete mentions to "manually skip that test" for the `keys/*` jobs, since this is now controlled by a manifest entry.

This looks like this on my device when running the 24.04 desktop full test plan:

image

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1321

## Tests

Tested locally on my DUT by running the 24.04 full test plan, and selecting the jobs in the Hotkey category. The section shown above appears in the System Manifest screen.

### DIFF
--- a/providers/base/units/keys/jobs.pxu
+++ b/providers/base/units/keys/jobs.pxu
@@ -3,16 +3,15 @@ category_id: com.canonical.plainbox::keys
 id: keys/lock-screen
 estimated_duration: 30.0
 imports: from com.canonical.plainbox import manifest
-requires: manifest.has_special_keys == 'True'
+requires: manifest.has_key_lock_screen == 'True'
 flags: also-after-suspend
 command: lock_screen_watcher.py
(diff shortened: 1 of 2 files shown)


hanhsuan:
There are more and more manifest. Is there an easy way to create manifest content for launcher without running the test plan?

Hook25 (approved):
From a checkbox point of view, fantastic, lets get a second opinion from someone from QA

LiaoU3 (approved):
It is great to seperate the manifest into several specific items!

PR #1393, opened 2024-08-06
EDID test fails at maximum resolution (BugFix)

p-gentili:
## Description

Another PR dedicated to the EDID job. It turns out that the new implementation (https://github.com/canonical/checkbox/pull/1286) shed some light on the other related problem, where older machines could read the maximum allowed resolution.

In case where the HW doesn't support an available mode from the monitor EDID, no preferred mode is available (which makes sense). This PR takes care of such scenario, attempting to configure the monitor with the maximum available resolution.
1. in case the maximum allowed resolution still matches what the test requested, the assertion over the actual configuration is performed as usual. Common example is with unsupported refresh rate.
2. in case the maximum allowed resolution is lower than requested, the testcase can be skipped since we can assume the HW does not support it. Common example is 2K resolution with HDMI < 1.3

## Resolved issues

Resolves [ZAP-678](https://warthogs.atlassian.net/browse/ZAP-678)

## Documentation

N/A

## Tests
Tested on some machines.

Previous behaviour:
- https://certification.canonical.com/hardware/202212-31024/submission/384021/test/176564/result/42874742/
- https://certification.canonical.com/hardware/202211-30885/submission/365364/test/176564/result/40446241/
- https://certification.canonical.com/hardware/202212-30973/submission/365373/test/176564/result/40447361/

Running from source:
```
Testing EDID cycling on HDMI-4
switching EDID to 1280x1024
PASS
switching EDID to 2560x1440
SKIP, max available was 2048x1152
switching EDID to 1920x1080
PASS
```

[ZAP-678]: https://warthogs.atlassian.net/browse/ZAP-678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/checkbox-support/checkbox_support/dbus/gnome_monitor.py
+++ b/checkbox-support/checkbox_support/dbus/gnome_monitor.py
@@ -68,7 +68,7 @@ class MonitorConfigGnome(MonitorConfig):
     def set_extended_mode(self) -> Dict[str, str]:
         """
         Set to extend mode so that each monitor can be displayed
-        at maximum resolution.
+        at preferred, or if missing, maximum resolution.
 
         :return configuration: ordered list of applied Configuration
         """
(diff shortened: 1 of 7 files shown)


Hook25 (dismissed):
See if you like the change proposed below but it makes sense

Hook25 (approved):

PR #1401, opened 2024-08-08
Fix the wrong namespace being used in checkbox-ce-oem launchers (BugFix)

zongminl:
## Description
Fix the wrong namespace being used in launchers

## Resolved issues
The current namespace being used in the launchers of checkbox-ce-oem snap are still `com.canonical.qa.ceoem` which should be using `com.canonical.contrib` already, this causes no test plans to be found when using test-runner launchers to start checkbox-ce-oem


LiaoU3 (approved):
Thank you for your hard work!
LGTM

PR #1415, opened 2024-08-12
Handle warnings in dkms status (BugFix)

jocave:
## Description

In some SRU runs with encountered `dkms status` output which included a warning. The dkms source seems to imply this is could be of interest to a user but not an error so therefore handling it in the test without a failures

## Tests

Extra unit test added and tested on machine dell-latitude7640-c31002

### DIFF
--- a/providers/sru/bin/dkms_build_validation.py
+++ b/providers/sru/bin/dkms_build_validation.py
@@ -65,6 +65,9 @@ def parse_dkms_status(dkms_status: str, ubuntu_release: str) -> List[Dict]:
         details, fullstatus = line.split(": ")
         if " " in fullstatus:
             (status, rest) = fullstatus.split(maxsplit=1)
+            logger.warning("dkms status included warning:")
+            logger.warning(" module: {}".format(details))
+            logger.warning(" message: {}".format(rest))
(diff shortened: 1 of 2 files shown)


providers/sru/bin/dkms_build_validation.py
```diff
@@ -62,7 +62,11 @@ def parse_dkms_status(dkms_status: str, ubuntu_release: str) -> List[Dict]:
     """
     kernel_info = []
     for line in dkms_status.splitlines():
-        details, status = line.split(": ")
+        details, fullstatus = line.split(": ")
+        if " " in fullstatus:
+            (status, rest) = fullstatus.split(maxsplit=1)
```
Hook25, line 67: This `rest` is not propagated anywhere, can we log it somehow?

Hook25 (dismissed):
LGTM, consider printing or dumping to stderr that `rest` if you want

providers/sru/bin/dkms_build_validation.py
jocave: Logged the message

Hook25 (approved):

PR #1426, opened 2024-08-19
Handle no installed dkms modules (BugFix)

jocave:
## Description

The "test_main_*" tests did not actually exercise anything because all calls where mocked and there wasnt any input provided.

The case where there was no dkms modules at all the system was not handled. Add a call_count on test main to ensure this addition wasnt skipping the apt log checking.

## Tests

Additional unittests with input from more release


Hook25 (approved):
+1, tyvm for the new tests as well

PR #1428, opened 2024-08-21
Fix metabox (infra)

Hook25:
## Description

This fixes two issues with metabox + 1 issue with a recently added test plan in the metabox provider
1. Metabox loads scenarios using a grossly outdated api that is now removed, this doesn't impact us at the moment but will very soon (discovered this because I tried to update the action to use ubuntu 24)
2. Action has a few workarounds that are needed (now more) that were un-documented. Now they are explained in the comments
3. The testplan landed was missing the "target" from the template_id and the bootstrap_include resource to make it run (the resource was pulled, but not during bootstrap, leading the template to not being initailized)

Root cause of the initial issue: setuptools updated and is now broken due to some incompatibility (I don't really know which) with importlib_metadata

## Resolved issues

Fixes: CHECKBOX-1536

## Documentation

This adds a few new comments to the workflow to explain why it is the way it is. 

## Tests

Passing test: https://github.com/canonical/checkbox/actions/runs/10487592300/job/29049421985

### DIFF
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -42,7 +42,7 @@ jobs:
       run:
         working-directory: metabox
     needs: metabox_run_required
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     env:
       # Workaround to get loguru colored output
       # See https://github.com/Delgan/loguru/issues/604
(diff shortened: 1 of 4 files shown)


pieqq (approved):
+1. Thanks for saving the day! :fire_extinguisher:  :fire:

PR #1430, opened 2024-08-21
Update audio playback/capture manifest definitions (bugfix)

### LINKED ISSUE #1329: The audio/detect_sinks based on incorrect manifest entry
### Bug Description

The `audio/detect_sinks` test is currently based on the `has_audio_playback` manifest entry. This seems to be incorrect, as this machine has a playback and passes all other audio tests but fails this one.

I believe this test should be based on another manifest entry.

Please let us know if you are making this change, so that we can organise the effort to update the manifest for this entry, as it will likely affect other jobs running in the lab.

### To Reproduce

1. Launch checkbox SRU test plan on a ASRock iEP-5000G Industrial PC (202211-30899)

### Environment

- OS: Ubuntu 22.04.4
- Checkbox Type: Deb
- Checkbox Version: 4.0.1~dev2~ubuntu22.04.1
- Hardware tested: Audio

### Relevant log output

```shell
Daemon is pulseaudio, use pulseaudio function
No available sinks found
```

### Additional context

_No response_

pieqq:
## Description

Make it clearer what `has_audio_playback` and `has_audio_capture` stand for.

These two manifest entries have been there for a long time, but it was never clear what they meant exactly. This led to issues such as the recent #1329. By having a more clear definition, testers know whether or not they should set these manifest entries or not.

image

## Resolved issues

Fix #1329

### DIFF
--- a/providers/base/units/audio/manifest.pxu
+++ b/providers/base/units/audio/manifest.pxu
@@ -7,13 +7,13 @@
 unit: manifest entry
 id: has_audio_playback
 _prompt: Does this machine have the following audio features?
-_name: Audio playback
+_name: Audio playback: Machine can emit sound. (For example, a desktop PC probably requires speakers connected to it)
 value-type: bool
 
 unit: manifest entry
 id: has_audio_capture
 _prompt: Does this machine have the following audio features?
-_name: Audio capture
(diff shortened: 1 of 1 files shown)


zongminl (approved):
Looks good to me.

Issue #815, opened 2023-11-07
CPU Clocktest Failed on HPE XD 665


mreed8855:
### Bug Description

On the HPE XD 665, the clocktest failed with the following error:

Testing for clock jitter on 96 cpus
sched_setaffinity: Invalid argument

There is a similar issue on the S390 but its unknown if they are related.

https://github.com/canonical/checkbox/issues/422

### To Reproduce

1. Install certification-canonical-server
2. Certify-2204

### Environment

22.04.2
5.15.0-87-generic

### Relevant log output

```shell
Error message:

Testing for clock jitter on 96 cpus
sched_setaffinity: Invalid argument
```

bladernr:
Need to add info about the CPUs in use here. This failed on s390 for architecture reasons. But this works on any other amd64 CPUs. So which CPUnamd which CPU family as a start. 

Then we need someone to look at the fairly simple source code and understand what the deal is with sched_setaffinity on this particular platform.

rodwsmith:
IBM claims to have found the problem, and a solution. Here's their explanation:

> Hi, thanks for the pointer to the source code. I figured out what is wrong.
> 
> The test program retroeves the number of cpus using "num_cpus = sysconf(_SC_NPROCESSORS_CONF);". After an
> upstream change in the kernel between 22.04 and 24.04, this returns now the number of physical cpus available.
> But these cpus do not need to be available to the system right now. Looping over num_cpus, the call to
> sched_setaffinity results in an error EINVAL which is explained with:
>
> EINVAL The affinity bit mask mask contains no processors that are currently physi‐
> cally on the system and permitted to the thread according to any restrictions
> that may be imposed by cpuset cgroups or the "cpuset" mechanism described in
> cpuset(7).
> 
> The correct way to retrieve the available number of cpus would be "num_cpus = sysconf(_SC_NPROCESSORS_ONLN);", as described here:
> 
> man sysconf
> - _SC_NPROCESSORS_CONF
> The number of processors configured. See also get_nprocs_conf(3).
> 
> - _SC_NPROCESSORS_ONLN
> The number of processors currently online (available). See also get_nprocs_conf(3).
> 
> With that change the test program runs as expected, and returns the following output:
> 
> root@a35lp52:~# gcc clocktest.c
> root@a35lp52:~# ./a.out
> Testing for clock jitter on 16 cpus
> PASSED: largest jitter seen was 0.000385
> 
> Testing clock direction for 5 minutes...
[...]

PR #1434, opened 2024-08-23
Fix for Clocktest Failure (Bugfix)

mreed8855:
The test program retroeves the number of cpus using "num_cpus = sysconf(_SC_NPROCESSORS_CONF);". After an
upstream change in the kernel between 22.04 and 24.04, this returns now the number of physical cpus available.
But these cpus do not need to be available to the system right now. Looping over num_cpus, the call to
sched_setaffinity results in an error EINVAL which is explained with:

EINVAL The affinity bit mask mask contains no processors that are currently physi‐
cally on the system and permitted to the thread according to any restrictions
that may be imposed by cpuset cgroups or the "cpuset" mechanism described in
cpuset(7).

The correct way to retrieve the available number of cpus would be "num_cpus = sysconf(_SC_NPROCESSORS_ONLN);", as described here:

man sysconf

   _SC_NPROCESSORS_CONF
    The number of processors configured. See also get_nprocs_conf(3).

    _SC_NPROCESSORS_ONLN
    The number of processors currently online (available). See also get_nprocs_conf(3).

    A Fix for issue https://github.com/canonical/checkbox/issues/815

### DIFF
--- a/providers/base/src/clocktest.c
+++ b/providers/base/src/clocktest.c
@@ -31,7 +31,7 @@ int test_clock_jitter(){
     unsigned cpu, num_cpus, iter;
     int failures = 0;
 
-    num_cpus = sysconf(_SC_NPROCESSORS_CONF);
+    num_cpus = sysconf(_SC_NPROCESSORS_ONLN);
     if (num_cpus == 1) {
         printf("Single CPU detected. No clock jitter testing necessary.\n");
         return 0;


mreed8855:
This still needs to be tested, I am creating the PR to see what other checks could fail.

Hook25 (approved):

PR #1441, opened 2024-08-29
Fix kill process issue in timeout.py (Bugfix)

eugene-yujinwu:
## Description
When the test case failed with timeout, it will reported as this:

```
INSERT NOW


Timeout: 30 seconds
/bin/sh: 1: kill: Illegal number: -
Traceback (most recent call last):
  File "/usr/bin/checkbox-support-run_watcher", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 93, in _f
    return run_with_timeout(f, timeout_s, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 78, in run_with_timeout
    raise TimeoutError("Task unable to finish in {}s".format(timeout_s))
TimeoutError: Task unable to finish in 30s
------------------------------------------------------------------------- >8 ---
Outcome: job failed
```

Something like this: **_/bin/sh: 1: kill: Illegal number: -:_**  will be reported.
It seems the timeout.py will try to use  `subprocess.run("kill -9 -- -{}".format(process.pid), shell=True)` to kill the process, but the subprocess doesn't work with "--" well.

The fix will try use os.kill to kill the process.

## Resolved issues

```
==============[ Running job 2 / 2. Estimated time left: 0:02:00 ]===============
-----------------[ USB 3.0 storage device insertion detected ]------------------
ID: com.canonical.certification::usb3/insert
Category: com.canonical.plainbox::usb
Purpose:

Check system can detect insertion of a USB 3.0 storage device.
NOTE: Make sure the USB storage device has a partition before starting
the test.

Steps:

1. Press continue.
2. Wait until the message "INSERT NOW" is displayed on the screen.
3. Connect USB 3.0 storage device.

Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]: 
... 8< -------------------------------------------------------------------------


INSERT NOW


Timeout: 30 seconds
/bin/sh: 1: kill: Illegal number: -
Traceback (most recent call last):
  File "/usr/bin/checkbox-support-run_watcher", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 93, in _f
    return run_with_timeout(f, timeout_s, *args, **kwargs)
[...]

PR #1442, opened 2024-08-29
Fix race condition timeout decorator (bugfix)

Hook25:
## Description

The timeout decorator seldomly fails to receive either a value or an exception from the subprocess. I am currently unable to reproduce this issue in any way but I've seen it happen on a very slow device running core20. Either way this improves the implementation in several key ways:
1. Don't check `empty` before getting from the queue: this is an anti-pattern and not guaranteed to work
2. Don't put a value we can determine is un-picklable in the queue: this crashes the writer thread of the queue with no error for `Queue`. 
3. Handle the case where the value is not picklable: this was handled the same as a normal exception, but it is a special case where we should really crash the program with a clear message
4. Bounded wait for gets: If the parent fails to receive anything after the child has died, we can't block the process on get. Rather what we do is raise a SystemExit to crash the parent and explain what happened
5. Best effort retry on get: I'm not sure if this is what is happening but given that the decoding is done in a separate thread, it could be that the first get is simply being called before it is done, so lets retry. This costs us (in those situations) at most 200ms, we can afford that
6. Close the queue in child process: This should also make the flush of the child thread more deterministic in some situations (or so some claim, from the history of the queue implementation, it is unclear to me how this could be the case)

This also contains a driveby fix to the issue described in this PR (removal of the `--` before the pid): https://github.com/canonical/checkbox/pull/1441

And a new test to prove that the solution in the decorator works (while using the builtin kill or any signal wont, to the best of my knowledge)

## Resolved issues

Fixes: CHECKBOX-1544

## Documentation

New comments to explain why some things are done (including why `subprocess.run` is used)

## Tests

New test that proves the kill behaviour works. Also new coverage for all the new lines

> Note: some lines appear uncovered because the test runs them in a separate process

### DIFF
--- a/checkbox-support/checkbox_support/helpers/timeout.py
+++ b/checkbox-support/checkbox_support/helpers/timeout.py
@@ -62,7 +62,7 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
                 exception_queue.put(
                     SystemExit(
                         "Function tried to return non-picklable value "
-                        "but this is not supported by the timeout decorator"
+                        "but this is not supported by the timeout decorator.\n"
                         "Returned object:\n" + repr(result)
                     )
                 )
@@ -111,7 +111,7 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
     with suppress(Empty):
         raise exception_queue.get_nowait()
 
-    # unpicklig is done in a separate thread, could it be un-scheduled yet?
+    # unpickling is done in a separate thread, could it be un-scheduled yet?
     with suppress(Empty):
         return result_queue.get(timeout=0.1)
     with suppress(Empty):
@@ -119,7 +119,7 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
 
     # this should never happen, lets crash the program if it does
     raise SystemExit(
-        "Function failed to propagate either a value or an exception\n"
+        "Function failed to propagate either a value or an exception.\n"
         "It is unclear why this happened or what the underlying function "
         "did."
     )
--- a/checkbox-support/checkbox_support/helpers/timeout.py
+++ b/checkbox-support/checkbox_support/helpers/timeout.py
@@ -23,13 +23,22 @@ Utility class that provides functionalities connected to placing timeouts on
 functions
 """
 import os
+import pickle
 import traceback
 import subprocess
 
+from queue import Empty
 from functools import partial
-from contextlib import wraps
 from unittest.mock import patch
-from multiprocessing import Process, SimpleQueue
+from contextlib import wraps, suppress
+from multiprocessing import Process, Queue
+
+
+def is_picklable(value):
+    with suppress(pickle.PicklingError), suppress(AttributeError):
+        _ = pickle.dumps(value)
+        return True
+    return False
 
 
 def run_with_timeout(f, timeout_s, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


eugene-yujinwu:
I test the timeout on my side with this fix, the -- issue is gone.  The test log is as below:

> 1. Press continue 
> 2. Wait until the message "INSERT NOW" is printed on the screen
> 3. Connect USB 2.0 storage device
> 
> Pick an action
>     => press ENTER to continue
>   c => add a comment
>   s => skip this job
>   q => save the session and quit
> [csq]: 
> ... 8< -------------------------------------------------------------------------
> 
> 
> INSERT NOW
> 
> 
> Timeout: 30 seconds
>   File "/usr/bin/checkbox-support-run_watcher", line 8, in <module>
>     sys.exit(main())
>   File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 137, in _f
>     return run_with_timeout(f, timeout_s, *args, **kwargs)
>   File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 107, in run_with_timeout
>     raise TimeoutError("Task unable to finish in {}s".format(timeout_s))
> TimeoutError: Task unable to finish in 30s
> ------------------------------------------------------------------------- >8 ---
> Outcome: job failed
> Finalizing session that hasn't been submitted anywhere: checkbox-run-2024-08-30T02.55.38
> ==================================[ Results ]===================================
>  ☒ : USB 2.0 storage device insertion detected

eugene-yujinwu:
Some more test :

Python 3.10.12 (main, Jul 29 2024, 16:56:48) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import subprocess
>>> ppid=683324
>>> subprocess.run("kill -9  -{}".format(ppid), shell=True)
/bin/sh: 1: kill: No such process

CompletedProcess(args='kill -9  -683324', returncode=1)

checkbox-support/checkbox_support/tests/test_timeout.py
```diff
@@ -137,38 +139,84 @@ def test_unpicklable_return_raises(self):
...
         with self.assertRaises(SystemExit):
             k()
+
+    def test_timeout_kills_subprocess_tree_on_timeout(self):
+        """
+        This tests that the timeout decorator not only kills the direct child
+        (function under test) but also any sub-process it has spawned. This
+        is done because one could `subprocess.run` a long-lived process from
+        the function and it could cause mayhem (and block the test session as
+        Checkbox waits for all childs to be done)
```
pieqq, line 165:
```suggestion
        Checkbox waits for all children to be done)
```

checkbox-support/checkbox_support/helpers/timeout.py
```diff
@@ -23,13 +23,22 @@
...
+from queue import Empty
 from functools import partial
-from contextlib import wraps
 from unittest.mock import patch
-from multiprocessing import Process, SimpleQueue
+from contextlib import wraps, suppress
+from multiprocessing import Process, Queue
+
+
+def is_picklable(value):
```
pieqq, line 37: I had to [dive in the multiprocessing module documentation](https://docs.python.org/3/library/multiprocessing.html#pipes-and-queues) to understand why you were checking the "picklability" of things.

It might be useful to provide a one-sentence docstring to explain the need for this function (and maybe a link to the stdlib doc).

checkbox-support/checkbox_support/helpers/timeout.py
```diff
@@ -40,45 +49,80 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
...
-                        "".join(
-                            [
-                                "Function failed but the timeout decorator is "
-                                "unable to propagate this un-picklable "
-                                "exception:\n"
-                            ]
-                            + traceback.format_exception(type(e), e, None)
-                        )
+                        "Function tried to return non-picklable value "
+                        "but this is not supported by the timeout decorator"
```
pieqq, line 65:
```suggestion
                        "but this is not supported by the timeout decorator.\n"
```

checkbox-support/checkbox_support/helpers/timeout.py
```diff
@@ -40,45 +49,80 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
...
-    if not exception_queue.empty():
-        raise exception_queue.get()
-    return result_queue.get()
+
+    with suppress(Empty):
+        return result_queue.get_nowait()
+    with suppress(Empty):
+        raise exception_queue.get_nowait()
+
+    # unpicklig is done in a separate thread, could it be un-scheduled yet?
```
pieqq, line 114:
```suggestion
    # unpickling is done in a separate thread, could it be un-scheduled yet?
```

checkbox-support/checkbox_support/helpers/timeout.py
```diff
@@ -40,45 +49,80 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
...
+    with suppress(Empty):
+        return result_queue.get(timeout=0.1)
+    with suppress(Empty):
+        raise exception_queue.get(timeout=0.1)
+
+    # this should never happen, lets crash the program if it does
+    raise SystemExit(
+        "Function failed to propagate either a value or an exception\n"
```
pieqq, line 122:
```suggestion
        "Function failed to propagate either a value or an exception.\n"
```

checkbox-support/checkbox_support/tests/test_timeout.py
```diff
@@ -137,38 +139,84 @@ def test_unpicklable_return_raises(self):
...
+        """
+
+        def inner(pid_pipe):
+            pid_pipe.send(os.getpid())
+            pid_pipe.close()
+            time.sleep(1e4)
```
pieqq, line 171: This line gave me goose bumps. What happens the day the timeout decorator is broken and this line is executed while running unit tests? It might block CI runs, no?

checkbox-support/checkbox_support/tests/test_timeout.py
```diff
@@ -137,38 +139,84 @@ def test_unpicklable_return_raises(self):
...
+            outer_p.start()
+            outer_p.join()
+
+        read, write = multiprocessing.Pipe()
+        with self.assertRaises(TimeoutError):
+            f(write)
+        with self.assertRaises(OSError):
+            pid = read.recv()
+            # give the process a few ms to wind down
+            time.sleep(0.01)
```
pieqq, line 190: Is it because of the note in [the multiprocessing doc](https://docs.python.org/3/library/multiprocessing.html#pipes-and-queues):

> After putting an object on an empty queue there may be an infinitesimal delay before the queue’s [empty()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.empty) method returns [False](https://docs.python.org/3/library/constants.html#False) and [get_nowait()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.get_nowait) can return without raising [queue.Empty](https://docs.python.org/3/library/queue.html#queue.Empty).

?

checkbox-support/checkbox_support/tests/test_timeout.py
Hook25: if the test is supposed to fail, it will fail. I'm pretty sure it wont hang the session, it could leak an inner process on failure, but there is no workaround to that, the test is to test that it will actually kill the whole tree.

checkbox-support/checkbox_support/tests/test_timeout.py
Hook25: No, this is a pipe, not a queue! This sleep is because we sent a signal to the subprocess, it could take a few ms for it to be fully killed and removed from the valid pid list. Namely, 

we kill the tree -> tree starts winding down -> we recv the inner pid 

If we check right away, the tree may still be winding down, so we interrupt for an instant. Then we try to see if the inner process still exists.

pieqq (approved):
Thanks for the feedback and the fixes!

checkbox-support/checkbox_support/helpers/timeout.py
Hook25: Done

PR #1447, opened 2024-09-03
Consider ID_INPUT_POINTINGSTICK as touchpad (BugFix)

khfeng:
Most resistive touchpads don't support ABS coordination nor multitouch. When both capabilities are absent, the touchpad gets tagged as ID_INPUT_MOUSE.

Since there's no such device as I2C mouse, if the touchpad is on I2C bus, it will also get tagged as ID_INPUT_POINTINGSTICK.

To really distinguish a true pointing stick and a resistive touchpad, the parser needs to find the slibling node created by HID Report Descriptor and ensure it's a "touchpad" (i.e. ABS or MT capability). It's not easy to traverse to a sibling node in the current parser, so simply treat ID_INPUT_POINTINGSTICK as resistive touchpad to resolve the issue.

### DIFF
--- a/checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
+++ b/checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
@@ -355,9 +355,9 @@ class TestUdevadmParser(TestCase, UdevadmDataMixIn):
         self.assertEqual(self.count(devices, "VIDEO"), 2)
         self.assertEqual(self.count(devices, "AUDIO"), 2)
         self.assertEqual(self.count(devices, "KEYBOARD"), 1)
-        self.assertEqual(self.count(devices, "TOUCHPAD"), 1)
+        self.assertEqual(self.count(devices, "TOUCHPAD"), 2)
         self.assertEqual(self.count(devices, "CARDREADER"), 1)
-        self.assertEqual(self.count(devices, "MOUSE"), 1)
+        self.assertEqual(self.count(devices, "MOUSE"), 0)
         self.assertEqual(self.count(devices, "CAPTURE"), 1)
         self.assertEqual(self.count(devices, "BLUETOOTH"), 1)
         self.assertEqual(self.count(devices, "WIRELESS"), 2)
@@ -1124,6 +1124,10 @@ class TestUdevadmParser(TestCase, UdevadmDataMixIn):
         devices = self.parse("CRYPTO_FDE", with_partitions=True)
         self.assertEqual(len(devices), 93)
         self.assertEqual(self.count(devices, "PARTITION"), 1)
+
+    def test_RESISTIVE_TOUCHPAD(self):
+        devices = self.parse("RESISTIVE_TOUCHPAD")
+        self.assertEqual(self.count(devices, "TOUCHPAD"), 1)
 
     def test_two_dms_one_with_ubutnu_save(self):
         """
--- a/checkbox-support/checkbox_support/parsers/tests/udevadm_data/RESISTIVE_TOUCHPAD.txt
+++ b/checkbox-support/checkbox_support/parsers/tests/udevadm_data/RESISTIVE_TOUCHPAD.txt
@@ -1,0 +1,531 @@
+P: /devices/pci0000:00/0000:00:15.0
+M: 0000:00:15.0
+R: 0
+U: pci
+V: intel-lpss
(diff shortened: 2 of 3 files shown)


checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
```diff
@@ -355,9 +355,9 @@ def test_TOSHIBA_NVME(self):
         self.assertEqual(self.count(devices, "VIDEO"), 2)
         self.assertEqual(self.count(devices, "AUDIO"), 2)
         self.assertEqual(self.count(devices, "KEYBOARD"), 1)
-        self.assertEqual(self.count(devices, "TOUCHPAD"), 1)
+        self.assertEqual(self.count(devices, "TOUCHPAD"), 2)
         self.assertEqual(self.count(devices, "CARDREADER"), 1)
-        self.assertEqual(self.count(devices, "MOUSE"), 1)
+        self.assertEqual(self.count(devices, "MOUSE"), 0)
```
pieqq, line 360: This doesn't seem right.

I went and check what device this udev DB came from, and it is a [Dell Latitude 5580](https://www.pcmag.com/reviews/dell-latitude-5580) which comes with a trackpoint, similar to the ones on the Lenovo laptops.

I asked one of our colleagues and they provided a sample of a Lenovo unit with such a trackpoint:

```
4649 P: /devices/pci0000:00/0000:00:1f.4/i2c-0/0-0015/input/input8
4650 L: 0
4651 E: DEVPATH=/devices/pci0000:00/0000:00:1f.4/i2c-0/0-0015/input/input8
4652 E: SUBSYSTEM=input
4653 E: PRODUCT=18/4f3/67/0
4654 E: NAME="Elan TrackPoint"
4655 E: PROP=21
4656 E: EV=7
4657 E: KEY=70000 0 0 0 0
4658 E: REL=3
4659 E: MODALIAS=input:b0018v04F3p0067e0000-e0,1,2,k110,111,112,r0,1,amlsfw
4660 E: USEC_INITIALIZED=3293319
4661 E: ID_INPUT=1
4662 E: ID_INPUT_POINTINGSTICK=1
4663 E: ID_INPUT_MOUSE=1
4664 E: ID_SERIAL=noserial
4665 E: ID_PATH=pci-0000:00:1f.4
4666 E: ID_PATH_TAG=pci-0000_00_1f_4
4667 E: ID_FOR_SEAT=input-pci-0000_00_1f_4
4668 E: TAGS=:seat:
```

And, as confirmed with them, in this case, when ID_INPUT_POINTINGSTICK=1 and ID_INPUT_MOUSE=1, the TrackPoint should be a mouse, not a touchpad. (otherwise the touchpad test cases would be run twice when running Checkbox on this device, which doesn't really make sense)

pieqq (changes requested):
See below inline comment: it seems introducing this fix would create a regression on devices with a trackpoint, which is currently recognizes as a mouse, but would start being recognized as a touchpad if we landed your change.

Is there a better way to identify the model you are focusing on, to make sure it's detected as a touchpad nonetheless?

checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
khfeng: The question is, why should a trackpoint be a mouse?

khfeng:
> See below inline comment: it seems introducing this fix would create a regression on devices with a trackpoint, which is currently recognizes as a mouse, but would start being recognized as a touchpad if we landed your change.
> 
> Is there a better way to identify the model you are focusing on, to make sure it's detected as a touchpad nonetheless?

What makes a pointingstick a mouse?

checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
khfeng: Is it possible to add a separate test case for `pointingsitck`? If that's present then it's okay to have no touchpad?

fernando79513:
This PR impacted one of the devices that have been recently added to the SRU.
https://github.com/canonical/hwcert-jenkins-jobs/pull/796
Is the conversation still ongoing?
Was there any progress here?
What should be the next steps?
@khfeng @pieqq

Hook25:
I'm closing this due to lack of updates. Feel free to re-open this is you still think this is needed

Issue #1272, opened 2024-06-05
SD Card will not be tested in `disk/read_preformance_.*`


LiaoU3:
### Bug Description

If a DUT has SD Card installed with system, the job [disk/read_performance_.*](https://github.com/canonical/checkbox/blob/c894f519ee728ab7bb34b10d39387bd9bd57e7d4/providers/base/units/disk/jobs.pxu#L70C5-L70C27) will not be run on the SD Card.

The root cause is because the [template-filter](https://github.com/canonical/checkbox/blob/c894f519ee728ab7bb34b10d39387bd9bd57e7d4/providers/base/units/disk/jobs.pxu#L67) require `device.category == 'DISK'`, however, the output from [device](https://github.com/canonical/checkbox/blob/c894f519ee728ab7bb34b10d39387bd9bd57e7d4/providers/resource/jobs/resource.pxu#L125), it says the category of SD Card is `CARDREADER` as below.
```
path: /devices/platform/axi/ff9d0000.usb/fe200000.usb/xhci-hcd.1.auto/usb1/1-1/1-1.1/1-1.1:1.0/host0/target0:0:0/0:0:0:0
name: sda
bus: scsi
category: CARDREADER
driver: sd
product: Ultra HS-COMBO
vendor: Generic
product_slug: Ultra_HS-COMBO
vendor_slug: Generic
```

Submission: https://certification.canonical.com/hardware/202202-29984/submission/371552/

### To Reproduce

1. Prepare a machine with SD Card installed
2. checkbox.checkbox-cli list-bootstrapped com.canonical.certification::disk-cert-automated

### Environment

 - Checkbox Type: Snap
 - Checkbox Version: 4.0.0-dev264

### Relevant log output

```shell
ubuntu@kria:~$ checkbox-limerick-classic.checkbox-cli list-bootstrapped com.canonical.certification::disk-cert-automated
$PROVIDERPATH is defined, so following provider sources are ignored ['/snap/checkbox-limerick-classic/x1/providers/checkbox-provider-limerick', '/home/ubuntu/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop'] 
Skipped file: /snap/checkbox24/current/providers/checkbox-provider-base/units/stress/suspend_cycles_reboot.md
com.canonical.certification::executable
com.canonical.certification::disk/detect
com.canonical.plainbox::manifest
com.canonical.certification::disk/check-software-raid


ubuntu@kria:~$ lsblk 
[...]

PR #1451, opened 2024-09-04
Sd card not tested (BugFix)

fernando79513:
## Description

This ticket introduces some changes to the disk tests.
The first of them is to include SD cards into the "DISK" category so we can run disk performance tests in devices that have root mounted on an SD card. This behavior was purposely ignored before.
```
if self.driver:
...
    # Only consider eMMC internal drives as DISK.
    # As it's not possible to distinguish them from simple MMC
    # removable storage, only those with a partition mounted as /
    # will be considered.
    # To avoid categorizing SD cards as main storage and run heavy
    # tests on them (e.g on a pandaboard) only devices with the udev
    # property MMC_TYPE == MMC are accepted.
    if self.driver.startswith("mmc"):
        if self._mmc_type == "MMC" and find_pkname_is_root_mountpoint(
            self.name, self._lsblk
        ):
            return "DISK"
        else:
            return "CARDREADER"
```
But we have decided to include also SD cards into the "DISK" category so the disk tests are run also in these devices
We should keep in mind that these tests, particularly the storage test, put stress on the SD cards, which could potentially reduce their lifespan.

The second one is to ignore some disks for certain types of tests:
 - Disks with less than 2MB of memory are ignored for reading tests
 - `mtdblock` disks are ignored for storage tests.

We discussed moving the disk-detecting logic to a separate resource instead of using "udevadm.py", but I think we can do that in a follow-up ticket.

## Resolved issues

#1272 
https://warthogs.atlassian.net/browse/CHECKBOX-1458

## Documentation

N/A

## Tests

The tests have been run in a laptop and in a Xilinx device with the root mounted in the SD card:
 - Laptop: [submission_2024-09-04T09.49.17.777150.zip](https://github.com/user-attachments/files/16868355/submission_2024-09-04T09.49.17.777150.zip)
 - Xilinx: https://certification.canonical.com/hardware/202202-29984/submission/391263/


providers/base/units/benchmarks/jobs.pxu
```diff
@@ -8,7 +8,14 @@ template-id: benchmarks/disk/hdparm-read_name
...
 estimated_duration: 15.0
 user: root
-command: hdparm -t /dev/{name}
+command: 
+ # Check if the disk size is smaller than the threshold (2MB)
+ disk_size=$(blockdev --getsize64 /dev/{name})
+ if [ "$disk_size" -gt 2097152 ]; then
+    hdparm -t /dev/{name}
+ else
+    echo "Disk size is smaller than the threshold (2MB), ignoring the test"
```
Hook25, line 17: I don't want to be pedantic, but this doesn't pass the test. It makes it pass. To get the test to be skipped you can either write something smart in the resource expression or write here in 
```
    echo "{ 'outcome' : 'skip', 'comment' : 'Disk size is smaller than the threshold (2MB), ignoring the test'}" > $PLAINBOX_SESSION_SHARE/__result
```

Hook25 (approved):
I don't know if you want to use that feature, but we do have it

providers/base/units/benchmarks/jobs.pxu
fernando79513: I didn't know that that feature existed. It makes more sense indeed to skip these tests. I will do it like that.

fernando79513: I agree with you that the skipped tests should be marked as skipped, not passed. I looked into it last friday but it was not as easy as setting the  __result file, since we are only reading that one when we resume the session

PR #1460, opened 2024-09-09
Include image information in Checkbox submission (New)

Hook25:
## Description

New job in pc-saity called com.canonical.certification::info/image-url calls bash script called get-image-url.sh. Probably needs some work to port it to the system_information section of the submission file.

This creates a new system_information collector that collects the URL if possible (namely, if ubuntu-report is installed on the machine and if the dcd string is there and valid). This safely fails in unexpected situations reporting the error. This also stores (in the stderr of the collector) the full report non-destructively, so in any situation we can go back and re-interpret the data

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1541

## Documentation

N/A

## Tests

This also adds appropriate unit tests

### DIFF
--- a/checkbox-ng/plainbox/impl/session/system_information.py
+++ b/checkbox-ng/plainbox/impl/session/system_information.py
@@ -253,6 +253,18 @@ class InxiCollector(Collector):
         )
 
 
+class ImageInfoCollector(Collector):
+    COLLECTOR_NAME = "image_info"
+
+    def __init__(self):
+        super().__init__(
+            collection_cmd=[
+                str(vendor.IMAGE_INFO),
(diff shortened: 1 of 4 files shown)


pieqq (approved):
Looks good to me!

Does it mean we can remove the `get-image-url.sh` script and associated job in the sanity provider?

Hook25:
I wouldn't remove it for now, as that will disrupt them and this is intended to collect the information and not give immediate feedback to the user (as in, it wont fail publicly in the log, prompting you to maybe update the string and rerun it), so it may not fit their purpose

PR #1466, opened 2024-09-11
Workflow to dispatch jobs to the lab (infra)

Hook25:
## Description

We need to be able to run a testplan (or subset of a testplan) in the lab given a checkbox revision. This makes reviewing additions easier and faster. 

This PR introduces a new workflow (that is only half of this quest) to actually do the building/dispatching part where the machine is provisioned and checkbox is installed from source (leveraging debian packaging for now) to the desired revision. 

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1532

## Documentation

Action is documented where needed with comments (especially in the matrix section)

## Tests

Green run in the lab only running wifi tests on a machine provisioned with desktop jammy:
- https://github.com/canonical/checkbox/actions/runs/10920469454/job/30310381300

Using pipx
- https://github.com/canonical/checkbox/actions/runs/10941251260/job/30375463873

Renamed launchers:
- https://github.com/canonical/checkbox/actions/runs/10956261911/job/30421915057

Command to trigger another at will:
```
gh workflow run dispatch_lab_job.yaml --ref lab_job_dispatch -f 'matrix_to_create=[{ data_source: "distro: desktop-22-04-2-uefi", queue: "202012-28526", match: ".*wireless.*", test_plan: "com.canonical.certification::sru" }]'
```


tools/lab_dispatch/build_install_deb.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
```
boukeas, line 1: This might be controversial but I will go ahead and ask it: what is the advantage of this being a Python program instead of a bash script? The core of the code runs bash commands so I personally think it would actually be shorter and more readable if this were a bash script.

tools/lab_dispatch/resources/service_file.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
```
boukeas, line 1: There might be a reason for this filename that I don't know about, but would it be possible to rename the file to something like `checkbox-agent.service` or `checkbox-ng.service` or anything that is suggestive of which service it describes?

tools/lab_dispatch/generic_source.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    python3 -m venv --system-site-packages checkbox_venv
+    . checkbox_venv/bin/activate
+    pip install --upgrade pip setuptools
+    pip install checkbox/checkbox-ng
+    # retrieve manifest
+    MANIFEST_FILE=manifest.conf
+    fetch_manifest --manifest_file manifest.conf $CID $HEXR_DEVICE_SECURE_ID
+    if [ $? -ne 0 ]; then
+        echo "Using default manifest"
```
boukeas, line 58: Ha, this is funny. You might be interested in [this PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/583) (currently in draft). I think that, given the current status of things, it would be a good idea to deprecate the use of a default manifest.

tools/lab_dispatch/generic_source.yaml
```diff
@@ -0,0 +1,78 @@
...
+    _run python3 build_install_deb.py --clean checkbox/checkbox-ng \
+        checkbox/checkbox-support checkbox/providers/resource \
+        checkbox/providers/base checkbox/providers/sru
+    _run sudo systemctl restart checkbox-ng.service
+
+    git clone https://github.com/canonical/checkbox.git
+    git -C checkbox checkout $CHECKBOX_REVISION
+
+    python3 -m venv --system-site-packages checkbox_venv
+    . checkbox_venv/bin/activate
```
boukeas, line 51: Fine to reject this remark but I personally prefer the explicit `source` from `.`

tools/lab_dispatch/generic_source.yaml
```diff
@@ -0,0 +1,78 @@
...
+    pip install --upgrade pip setuptools
```
boukeas, line 52: This is really interesting. After some [conversations I had with Paul on this PR](https://github.com/canonical/hwcert-jenkins-tools/pull/36) the installer actually installs `pipx` and `python3-venv` as deb packages, i.e. not through `pip`. We might discuss which of the two approaches is best but what I am essentially saying is that you could probably use whatever the installer has already installed.

And I am wondering whether you can directly call `pipx install checkbox/checkbox-ng` without explicitly activating the virtual environment (as `pipx` will handle that for you). 

**Update**: Actually, `install_checkbox_agent_source` [does exactly that](https://github.com/canonical/hwcert-jenkins-tools/blob/61e32fe28d0060fabfe3621af2f43713d6f5398e/scriptlets/install_checkbox_agent_source#L48). Would you be able to use that scriptlet directly?

tools/lab_dispatch/generic_source.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    wait_for_ssh
+
+    _put $RESOURCES_PATH/build_install_deb.py :
+    _run git clone https://github.com/canonical/checkbox.git
+    _run git -C checkbox checkout $CHECKBOX_REVISION
+    _run python3 build_install_deb.py --clean checkbox/checkbox-ng \
+        checkbox/checkbox-support checkbox/providers/resource \
+        checkbox/providers/base checkbox/providers/sru
+    _run sudo systemctl restart checkbox-ng.service
```
boukeas, line 45: If you like, you can omit `.service`.

.github/workflows/dispatch_lab_job.yaml
```diff
@@ -1,14 +1,59 @@
...
+      run:
+        working-directory: tools/lab_dispatch
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install dependencies
+        run: |
+          sudo apt install gettext
       - name: Dispatch test in the lab and monitor it
```
boukeas, line 39: I would suggest using "Dispatch test in the lab and monitor it" in the next step. This one sounds more like "Build resources" or something like that.

tools/lab_dispatch/generic_source.yaml
```diff
@@ -0,0 +1,78 @@
...
+  $INPUT_DATA_SOURCE
+test_data:
+  attachments:
+    - local: "tools/lab_dispatch/resources/manifest.conf"
+      agent: "manifest.conf"
+    - local: "tools/lab_dispatch/resources/input.conf"
+      agent: "input.conf"
+    - local: "tools/lab_dispatch/resources/service_file.service"
+      agent: "service_file.service"
+    - local: "tools/lab_dispatch/build_install_deb.py"
```
boukeas, line 14: Not to action right now, but if there's any chance this `build_install_deb` could be useful in non-Checkbox workflows, then it might belong to the tools repo and eventually move there.

tools/lab_dispatch/resources/input.conf_template
```diff
@@ -0,0 +1,40 @@
...
+forced = yes
+
+[test selection]
+forced = true
+match=$INPUT_MATCH
+
+[ui]
+type = silent
+
+[environment]
```
boukeas, line 16: I have a couple of suggestions here:
- In the jobs I am aware of, the `[environment]` section is split into a separate file (which is _annoyingly_ named `checkbox.conf`). And, at some point in the near future, you may be receiving that separate environment file from an endpoint or a different source (i.e. not storing it in your repo) because there needs to be a single source of truth. So, in the interest of consistency with other jobs _and_ perhaps seeking to minimise future changes, it might be a good idea to split this file into the "head" and the environment section.
- Continuing from the previous suggestion, I am not too crazy about the name `input.conf_template`. I understand why you named it that way but I don't think it reflects the contents of the file. So, if you agree with splitting the file, my suggestions would be:
  - `head.template.conf` (later turning into `head.conf` after you've instantiated the input)
  - `environment.template.conf (later turning into `environment.partial.conf` after you've instantiated the input and `environment.conf` after the next instantiation on the agent)

tools/lab_dispatch/resources/service_file.service
Hook25: no reason, turning it into checkbox-ng.service

Hook25: Actually, removing this, as this is no longer needed. It was needed when I wanted to provision fully from source, but now that I use debs, I get the service "for free"

tools/lab_dispatch/generic_source.yaml
Hook25: for this PR I think it makes sense to have a default or if we even need it, a way to actually provide a manifest via a parameter because this is a testing workflow, so using it for new queues, or to test, for example, new manifest keys, is very much on the table. This could be a very interesting follow up direction actually.

Hook25: as for your pr, I agree that in sru it should fail and blow up if the manifest doesn't exist, else you will spend time chasing ghosts sometimes (or wondering why the test coverage is so low)

tools/lab_dispatch/generic_source.yaml
Hook25: ok

tools/lab_dispatch/generic_source.yaml
Hook25: no because it checks out to a revision, so I cant use the scriplet. I can use pipx, I used normal pip and a venv because I'm just more used to that. I can try this.

tools/lab_dispatch/generic_source.yaml
Hook25: I always include the extension out of habit

.github/workflows/dispatch_lab_job.yaml
Hook25: (I pulled out the actual thing the step did without changing the name, fixing!)

tools/lab_dispatch/resources/input.conf_template
Hook25: I think this is more complicated (as I need to do 2 substitutions or stack first and then substitute) and I don't see a benefit with it. Once we have an environment endpoint, we can just remove the envvars from the file and shift it to a fetch like we do with the manifests. I like the fact that the conf file is relatively small and readable, one can open it and modify it (for example to get some temporary testing done) with relative ease. I don't want to be in the situation where you are left wondering where are the values set and why.

tools/lab_dispatch/generic_source.yaml
Hook25: I don't see a plausible usage of this outside of checkbox as it is very checkbox-oriented right now, and will become more so in the future. I would leave this here. If we find we need it somewhere else, we can move it.

tools/lab_dispatch/build_install_deb.py
boukeas: I still think the Python script is unreadable but it's your call.

tools/lab_dispatch/resources/input.conf_template
boukeas: Would you consider renaming this single configuration file to something different? Right now, the naming trajectory is:

`input.conf_template` --> (partially instantiated in the runner)
`input.conf` --> (fully instantiated in the agent)
`test_conf.conf` --> (stacked with manifest)
`checkbox.conf`

_One_ suggestion would be:

`checkbox.no-manifest.template.conf` --> (partially instantiated in the runner)
`checkbox.no-manifest.partial.conf` --> (fully instantiated in the agent)
`checkbox.no-manifest.conf` --> (stacked with manifest)
`checkbox.conf`

The naming decisions for these files are supposed to indicate what each configuration fragment contains. And so far the names we have selected in our SRU and snap-testing jobs fall very short of that. Anyway, again up to you.

boukeas (approved):
Molto bene.

tools/lab_dispatch/resources/input.conf_template
Hook25: Good idea, done

boukeas (approved):

Issue #1456, opened 2024-09-06
sdhc-insert/ remove jobs not working for USB CARDREADER component


scotthu27:
### Bug Description

The SD card reader in some PC platforms is a combo card reader combines USB & SDHC.
```
path: /devices/pci0000:00/0000:00:14.0/usb1/1-7/1-7:1.0/host0/target0:0:0/0:0:0:0
name: sda
bus: scsi
category: CARDREADER
driver: sd
product: SD/MMC/MS PRO
vendor: Generic-
product_slug: SD_MMC_MS_PRO
vendor_slug: Generic-
```

The case not working since [Removed udisks2 dependencies from mediacard](https://github.com/canonical/checkbox/pull/1291/commits/edb595a081e91365276c58905d9fbcb39cd91590) changed

### To Reproduce

1. checkbox-cli run com.canonical.certification::mediacard/sdhc-insert

### Environment

Checkbox Type: Debian
Checkbox Version: 4.2.0~dev59~ubuntu24.04.1

### Relevant log output

```shell
$ /usr/bin/checkbox-support-run_watcher insertion mediacard


INSERT NOW


Timeout: 30 seconds
Traceback (most recent call last):
  File "/usr/bin/checkbox-support-run_watcher", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 144, in _f
    return run_with_timeout(f, timeout_s, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/timeout.py", line 114, in run_with_timeout
    raise TimeoutError("Task unable to finish in {}s".format(timeout_s))
TimeoutError: Task unable to finish in 30s


$ /usr/bin/checkbox-support-run_watcher insertion usb2


INSERT NOW


Timeout: 30 seconds
high_speed_usb was inserted. Controller: xhci_hcd, Number: 15
usable partition: sda1
USB2 insertion test passed.
no env var PLAINBOX_SESSION_SHARE


ubuntu@localhost:~$ lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
loop0         7:0    0     4K  1 loop /snap/bare/5
loop1         7:1    0  32.7M  1 loop /snap/bugit/34
[...]

PR #1477, opened 2024-09-16
Fixed issue with usb string comparison (BugFix)

fernando79513:
## Description

Taking a look at https://github.com/canonical/checkbox/issues/1456 I saw that there is an error in the "insertion" condition.
Explanation
The line:

python
Copy code
if "USB Mass Storage device detected" or "uas" in line_str:
does not work as intended because of the way Python evaluates conditional expressions. Here's what's happening step by step:

Expression Evaluation Order:

Python evaluates "USB Mass Storage device detected" first.
Since non-empty strings are considered True in a boolean context, "USB Mass Storage device detected" evaluates to True.
Therefore, the entire expression before the or operator is True.

## Resolved issues

N/A

## Documentation

N/A

## Tests

Run Storage tests (22.04 laptop)
`checkbox-cli run com.canonical.certification::usb3-cert-manual`

### DIFF
--- a/checkbox-support/checkbox_support/scripts/run_watcher.py
+++ b/checkbox-support/checkbox_support/scripts/run_watcher.py
@@ -249,7 +249,7 @@ class USBStorage(StorageWatcher):
                 ).group(1)
 
         # Look for insertion action
-        if "USB Mass Storage device detected" or "uas" in line_str:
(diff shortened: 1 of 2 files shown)


pieqq (approved):
Thanks for finding this issue and providing a quick fix for it!

eugene-yujinwu:
my bad! Thanks for the fix.

Issue #1382, opened 2024-08-01
`stress-tests/hibernate-{s4_iterations}-cycles-log-check` shouldn't be executed while `stress-tests/hibernate_{{ s4_iterations }}_cycles` is skipped


hanhsuan:
### Bug Description

While stress testing, the  `stress-tests/hibernate-{s4_iterations}-cycles-log-check` is executed while `stress-tests/hibernate_{{ s4_iterations }}_cycles` is skipped. However, There is no log and `stress-tests/hibernate-{s4_iterations}-cycles-log-check` is failed always.

### To Reproduce

Run client-cert-iot-server-24-04-stress test plan

### Environment

- OS Ubuntu server 24.04
- Checkbox Type Snap
- Checkbox Version 4.0.0.dev333

### Relevant log output

[submission](https://certification.canonical.com/hardware/202405-34047/submission/384085/test-results/?term=hibernate)

zongminl:
Fixed by #1483

PR #1483, opened 2024-09-18
Remove hibernate from iot test plan (BugFix)

hanhsuan:
## Description

There is no more hibernate test requirement in the iot project and `stress-full` test plan is only included in the iot related test plans. Therefore, removing this test from stress-full test plan won't impact other fields.

## Resolved issues

#1382

### DIFF
--- a/providers/base/units/stress/test-plan.pxu
+++ b/providers/base/units/stress/test-plan.pxu
@@ -130,7 +130,6 @@ nested_part:
     warm-boot-stress-test
     cold-boot-stress-test
     suspend-cycles-stress-test
-    hibernate-stress-test
     stress-ng-automated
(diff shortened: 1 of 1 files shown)


fernando79513:
Should we also remove the actual test plan?
https://github.com/canonical/checkbox/blob/main/providers/base/units/stress/test-plan.pxu#L227

tsunghanliu:
From SWE perspective, the hibernate function is most not available on IoT device. So +1 to remove it from the stress test plan.

If any device needs this, it can be added into the project checkbox.

Hook25 (approved):
Given that this was agreed by all iot maintainers and that this doesnt have any impact outside of IoT, +1

We can always re-insert it on a per-project basis.

PR #1489, opened 2024-09-19
Add manifest units to expand output (New)

Hook25:
## Description

We need a way to programmatically fetch the list of manifest entries needed to run a test plan. This makes it easier to automatically verify manifests in CI, track changes to testplans manifest entries and more. This adds manifest units to the `expand` command filtering out those that aren't used in the testplan.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1555

## Documentation

This adds plenty comments to the expander.

## Tests

WIP

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/subcommands.py
+++ b/checkbox-ng/checkbox_ng/launcher/subcommands.py
@@ -1393,7 +1393,8 @@ class Expand:
                 obj["certification-status"] = (
                     self.get_effective_certification_status(unit)
                 )
-                obj["template-id"] = unit.template_id
+                if unit.template_id:
+                    obj["template-id"] = unit.template_id
             obj_list.append(obj)
 
         obj_list.sort(key=lambda x: x.get("template-id", x["id"]) or x["id"])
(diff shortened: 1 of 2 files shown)


checkbox-ng/checkbox_ng/launcher/subcommands.py
```diff
@@ -1340,31 +1371,47 @@ def invoked(self, ctx):
...
         obj_list = []
-        for unit in jobs_and_templates_list:
+        for unit in units_to_print:
             obj = unit._raw_data.copy()
             obj["unit"] = unit.unit
             obj["id"] = unit.id  # To get the fully qualified id
-            obj["certification-status"] = (
-                self.get_effective_certification_status(unit)
-            )
-            if unit.template_id:
```
pieqq, line 1356: Did you remove this condition intentionally? While playing with your branch, I compared the JSON output between `main` and your branch and noticed a lot of `template_id: null` in there.

pieqq (approved):
Thanks for the comments in `_get_relevant_manifest_units`, otherwise that would be pretty complicated to parse!

I've played with it a little bit and it seems to do what it's advertised for.

I only have one inline comment about the fact that now every job has a `template_id` attribute, even if it's `null`.

checkbox-ng/checkbox_ng/launcher/subcommands.py
Hook25: no I've made a mistake, the check for none template_id is still needed

pieqq (approved):
Thanks!

PR #1494, opened 2024-09-23
Update gadget interface validation test plan (BugFix)

stanley31huang:
update ce-oem-automated test plan and gadget interface test cases

## Description
There are two major changes in this PR
1. include the `ce-oem-gadget-automated` into the nested part of `ce-oem-automated`
2. remove the `also-after-suspend` flag from the gadget interface tests.

## Resolved issues
N/A

## Documentation
N/A

## Tests
Passed the provider validation locally.

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/gadget/jobs.pxu
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/gadget/jobs.pxu
@@ -52,7 +52,6 @@ requires:
   lsb.distributor_id == 'Ubuntu Core'
 category_id: com.canonical.certification::gadget
 estimated_duration: 5s
-flags: also-after-suspend
 command:
   check_gadget_interface.py --type {type} --name {name} --interface {interface} # --attrs "{attrs}"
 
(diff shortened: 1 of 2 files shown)


LiaoU3 (commented):
Why don't we want to run the after suspend tests?

stanley31huang:
> Why don't we want to run the after suspend tests?

IIRC, the snap interfaces are initialized when the snap is installed, so we should be safe to ignore those tests in Post-suspend stage.

LiaoU3:
> > Why don't we want to run the after suspend tests?
> 
> IIRC, the snap interfaces are initialized when the snap is installed, so we should be safe to ignore those tests in Post-suspend stage.

I don't get it. What will happen if we test them after suspend?

stanley31huang:
> > > Why don't we want to run the after suspend tests?
> > 
> > 
> > IIRC, the snap interfaces are initialized when the snap is installed, so we should be safe to ignore those tests in Post-suspend stage.
> 
> I don't get it. What will happen if we test them after suspend?

I mean the snap interfaces are initialized when it is installed. Thus, the interface will not be affected during the suspend and resume process (unlike the kernel driver, which may experience changes or need reinitialization) . So we don't need to have the after-suspend tests.

LiaoU3 (approved):
+1

PR #1495, opened 2024-09-23
Fix envvars + propagation in lab pipeline (infra)

Hook25:
## Description

The lab pipeline was failing on 24.04 because one of the packages was using the interactive frontend to fetch some value from the user. This was because the `apt` command was launched without the `-E` flag, therefore the environment was not inherited from the parent shell, clearing the cusom environment. Additionally, the `DEBIAN_FRONTEND` envvar was unset, leaving to the specific workflow the chore to set it. This fixes both of these issues.

Minor: this also adds grouping to the installer output, making it easier to read
Minor: this also enables exit on error, as this makes the output easier to read on failure

## Resolved issues

Fixes: pipeline currently not working on noble (not working if any dependency asks for manual interaction)
Fixes: CHECKBOX-1579

## Documentation

N/A

## Tests

Failing run: https://github.com/canonical/checkbox/actions/runs/10958861773/job/30429994853
Passing run post patch: https://github.com/canonical/checkbox/actions/runs/10989122085/job/30506715951

After edit inline DEBIAN_FRONTED
- https://github.com/canonical/checkbox/actions/runs/11012002495/job/30577294745
- https://github.com/canonical/checkbox/actions/runs/11012200039/job/30577891677


tools/lab_dispatch/build_install_deb.py
```diff
@@ -16,37 +34,62 @@ def parse_args():
 
 
 def prepare_repo(repo_root, package_path):
-    shutil.move(package_path / "debian", repo_root)
+    shutil.move(str(package_path / "debian"), str(repo_root))
```
boukeas, line 37: Is Python 3.8 used on our runners?

tools/lab_dispatch/build_install_deb.py
```diff
@@ -16,37 +34,62 @@ def parse_args():
...
     subprocess.check_call(
-        ["sudo", "apt-get", "-y", "build-dep", "."], cwd=repo_root
+        [
+            "sudo",
+            "-E",
+            "apt-get",
+            "-y",
+            "build-dep",
+            ".",
+        ],
```
boukeas, line 53: So we are running the equivalent of:
```
DEBIAN_FRONTEND=noninteractive sudo -E apt-get...
```
Would it work if we instead ran the equivalent of:
```
sudo DEBIAN_FRONTEND=noninteractive apt-get...`
```
This is usually the suggestion on how to set the non-interactive mode (probably because it's a little safer) and it is consistent with how we usually install packages in other scripts.

tools/lab_dispatch/build_install_deb.py
Hook25: This runs on the DUT, which are bionic+, so yes, 3.8 it is (but our agents do run python3.8 as well)

tools/lab_dispatch/build_install_deb.py
Hook25: I don't know if providing the envvars to the Popen args is a good practice actually, let me investigate. This is not great but it is ran on a throw away instance, so it shouldn't cause many issues either

tools/lab_dispatch/build_install_deb.py
boukeas: Thanks. It seems `shutil.move` started accepting `pathlib.Path`s as arguments starting from 3.9, so we can't avoid the conversion to strings.

tools/lab_dispatch/build_install_deb.py
Hook25: done

PR #1498, opened 2024-09-24
Fix the combination of backup_loc path (Bugfix)

seankingyang:
## Description
Fix the path combination fo `backup_loc`.
The reason is `Path` can't combine the path start with "/", so need to make the `basedir` is the relative path with "/".

## Resolved issues
It will break the script if the machine connect with the wifi AP before test when it save the connection's configuration.
This will be 100% fail when  the above situation.

## Tests
https://certification.canonical.com/hardware/202409-35379/submission/396113/test-results/pass/

### DIFF
--- a/providers/base/bin/wifi_nmcli_backup.py
+++ b/providers/base/bin/wifi_nmcli_backup.py
@@ -78,7 +78,7 @@ def save_connections(keyfile_list):
             continue
 
         print("  Found file {}".format(f))
-        basedir = Path(f).parent
+        basedir = Path(f).parent.relative_to("/")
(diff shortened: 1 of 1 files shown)


eugene-yujinwu:
This PR fix the bug https://bugs.launchpad.net/sutton/+bug/2081247. The test result after apply the fix: https://certification.canonical.com/hardware/202409-35352/submission/396157/test-results/

eugene-yujinwu:
LGTM

fernando79513 (commented):
LGTM +1!

fernando79513 (approved):

PR #1160, opened 2024-04-09
Add template_id to the Checkbox submission JSON schema (infra)

pieqq:
## Description

Based on checkbox.json update done in #1107.

PR #1520, opened 2024-10-01
Add info to rejected jobs list in `submission.json` (new)

pieqq:
## Description

Developments in Checkbox and C3 require more information being shared between the two.

C3 uses the `submission.json` file generated by Checkbox when creating the submission archive. This JSON file contains a `rejected-jobs` section that, until now, only contained the `full_id` of the jobs being de-selected.

With this PR, the exported submission.json includes more information in the `rejected-jobs` array, similarly to what `results` contain. This includes partial id and full id, plugin (which allows us to know what kind of job this is, including attachment, resource, etc.), and effective category and certification status (blocker or non-blocker).

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1597

## Tests

Ran the following test with the version in `main` and the version in this branch:

1. Run `checkbox-cli`
2. Select the 24.04 desktop test plan
3. Deselect all the jobs (by pressing `D`)
4. Run the tests

Compare the number of rejected jobs in both cases to make sure we have the same outcome:

```
$ jq '.["rejected-jobs"] | length' submission_before_patch.json
1217
$ jq '.["rejected-jobs"] | length' submission_after_patch.json
1217
```

Here is a sample submission JSON file from a run on my laptop: [submission_after_effective.json](https://github.com/user-attachments/files/17211303/submission_after_effective.json)

Note that, exactly like jobs that have been run (present in `results` section), the jobs in `rejected-jobs` include the effective certification_status and category. For instance, the `camera/detect` job is marked as a cert-blocker:

```json
        {
            "id": "camera/detect",
            "full_id": "com.canonical.certification::camera/detect",
            "name": "This Automated test attempts to detect a camera.",
            "plugin": "shell",
            "certification_status": "blocker",
            "category": "",
            "category_id": "com.canonical.plainbox::camera",
            "template_id": null
        },
```

The rejected-jobs also include attachment and resource jobs:

```json
        {
            "id": "camera/multiple-resolution-images-attachment_video0",
            "full_id": "com.canonical.certification::camera/multiple-resolution-images-attachment_video0",
            "name": "Attach an image from the multiple resolution images test for Integrated_Webcam_FHD__Integrat",
            "plugin": "attachment",
            "certification_status": "non-blocker",
            "category": "",
            "category_id": "com.canonical.plainbox::camera",
            "template_id": "com.canonical.certification::camera/multiple-resolution-images-attachment_name"
        },
...
        {
            "id": "cdimage",
            "full_id": "com.canonical.certification::cdimage",
            "name": "Collect information about installation media (casper)",
            "plugin": "resource",
            "certification_status": "non-blocker",
            "category": "Gathers information about the DUT",
            "category_id": "com.canonical.certification::information_gathering",
            "template_id": null
        },
```

### DIFF
--- a/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
+++ b/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
@@ -108,16 +108,16 @@
     {%- endfor %}
     ],
     "rejected-jobs": [
-    {%- for job_id, job_state in job_state_map|dictsort if job_id in state.metadata.rejected_jobs %}
+    {%- for job_id in state.metadata.rejected_jobs %}
         {
             "id": "{{ job_id|strip_ns }}",
             "full_id": "{{ job_id }}",
-            "name": "{{ job_state.job.tr_summary() }}",
-            "plugin": "{{ job_state.job.plugin }}",
-            "certification_status": "{{ job_state.effective_certification_status }}",
-            "category": "{{ category_map[job_state.effective_category_id] }}",
-            "category_id": "{{ job_state.effective_category_id }}",
-            "template_id": {{ job_state.job.template_id | jsonify | safe }}
+            "name": "{{ job_state_map[job_id].job.tr_summary() }}",
+            "plugin": "{{ job_state_map[job_id].job.plugin }}",
+            "certification_status": "{{ job_state_map[job_id].effective_certification_status }}",
+            "category": "{{ category_map[job_state_map[job_id].effective_category_id] }}",
+            "category_id": "{{ job_state_map[job_id].effective_category_id }}",
+            "template_id": {{ job_state_map[job_id].job.template_id | jsonify | safe }}
         }{%- if not loop.last -%},{%- endif %}
     {%- endfor %}
     ],
--- a/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
+++ b/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
@@ -108,9 +108,15 @@
     {%- endfor %}
     ],
     "rejected-jobs": [
-    {%- for job_id in state.metadata.rejected_jobs %}
+    {%- for job in state.metadata.rejected_jobs %}
         {
-            "full_id": "{{ job_id }}"
+            "id": "{{ job['partial_id'] }}",
+            "full_id": "{{ job['id'] }}",
+            "name": "{{ job['name'] }}",
+            "category": "{{ category_map[job['category_id']] }}",
+            "category_id": "{{ job['category_id'] }}",
+            "plugin": "{{ job['plugin'] }}",
+            "template_id": {{ job['template_id'] | jsonify | safe }}
         }{%- if not loop.last -%},{%- endif %}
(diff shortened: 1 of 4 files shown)


andrejvelichkovski:
Just verified that C3 doesn't parse this list to any more details. I think this will be a useful addition to help us validate the testing coverage that wont break C3 compatibility.

pieqq:
Turning this into draft again as I think there is a way to export effective category and effective certification status, which could be helpful.

andrejvelichkovski:
The changes look good to me! Are we now also adding the `plugin` field to each test result, perhaps we can parse this into C3 as well?

pieqq:
> The changes look good to me! Are we now also adding the `plugin` field to each test result, perhaps we can parse this into C3 as well?

The problem is we have a weird mapping between Checkbox and C3. [In Checkbox,](https://canonical-checkbox.readthedocs-hosted.com/en/stable/reference/units/job.html#job-fields), a job can have the following `plugin` options:

- manual
- shell
- user-interact
- user-interact-verify
- attachment
- resource

I think in C3, this is emulated by the `type` field in Test model, which concatenate this (to be confirmed).

Similarly, in C3, the status of a TestResult is narrowed down to `skip/fail/pass`, whereas in Checkbox there are other possibilities (`crashed`, `not-supported`, etc.). I don't know why the raw information is not sent to C3, and C3 then has the logic to do any aggregation it would see fit.

But yes, we could start saving this info in C3 as well.

submission-schema/schema.json
Hook25, line 1: Does this contain more than what is just done in this pr?

Hook25 (dismissed):
See if it makes sense, else you can land it as is

checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
```diff
@@ -108,9 +108,16 @@
     {%- endfor %}
     ],
     "rejected-jobs": [
-    {%- for job_id in state.metadata.rejected_jobs %}
+    {%- for job_id, job_state in job_state_map|dictsort if job_id in state.metadata.rejected_jobs %}
```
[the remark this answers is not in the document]
pieqq: I removed the dictsort and only iterate on rejected jobs list (fixed in https://github.com/canonical/checkbox/pull/1520/commits/2a223c7acb31c43fd62226b2d96f538adbfc5ae6). That said, I basically copy-pasted what's been done for all the other arrays (`results`, `attachment-results`, `resource-results`), so we may want to reconsider doing this for these arrays as well, especially if it's really problematic on slow devices.

submission-schema/schema.json
pieqq: I initially made a pretty bad implementation of a solution, which modified some Python files, and I realized in the process we didn't have any unit tests, so I added some. Then I realized there was a much simpler implementation (simply modifying the `checkbox.json` template) and went ahead with this, but I left the unit tests anyway as it's useful.

pieqq: OK now I realized this comment is about `schema.json`.

This is because it includes changes originally made for an older PR that was never merged (#1160 ), and that are now superseded by this PR.

Issue #1371, opened 2024-07-26
Message "Timer is not a type" be got while executing QT related cases through control (remote) by using snap 


baconYao:
### Bug Description

No QT window be shown on screen.

Log while running the `touchscreen/multitouch-zoom` job

```bash
file:///snap/checkbox22/current/providers/checkbox-provider-base/data/touch_zoom_test.qml:74 Timer is not a type
```

Related issue: https://github.com/canonical/checkbox/issues/728

### To Reproduce

1. Install Checkbox on DUT via the following commands
    ```
    $ sudo snap install checkbox22 --beta
    $ sudo snap install checkbox --beta --classic
    ```
2. Run Checkbox Snap via remote on host
    ```
    $ checkbox.checkbox-cli remote <IP of DUT>
    ```
3. Choose `Desktop Preload Certification Tests for 22.04 (Manual Tests)` test plan
4. Run  `touchscreen/multitouch-rotate` and `touchscreen/multitouch-zoom` jobs

### Environment

OS: Desktop
Checkbox Type: Snap
```
checkbox                   4.0.0-dev329      7644   22.04/edge       ce-certification-qa  classic
checkbox22                 4.0.0-dev329      1005   latest/edge      ce-certification-qa  -
```

Hook25:
As per our discussion: this may be related to the QT version installed. The device is running arm64, we need to double check this

baconYao:
### Debug information in `checkbox.shell`

```bash
(checkbox-shell) ubuntu@mtk-genio:~$ export QT_DEBUG_PLUGINS=1
(checkbox-shell) ubuntu@mtk-genio:~$ qmlscene -qt5 ./touch_zoom_test.qml
QFactoryLoader::QFactoryLoader() checking directory path "/snap/checkbox22/1057/usr/lib/aarch64-linux-gnu/qt5/plugins/platforms" ...
QFactoryLoader::QFactoryLoader() looking at "/snap/checkbox22/1057/usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqeglfs.so"
Found metadata in lib /snap/checkbox22/1057/usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqeglfs.so, metadata=
{
    "IID": "org.qt-project.Qt.QPA.QPlatformIntegrationFactoryInterface.5.3",
    "MetaData": {
        "Keys": [
            "eglfs"
        ]
    },
    "archreq": 0,
[...]

PR #1528, opened 2024-10-03
Fix touchscreen tests on ARM64 via remote (bugfix)

Hook25:
## Description

When running the touchscreen tests via remote, the user is promted to rotate the rectangle on a window but the `user-interact-verify` job (silently) fails to start the window at all. When the stderr is un-silenced, the following error is printed in the console:
```
file:///snap/checkboxXX/[...]/touch_zoom_test.qml:74 Timer is not a type
```
Timer is a builtin type of QT and this happens for any other type (removing the timer just makes it fail on the next builtin type). 

After much investigation, and thanks to the reported fact that the test starts fine if launched from the `checkbox.shell` or locally, I found that the problem is the missing envvar in this PR. This fixes the propagation of the envvar. I have whitelisted it instead of explicitly requiring it in the test because I believe that one doesn't have to understand exactly how QT works when writing a test, so they don't need to reason about which QT envvars are needed. 

## Resolved issues

Fixes: CHECKBOX-1511
Fixes: https://github.com/canonical/checkbox/issues/1371

## Documentation

N/A

## Tests

I have built and ran the test via remote with the following launcher. The window now shows up even via remote when before it did not.
```
    #!/usr/bin/env checkbox-cli
    [launcher]
    launcher_version = 1
    stock_reports = text
    [test plan]
    unit = com.canonical.certification::touchscreen-cert-manual
    forced = yes
    [test selection]
    forced=yes
    exclude=com.canonical.certification::touchscreen/drag-n-drop
    match = com.canonical.certification::touchscreen/multitouch-rotate
    [manifest]
    com.canonical.certification::has_touchscreen=yes
```

To test it yourself use [this checkbox24 snap](https://drive.google.com/file/d/135Ftfy7wj1wCCUq7S9BzBKPeAEE8kx9Y/view?usp=sharing), built from this branch (or build it yourself on arm64):
1. Install the snap
2. Install the frontend
3. Run remote with the above launcher
4. Press Return
5. A window with a blue rectangle should show up on the DUT

> Note: This is a core24 runtime snap. Remember CheckboxXX core only work on the XX version (so this doesn't work on jammy, only noble), if you want to test it on jammy I can build a core22 snap for you, just let me know

### DIFF
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -539,6 +539,7 @@ def get_execution_environment(job, environ, session_id, nest_dir):
                 "QT_QPA_PLATFORM",
                 "QT_QPA_PLATFORMTHEME",
                 "QT_DEBUG_PLUGINS",
+                "QML2_IMPORT_PATH",
             ]
             for key, value in env.items():
                 if key in copy_vars or key.startswith("SNAP"):
@@ -640,6 +641,7 @@ def get_differential_execution_environment(
             "QT_QPA_PLATFORM",
             "QT_QPA_PLATFORMTHEME",
             "QT_DEBUG_PLUGINS",
+            "QML2_IMPORT_PATH",
         ]
         for key, value in base_env.items():
             if key in copy_vars or key.startswith("SNAP"):


Hook25:
Proof of this working on the original hardware in the lab:
image

baconYao:
LGTM

baconYao (approved):
Fantastic! LGTM +1

PR #1531, opened 2024-10-04
Add amd gpgpu tests (New)

pedro-avalos:
## Description

- Added a wrapper for ROCm Validation suite that checks the result of the tests
- Added AMD GPGPU test jobs using ROCm Validation Suite

## Resolved issues

Addresses [CHECKBOX-969](https://warthogs.atlassian.net/browse/CHECKBOX-969) and [CHECKBOX-1539](https://warthogs.atlassian.net/browse/CHECKBOX-1539).

## Documentation

Coverage tests are included with this PR.

## Tests

I tested this on my desktop, which has an AMD GPU (`AMD Radeon RX 6750 XT`). NOTE: For some AMD GPUs, you may need to change the environment variable `HSA_OVERRIDE_GFX_VERSION` for `rvs` to recognize that your GPU is compatible with some ROCm/HIP functions. For example, for my GPU, I had to edit the value to `10.3.0`.

[CHECKBOX-969]: https://warthogs.atlassian.net/browse/CHECKBOX-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHECKBOX-1539]: https://warthogs.atlassian.net/browse/CHECKBOX-1539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/gpgpu/tests/test_rvs.py
+++ b/providers/gpgpu/tests/test_rvs.py
@@ -29,25 +29,11 @@ from rvs import (
     PassFailModuleRunner,
     main,
     parse_args,
-    which_rvs,
 )
 
 
 @patch("rvs.logging")
 class TestMain(TestCase):
-    @patch(
-        "subprocess.run",
-        return_value=MagicMock(returncode=0, stdout="/bin/rvs"),
-    )
-    def test_which_rvs_found(self, run_mock, logging_mock):
-        p = which_rvs()
-        self.assertEqual(p, Path(run_mock().stdout.strip()))
-
-    @patch("subprocess.run", side_effect=MagicMock(returncode=1))
-    def test_which_rvs_fallback(self, run_mock, logging_mock):
-        p = which_rvs()
-        self.assertEqual(p, RVS_BIN)
-
     @patch("sys.stderr", new_callable=io.StringIO)
     @patch("sys.argv", ["rvs.py", "--list-modules"])
     def test_parse_args_list_modules(self, logging_mock, stderr_mock):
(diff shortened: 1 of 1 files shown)


pedro-avalos:
One design question I'd like some feedback on is how `rvs.py` takes the modules as positional arguments. Since we're likely to just use these one at a time in Checkbox jobs, should these be subparsers instead? If they are subparsers, then we could conceivably add the module configuration as options for those subparsers. The more I think about it, they should be subparsers. Let me know what you think!

fernando79513:
> akes the modules as positional arguments. Since we're likely to just use these one at a time in Checkbox jobs, should these be subparsers instead? If they are subparsers, then we could conceivably add the module configuration as options for those subparsers. The more I think about it, they should be subparsers. Let me know what you think!

I think the configs are complex enough to be kept into separated config files. I would restrict the `modules` parameter to run only one module, and I think you could simplify the argparser by setting `choices` for the possible modules:
```
    parser.add_argument(
        "modules",
        metavar="MOD",
        nargs="*",
        choices=RVS_MODULES.keys(),
        type=str,
        help="RVS modules to run [{}]".format(", ".join(RVS_MODULES)),
    )
```

fernando79513 (changes requested):
Some small changes. You don't need to handle the return values if you are not going to use them. Consider just raising a `SystemExit()` with the message you want to simplify the code.

providers/gpgpu/bin/rvs.py
```diff
@@ -0,0 +1,223 @@
...
+    proc = subprocess.run(
+        ["which", "rvs"],
+        check=False,
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    return Path(proc.stdout.strip()) if proc.returncode == 0 else RVS_BIN
+
+
+class ModuleRunner:
```
fernando79513, line 50: I think the whole ModuleRunner class could be simplified and made more explicit, getting rid of all the `subprocess.run` wrappers and the complexity in the validators:
```
class ModuleRunner:
    """This class represents the base module runner."""

    def __init__(self, rvs: Path, config: Path) -> None:
        """Initializes the module runner."""
        self.config = config
        self.rvs = rvs or Path(shutil.which("rvs") or RVS_BIN)

    def run(self, module: str):
        logging.debug("%s: RUNNING", module)

        result = subprocess.run(
            [self.rvs, "-c", self.config],
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            universal_newlines=True,
        )

        if result.returncode != 0:
            logging.error(result.stdout)
            logging.error(result.stderr)
            raise SystemExit("%s: FAILURE - failed to execute" % module)

        # Log the error output as debug
        if result.stderr:
            logging.debug(result.stderr)

        if hasattr(self, "_validate_output"):
            if not self._validate_output(result.stdout, module):
                logging.error(result.stdout)
                raise SystemExit("%s: FAILURE - validation failed" % module)
            
        logging.info(result.stdout)
        logging.debug("%s: SUCCESS", module)


class PassFailModuleRunner(ModuleRunner):
    """This class represents a module runner that passes or fails."""

    def _validate_output(self, output: str, module: str):
        # Find any of the common success messages in stdout
        for line in output.splitlines():
            if any(
                re.search(pass_str, line)
                for pass_str in [
                    r"%s true" % re.escape(module),
                    r"pass: TRUE",
                    r"GFLOPS \d+ Target GFLOPS: \d+ met: TRUE",
                ]
            ):
                return False
        return True


class MemModuleRunner(ModuleRunner):
    """This class represents the memory test module runner."""

    def _validate_output(self, output: str, module: str):
        return not all(
            "mem Test %s : PASS" % i in output.splitlines()
            for i in range(1, 12)
        )
```

fernando79513 (approved):
LGTM +1!

Issue #1533, opened 2024-10-07
Incorrect ping parameter on wirless open ac connection test


omar-selo:
### Bug Description

During SRU snap testing, one of the devices kept [failing](https://certification.canonical.com/hardware/201802-26087/submission/397940/test-results/fail/?term=open_ac) `wireless/wireless_connection_open_ac_np_wlp1s0` test with the logs provided below. Zooming in on the ping test we see the following logs:
```
## Perform a ping test
Got gateway address: 192.168.144.1 (TP-LINK TECHNOLOGIES CO.,LTD.)
Ping result: {'transmitted': 0, 'received': 0, 'pct_loss': 0, 'cause': "Failed with exception: Command '['ping', '192.168.144.1 (TP-LINK TECHNOLOGIES CO.,LTD.)', '-c', '5', '-w', '10', '-Iwlp1s0']' returned non-zero exit status 2.\nstdout: \nstderr: ping: 192.168.144.1 (TP-LINK TECHNOLOGIES CO.,LTD.): Name or service not known\n"}
Connection test failed
```

This suggests that the test is trying to ping the string `'192.168.144.1 (TP-LINK TECHNOLOGIES CO.,LTD.)'` instead of just `192.168.144.1`.

### To Reproduce

Run http://10.102.156.15:8080/job/cert-dawson-i-uc20-snapd-beta/ job and observer the result of `wireless/wireless_connection_open_ac_np_wlp1s0` test

### Environment

- OS UC20
- checkbox Version: 4.2.0-dev99-9568
- checkbox20 Version: 4.2.0-dev99-2054

### Relevant log output

```shell
Interface eno1 using module e1000e
  Parameters:
    copybreak: 256

Checking kernel ring buffer for e1000e messages:
kern  :info  : [Wed Oct  2 08:37:55 2024] e1000e: EEE TX LPI TIMER: 00000011
kern  :info  : [Wed Oct  2 08:38:01 2024] e1000e: eno1 NIC Link is Up 1000 Mbps Full Duplex, Flow Control: None


Interface wlp1s0 using module iwlwifi
  Parameters:
    lar_disable: N
    fw_monitor: N
    11n_disable: 0
    power_save: N
    swcrypto: 0
    antenna_coupling: 0
    power_level: 0
    amsdu_size: 0
    uapsd_disable: 3
    fw_restart: Y
    led_mode: 0
    disable_11ax: N
    enable_ini: N
    disable_11ac: N
    bt_coex_active: Y
[...]

PR #1536, opened 2024-10-09
Fix wifi netplan gateway parser (Bugfix)

seankingyang:
## Description
Correct the gateway parser.

In UC20, the systemd version is 245, and the formate is difference with the newer systemd version:
```
ubuntu@ubuntu:~$ networkctl --version
systemd 245 (245.4-4ubuntu3.23)

ubuntu@ubuntu:~$ networkctl status --no-pager --no-legend wlp1s0 
● 3: wlp1s0                                                              
             Link File: /usr/lib/systemd/network/99-default.link         
          Network File: /run/systemd/network/10-netplan-wlp1s0.network   
                  Type: wlan                                             
                 State: routable (configured)  
                  Path: pci-0000:01:00.0                                 
                Driver: iwlwifi                                          
                Vendor: Intel Corporation                                
                 Model: Wireless 8265 / 8275 (Dual Band Wireless-AC 8265)
            HW Address: b4:6b:fc:33:40:e6 (Intel Corporate)              
                   MTU: 1500 (min: 256, max: 2304)                       
     WiFi access point: cert-ac-open-tel-l3-01 (ae:84:c6:62:8d:12)       
  Queue Length (Tx/Rx): 1/1                                              
               Address: 192.168.153.146 (DHCP4)                          
                        fe80::b66b:fcff:fe33:40e6                        
               Gateway: 192.168.144.1 (TP-LINK TECHNOLOGIES CO.,LTD.)    
                   DNS: 192.168.144.1                                    
     Activation Policy: up                                               
   Required For Online: yes       
```

In UC22, the systemd version is 249:
```
ceqa@ubuntu:~$ networkctl --version
systemd 249 (249.11-0ubuntu3.12)

ceqa@ubuntu:~$ networkctl status --no-pager --no-legend wlp1s0 
● 3: wlp1s0                                                                    
                     Link File: /usr/lib/systemd/network/99-default.link
                  Network File: /run/systemd/network/10-netplan-wlp1s0.network
                          Type: wlan
                         State: routable (configured)
                  Online state: online                                         
                          Path: platform-112f0000.pcie-pci-0000:01:00.0
                        Driver: mt7921e
                        Vendor: MEDIATEK Corp.
                    HW Address: 2c:3b:70:3f:d2:f3
                           MTU: 1500 (min: 256, max: 2304)
                         QDisc: noqueue
  IPv6 Address Generation Mode: eui64
             WiFi access point: cert-ac-open-tel-l3-01 (ae:84:c6:62:8d:12)
          Queue Length (Tx/Rx): 1/1
                       Address: 192.168.156.224 (DHCP4 via 192.168.144.1)
                                fe80::2e3b:70ff:fe3f:d2f3
                       Gateway: 192.168.144.1
                           DNS: 192.168.144.1
             Activation Policy: up
           Required For Online: yes
               DHCP4 Client ID: IAID:0xff4d6f5/DUID
             DHCP6 Client DUID: DUID-EN/Vendor:0000ab11719fc417932a6a1c0000

```

## Resolved issues

Incorrect ping parameter on wirless open ac connection test #1533


providers/base/bin/wifi_client_test_netplan.py
```diff
@@ -315,12 +316,32 @@ def print_route_info():
...
+        return True
+    except ValueError:
+        return False
+
+
 def get_gateway(interface, renderer):
-    gateway = None
     info = get_interface_info(interface, renderer)
-    gateway = info.get("gateway", None)
+    gateway = info.get("gateway")
```
Hook25, line 340: so that we can always split
```suggestion
    gateway = info.get("gateway") or ""
```

providers/base/bin/wifi_client_test_netplan.py
```diff
@@ -315,12 +316,32 @@ def print_route_info():
...
+    ip_part = gateway.split()[0]  # Get the first part before any space
+    return ip_part if __validate_ip(ip_part) else ""
+
+
+def __validate_ip(ip):
+    try:
+        ipaddress.ip_address(ip)
+        return True
+    except ValueError:
+        return False
```
Hook25, line 335: I don't see why we should create so many functions, I would create just one. Note that this removes the `if not xxx` as below I think we should always ""
```suggestion
def _validate_gateway_ip(gateway):
    # Check if the gateway is a valid IP address
    # Examples:
    # 192.168.144.1 (TP-Link 123)
    # 192.168.144.1
    ip_address = gateway.split()[0]  # Get the first part before any space
    try:
      ipaddress.ip_address(ip_address)
      return ip_address
    except ValueError:
      return ""
```

Hook25 (changes requested):
I would not create the additional function

providers/base/bin/wifi_client_test_netplan.py
seankingyang: Thank for suggestion~~

Hook25 (approved):

PR #1541, opened 2024-10-15
Add manifest entry for dangerous grade images (new)

### LINKED ISSUE #1479: drop image/model-grade test for platforms that we test with dangerous images
### Enhancement Proposal

This example test: http://10.102.156.15:8080/job/cert-cm4l-arm64-pi-kernel-24-beta/7/testReport/com.canonical/certification__image_model-grade/com_canonical_certification__image_model_grade/
most likely shouldn't be run. This discussion seconds the idea: https://chat.canonical.com/canonical/pl/fnyu6xj5yty39kbapkr69zohoy

pieqq:
## Description

In our testing lab, we need to provision some devices with Ubuntu Core
images that are set to use the 'dangerous' grade.

This will fail the `image/model-grade` test.

Add a new `dangerous_grade_core_image` manifest to decide whether or not
we should be running this test.

It will be run by default (when there is no manifest entry, its value is
set to 'False'), but by setting it to True, the test will be skipped.

image

## Resolved issues

Fix #1479 
Fix CHECKBOX-1572

### DIFF
--- a/providers/base/units/image/jobs.pxu
+++ b/providers/base/units/image/jobs.pxu
@@ -78,8 +78,10 @@ flags: preserve-locale
 id: image/model-grade
 category_id: image
 _summary: Check that the model grade is correctly set
+imports: from com.canonical.plainbox import manifest
 requires:
  lsb.distributor_id == "Ubuntu Core" and int(lsb.release) >= 20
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
+1, ty for coming up with a name

PR #1542, opened 2024-10-15
Fix GPGPU provider distribution (New)

pedro-avalos:
## Description

Currently, the GPGPU provider depends on the user running a setup script after installing the provider. This script sets up the NVIDIA/AMD repositories to get the required packages from upstream (i.e. CUDA toolkit or ROCm). This has worked so far, but it is negatively affecting the way we test the provider, especially with automated setups.

## Resolved issues

Resolves CHECKBOX-1519

### DIFF
--- a/checkbox-core-snap/series16/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series16/snap/snapcraft.yaml
@@ -57,6 +57,12 @@ package-repositories:
     ppa: checkbox-dev/stable
   - type: apt
     ppa: colin-king/stress-ng
+  - type: apt
+    architectures: [amd64]
+    formats: [deb]
+    path: /
+    key-id: EB693B3035CD5710E231E123A4B469963BF863CC
+    url: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64
 
 parts:
   version-calculator:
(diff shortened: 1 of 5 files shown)


pedro-avalos:
Ok, the next step should be to get the repository setup part of the `gpu-setup` script as a part of the deb packaging step. I'm thinking maybe a `postinst` script should do the trick?

providers/gpgpu/units/jobs.pxu
```diff
@@ -4,31 +4,31 @@ plugin: shell
 estimated_duration: 300
 requires:
     graphics_card.vendor == 'NVIDIA Corporation'
-    package.name == 'cuda-toolkit'
+    snap.name == 'gpu-burn'
     uname.machine == 'x86_64'
 _summary: NVIDIA GPGPU stress testing
-command: gpu_burn.sh
+command: gpu-burn 14400 |  grep -v -e '^[[:space:]]*$' -e "errors:" -e "Summary at"
```
Hook25, line 10: I see that this partially replaces the old script but please avoid hiding the output of commands (I know, this is not new code, but lets fix it). Please `tee` the output so that it can be read from the session and `set -eo pipefail` for now

Hook25 (changes requested):
I have one commands change (pretty minor) and a major thing to ask, please create a sister PR in this repo and update the line I'm linking, else landing this will break our canary autopromotion script! 

https://github.com/canonical/hwcert-jenkins-tools/blob/main/version-published/checkbox-canary.yaml#L68

Hook25 (approved):
+1, fantastic job, really well done

PR #1546, opened 2024-10-17
Fix missing `subprocess.PIPE` in reboot_check_test.py (Bugfix)

tomli380576:
## Description

The new reboot check test was missing `stdout=subprocess.PIPE` in calls to `subprocess.run` which was causing it to return None for `<CompletedProcess>.stdout`

## Resolved issues
#1548 

## Tests

1. The original test cases pass
2. Side loaded on 202406-34083 and full wb/cb tests can be run


Hook25 (approved):
+1, lgtm

PR #1553, opened 2024-10-18
Remove require for rt-tests (BugFix)

LiaoU3:
## Description
Since we have package `rt-tests` in [snapcraft](https://github.com/canonical/checkbox/blob/90acaebc6997387906e7ae832fef0dbe22b17809/checkbox-core-snap/series24/snap/snapcraft.yaml#L303) and [packaging](https://github.com/canonical/checkbox/blob/90acaebc6997387906e7ae832fef0dbe22b17809/providers/base/units/rt-performance-tests/packaging.pxu#L5), there is no need to require the package in the tests.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A

### DIFF
--- a/providers/base/units/rt-performance-tests/jobs.pxu
+++ b/providers/base/units/rt-performance-tests/jobs.pxu
@@ -9,8 +9,6 @@ _purpose:
     that the maximum latency did not exceed the threshold and that there are no
     overflows (scheduling misses) for each thread.
 plugin: shell
-requires:
-    package.name == 'rt-tests'
 user: root
 environ: CYCLICTEST_DURATION
 command:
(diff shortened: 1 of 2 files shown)


pieqq:
Why is this removal needed? Having `package.name` can ensure the test runs only when that package is available (for instance, when running using a virtual environment).

LiaoU3:
@pieqq It is not working on the snap version of checkbox, since we package `rt-tests` into the stage-packages, it is not going to be shown in the resource job `package`, due to the fact that `package` only list the debian package installed on the local

zongminl (changes requested):
Please use snap resource job instead if this is a snap package to be checked

pieqq:
> Please use snap resource job instead if this is a snap package to be checked

That, or the `executable` resource.

LiaoU3:
`executable` makes more sense

zongminl (approved):
LGTM

pieqq (approved):
LGTM. Thanks!

PR #1554, opened 2024-10-21
[checkbox-ceoem] Modify optee test to allow different xtest snap (Bugfix)

rickwu666666:
## Description
Modify the optee test since we have platform that tee-supplicant and xtest running on different snap. We also add environment for both xtest and tee-supplicant for debug perpose.

## Tests
sideload result without manifest:
https://certification.canonical.com/hardware/202408-34281/submission/400476/


rickwu666666:
@stanley31huang As discussed offline, I've removed the logic of enable tee-supplicant.

stanley31huang (approved):
LGTM

PR #1555, opened 2024-10-21
Fix the parser so that it doesn't loop forever (bugfix)

Hook25:
## Description

When validating units, Checkbox tries to parse fields to see if there are any errors with them. Turns out that if a field contains an un-terminated string`"...`, that will cause the parser to always loop forever or crash with an assertion error. This was affecting a test run today due to a non-sluggified name getting templated into a job. This manifested itself as an infinited crash-restore loop due to also an un-related error (if the session gets stuck on an infinite loop, the controller tries to reconnect eventually, Checkbox refuses to "recover" during bootstrapping, restarting the session)

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1622

## Documentation

This also documents how the parser works because it is quite something

## Tests

This also adds the tests I used to reproduce the issue

### DIFF
--- a/checkbox-ng/plainbox/impl/test_xparsers.py
+++ b/checkbox-ng/plainbox/impl/test_xparsers.py
@@ -1,8 +1,9 @@
 # This file is part of Checkbox.
 #
-# Copyright 2012-2015 Canonical Ltd.
+# Copyright 2012-2024 Canonical Ltd.
 # Written by:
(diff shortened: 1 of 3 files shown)


fernando79513 (approved):
Great investigation, what a piece of code!
LGTM+1!

PR #1558, opened 2024-10-23
Add reference to manifest entries in expand subcommand help (bugfix)

pieqq:
## Description

Following https://github.com/canonical/checkbox/pull/1489, the `expand` subcommand now also lists manifest entries. This is now reflected in the subcommand help.

## Documentation

```
$ checkbox-cli expand --help
usage: checkbox-cli [-h] [-f FORMAT] TEST_PLAN

Expand a given test plan: display all the jobs, templates and manifest entries that are defined in this test plan and that  would be executed if ran. This is useful to visualize the full list of units called for complex test plans that consist of many nested parts with different 'include' and 'exclude' sections.

NOTE: the elements listed here are not sorted by execution order. To see the execution order, please use the 'list-bootstrapped' command instead.

positional arguments:
  TEST_PLAN             test-plan id to expand

options:
  -f FORMAT, --format FORMAT
                        output format: 'text' or 'json' (default: text)

```

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/subcommands.py
+++ b/checkbox-ng/checkbox_ng/launcher/subcommands.py
@@ -1306,11 +1306,11 @@ class Expand:
     def register_arguments(self, parser):
         parser.formatter_class = RawDescriptionHelpFormatter
         parser.description = (
-            "Expand a given test plan: display all the jobs and templates "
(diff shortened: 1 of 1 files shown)


Hook25 (approved):
thanks

PR #1562, opened 2024-10-25
Mock os.makedirs in launcher return codes (bugfix)

Hook25:
## Description

Test fails on launchpad because it doesn't mock the stage. Stage tries to save the C3 URL and crashes on PermissionDenied. Also, very bad idea to create the directory while running tests!

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


boukeas (approved):
+1

PR #1564, opened 2024-10-28
Add v4l-utils dependency (New)

tomli380576:
## Description

Added `v4l-utils` dependency to the debian control file.

## Tests

Verified that the package exists and can be installed with `sudo apt` in ubuntu 16, 18, 20 virtual machines and real machines running ubuntu 22, 24


tomli380576:
sry included 1 extra file

PR #1567, opened 2024-10-29
Append to the submission log (bugfix)

Hook25:
## Description

The submission url is part of more return values of just the C3 exporter. This was not intended intiially and makes it so the permanence on disk of the url depends on the order of exporters, which breaks the feature. Given this information though, this modifies the transport so that it appends to the log file and changes its name. This is a good idea because that way we can always to automatic manipulation of these files without having to wonder how they are called.

Namely, if a submission$NAME is created, to get the files that were generated as a part of that sumission one has to just consult the submission$NAME.submission.log file, which will contain all URIs, one per line. This is especially convenient if one happens to close the window after the submission is done.

This preserves the original intent to be able to recover the submission url from file without having to parse the checkbox-cli output.

## Resolved issues

N/A

## Documentation

N/A

## Tests

The feature was already tested

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/stages.py
+++ b/checkbox-ng/checkbox_ng/launcher/stages.py
@@ -710,7 +710,7 @@ class ReportsStage(CheckboxUiStage):
                                     url, path
                                 )
                             )
-                            f.writelines([url, "\n"])
+                            f.write(url + "\n")
                 except TransportError as exc:
                     _logger.warning(
                         _("Problem occured when submitting '%s' report: %s"),
--- a/checkbox-ng/checkbox_ng/launcher/stages.py
+++ b/checkbox-ng/checkbox_ng/launcher/stages.py
@@ -701,14 +701,16 @@ class ReportsStage(CheckboxUiStage):
                     result = result or {}
                     url = result.get("url") or result.get("status_url")
                     if url:
-                        path = self._get_submission_file_path(".c3_url.log")
-                        with open(path, "w+") as f:
+                        path = self._get_submission_file_path(
+                            ".submission.log"
+                        )
+                        with open(path, "a+") as f:
                             print(
-                                "Submission url ({}) saved also to: {}".format(
-                                    path, url
+                                "Submission url: {} (URL added to: {})".format(
+                                    url, path
                                 )
                             )
-                            f.write(url)
+                            f.writelines([url, "\n"])
                 except TransportError as exc:
                     _logger.warning(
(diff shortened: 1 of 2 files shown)


checkbox-ng/checkbox_ng/launcher/stages.py
```diff
@@ -701,14 +701,16 @@ def _export_results(self):
...
-                        with open(path, "w+") as f:
+                        path = self._get_submission_file_path(
+                            ".submission.log"
+                        )
+                        with open(path, "a+") as f:
                             print(
-                                "Submission url ({}) saved also to: {}".format(
-                                    path, url
+                                "Submission url: {} (URL added to: {})".format(
+                                    url, path
```
pieqq, line 710: `Submission url:` adds a lot of verbosity for no practical gain, since it's gonna be printed for every exporter.

I've played a bit and came with this suggestion:

```suggestion
                                "[{}] {} (URL added to: {})".format(
                                    params["transport"], url, path
```

It leads to this kind of things:

```
[html_file] file:///home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.html (URL added to: /home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.submission.log)
[junit_file] file:///home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.junit.xml (URL added to: /home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.submission.log)
[tar_file] file:///home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.tar.xz (URL added to: /home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.submission.log)
Do you want to submit 'upload to certification-staging' report?
  y => yes
  n => no
[yn]: y
Enter submission description (press Enter to skip): ...
Enter secure-id:...
[c3] https://certification.staging.canonical.com/submissions/status/37897 (URL added to: /home/pieq/.local/share/checkbox-ng/submission_2024-10-29T15.25.16.504567.submission.log)
```

pieqq (changes requested):
I would get rid of `Submission url:` string since it's not really useful, and maybe replace it with something (see inline comment).

checkbox-ng/checkbox_ng/launcher/stages.py
Hook25: good idea, done

pieqq (approved):
+1

PR #1566, opened 2024-10-29
Add v4l-utils snap dependency (New)

tomli380576:
## Description

Added v4l-utils dependency for the snap version of checkbox

## Tests

Built the `checkbox22` and `checkbox24` snaps by following the [build guide](https://github.com/canonical/checkbox/tree/main/checkbox-core-snap) and installed them locally on machines running ubuntu 22 & 24 respectively. `v4l2-compliance` and `v4l2-ctl` commands exist inside `checkbox.shell` even though the host machine doesn't have it installed.

PR #1569, opened 2024-10-31
Implement V4L2 compliance parser (New)

tomli380576:
## Description

This PR implements a V4L2 compliance test parser in checkbox support for test cases to validate that the ioctl requests they need to use are actually supported. 

Requires #1566  

## Documentation

Test cases can `from checkbox_support.parsers.v4l2_compliance import parse_v4l2_compliance` and call the function for a device, such as `parse_v4l2_compliance('/dev/video0')`

## Tests

Unit tests


Hook25 (changes requested):
Good job here, sorry it took me a while to review it. Consider my comments below, the only one that is blocking is the one about `pkg_resource`, because soon it will go away, and when it does I will have to patch half of checkbox to add the workaroud I suggested.

checkbox-support/checkbox_support/parsers/v4l2_compliance.py
```diff
@@ -0,0 +1,196 @@
...
+    "VIDIOC_S_HW_FREQ_SEEK": ["VIDIOC_S_HW_FREQ_SEEK"],
+    "VIDIOC_ENUMAUDIO": ["VIDIOC_ENUMAUDIO"],
+    "VIDIOC_G/S/ENUMINPUT": [
+        "VIDIOC_G_SELECTION",
+        "VIDIOC_ENUMINPUT",
+        "VIDIOC_S_INPUT",
+    ],
+    "VIDIOC_G/S_AUDIO": ["VIDIOC_G_AUDIO", "VIDIOC_S_AUDIO"],
+    "VIDIOC_G/S_MODULATOR": ["VIDIOC_G_MODULATOR", "VIDIOC_S_MODULATOR"],
+    "VIDIOC_G/S_FREQUENCY": ["VIDIOC_G_FREQUENCY", "VIDIOC_S_FREQUENCY"],
```
fernando79513, line 34: I think this one is duplicated
```suggestion
```

tomli380576: Thanks for the catch!

checkbox-support/checkbox_support/parsers/v4l2_compliance.py
```diff
@@ -0,0 +1,196 @@
...
+        elif line.endswith(": OK (Not Supported)"):
+            name, is_ioctl_name = get_test_name_from_line(line)
+            if is_ioctl_name:
+                for ioctl_name in TEST_NAME_TO_IOCTL_MAP.get(name, []):
+                    details["not_supported"].append(ioctl_name)
+        elif line.endswith(": FAIL"):
+            name, is_ioctl_name = get_test_name_from_line(line)
+            if is_ioctl_name:
+                for ioctl_name in TEST_NAME_TO_IOCTL_MAP.get(name, []):
+                    details["failed"].append(ioctl_name)
```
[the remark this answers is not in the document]
tomli380576: Ohh that's a lot cleaner ty

checkbox-support/checkbox_support/parsers/tests/test_v4l2_compliance.py
```diff
@@ -0,0 +1,64 @@
...
+import unittest as ut
+from unittest.mock import patch, MagicMock
+from pkg_resources import resource_filename
+
+
+def read_file_as_str(name: str):
+    resource = "parsers/tests/v4l2_compliance_data/{}.txt".format(name)
+    filename = resource_filename("checkbox_support", resource)
+    with open(filename) as f:
+        return f.read()
```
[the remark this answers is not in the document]
tomli380576: Ah I see, I used pkg_resource because I saw another parser test using that to get the test input files. (Also does this mean we are moving towards python 3.9? :D)

checkbox-support/checkbox_support/parsers/v4l2_compliance.py
```diff
@@ -0,0 +1,207 @@
...
+            if name_match is not None:
+                clean_name = (
+                    name_match.group(1)
+                    .replace(":", "")
+                    .replace("(not using libv4l2)", "")
+                    .strip()
+                )
+                if clean_name != "":
+                    summary["device_name"] = clean_name
+            continue
```
Hook25, line 190: The regex is not needed here, match will always match because the line startswith the match and `.*` always matches, regardless, we can do something way simpler here:
```suggestion
        if not summary["device_name"] and line.startswith(
            "Compliance test for"
        ):
            # try to see if there's a device name at the top
            # if not, report None
            clean_name = (
                line
                .replace("Compliance test for ", "")
                .replace(":", "")
                .replace("(not using libv4l2)", "")
                .strip()
            )
            if clean_name:
                summary["device_name"] = clean_name
            continue
```

Hook25 (dismissed):
lgtm, see if the following makes sense to you

PR #1576, opened 2024-11-04
Add cycler method for cycling change the resolution and transform (New)

hanhsuan:
## Description

According to the #855 , parsing the output from gnome_randr is not acceptable. Therefore, add cycler method for cycling change the resolution and transform to instead of modifying the old gnome_randr_cycle.py.

## Resolved issues

[The private bug](https://bugs.launchpad.net/civet-cat/+bug/2036172)

### DIFF
--- a/checkbox-support/checkbox_support/dbus/gnome_monitor.py
+++ b/checkbox-support/checkbox_support/dbus/gnome_monitor.py
@@ -107,7 +107,7 @@ class MonitorConfigGnome(MonitorConfig):
 
     def cycle(
         self,
-        res: bool = True,
+        resolution: bool = True,
         transform: bool = False,
         resolution_filter: Callable[[List[Mode]], List[Mode]] = None,
         action: Callable[..., Any] = None,
@@ -117,7 +117,7 @@ class MonitorConfigGnome(MonitorConfig):
         Automatically cycle through the supported monitor configurations.
 
         Args:
-            res: Cycling the resolution or not.
+            resolution: Cycling the resolution or not.
 
             transform: Cycling the transform or not.
 
@@ -136,7 +136,7 @@ class MonitorConfigGnome(MonitorConfig):
         # ["normal": 0, "left": 1, "inverted": 6, "right": 3]
         trans_list = [0, 1, 6, 3] if transform else [0]
 
-        # for multiple monitors, we need to create res combination
+        # for multiple monitors, we need to create resolution combination
         state = self._get_current_state()
         for monitor, modes in state[1].items():
             monitors.append(monitor)
@@ -178,7 +178,7 @@ class MonitorConfigGnome(MonitorConfig):
                 self._apply_monitors_config(state[0], logical_monitors)
                 if action:
                     action(uni_string, **kwargs)
-            if not res:
+            if not resolution:
                 break
         # change back to preferred monitor configuration
         self.set_extended_mode()
--- a/checkbox-support/checkbox_support/dbus/gnome_monitor.py
+++ b/checkbox-support/checkbox_support/dbus/gnome_monitor.py
@@ -24,7 +24,6 @@ Original script that inspired this class:
 from collections import namedtuple
 from typing import Dict, List, Tuple, Set, Callable, Any
 from gi.repository import GLib, Gio
-from fractions import Fraction
 import itertools
 
 from checkbox_support.monitor_config import MonitorConfig
@@ -109,9 +108,8 @@ class MonitorConfigGnome(MonitorConfig):
     def cycle(
         self,
         res: bool = True,
-        max_res_amount: int = 5,
         transform: bool = False,
-        log: Callable[..., Any] = None,
+        resolution_filter: Callable[List[Mode], List[Mode]] = None,
         action: Callable[..., Any] = None,
         **kwargs
     ):
@@ -121,13 +119,11 @@ class MonitorConfigGnome(MonitorConfig):
         Args:
             res: Cycling the resolution or not.
 
-            max_res_amount: Limit the number of tested configurations
-                to avoid an unnecessarily large test matrix.
-
             transform: Cycling the transform or not.
 
-            log: For logging, the string is constructed by
-                 [monitor name]_[resolution]_[transform]_.
+            resolution_filter: For filtering resolution then returning needed,
+                    it will take List[Mode] as parameter and return
+                    the same data type
 
             action: For extra steps for each cycle,
                     the string is constructed by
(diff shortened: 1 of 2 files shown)


checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
...
+                    m = m_list.pop(0)
+                    uni_string += "{}_{}_{}_".format(
+                        monitor,
+                        m.resolution,
+                        {
+                            0: "normal",
+                            1: "left",
+                            3: "right",
+                            6: "inverted",
+                        }.get(trans),
```
p-gentili, line 166: Let's just make `trans_list` a tuple of `("normal", ... )` instead. No need for this map.

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
...
+        self,
+        res: bool = True,
+        max_res_amount: int = 5,
+        trans: bool = False,
+        log: Callable[..., Any] = None,
+        action: Callable[..., Any] = None,
+        **kwargs
+    ):
+        """
+        Cycling change the monitors configurations automatically
```
p-gentili, line 119: For clearness
```suggestion
        Automatically cycle through the supported monitor configurations.
```

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
         self._apply_monitors_config(state[0], extended_logical_monitors)
         return configuration
 
+    def cycler(
+        self,
+        res: bool = True,
+        max_res_amount: int = 5,
+        trans: bool = False,
```
p-gentili, line 113: Can you name this flag `transform` instead?

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
...
+        **kwargs
+    ):
+        """
+        Cycling change the monitors configurations automatically
+
+        Args:
+            res: Cycling the resolution or not.
+
+            max_res_amount: The supported resolution might be a lot
+                            and this could limited test amount.
```
p-gentili, line 125: For clearness
```suggestion
            max_res_amount: Limit the number of tested configurations
                to avoid an unnecessarily large test matrix.
```

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
...
+                if log:
+                    log(uni_string, **kwargs)
+                if action:
+                    action(uni_string, **kwargs)
+            if not res:
+                break
+        # change back to preferred monitor configuration
+        self.set_extended_mode()
+
+    def _resolution_filter(self, modes: List[Mode], max_res_amount: int):
```
p-gentili, line 191: Can you add a docstring explaining the rationale used to filter out some modes?

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
...
+            modes_list.append(self._resolution_filter(modes, max_res_amount))
+        mode_combination = list(itertools.product(*modes_list))
+
+        for mode in mode_combination:
+            for trans in trans_list:
+                logical_monitors = []
+                position_x = 0
+                m_list = list(mode)
+                uni_string = ""
+                for monitor in monitors:
```
p-gentili, line 156: I hope I got this right. Instead of creating `m_list` and iterating through it via `.pop(0)` why don't we instead use something like:
```
for monitor, m in zip(monitors, mode):
```

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
         self._apply_monitors_config(state[0], extended_logical_monitors)
         return configuration
 
+    def cycler(
```
p-gentili, line 109: I would just call it cycle, which is the action taken by this function. `cycler` sounds more like a class name to me.
```suggestion
    def cycle(
```

p-gentili (changes requested):
Thanks for working on this! A few suggestions below, mostly for clearness.

Do you have a test job we can use for testing?

hanhsuan:
> Thanks for working on this! A few suggestions below, mostly for clearness.
> 
> Do you have a test job we can use for testing?

The example code is

```python
#!/usr/bin/env python3

from checkbox_support.monitor_config import MonitorConfig
from checkbox_support.helpers import display_info
import subprocess
import time

def logger(log: str, **kwargs):
    print(log)

def action(filename, **kwargs):
    time.sleep(5)
    subprocess.check_output(["sudo", "keyboard_mouse"])
    subprocess.check_output(["gnome-screenshot", "-f", filename])

def cycle_test():
    try:
        monitor_config = display_info.get_monitor_config()
    except ValueError as exc:
        raise SystemExit("Current host is not support")

    monitor_config.cycle(res=True, max_res_amount=5, trans=True, log=logger, action=action)

if __name__ == "__main__":
    cycle_test()
```

the source code of keyboard_mouse command is in https://github.com/canonical/checkbox/pull/1582

checkbox-support/checkbox_support/dbus/gnome_monitor.py
hanhsuan: The transform in the `logical_monitors` should be `int`. I don't understand what should I do to use string tuple to archive it.

checkbox-support/checkbox_support/dbus/gnome_monitor.py
```diff
@@ -104,6 +106,107 @@ def set_extended_mode(self) -> Dict[str, str]:
...
+        trans: bool = False,
+        log: Callable[..., Any] = None,
+        action: Callable[..., Any] = None,
+        **kwargs
+    ):
+        """
+        Cycling change the monitors configurations automatically
+
+        Args:
+            res: Cycling the resolution or not.
```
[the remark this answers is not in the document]
hanhsuan: This function is used to replace `id: graphics/{index}_resolution-change_{product_slug}` and `id: graphics/{index}_rotation_{product_slug}`. Therefore, set the `res` to `false` could cycle rotations only. However, you are right, it could be controlled by `max_res_amount`. How about changing the resolution_filter to a callable function ?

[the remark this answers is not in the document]
p-gentili: Oh ok, considering we can still cycle through transformation it makes sense to have the switch. I'd just recommend using the `resolution` name, to make it clearer.

p-gentili (approved):
LGTM, thanks for all the changes!

PR #1577, opened 2024-11-04
Add LXD GPU passthrough tests (New)

pedro-avalos:
## Description

- Created `gpu_passthrough.py` within GPGPU provider
- Added LXD container tests for GPU passthrough setups.
- Created `LXD` and `LXDVM` classes that can be used to wrap LXD container or LXD virtual machine
  - These could probably go into checkbox-support ?

## Resolved issues

- Addresses CHECKBOX-1451 and CHECKBOX-970.

## Documentation

n/a

## Tests

Tested locally on a laptop with NVIDIA GPU. Tested on `torchtusk` as well

Submission from `torchtusk`: https://certification.canonical.com/submissions/status/293943

### DIFF
--- a/providers/gpgpu/bin/gpu_passthrough.py
+++ b/providers/gpgpu/bin/gpu_passthrough.py
@@ -74,8 +74,10 @@ GPU_VENDORS = {
                 "security.secureboot=false",
             ],
             "config_cmds": [
+                "apt-get -q update",
+                "apt-get -q upgrade",
                 "apt-get -q install -y ubuntu-drivers-common",
-                "ubuntu-drivers install",
+                "ubuntu-drivers install --gpgpu",
             ],
         },
     },
--- a/providers/gpgpu/bin/gpu_passthrough.py
+++ b/providers/gpgpu/bin/gpu_passthrough.py
@@ -27,8 +27,7 @@ import subprocess
 import time
 import urllib
 import uuid
-from dataclasses import dataclass
-from typing import List, Optional
+from typing import Dict, List, Optional
 from urllib.parse import urlparse
 
 try:
@@ -53,32 +52,22 @@ except ImportError:
 ARCH = os.uname().machine
 
 
-@dataclass
-class Repository:
-    """Represents an apt repository."""
-
-    name: str
-    repo_line: str
-    gpg_url: str
-    gpg_fingerprint: str
-    pinfile: Optional[str] = None
-
-
+# XXX: The repository part of this should go away if we package mixbench
 NVIDIA_URL = "https://developer.download.nvidia.com/compute/cuda/repos/ubuntu{}/{}".format(  # noqa: E501
     RELEASE.replace(".", ""), ARCH
 )
 GPU_VENDORS = {
     "nvidia": {
         "repos": [
-            Repository(
-                name="cuda",
-                repo_line="{} /".format(NVIDIA_URL),
-                gpg_url="{}/3bf863cc.pub".format(NVIDIA_URL),
-                gpg_fingerprint="EB693B3035CD5710E231E123A4B469963BF863CC",
-                pinfile="{}/cuda-ubuntu{}.pin".format(
+            {
+                "name": "cuda",
+                "repo_line": "{} /".format(NVIDIA_URL),
+                "gpg_url": "{}/3bf863cc.pub".format(NVIDIA_URL),
+                "gpg_fingerprint": "EB693B3035CD5710E231E123A4B469963BF863CC",
+                "pinfile": "{}/cuda-ubuntu{}.pin".format(
                     NVIDIA_URL, RELEASE.replace(".", "")
                 ),
-            )
+            }
         ],
     },
 }
@@ -292,26 +281,32 @@ class LXDVM(LXD):
 
 
 # XXX: If we package the mixbench program, this wouldn't be needed
-def add_apt_repo(instance: LXD, repo: Repository):
+def add_apt_repo(instance: LXD, repo: Dict):
     """Adds an APT repository to a LXD instance."""
-    logging.debug("Downloading GPG key from %s", repo.gpg_url)
+    name = repo["name"]
+    repo_line = repo["repo_line"]
+    gpg_url = repo["gpg_url"]
+    gpg_fingerprint = repo["gpg_fingerprint"]
+    pinfile = repo["pinfile"]
+
+    logging.debug("Downloading GPG key from %s", gpg_url)
     temp_keyring = "./tmp.gpg"
-    gpg_dest = "/usr/share/keyrings/{}.gpg".format(repo.name)
+    gpg_dest = "/usr/share/keyrings/{}.gpg".format(name)
     instance.run(
-        "wget -O {}.gpg '{}'".format(repo.name, repo.gpg_url),
+        "wget -O {}.gpg '{}'".format(name, gpg_url),
         on_guest=True,
         check=True,
     )
     instance.run(
         "gpg --no-default-keyring --keyring {} --import {}.gpg".format(
-            temp_keyring, repo.name
+            temp_keyring, name
         ),
         on_guest=True,
         check=True,
     )
     instance.run(
         "gpg --no-default-keyring --keyring {} --fingerprint {}".format(
-            temp_keyring, repo.gpg_fingerprint
+            temp_keyring, gpg_fingerprint
         ),
         on_guest=True,
         check=True,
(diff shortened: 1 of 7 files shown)


providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+                proc = self.run(
+                    "lxc image copy {}{} local: --alias {}".format(
+                        self.remote, RELEASE, self.image_alias.hex
+                    )
+                )
+                if proc.returncode == 0:
+                    return
+                logging.error("Error encountered while importing images")
+                logging.error("Attempting up to %d times", retries)
+            raise RuntimeError("Images could not be inserted")
```
Hook25, line 192: Here lets use the same retry helper as a function, also, 2 is enough? I would make it 5 just because
```suggestion
            from checkbox_support.helpers.retry import run_with_retry
            run_with_retry(
                self.run, 
                2, 
                2, 
                "lxc image copy {}{} local: --alias {}".format(
                    self.remote, RELEASE, self.image_alias.hex
                )
            )
```

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def init_lxd(self):
+        """Initializes LXD."""
+        if self.run("lxd waitready --timeout=5").returncode != 0:
+            self.run("lxd init --auto", check=True)
```
Hook25, line 197: We switched to exceptions, so lets change this as well:
```suggestion
        try:
            self.run("lxd waitready --timeout=5")
        except CalledProcessError:
            self.run("lxd init --auto")
```

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+        self.insert_images()
+
+    def cleanup(self):
+        """Cleans up instance."""
+        self.run("lxc image delete {}".format(self.image_alias.hex))
+        self.run("lxc delete --force {}".format(self.name))
```
Hook25, line 203:
```suggestion
        self.run("lxc image delete {}".format(self.image_alias.hex), ignore_errors=True)
        self.run("lxc delete --force {}".format(self.name), ignore_errors=True)
```

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def launch(self, options: Optional[List] = None):
+        """Sets up and creates the instance."""
+        if not options:
+            options = []
+
+        cmd = "lxc launch {} {}".format(self.image_alias.hex, self.name)
+        if options:
+            cmd = "{} {}".format(cmd, " ".join(options))
+        self.run(cmd, check=True)
```
Hook25, line 213:
```suggestion
    def launch(self, options: Optional[List] = []):
        """Sets up and creates the instance."""
        cmd = ["lxc", "launch", self.image_alias.hex, self.name]
        cmd += options
        self.run(shlex.join(cmd))
```

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+        if self.template and self.image:
+            logging.debug("Importing images into LXD")
+            self.run(
+                "lxc image import {} {} --alias {}".format(
+                    self.template, self.image, self.image_alias.hex
+                ),
+                check=True,
+            )
+
+    def launch(self, options=None):
```
Hook25, line 257: Consider changing this as above

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def add_device(self, device: str, device_type: str, options=None):
+        # Hot plugging is only supported on containers
+        self.stop(force=True)
+        super().add_device(device, device_type, options)
+        self.start()
+
+
+# XXX: If we package the mixbench program, this wouldn't be needed
+def add_apt_repo(instance: LXD, repo: Dict):
```
Hook25, line 284: I suppose this is no longer needed from the comment and your current progress?

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+        check=True,
+    )
+
+    logging.debug("Updating APT cache")
+    instance.run("apt-get -q update", on_guest=True, check=True)
+
+
+# TODO: Package test program (e.g., as a snap) to simplify this logic
+#       If we package mixbench, then this function goes away entirely
+def build_gpu_test(instance: LXD, vendor: str):
```
Hook25, line 354: Same as above

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+    try:
+        instance.init_lxd()
+        options = []
+        if args.vendor == "nvidia":
+            options = ["-c nvidia.runtime=true"]
+        instance.launch(options=options)
+        instance.add_device("gpu", "gpu", options=["pci={}".format(args.pci)])
+
+        logging.info("Waiting for network to be up")
+        time.sleep(20)
```
Hook25, line 447: Please don't, try to "discover" if the network is up, as that may take 1ms -> 20s

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+        )
+        raise SystemExit(1)
+
+
+def test_lxd_gpu(args):
+    """Tests GPU passthrough with a LXD container."""
+    logging.info("Executing LXD GPU passthrough test")
+
+    instance = LXD(args.template, args.rootfs)
+    try:
```
Hook25, line 438: This makes me wonder if both LXD and LXDVM should be context managers, that would make it easier to read this and use the functionality in general. Something like
```
with LXD(...) as instance: 
    ...
```

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+    finally:
+        instance.cleanup()
+
+
+def test_lxdvm_gpu(args):
+    """Tests GPU passthrough with a LXD virtual machine."""
+    logging.info("Executing LXD VM GPU passthrough test")
+
+    instance = LXDVM(args.template, args.image)
+    try:
```
Hook25, line 460: Same as above

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+            instance.run(
+                "apt-get -q install -y ubuntu-drivers-common",
+                on_guest=True,
+                check=True,
+            )
+            instance.run("ubuntu-drivers install", on_guest=True, check=True)
+        instance.restart()
+
+        logging.info("Waiting for network to be up")
+        time.sleep(20)
```
Hook25, line 482: Same as above

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        dest="log_level",
+        action="store_const",
+        default=logging.INFO,
+        const=logging.DEBUG,
+        help="Increase logging level",
+    )
```
Hook25, line 503: I'm not sure I see the value of this, why would we hide the output of most commands most of the times?

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -0,0 +1,576 @@
...
+                "gpg_fingerprint": "EB693B3035CD5710E231E123A4B469963BF863CC",
+                "pinfile": "{}/cuda-ubuntu{}.pin".format(
+                    NVIDIA_URL, RELEASE.replace(".", "")
+                ),
+            }
+        ],
+    },
+}
+GPU_RUNS = 20
+GPU_THRESHOLD_SEC = 12.0  # TODO: finetune
```
Hook25, line 75: Please document that these can be overridden via an envvar or a parameter to the script

providers/gpgpu/units/test-plan.pxu
```diff
@@ -23,3 +23,14 @@ include:
     gpgpu/rvs-pebb
     gpgpu/rvs-pbqt
     gpgpu/rvs-babel
+
+id: gpgpu-virtualization
```
Hook25, line 27: Why is this called gpgpu virtualization and not passthrough?

Hook25: ditto for the other fields

providers/gpgpu/bin/gpu_passthrough.py
pedro-avalos: Resolved, but we should not be using mutable default arguments.

providers/gpgpu/bin/gpu_passthrough.py
pedro-avalos: Yeah, last time I tried to use ping within the system, it did not work. Maybe inetutils was not installed by default?

providers/gpgpu/bin/gpu_passthrough.py
pedro-avalos: Oh, that would be nice. I'll `try` to do that (pun intended)

providers/gpgpu/bin/gpu_passthrough.py
Hook25: You should not mutate default arguments in my opinion, I don't agree one should completely abandon the thought of default values if they are not immutable and implement them manually via a default value + if else.

providers/gpgpu/bin/gpu_passthrough.py
pedro-avalos: I've added a few more `logging.info` messages to show what step the program is on.
I also increased the level for some error messages that used to be debug before. If you think I should do more, let me know.

providers/gpgpu/units/test-plan.pxu
```diff
@@ -23,3 +23,13 @@ include:
...
+
+id: gpgpu-passthrough
+unit: test plan
+_name: GPGPU Virtualization Passthrough Testing
+_description:
+ Automated Tests for GPGPU Passthrough (non-graphical)
+include:
+    gpgpu/lxd-nvidia-gpu-passthrough-index
+bootstrap_include:
+    graphics_card
```
Hook25, line 35:
```suggestion
    graphics_card
    executable
    package
    snap
```

Hook25 (approved):
Consider adding these in the boostrap include section, they wont be run only if the template doesn't expand

providers/gpgpu/units/test-plan.pxu
pedro-avalos: Done!

Hook25 (approved):

PR #1581, opened 2024-11-06
Simplify GPGPU jobs by using siblings field (New)

pedro-avalos:
## Description

- Simplifies and reduces the line length for the jobs defined in the GPGPU provider


fernando79513 (approved):
LGTM +1!
It looks cleaner now.

PR #1582, opened 2024-11-07
Add the keyboard mouse random executable (New)

GabrielChenCC:
## Description

We discovered this problem during last year's testing: https://bugs.launchpad.net/sutton/+bug/2045249 . We found this bug by manually coordinating the random use of the mouse and keyboard during screen rotation. Now I use this C code to implement the random automatic behavior of the mouse and keyboard. It can be used in conjunction with another PR https://github.com/canonical/checkbox/pull/1576 submitted by my colleague Hanhsuan to complete the random movement of the mouse and keyboard at different resolutions while rotating the screen.

Note: The keyboard and mouse random C code needs to execute via 'sudo'. We have also tried to use Python's uinput to avoid using sudo permissions, but by looking up the document https://python-evdev.readthedocs.io/en/latest/tutorial.html#injecting-input, Python cannot avoid sudo either, and C code is available for the rotation test case without introducing external dependence. So I choose C code, it is more suitable for the rotation test case.  

## Resolved issues

https://warthogs.atlassian.net/browse/OEMQA-5102 
https://bugs.launchpad.net/sutton/+bug/2045249

## Tests

$ gcc keyboard-mouse_test.c -o demo 
$ sudo ./demo 
The demo file works as design

Add keyboard-mouse_test.c to Makefile and compile without any errors

### DIFF
--- a/providers/base/src/EXECUTABLES
+++ b/providers/base/src/EXECUTABLES
@@ -3,3 +3,4 @@ clocktest
 threaded_memtest
 ptrace_test
 testptp
+keyboard-mouse_test
--- a/providers/base/src/Makefile
+++ b/providers/base/src/Makefile
@@ -1,5 +1,5 @@
 .PHONY:
-all: alsa_test clocktest threaded_memtest ptrace_test testptp
+all: alsa_test clocktest threaded_memtest ptrace_test testptp keyboard-mouse_test
 
 ifeq ($(shell uname -m), x86_64)
 all: testptp
@@ -15,5 +15,6 @@ clocktest: CFLAGS += -D_POSIX_C_SOURCE=199309L -D_DEFAULT_SOURCE
 clocktest: LDLIBS += -lrt
 alsa_test: CXXFLAGS += -std=c++11
 alsa_test: LDLIBS += -lasound -pthread
+keyboard-mouse_test:
 
 CFLAGS += -Wall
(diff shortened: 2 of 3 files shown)


pieqq:
I think your PR is missing the C code files?

pieqq (changes requested):
I noticed there is an `EXECUTABLES` file in `src/` that contains the name of the binaries that are built by the base provider.

I'm not sure what this does exactly, but I know [Checkbox uses the content of this file somehow](https://github.com/canonical/checkbox/blob/4305ba10c5e2c7ed4053478a23c8ac6a24d88079/checkbox-ng/plainbox/impl/secure/providers/v1.py#L535-L543).

Could you add the name of the generated binary (`keyboard-mouse_test`) to it?

Otherwise, I've successfully ran the `./manage.py build` command, and the binary is available in `build/bin/`, and I could run it without `sudo` (on Noble), so it's good to land once this change is made.

GabrielChenCC:
> I noticed there is an `EXECUTABLES` file in `src/` that contains the name of the binaries that are built by the base provider.
> 
> I'm not sure what this does exactly, but I know [Checkbox uses the content of this file somehow](https://github.com/canonical/checkbox/blob/4305ba10c5e2c7ed4053478a23c8ac6a24d88079/checkbox-ng/plainbox/impl/secure/providers/v1.py#L535-L543).
> 
> Could you add the name of the generated binary (`keyboard-mouse_test`) to it?
> 
> Otherwise, I've successfully ran the `./manage.py build` command, and the binary is available in `build/bin/`, and I could run it without `sudo` (on Noble), so it's good to land once this change is made.

Thank you for your suggestion. This file is indeed something I didn't notice. I have added the corresponding fields to this file. Please try again for me. Thank you.

PR #1584, opened 2024-11-07
Remove usb-c/hid job (New)

pieqq:
This job was originally written when support for USB-C ports was introduced. However, it doesn't really make sense to have this, given that:

- there is a `usb-c/c-to-a-adapter/hid` job
- when a HID is used over USB-C, it uses the USB 2.0 bus used by USB-A HID

This job is a cert-blocker and impacts issuance of Certificate for some projects, but there is not much value in having it in the certification program.

## Tests

```
$ ./manage.py validate
(...)
```


seankingyang:
LGTM~~

Hook25 (approved):

PR #1585, opened 2024-11-07
Add manifest gate for obex tests (bugfix)

Hook25:
## Description

The obex test was introduced with the only gate being the presence of the software to test it. This is problematic because the software will always be installed in the snap, but not all platforms support this extension. Additionally one needs the hardware to test this capability as well. For this reason we have decided to introduce a new manifest to gate the execution of this test because it is currently failing in sru (we had to manually exclude it) and in snap testing (where it is non-trivial to exclude). 

This also removes the test from the exclusion list, this way once we have the hardware to do sru obex tests we just need to enable it via the manifest without changing the test plans

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1646

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/providers/base/units/bluetooth/manifest.pxu
+++ b/providers/base/units/bluetooth/manifest.pxu
@@ -7,3 +7,8 @@ unit: manifest entry
 id: has_bt_smart
 _name: A Bluetooth Module with Smart (4.0 or later) Support
 value-type: bool
+
+unit: manifest entry
+id: has_bt_obex_support
+_name: A Bluetooth Module with OBEX Support
+value-type: bool
--- a/providers/base/units/bluetooth/manifest.pxu
+++ b/providers/base/units/bluetooth/manifest.pxu
@@ -10,5 +10,5 @@ value-type: bool
 
 unit: manifest entry
 id: has_bt_obex_support
-_name: A Bluetooth Module with OBEX Support
+_name: A Bluetooth Module with OBject EXchange (OBEX) Support
 value-type: bool
--- a/providers/base/units/bluetooth/jobs.pxu
+++ b/providers/base/units/bluetooth/jobs.pxu
@@ -434,10 +434,12 @@ id: bluetooth/bluetooth_obex_send
 depends: bluetooth/detect-output
 estimated_duration: 10.0
 environ: BTDEVADDR PLAINBOX_PROVIDER_DATA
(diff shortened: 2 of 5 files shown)


providers/base/units/bluetooth/manifest.pxu
```diff
@@ -7,3 +7,8 @@ unit: manifest entry
 id: has_bt_smart
 _name: A Bluetooth Module with Smart (4.0 or later) Support
 value-type: bool
+
+unit: manifest entry
+id: has_bt_obex_support
+_name: A Bluetooth Module with OBEX Support
```
pieqq, line 13: This is not the problem we are trying to solve. All the BT modules have OBEX support for what I know.

The problem is that we don't have the infrastructure to support testing OBEX everywhere in the lab. In order for the BT OBEX test to work, you need an OBEX server configured and close enough to the DUT to send and receive files to.

The manifest should be named something like `bt_obex_server_available` instead, but then we have to make it very clear that this is required for QA or ODM to run, otherwise they may just set this manifest to `False` and the related test(s) would be skipped...

pieqq (changes requested):
I think the problem is not the DUT support of OBEX feature, it's an infrastructure problem, so the naming of the manifest is incorrect, and anyway we need it to be very clear that this should be set to `True` when doing QA/ODM enablement testing.

providers/base/units/bluetooth/manifest.pxu
Hook25: The fact that we are going to abuse this manifest key for the purpose of skipping the test in SRU in my opinion doesn't make it less valuable. In the form I have created it, it is a question about a capability. We may decide to lie about the capability in our own testing for our own purposes, but if we create `bt_obex_server_available`, that requires the description to be updated to say to always set it to true. While if we call it what it actually is, I think it is more clear to everybody when to set it true (which is basically all the times)

pieqq, line 13:
```suggestion
_name: A Bluetooth Module with OBject EXchange (OBEX) Support
```

pieqq (approved):
After discussing with Max, agreed to the proposed change with a little change in the name.

pieqq (approved):

pieqq (approved):

Issue #1559, opened 2024-10-23
Increase the threshold for camera-quality_video.*


kevinyehk:
### Enhancement Proposal

Some images from the camera have higher noise, base on the conversion with @fernando79513, we could increase the threshold of the algorithm to pass the test.

fernando79513:
I look at some recomendations for rescaling the image and changing the interpolation method to `AREA` fixed the issue with this device, since it reduced the noise produced when rescaling the image. We are still failing with low quality images, but now the test pass with this device

fernando79513:
Original image:
![download](https://github.com/user-attachments/assets/79366f2f-a8ba-4f76-841d-255b6a7e6d62)

NEAREST interpolation
![resized_image_nearest](https://github.com/user-attachments/assets/34417071-d1d9-4e68-800a-11f7897b48f5)
The BRISQUE score is too high: 65.15044131116338 > 60

AREA interpolation
![resized_image_area](https://github.com/user-attachments/assets/9777cfb3-7512-4f58-93c0-559953fe5362)
BRISQUE score: 32.42782220073718

fernando79513:
Bad quality images:

![wrong_camera_color](https://github.com/user-attachments/assets/4ad77f81-cbca-4f6c-a50b-d0cfe1ce1d06)
wrong_camera_color.jpeg
The BRISQUE score is too high: 159.1849551827096 > 60

![2023-09-27-113128](https://github.com/user-attachments/assets/e6e01c79-d049-4214-9973-d3a4aa291d1c)
The BRISQUE score is too high: 80.10864111807274 > 60

![frame_0](https://github.com/user-attachments/assets/6879c74d-507c-470c-a9d1-b427966014a6)
Unable to compute BRISQUE score

Hook25:
Fixed in #1586

PR #1586, opened 2024-11-07
Changed resizing method from NEAREST to AREA (BugFix)

fernando79513:
## Description

The quality test was failing with an image with acceptable quality (see #1559). We have modified the interpolation algorithm so it does not accentuate the noise when downscaling the image, changing from NEAREST to AREA interpolation.

## Resolved issues

#1559 

## Documentation

N/A

## Tests

Tested with the failing image (see #1559)
Tested on the failing device. The score went from:
`The BRISQUE score is too high: 72.13535618808396 > 60`
to
`BRISQUE score: 32.9690039977346`


Hook25 (approved):
ty for the example pictures

PR #1588, opened 2024-11-08
Fixing type errors in prime_offload_tester.py (Bugfix)

tomli380576:
## Description

This PR fixes an issue where if `prime_offload_tester.py` was run with arguments like `python3 prime_offload_tester.py -d i915 -p 0000:00:02.0` and the command to check is `glxgears`, the test always passes because it was checking if `"glxgears"[0]` is one of the clients of said GPU. Since "glxgears"[0] == 'g', it always appears in the `/sys/kernel/debug/dri/<card_pci_id>/clients`

Type annotations were also replaced with ones supported in python 3.5


Hook25 (approved):
LGTM

PR #1589, opened 2024-11-08
Update the setup lxd action (infra)

Hook25:
## Description

Currently the metabox action is broken due to the runner user being unable to communicate with the lxd socket. This may be due to many reasons but from the debug log I can't tell how it sets up the group for the normal user, so I'm unable to understand how it ever worked before. Regardless, updating the action (the tag is 2 years old) seems to fix the issue.

## Resolved issues

Fixes: CHECKBOX-1649

## Documentation

N/A

## Tests

failing pre patch: https://github.com/canonical/checkbox/actions/runs/11742516038/job/32714410426
passing after patch: https://github.com/canonical/checkbox/actions/runs/11743071203/job/32716219919

### DIFF
--- a/.github/workflows/metabox.yaml
+++ b/.github/workflows/metabox.yaml
@@ -52,7 +52,7 @@ jobs:
       - name: Checkout Checkbox monorepo
         uses: actions/checkout@v4
       - name: Setup LXD
-        uses: canonical/setup-lxd@v0.1.1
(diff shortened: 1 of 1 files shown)


fernando79513 (approved):
LGTM +1!
Let's use this for now and we will move to the new tag once it's created in `setup-lxd`

PR #1597, opened 2024-11-12
Fix suspend time integrated gpu (BugFix)

fernando79513:

### DIFF
--- a/providers/base/units/suspend/suspend-graphics.pxu
+++ b/providers/base/units/suspend/suspend-graphics.pxu
@@ -266,7 +266,12 @@ depends:
   suspend/suspend_advanced_auto
  {%- endif %}
 estimated_duration: 1.2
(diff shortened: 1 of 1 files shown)


Hook25 (changes requested):
Do you really need the new jinja thing? it is checking if the file exists either way, cant we 
```
if [ -e "$PLAINBOX_SESSION_SHARE/{{ index }}_suspend_single.log" ]; then
    cat "$PLAINBOX_SESSION_SHARE/{{ index }}_suspend_single.log"
elif [ -e "$PLAINBOX_SESSION_SHARE/suspend_single.log" ]; then
    cat "$PLAINBOX_SESSION_SHARE/suspend_single.log"
else
    echo "No suspend log files found"
fi
    
```

Hook25 (approved):

PR #1601, opened 2024-11-14
Refactor DSS provider tests (New)

motjuste:
## Description

The tests for DSS provider in `contrib` have been refactored so that they use individual scripts now available in the relevant `bin` folder instead of scripts written inline in the jobs' command.  The tests end up checking the same things, but are now better implemented.

The shell scripts were extracted out into one of multiple ones that roughly group related tests, including those generic to DSS, those for setting up Intel GPUs, those for using ITEX, and those for using IPEX.  Furthermore, larger Python scripts that were previously written in-line have been factored out into individual ones.  Please note that these Python scripts are not meant to be run by the jobs directly, instead they are submitted via the job to DSS to be run in the appropriate Kubernetes Pods, hence the Python dependencies of these scripts are not relevant to the provider itself.

## Resolved issues

- [CHECKBOX-1587](https://warthogs.atlassian.net/browse/CHECKBOX-1587)

## Documentation

No updates to Checkbox documentation.

## Tests

Manual testing with the following Testflinger job spec had a [successful run](https://testflinger.canonical.com/jobs/bffec1d2-58a3-4e46-81d0-a3d8924ebfd1).

<details>
<summary>Testflinger job-spec</summary>

```yaml
job_queue: dell-precision-3470-c30322
provision_data:
  distro: jammy
test_data:
  test_cmds: |
    ssh -t ubuntu@$DEVICE_IP '
      export DEBIAN_FRONTEND=noninteractive
      sudo apt update
      sudo apt -y install git
      git clone -b CHECKBOX-1587-refactor-dss-provider https://github.com/canonical/checkbox.git
      cd checkbox/contrib/checkbox-dss-validation
      sudo snap install --classic snapcraft
      sudo snap install checkbox22
      lxd init --auto
      snapcraft
      sudo snap install --dangerous --classic ./checkbox-dss_*_amd64.snap
      checkbox-dss.install-deps
      checkbox-dss.validate-intel-gpu
    '
```

</details>

Furthermore, the relevant GitHub workflow was also manually run with results available [here](https://github.com/canonical/checkbox/actions/runs/11844409332)

[CHECKBOX-1587]: https://warthogs.atlassian.net/browse/CHECKBOX-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/testflinger-contrib-dss-regression.yaml
+++ b/.github/workflows/testflinger-contrib-dss-regression.yaml
@@ -40,7 +40,7 @@ jobs:
           -e "s|REPLACE_QUEUE|${{ matrix.queue }}|" \
           -e "s|REPLACE_PROVISION_DATA|${PROVISION_DATA}|" \
           -e "s|REPLACE_DSS_CHANNEL|${{ matrix.dss_channel }}|" \
-          ${GITHUB_WORKSPACE}/testflinger/job-def.yaml > \
+          ${GITHUB_WORKSPACE}/contrib/checkbox-dss-validation/testflinger/job-def.yaml > \
           ${GITHUB_WORKSPACE}/job.yaml
       - name: Build job file from template with oemscript provisioning
         if: ${{ matrix.queue == 'dell-precision-5680-c31665' }}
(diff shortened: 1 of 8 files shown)


pieqq (approved):
Excellent work, and excellent PR! It's clear, well documented and tested, and each change is its own commit, which makes review much easier!

You did a good job of separating the pieces of shell from the job definitions into their own scripts (and even did a more than welcome refactor on that).

It's the first time I see a Python script being used as data to be fed to a test script. Usually, for test data, we use the `data/` directory in the provider (instead of `bin/`), and this can be called via the `$PLAINBOX_PROVIDER_DATA` environment variable (see an example [here,](https://github.com/canonical/checkbox/blob/1cd8e52e7177c314219fb89cbc3c8c10b7eee555/providers/base/units/touchscreen/jobs.pxu#L75) for instance). That said, since this is a Python script, I'm not sure where it should live, so I think your implementation is OK (even though it means a bit of stretching the Bash script to find the data back).

Good job!

Issue #1557, opened 2024-10-23
Cannot export the submission tar file after finishing the session when the test plan is not whole


weizhenwu:
### Bug Description

Checkbox cannot export the submission tar file after finishing the session.
The errors below will be reported and the session will stay incomplete.
It happens only when I don't run the whole test plan but select partial tests.

* The latest version to work without these errors is 4.2.0-dev99

### To Reproduce
1. install following
```
snap install checkbox22 --beta
snap install checkbox-carmel-classic --classic --edge
snap install checkbox-ce-oem --channel=22.04/edge --classic
```
2. Launch checkbox `checkbox-carmal-classic.checkbox-cli`
3. Select `carmel-classic-automated`
4. Select only camera tests and start testing
5. Press F to export to submission after finishing

### Environment

OS: ubuntu server 22.04
Checkbox type: snap
Checkbox version: 4.2.0-dev121
DUT: 202409-35561

### Relevant log output

```shell
ERROR:plainbox.bug:Undeclared exception UndefinedError raised from export_to_transport
ERROR:checkbox-ng.launcher.stages:Problem with a '2_tar_file' report using 'com.canonical.plainbox::tar' exporter sent to '/home/ubuntu/.local/share/checkbox-ng/submission_2024-10-23T03.08.00.817432.tar.xz' transport. Reason 'dict object' has no attribute 'com.canonical.contrib::ce-oem-info/snapd_installation_time'
Traceback (most recent call last):
  File "/snap/checkbox22/current/lib/python3.10/site-packages/checkbox_ng/launcher/stages.py", line 669, in _export_results
    result = self.sa.export_to_transport(
  File "/snap/checkbox22/current/lib/python3.10/site-packages/plainbox/impl/decorators.py", line 153, in wrapper
    raise exc
  File "/snap/checkbox22/current/lib/python3.10/site-packages/plainbox/impl/decorators.py", line 145, in wrapper
    return func(*args, **kwargs)
  File "/snap/checkbox22/current/lib/python3.10/site-packages/plainbox/impl/session/assistant.py", line 1823, in export_to_transport
    exporter.dump_from_session_manager(self._manager, exported_stream)
  File "/snap/checkbox22/current/lib/python3.10/site-packages/plainbox/impl/exporter/tar.py", line 74, in dump_from_session_manager
    exporter.dump_from_session_manager(manager, _s)
[...]

PR #1605, opened 2024-11-18
Restore all jobs in the job state map (bugfix)

Hook25:
## Description

This fixes the bug with the session that when a test is de-selected, any resume of the session will make the submission.json generation fail. This was because the job_state_map is not entirely re-loaded after the resume, but only the tests that were selected are correctly put in the map. 

Additionally, the validation of tests would not cover rejected tests (which is dangerous, as a test may be deselected for a reason, and that reason may not hold anymore). 

Finally there was a bug in the session resume where tests were only incidentally not re-run as the rejected job list was not re-evaluated after resume.

Minor: this fixes a bug with `ExpectNot`, as it wouldn't work if the checkbox controller didnt finish running, now it does

## Resolved issues

Fixes: CHECKBOX-1627
Fixes: https://github.com/canonical/checkbox/issues/1557

## Documentation

N/A

## Tests

This adds a new metabox scenario for the regression

### DIFF
--- a/checkbox-ng/plainbox/impl/session/test_resume.py
+++ b/checkbox-ng/plainbox/impl/session/test_resume.py
@@ -2033,11 +2033,13 @@ class SessionJobsAndResultsResumeTests(TestCaseWithParameters):
 
     def test_simple_session_with_rejected(self):
         """
-        verify that _restore_SessionState_jobs_and_results() works when
-        faced with a representation of a simple session (no generated jobs
-        or anything "exotic").
+        verify that _restore_SessionState_jobs_and_results() also resumes in
+        the job_state_map jobs that were manually excluded from the user
+        (rejected jobs)
         """
         job = make_job(id="job")
+        # this job was de-selected by the user manually in the test selection
+        # screen
         job1 = make_job(id="job1")
         session_repr = {
             "jobs": {job.id: job.checksum, job1.id: job1.checksum},
--- a/checkbox-ng/plainbox/impl/session/test_resume.py
+++ b/checkbox-ng/plainbox/impl/session/test_resume.py
@@ -2031,6 +2031,41 @@ class SessionJobsAndResultsResumeTests(TestCaseWithParameters):
         # as specific tests for restoring results are written elsewhere
         self.assertEqual(session.job_state_map[job.id].result.outcome, "pass")
 
+    def test_simple_session_with_rejected(self):
+        """
+        verify that _restore_SessionState_jobs_and_results() works when
+        faced with a representation of a simple session (no generated jobs
+        or anything "exotic").
+        """
+        job = make_job(id="job")
+        job1 = make_job(id="job1")
+        session_repr = {
+            "jobs": {job.id: job.checksum, job1.id: job1.checksum},
+            "results": {
+                job.id: [
+                    {
+                        "outcome": "pass",
+                        "comments": None,
+                        "execution_duration": None,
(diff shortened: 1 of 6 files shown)


checkbox-ng/plainbox/impl/session/test_resume.py
```diff
@@ -2031,6 +2031,41 @@ def test_simple_session(self):
         # as specific tests for restoring results are written elsewhere
         self.assertEqual(session.job_state_map[job.id].result.outcome, "pass")
 
+    def test_simple_session_with_rejected(self):
+        """
+        verify that _restore_SessionState_jobs_and_results() works when
+        faced with a representation of a simple session (no generated jobs
+        or anything "exotic").
```
pieqq, line 2038: This is the same comment as `test_simple_session`, I guess you may want to update it to mention the rejected jobs.

checkbox-ng/plainbox/impl/session/test_resume.py
```diff
@@ -2031,6 +2031,41 @@ def test_simple_session(self):
...
+        session = SessionState([job, job1])
+        helper._restore_SessionState_jobs_and_results(session, session_repr)
+        # Session still has one job in it
+        self.assertEqual(session.job_list, [job, job1])
+        # Resources don't have anything (no resource jobs)
+        self.assertEqual(session.resource_map, {})
+        # The result was restored correctly. This is just a smoke test
+        # as specific tests for restoring results are written elsewhere
+        self.assertEqual(session.job_state_map[job.id].result.outcome, "pass")
+        self.assertIn(job1.id, session.job_state_map)
```
pieqq, line 2067: Add a comment like `# job1 is a rejected job, but it should still be restored in the job_state_map after the session is resumed`.

pieqq (changes requested):
Great stuff! Thanks for spending the time to figure out this issue.

For the unit tests, it looks like `SessionJobsAndResultsResumeTests` is parametrized somehow, so it might be worth making sure the right version of `SessionResumeHelper` is tested.

Otherwise, I really like the Metabox test, it's very clear.

A few other comments in the unit tests, but otherwise I'd say it's good to land!

checkbox-ng/plainbox/impl/session/test_resume.py
Hook25: done

checkbox-ng/plainbox/impl/session/test_resume.py
Hook25: done

pieqq (approved):
Awesome! Thanks!

metabox/metabox/scenarios/restart/agent_respawn.py
```diff
@@ -198,3 +199,43 @@ class ResumePreservesRejectedJobsStateMap(Scenario):
...
     ]
+
+
+@tag("resume", "manual", "regression")
+class RemoteResumePreservesRejectedJobsStateMap(Scenario):
+    """
+    Check that the job_state_map survives both manual closure and restarts
+
+    This differs from Local because in remote the controller waits for the
+    agent to come back, we loose the output of the rebooting job and we don't
```
pieqq, line 210:
```suggestion
    agent to come back, we lose the output of the rebooting job and we don't
```

pieqq (dismissed):
Awesome stuff! I just found a little typo (loose/lose), but otherwise it's good to land.

PR #1607, opened 2024-11-19
Add the missing sort for lsusb in reboot_check_test.py and fix unused device comparison result (BugFix)

tomli380576:
## Description

- Added a sort step after calling checkbox-support-lsusb
- Suppress FWTS progress messages to make stderr cleaner and errors more visible
- Made the comparison output more verbose when a comparison fails
- Fixed an issue where device comparison result was not updated after calling `compare_device_lists`

## Resolved issues

1. The original shell script had an explicit `checkbox-support-lsusb | sort` step that was missing in the python version. Adding this back to preserve the same behavior.
2. Device comparison result was never updated. This becomes a problem if the only issue on the DUT is device comparison

## Tests

Unit tests

Sideloaded the fix on 202312-33267 and plugged in an USB stick during reboot to force a device difference. The fix correctly reports that the device config changed and failed the test case

### DIFF
--- a/providers/base/bin/reboot_check_test.py
+++ b/providers/base/bin/reboot_check_test.py
@@ -106,7 +106,10 @@ class DeviceInfoCollector:
             expected = "{}/{}_log".format(expected_dir, device)
             actual = "{}/{}_log".format(actual_dir, device)
             if not filecmp.cmp(expected, actual):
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
+1, lgtm

PR #1621, opened 2024-11-26
Print time stamps and drm connection status values in reboot checks (New)

tomli380576:
## Description

We were having some difficulty narrowing down the cause when checkbox reports the DUT was not doing hardware rendering. It was kinda hard to tell if it was due to:
 - checkbox running so early in the boot process that none of the graphics related steps were even started by Ubuntu 
 - the DUT is actually broken 
 - display connection was moved to another output (like `cardUnknown`) 

This PR adds a timestamp output and lists all the drm connection status during the test

## Resolved issues

No real issue here, this is more like a QoL change for QA

## Documentation

Sample output:
image

## Tests

Original unit tests pass

### DIFF
--- a/providers/base/bin/reboot_check_test.py
+++ b/providers/base/bin/reboot_check_test.py
@@ -9,6 +9,7 @@ import filecmp
 import sys
 import typing as T
 from checkbox_support.scripts.image_checker import has_desktop_environment
+from datetime import datetime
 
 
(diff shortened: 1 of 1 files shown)


Hook25 (approved):

Hook25 (approved):
good idea, +1

PR #1629, opened 2024-11-29
Update issue template (infra)

pieqq:
Add two fields:

- Expected Result
- Actual Result

to make sure we better understand the expectations from the bug author.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -35,6 +35,26 @@ body:
     validations:
       required: true
   - type: textarea
+    id: expected_result
+    attributes:
+      label: Expected Result
+      description: >
+        What should happen?
+      placeholder: |
+        The job passes and all is green.
+    validations:
+      required: true
+  - type: textarea
+    id: actual_result
+    attributes:
+      label: Actual Result
+      description: >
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -34,6 +34,26 @@ body:
...
     validations:
       required: true
+  - type: textarea
+    id: expected_result
+    attributes:
+      label: Expected Result
+      description: >
+        What should happen?
+      placeholder: |
+        The job passes and all is green.
```
Hook25, line 44: The job passes because ...

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -34,6 +34,26 @@ body:
...
+    validations:
+      required: true
+  - type: textarea
+    id: actual_result
+    attributes:
+      label: Actual Result
+      description: >
+        What actually happens?
+      placeholder: |
+        The job fails and everything is red!
```
Hook25, line 54: The job fails because...

Hook25 (changes requested):
lets motivate the contributor to write why according to them something goes south

Hook25 (approved):

PR #1633, opened 2024-12-03
Add DSS integration tests for CPU and NVIDIA GPU (New)

motjuste:
## Description

Added integration tests from the [`data-science-stack` repo](https://github.com/canonical/data-science-stack).  These tests need to be run at the very end, and require that DSS is not initialized, because the integration tests themselves initialize and purge DSS and check for relevant success.  Integration tests for NVIDIA GPU also require that the relevant k8s plugin is setup.

The Git repo is cloned, the relevant virtual environment is created, and the tests are executed using `tox` from the virtual environment.

## Resolved issues

- [CHECKBOX-1667](https://warthogs.atlassian.net/browse/CHECKBOX-1667)

## Documentation

There are no changes to the documentation.

## Tests

The relevant DSS workflow run: https://github.com/canonical/checkbox/actions/runs/12143951564

<details>
<summary><em>to save ourselves some time...</em></summary>

The job linked above was actually run on a different branch, which consists of commits from this PR, and PR #1634 together.  This is because the workflow runs can take a very long time, and while their changes don't overlap, the two PRs are worth testing together.

</details>

The failures on testing DSS latest/edge are expected due to a [reported issue](https://github.com/canonical/data-science-stack/issues/191) and do not stem from these changes to Checkbox.  The one failure testing DSS latest/stable is during provisioning from Testflinger, before the Checkbox validations are even installed.

[CHECKBOX-1667]: https://warthogs.atlassian.net/browse/CHECKBOX-1667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
+++ b/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
@@ -369,7 +369,7 @@ requires:
 depends: dss/purge
 _summary: Check that all DSS integration tests for CPU pass
 estimated_duration: 1m
-command: check_dss_integration_tests.sh can_be_purged pass_on_cpu
+command: check_dss_integration_tests.sh pass_on_cpu
 
 id: dss_integration_tests/nvidia_gpu
 category_id: dss-regress
@@ -386,4 +386,4 @@ depends:
   nvidia_gpu_addon/validations_succeed
 _summary: Check that all DSS integration tests for NVIDIA GPUs
 estimated_duration: 1m
-command: check_dss_integration_tests.sh can_be_purged pass_on_nvidia_gpu
+command: check_dss_integration_tests.sh pass_on_nvidia_gpu
--- a/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
+++ b/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
@@ -340,3 +340,50 @@ depends: dss/create_tensorflow_cuda_notebook
 _summary: Check that the Tensorflow CUDA notebook can be removed
 estimated_duration: 1m
 command: check_dss.sh can_remove_notebook tensorflow-cuda
+
+# DSS integration test jobs ###################################################
+
+id: dss/purge
+category_id: dss-regress
+flags: simple
+imports: from com.canonical.certification import executable
+requires: executable.name == 'dss'
+depends: dss/initialize
+_summary: Check that DSS can be purged
+estimated_duration: 1m
+command: check_dss.sh can_be_purged
+
+# NOTE: DSS Integration tests expect DSS to not be initialized
+#  hence they also depend on the dss/purge job
+
+id: dss_integration_tests/cpu
+category_id: dss-regress
+flags: simple
+imports:
+  from com.canonical.certification import executable
+  from com.canonical.certification import package
+requires:
+  executable.name == 'dss'
+  executable.name == 'python3'
+  package.name == 'python3-venv'
+depends: dss/purge
+_summary: Check that all DSS integration tests for CPU pass
+estimated_duration: 1m
+command: check_dss_integration_tests.sh can_be_purged pass_on_cpu
+
+id: dss_integration_tests/nvidia_gpu
+category_id: dss-regress
+flags: simple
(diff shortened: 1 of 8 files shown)


contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
```diff
@@ -340,3 +340,50 @@ depends: dss/create_tensorflow_cuda_notebook
...
+  from com.canonical.certification import executable
+  from com.canonical.certification import package
+requires:
+  executable.name == 'dss'
+  executable.name == 'python3'
+  package.name == 'python3-venv'
+depends: dss/purge
+_summary: Check that all DSS integration tests for CPU pass
+estimated_duration: 1m
+command: check_dss_integration_tests.sh can_be_purged pass_on_cpu
```
pieqq, line 372: is `can_be_purged` expected here? It looks like it's not part of the `check_dss_integration_tests.sh` script.

contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
```diff
@@ -340,3 +340,50 @@ depends: dss/create_tensorflow_cuda_notebook
...
+requires:
+  executable.name == 'dss'
+  executable.name == 'python3'
+  package.name == 'python3-venv'
+depends:
+  dss/purge
+  nvidia_gpu_addon/validations_succeed
+_summary: Check that all DSS integration tests for NVIDIA GPUs
+estimated_duration: 1m
+command: check_dss_integration_tests.sh can_be_purged pass_on_nvidia_gpu
```
pieqq, line 389: is `can_be_purged` expected here? It looks like it's not part of the `check_dss_integration_tests.sh` script.

contrib/checkbox-dss-validation/checkbox-provider-dss/units/test-plan.pxu
```diff
@@ -35,7 +35,11 @@ include:
     dss/create_tensorflow_cuda_notebook
     cuda/tensorflow_can_use_cuda
     dss/remove_tensorflow_cuda_notebook
+    dss/purge
+    dss_integration_tests/cpu
+    dss_integration_tests/nvidia_gpu
```
pieqq, line 40: In their respective job defintions, both `dss_integration_tests/cpu` and `dss_integration_tests/nvidia_gpu` depend on `dss/purge`, meaning they will be run after `dss/purge` has been run, but it doesn't mean they will both start from a "clean" state. As they are ordered here, the CPU test will run first, and whatever state DSS is in, it will be used right after for the Nvidia GPU test. Is that accounted for?

pieqq (changes requested):
I think I spotted a typo, and I have a question inline below.

contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
motjuste: Thank you for your keen eye ... pushing a fix now.

contrib/checkbox-dss-validation/checkbox-provider-dss/units/test-plan.pxu
motjuste: yes ... the both integration tests reset the state back ... each of them first initializes DSS, runs some tests, then purges DSS.

contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
motjuste: This error was not caught because the exit code was still being set correctly when the options were not the ones that were supported.  I have added a fix to all the shell scripts where the exit code is set to 1 when unsupported options are provided.

motjuste:
The tests have been re-triggered: https://github.com/canonical/checkbox/actions/runs/12201552963

motjuste:
After discussion with @pieqq, this needs to be re-implemented as Python scripts instead of shell-scripts, so that they can be unit-tested.  That will be picked up in a separate PR in the future.

PR #1634, opened 2024-12-03
Add reboot and re-test to Testflinger job for DSS validation (New)

motjuste:
## Description

The updates are only to the job definition template used by the [GitHub workflow](https://github.com/canonical/checkbox/blob/main/.github/workflows/testflinger-contrib-dss-regression.yaml) for running the DSS validations on machines from Testflinger.  We reboot the machine and re-run the validations, with some delay in the middle to let things settle down.  

## Resolved issues

- [CHECKBOX-1669](https://warthogs.atlassian.net/browse/CHECKBOX-1669)

## Documentation

There are no changes to the documentation.

## Tests

The relevant DSS workflow run: https://github.com/canonical/checkbox/actions/runs/12143951564

<details>
<summary><em>to save ourselves some time...</em></summary>

The job linked above was actually run on a different branch, which consists of commits from this PR, and PR #1633 together.  This is because the workflow runs can take a very long time, and while their changes don't overlap, the two PRs are worth testing together.

</details>

The failures on testing DSS latest/edge are expected due to a [reported issue](https://github.com/canonical/data-science-stack/issues/191) and do not stem from these changes to Checkbox.  The one failure testing DSS latest/stable is during provisioning from Testflinger, before the Checkbox validations are even installed.

[CHECKBOX-1669]: https://warthogs.atlassian.net/browse/CHECKBOX-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrib/checkbox-dss-validation/testflinger/job-def.yaml
+++ b/contrib/checkbox-dss-validation/testflinger/job-def.yaml
@@ -59,3 +59,26 @@ test_data:
     ssh ubuntu@$DEVICE_IP '
       checkbox-dss.validate-with-gpu
     '
+
+    # Reboot the machine
+    ssh -t ubuntu@$DEVICE_IP '
+      SLEEP_SECS=15
+      >&2 echo "[INFO]: sleeping for ${SLEEP_SECS} seconds before rebooting."
+      sleep ${SLEEP_SECS}
+      >&2 echo "[INFO]: rebooting now.  Best of luck!"
+      sudo reboot
+    ' || >&2 echo "reboot started with exit code $?"
+
+    SLEEP_SECS=30
+    >&2 echo "[INFO]: sleeping for ${SLEEP_SECS} seconds for reboot to finish."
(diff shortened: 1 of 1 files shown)


pieqq (changes requested):
I don't think this is a good approach. You are running the same test plan twice, without having any way of checking at what point each of these tests have been run.

I am not sure how you plan to store the results. Are they only kept in Testflinger, or stored somewhere else (e.g. C3)?

Moreover, do you really want to re-run the whole test plan after rebooting, or would a subset of jobs be enough?

I would encourage you to write a dedicated test plan that includes a reboot job (this is a `noreturn` job, see Checkbox doc), with separate jobs depending on it run afterwards (see `siblings` in Checkbox documentation) to check that DSS is still alive and respond as expected to commands.

PR #1640, opened 2024-12-05
Enabling DSS testing on 1/track (Infra)

deusebio:
## Description

Since a new track has just been created ([reference](https://forum.snapcraft.io/t/create-new-track-for-data-science-stack/44203)) and released in snapstore, we now enable testing against `1/stable` and `1/candidate` instead of `latest/stable` (that should soon be deprecated)


motjuste (approved):
1/candidate is not released yet, but snapd then forwards to 1/stable, which is fine.

Issue #1537, opened 2024-10-09
Intel MIPI Camera can't capture the picture


seankingyang:
### Bug Description

```
-----------[ Webcam still image capture test for Intel_MIPI_Camera ]------------                                                                                                                                                              
ID: com.canonical.certification::camera/still_video0                                                                                                                                                                                          
Category: com.canonical.plainbox::camera                                                                                                                                                                                                      
Purpose:

This test will check that the Intel_MIPI_Camera works

Steps:

1. Click on Test to display a still image from the camera for ten seconds.

Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]: 
... 8< -------------------------------------------------------------------------
Error: Internal data stream error.
Format: NV12 (Y/UV 4:2:0)
Resolutions: 1280x720

Failed to capture image with fswebcam, using gstreamer
Error: Internal data stream error.

```

The mipi camera on this DUT can't capture the 640x480 (the script's default https://github.com/canonical/checkbox/blob/main/providers/base/bin/camera_test.py#L718-L726) which is not support on Intel MIPI camer.
And I try the gst cmd:
```
ubuntu@ubuntu:~$ gst-launch-1.0 v4l2src device=/dev/video0 num-buffers=10 ! "video/x-raw,width=640,height=480" ! jpegenc ! filesink location="640_480.jpeg"
Setting pipeline to PAUSED ...
Pipeline is live and does not need PREROLL ...
Pipeline is PREROLLED ...
Setting pipeline to PLAYING ...
New clock: GstSystemClock
ERROR: from element /GstPipeline:pipeline0/GstV4l2Src:v4l2src0: Internal data stream error.
Additional debug info:
../libs/gst/base/gstbasesrc.c(3127): gst_base_src_loop (): /GstPipeline:pipeline0/GstV4l2Src:v4l2src0:
streaming stopped, reason not-negotiated (-4)
Execution ended after 0:00:00.000824281
Setting pipeline to NULL ...
Freeing pipeline ...

ubuntu@ubuntu:~$ ls -al | grep jpeg
-rw-rw-r--  1 ubuntu ubuntu        0 Oct  9 13:53 640_480.jpeg
```
It will generate the empty size file.

Intel MIPI camera will use the `v4l2-relayd,.service` to transfer the fixed format and width/hight (1280x720).
```
$ systemctl status v4l2-relayd.service 
● v4l2-relayd.service - v4l2-relay daemon service
[...]

PR #1643, opened 2024-12-08
Fix black photo issue on intel MIPI cameras (Bugfix)

tomli380576:
## Description

The gstreamer pipeline in the current `camera_test.py` was having trouble with cameras that require a "boot up time" because it was taking the 1st buffer, which is blank for these cameras, that reaches filesink. This PR adds a valve element that dynamically opens while the pipeline is running to allow the source to stay open for a few seconds before the photo is taken. 

There was also an issue where `<CameraTest>._width` and `<CameraTest>._height` were never updated from 640x480 after we query the supported formats. This breaks on intel MIPI cameras because it only supports 1280x720 and it will complain about not-negotiated caps $\Rightarrow$ `Error: Internal data stream error` when `<Pipeline>.set_state(Gst.State.PLAYING)` is called. This is fixed as well but I'm not too sure if this was intentional. Do later tests depend on the image being 640x480?

## Resolved issues

#1537 

## Documentation

The number of seconds to delay can be set by `--wait-seconds <num_seconds>`. Default is 3.

## Tests

- Original unit tests.
- Validated on a few machines running 24.04 with intel MIPI cameras and the test runs normally. Machines with USB cameras are nice as usual and pass the test too.

### DIFF
--- a/providers/base/bin/camera_test.py
+++ b/providers/base/bin/camera_test.py
@@ -178,6 +178,8 @@ class CameraTest:
             by the driver, and see if they are valid.
     """
 
+    PHOTO_DELAY_SECONDS = 2
+
     def __init__(self, **kwargs):
         self._width = 640
         self._height = 480
@@ -483,7 +485,6 @@ class CameraTest:
         # Add source
         source = self.Gst.ElementFactory.make("v4l2src", "video-source")
         source.set_property("device", self.device)
-        source.set_property("num-buffers", 10)
         pipeline.add(source)
 
         # Add the caps element and include a format converter if needed to
@@ -516,8 +517,17 @@ class CameraTest:
             pipeline.add(caps)
             rgb_capture = caps
 
+        # Add valve, doc says drop=True is default
+        valve = self.Gst.ElementFactory.make("valve", "photo-valve")
+        assert valve, "Valve element could not be created"
+        valve.set_property("drop", True)
+        pipeline.add(valve)
+
         # Add encoder
         encoder = self.Gst.ElementFactory.make("jpegenc", "encoder")
+        # snapshot=True sends a EOS downstream 
+        # when the first buffer reaches jpegenc
+        encoder.set_property("snapshot", True)
         pipeline.add(encoder)
 
         # Add sink
@@ -527,8 +537,10 @@ class CameraTest:
 
         # Link elements
         source.link(caps)
-        rgb_capture.link(encoder)
+        rgb_capture.link(valve)
+        valve.link(encoder)
         encoder.link(sink)
+        # source ! rgbcapture ! valve ! encoder ! filesink
 
         # Connect the bus to the message handler
         bus = pipeline.get_bus()
@@ -539,8 +551,12 @@ class CameraTest:
         self.pipeline = pipeline
         self.pipeline.set_state(self.Gst.State.PLAYING)
 
-        # Add a timeout of 90 seconds to capture the image
+        # Add a global timeout of 90 seconds to capture the image
         self.timeout = self.GLib.timeout_add_seconds(90, self._on_timeout)
+        self.GLib.timeout_add_seconds(
+            self.PHOTO_DELAY_SECONDS,
+            lambda: valve.set_property("drop", False)
+        )
 
         # Start the main loop. If the loop finishes successfully, we will
         # remove the timeout. If the timeout is reached, we will stop the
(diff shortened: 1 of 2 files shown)


tomli380576:
Validated that `valve` exists on the default gstreamer installation on 16 and 18 so we should be good

tomli380576:
Just a quick note on multifilesink: I used it because with the way we are terminating the pipeline (time/num-buffers based EOS), filesink will write extra buffers after the initial valid buffer. Basically it made a file like this:
```
┌────┬────┬────┬────┐
│ F1 │ F2 │ F3 │ F4 │
└────┴────┴────┴────┘ 
```
where F1 is the actual photo we want, F2-F4 are the extra stuff that filesink kept writing until EOS was sent. This file is still open-able probably because the metadata only tells the image viewer to read up to F1 and just ignore the rest. This can make the image file excessively big if the image dimensions are big or the framerate is high. I found that 1280x720 photos taken at 30fps was taking up to 7Mb of space when it only needs like ~200Kb.

providers/base/bin/camera_test.py
```diff
@@ -505,30 +524,47 @@ def _capture_image_gstreamer(self, filename, width, height, pixelformat):
...
             pipeline.add(caps)
             rgb_capture = caps
 
-        # Add encoder
-        encoder = self.Gst.ElementFactory.make("jpegenc", "encoder")
-        pipeline.add(encoder)
+        # Add valve, note that we can't rely on the default value of drop
+        # it varies by installations
+        valve = self.Gst.ElementFactory.make("valve", "photo-valve")
+        assert valve, "Valve element could not be created"
```
fernando79513, line 543: Assertions should be used mainly for debugging purposes and could be turned off by some compilation options.
It's better practice to raise an exception:
```suggestion
        if not valve:
            raise RuntimeError("Valve element could not be created")
```

providers/base/bin/camera_test.py
```diff
@@ -505,30 +524,47 @@ def _capture_image_gstreamer(self, filename, width, height, pixelformat):
...
+            valve.link(sink)
+
+        else:
+            # Add encoder
+            encoder = self.Gst.ElementFactory.make("jpegenc", "encoder")
+            pipeline.add(encoder)
+            valve.link(encoder)
+            encoder.link(sink)
+
+        # source ! rgbcapture ! valve ! encoder ! filesink
```
fernando79513, line 567: I like this pipeline explanation, just add the right one for each one of the cases:
```suggestion
        if pixelformat == "MJPG":
            # source ! rgbcapture ! valve ! filesink
            valve.link(sink)

        else:
            # source ! rgbcapture ! valve ! encoder ! filesink
            # Add encoder
            encoder = self.Gst.ElementFactory.make("jpegenc", "encoder")
            pipeline.add(encoder)
            valve.link(encoder)
            encoder.link(sink)
```

providers/base/bin/camera_test.py
```diff
@@ -539,8 +575,32 @@ def _capture_image_gstreamer(self, filename, width, height, pixelformat):
...
+            90, self._on_timeout
+        )
+
+        def eos_timeout():
+            # this relies on the eos handler in _on_gst_message
+            self.timeout["eos_timeout"] = None
+            self.pipeline.send_event(self.Gst.Event.new_eos())
+
+        self.timeout["eos_timeout"] = self.GLib.timeout_add_seconds(
+            (self.photo_wait_seconds or 0) + 1, eos_timeout
```
fernando79513, line 589: Do we need to check here if `self.photo_wait_seconds` is a falsy value?
We already have an assertion in the __init__ method that compares that value to an integer (0)

fernando79513 (commented):
Good job here!
I think this implementation improves the current approach of ignoring the first frames and should be more robust in general.
I've left just some small comments here and there.

providers/base/bin/camera_test.py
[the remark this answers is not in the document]
tomli380576: Didn't know it's possible to turn off assertions :fearful: Thanks for the catch!

providers/base/bin/camera_test.py
tomli380576: Thanks for the catch, I think I had it in there before the assertion was included. So  yes it's not needed anymore.

fernando79513 (approved):
Good job!
LGTM +1

PR #1646, opened 2024-12-10
proviers/base: add requires for bluetooth4/beacon_eddystone_url_* (New)

hector-cao:
## Description

In order to pass, the test bluetooth4/beacon_eddystone_url_* needs:
- bluez package/snap (+ adequate connections for ubuntu core)

This PR add `requires` section to the test bluetooth4/beacon_eddystone_url_*

## Resolved issues

During Intel IOTG certification test run, we bumped into this test failure for server classic image,
It took us some time to realize that the bluez package is missing, this time spent can be saved with this
`requires` section

## Documentation

N/A

## Tests

The test runs successfully on 22.04 server image

### DIFF
--- a/providers/base/units/bluetooth/jobs.pxu
+++ b/providers/base/units/bluetooth/jobs.pxu
@@ -287,6 +287,11 @@ user: root
 flags: also-after-suspend
 category_id: com.canonical.plainbox::bluetooth
 estimated_duration: 10
+requires:
+  package.name == 'bluez' or snap.name == 'bluez'
+  {%- if __on_ubuntucore__ %}
+  connections.slot == 'bluez:service' and connections.plug == '{{ __system_env__["SNAP_NAME"] }}:bluez'
+  {% endif -%}
 
 unit: template
 template-resource: bluez-internal-rfcomm-tests
--- a/providers/base/units/bluetooth/jobs.pxu
+++ b/providers/base/units/bluetooth/jobs.pxu
@@ -285,7 +285,7 @@ command:
  checkbox-support-eddystone_scanner -D {{ interface }}
 plugin: shell
 user: root
-flags: also-after-suspend
+flags: also-after-suspend fail-on-resource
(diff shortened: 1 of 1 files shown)


providers/base/units/bluetooth/jobs.pxu
```diff
@@ -276,17 +276,23 @@ requires:
...
 command:
- checkbox-support-eddystone_scanner -D {interface}
+ checkbox-support-eddystone_scanner -D {{ interface }}
 plugin: shell
 user: root
 flags: also-after-suspend
 category_id: com.canonical.plainbox::bluetooth
 estimated_duration: 10
+requires:
+ package.name == 'bluez' or snap.name == 'bluez'
```
pieqq, line 292: If you use this statement, and the bluez snap/package is missing, the test will be skipped, which might be even harder to spot if this test was supposed to be executed to begin with.

Is this what you want to do?

pieqq (commented):
Hey! Thanks for this PR. I have a comment below.

providers/base/units/bluetooth/jobs.pxu
hector-cao: Thanks @pieqq,
- I see this statement is used for other tests in this jobs file, that is why I had the idea to do the same for this test
- You are right about the fact that we have to check all the skipped tests to make sure that are skipped for good reason, but without this statement, I also had a hard time to figure out that this test failed because `bluez` was missing, i have the feeling that it is a global discussion we might have because expressing dependency of a test on a binary availability is very common in our existing tests, so your concern is somehow global and not applicable to only this PR,

zongminl: Maybe also having a fail-on-resource flag in this job will get what you want

zongminl (approved):
+1, thank you.

pieqq (approved):
+1, thanks!

PR #1649, opened 2024-12-11
Workaround WebSocketClient race condition metabox (infra)

Hook25:
## Description

There is a race condition in the `WebSocketClient` library. This makes metabox crash (without finishing the test run) when the controller or local crashes before sending something to it. This fixes it

## Resolved issues

Partially fixes: CHECKBOX-1699

## Documentation

Big comment explaining why this is done

## Tests

N/A


pieqq (approved):
LGTM

PR #1654, opened 2024-12-12
Multishell support and add assertion on min. timeout (bugfix)

Hook25:
## Description

In the launchpad builder for Noble, the shell selected is `sh`. This is how I discovered that `sh`'s kill doesn't actually support the tree kill (`-PID` semantic). This adds a fallback mechanism (that tries to use known working shells instead of the default one), then tries the default one and if it fails tries one last time to at least kill the direct child.

This also fixes a race condition, when join timeout is 0, in very unlucky circumstances the subprocess may not be up after the join, leading to the kill failing and the subprocess leaking free. Also, a timeout of 0 doesn't make sense.

## Resolved issues

Fixes: CHECKBOX-1701

## Documentation

N/A

## Tests

Update tests

### DIFF
--- a/checkbox-support/checkbox_support/helpers/timeout.py
+++ b/checkbox-support/checkbox_support/helpers/timeout.py
@@ -130,7 +130,7 @@ def run_with_timeout(f, timeout_s, *args, **kwargs):
     process.join(timeout_s)
 
     if process.is_alive():
-        # this kills the whole process tree, not just the child
+        # this tries to kill the whole process tree, not just the child.
         kill_tree(process.pid)
         raise TimeoutError("Task unable to finish in {}s".format(timeout_s))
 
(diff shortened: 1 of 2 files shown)


checkbox-support/checkbox_support/helpers/timeout.py
```diff
@@ -110,7 +131,7 @@ def _f(*args, **kwargs):
 
     if process.is_alive():
         # this kills the whole process tree, not just the child
```
fernando79513, line 133: Unless shell is sh
```suggestion
        # this tries to kill the whole process tree, not just the child.
```

fernando79513 (dismissed):
Quite interesting discovery.
LGTM+1!

fernando79513 (approved):

PR #1658, opened 2024-12-17
Support float castable values instead of just floats (bugfix)

Hook25:
## Description

The new behaviour may introduce units that produce iologs in 0 time. This causes the validation after resume to crash and burn because 0 is not a float, but an integer. This fixes the issue by allowing any value that is float castable (which is slightly more than just integers, but I see no point in being strict about the timestamp of the iolog production!)

## Resolved issues

Fix: CHECKBOX-1709

## Documentation

N;/A

## Tests

N/A


pieqq (approved):
+1.

I wish this `_validate` function was a little bit more flexible to handle this internally, but on the other hand I don't want to start overengineering things here...

Issue #1663, opened 2024-12-19
test-store-config fails because of new template checks.


fernando79513:
### Bug Description

With the new template checks, we discovered an error with the "snappy/test-store-config-store" test. The test was trying to be created even if no store was declared.

```
13:22:10 ------[ Test that image is using the correct snappy store configuration. ]------
13:22:10 ID: com.canonical.certification::snappy/test-store-config-MISSING_PARAM_1
13:22:10 Category: Snapd
13:22:10 --------------------------------------------------------------------------------
13:22:10 Template 'com.canonical.certification::snappy/test-store-config-store' generated an invalid unit. Field 'command' has value 'echo "Expected store ID:"
13:22:10 $STORE_ID"
13:22:10 
13:22:10 Store ID in model assertion:"
13:22:10 {store}"
13:22:10 ORE_ID" == "{store}" ]' but the template-resource didn't generate any value for 'store'
13:22:10 --------------------------------------------------------------------------------
13:22:10 Outcome: job failed
```

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

Run:
`checkbox-cli run com.canonical.certification::snappy-snap-automated`

### Expected Result

The "snappy/test-store-config-.*" test is not run in devices that don't declare the store in the `model_resource`

### Actual Result

The job fails with a template error

### Environment

- Snap: core20
- Snap version: 20241206
- Snap revision: 2497
- Snap channel: beta
- Kernel version:	5.4.0-1120-raspi
- Checkbox-version: 4.3.0-dev36

PR #1664, opened 2024-12-19
Added a check in the job "test-store-config" for missing store field (BugFix)

fernando79513:
## Description

With the new template checks, we discovered an error with the "snappy/test-store-config-store" test. The test was trying to be created even if no store was declared.
```
CategorySnapdIO LogTemplate 'com.canonical.certification::snappy/test-store-config-store' generated an invalid unit. Field 'command' has value 'echo "Expected store ID:"
echo "$STORE_ID"
echo
echo "Store ID in model assertion:"
echo "{store}"
[ "$STORE_ID" == "{store}" ]' but the template-resource didn't generate any value for 'store'
```

Now we set the store field in the "model_resource" to "unknown" and filter that in the templated test
 
## Resolved issues

#1663 

N/A

## Tests

After running a reduced version of the snappy automated tests, the test-store-config test is not templated anymore

### DIFF
--- a/checkbox-support/checkbox_support/snap_utils/asserts.py
+++ b/checkbox-support/checkbox_support/snap_utils/asserts.py
@@ -65,6 +65,10 @@ def model_to_resource(model_assertion):
                     resource["{}_track".format(key)] = track
                 else:
                     resource[key] = val
+
+    # Check if the model has a store
+    if "store" not in resource:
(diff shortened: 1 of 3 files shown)


Hook25 (approved):

Hook25 (approved):
tyvm for fixing this! +1

PR #1668, opened 2024-12-20
Security scanning zizmor (infra)

Hook25:
## Description

This creates a new workflow to do sistematic vulnerability scanning of our PRs. This is not a silver bullet, as it will be ran only when new workflows are introduced or old ones are modified but its a good start.

This also adjusts all "problems" found by the tool so that it passes all the times now namely:
- checkout action -> do not persist token, we never pass the token implicitly around and this is an anti feature, disabled
- codecov action -> updated and mitigated a problem (codecov has a branch called v3, so we shouldn't use that reference because we trust an un-restricted branch that way)
- 

## Resolved issues

Fixes: CHECKBOX-1706

## Documentation

N/A

## Tests

Tested tox action to check that the new codecov action works: https://github.com/canonical/checkbox/actions/runs/12431986975

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - uses: psf/black@stable
         with:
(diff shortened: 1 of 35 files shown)


pieqq (approved):
LGTM. thanks!

Issue #1631, opened 2024-12-03
started unity_support_test check  test without waiting for the system to be stable during WB/CB cycles.


seankingyang:
### Bug Description

In debian checkbox, the checkbox-ng service will be start when network.targer is triggerred.
It will run the check before the gnome session is ready.
We should run the check after the user space service/ target is run and finish with status.
We can add the `systemd-analysis` until this cmd have the ourput, then we can run the check.

checkbox will resume the session at the Dec 03 00:11:04.
```
Dec 03 00:11:04 ubuntu systemd[1]: Started Checkbox Remote Service.
Dec 03 00:11:04 ubuntu sudo[1119]:     root : PWD=/ ; USER=root ; COMMAND=/usr/bin/true
Dec 03 00:11:04 ubuntu sudo[1119]: pam_unix(sudo:session): session opened for user root(uid=0) by (uid=0)
Dec 03 00:11:04 ubuntu sudo[1119]: pam_unix(sudo:session): session closed for user root
Dec 03 00:11:04 ubuntu sudo[1131]:     root : PWD=/ ; USER=root ; COMMAND=/usr/bin/true
Dec 03 00:11:04 ubuntu sudo[1131]: pam_unix(sudo:session): session opened for user root(uid=0) by (uid=0)
Dec 03 00:11:04 ubuntu sudo[1131]: pam_unix(sudo:session): session closed for user root
Dec 03 00:11:04 ubuntu checkbox-ng.service[913]: Config: daemon section name is deprecated. Use agent instead.
Dec 03 00:11:04 ubuntu checkbox-ng.service[913]: Resuming session: 'session_title-2024-12-02T07.07.50'
Dec 03 00:11:04 ubuntu checkbox-ng.service[913]: Skipped file: /usr/share/checkbox-provider-base/units/camera/README.rst
Dec 03 00:11:05 ubuntu checkbox-ng.service[913]: Skipped file: /usr/share/locale/aa/LC_MESSAGES/xdg-user-dirs.mo
Dec 03 00:11:05 ubuntu checkbox-ng.service[913]: Skipped file: /usr/share/locale/ab/LC_MESSAGES/iso_3166-1.mo

```

DISPLAY varialbe will be set at Dec 03 00:11:13
```
Dec 03 00:11:13 ubuntu /usr/libexec/gdm-x-session[2088]: dbus-update-activation-environment: setting DISPLAY=:0
[...]

PR #1671, opened 2024-12-23
Fix race condition in reboot_check_test's renderer test and add lsusb/lspci timeout (BugFix)

tomli380576:
## Description

- Added a method to wait for graphical.target before running unity_support_test.
- Added timeouts for lspci and lsusb

## Resolved issues

Race condition issue: https://github.com/canonical/checkbox/issues/1631 TLDR: Calling unity_support_test before graphical.target is reached will falsely report the test as failed since it can't open the display.

There was also an issue reported by Hanhsuan where lspci and lsusb can get stuck and never return, so I added a timeout for their `subprocess.run` calls.
- By "stuck" I mean lspci & lsusb are waiting indefinitely for devices to respond, which apparently can happen with broken drivers. This type of freeze can be handled by a basic timeout

## Documentation

The wait timeout can be controlled by `--graphical-target-timeout <timeout_seconds>`. This value is ignored when `-g` or `--graphics` are not specified. Default is 120 seconds

## Tests

Unit tests

### DIFF
--- a/providers/base/bin/reboot_check_test.py
+++ b/providers/base/bin/reboot_check_test.py
@@ -11,7 +11,6 @@ import typing as T
 from checkbox_support.scripts.image_checker import has_desktop_environment
 from datetime import datetime
 import time
-from checkbox_support.helpers.timeout import timeout
 
 
 # Checkbox could run in a snap container, so we need to prepend this root path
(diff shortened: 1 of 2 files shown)


providers/base/bin/reboot_check_test.py
```diff
@@ -302,6 +311,36 @@ def is_hardware_renderer_available(self) -> bool:
         print("[ ERR ] Software rendering detected", file=sys.stderr)
         return False
 
+    def wait_for_graphical_target(self, max_wait_seconds: int) -> bool:
```
Hook25, line 314: Please use the timeout decorator, also, why are you polling?

Hook25 (changes requested):

PR #1672, opened 2024-12-24
[checkbox-ce-oem] Add interface for docker provider for uc and checkbox24 for series 24 building (Bugfix)

LiaoU3:
## Description
Since maybe in the future, we will need to run the docker test via checkbox-ce-oem on core image, we have to define plugs to connect.
Besides, I completed the `TODO` for series 24 that I added back checkbox24 for build snaps and default providers.
According to #1285, we should have rename from plainbox to checkbox for checkbox snap, hence, there should be no exception for checkbox-ce-oem.

## Resolved issues
N/A

## Documentation
N/A

## Tests
I have checked that it is buildable and there is docker tests.


zongminl (approved):
+1, thank you

PR #1675, opened 2024-12-26
Modify description of manifest that used by pdu cold reboot test (Bugfix)

rickwu666666:
## Description
Modify description of manifest that used by pdu cold reboot test to more sepcific to the platform that not supported RTC wake and intend to power cycle via PDU.


stanley31huang (approved):
LGTM

PR #1676, opened 2025-01-02
Automatic GStreamer pipelines for testing cameras (New)

tomli380576:
## Description

This PR implements multiple GStreamer pipelines that tests if a camera is working properly. Given a source element and a Gst.Caps object, we can 

1. Take photos 
2. Record videos
3. Show a viewfinder

The test script in provider/base/bin also iterates through all the elements and their caps using the output of Gst.DeviceMonitor. 

## Resolved issues

This is part of the new camera tests that we plan to implement https://warthogs.atlassian.net/browse/OEMQA-4698 
- Originally this was for testing MIPI cameras, but this more generic version handles USB cameras as well.

## Documentation

Interpretation of the test result:
- PASS means the following
    1. Cameras are visible to GStreamer (so applications like cheese, snapshot, and firefox with pipewire support enabled can see them)
    2. Each camera can execute the subcommand (photo, video, viewfinder) correctly using each of their advertised capability and produce a readable photo/video file if the subcommand is not viewfinder.
    3. The camera pipeline correctly terminates and there's no deadlock/hang inside the source element. (unfortunately this can't be handled by checkbox's timeout yet) 

## Tests

Unit tests

Real hardware:
- 22.04 and 24.04 Laptops with USB cameras
- 18.04 and 20.04 Virtual machines with v4l2 pass through


tomli380576:
Reverting back to draft because GLib's main loop doesn't seem to like checkbox's timeout, it causes the loop to hang for some reason

Hook25:
Hello, 
I don't see why this can't be split into two separate PR, there are 2 files + 2 test files, that would make 2 PRs of ~1k lines. They are still huge but at least they are not 2k lines. Please do that and propose the PR again. 
Thanks

Issue #1630, opened 2024-12-02
Invalid check of unity_support_test for Genio Devices (ARM Mali GPU)


baconYao:
### Problem

Recently, the cold/warm boot stress tests failed on Genio and ADVANTECH - RSB-3810 devices in SRU and enabledment phase.
  - [ADVANTECH - RSB-3810](http://10.102.156.15:8080/job/cert-odm-sru-jammy-desktop-advantech-rsb-3810/8/testReport/com.canonical/certification__cold-boot-loop-test1/com_canonical_certification__cold_boot_loop_test1/) 
  - [Genio 1200](https://certification.canonical.com/hardware/202307-31859/submission/402762/test-results/fail/?term=loop-test)
  - [Genio 700](https://certification.canonical.com/hardware/202409-35342/submission/402766/test-results/fail/?term=loop-test)
  - [Genio 510](https://certification.canonical.com/hardware/202312-33205/submission/402764/test-results/fail/?term=loop-test)

The failed output like below and it's caused by checking `unity support test`

```bash
Test: Scan kernel log for errors and warnings.                              
                                                                     :   0.0% /
  Kernel log error check.                                            :   0.0% -
  Kernel log error check.                                            :  50.0% \
                                                                               
  Kernel log error check.                                      
Test: Scan kernel log for Oopses.                                           
                                                                     :  50.0% |
  Kernel log oops check.                                             :  50.0% /
  Kernel log oops check.                                             : 100.0% -
                                                                               
  Kernel log oops check.                                  2 passed
Running 2 tests, results appended to /var/tmp/checkbox-ng/sessions/session_title-2024-11-20T08.50.24.session/session-share/cold_reboot_cycle1/fwts_klog_oops.log
No errors detected
[ ERR ] unity support test returned 1
Comparing devices in (expected) /var/tmp/checkbox-ng/sessions/session_title-2024-11-20T08.50.24.session/session-share/before_reboot against (actual) /var/tmp/checkbox-ng/sessions/session_title-2024-11-20T08.50.24.session/session-share/cold_reboot_cycle1...
[ OK ] Devices match!
[ OK ] fwts checks passed!
[ OK ] Didn't find any failed system services!
These nodes ['renderD128', 'card0-eDP-1', 'card0-DSI-1', 'card0'] exist
card0-eDP-1 is connected to display!
Checking $DISPLAY=:0
```

### Reason

There were two commits to enhance and fix the validation for cold/warm stress test

[...]

PR #1677, opened 2025-01-03
Use glmark2 for hardware rendering tests instead of unity_support_test (BugFix)

tomli380576:
## Description

This PR uses the `glmark2`, `glmark2-wayland` and their es2 variants to determine whether the DUT is using hardware rendering instead of `unity_support_test`. It checks the output of glmark2 for a GL_RENDERER string and uses the same logic as unity_support_test to see if it's produced by a software renderer. 

Requires #1671 
Requires https://github.com/canonical/checkbox/pull/1880 for the glmark2-es2-wayland command
## Resolved issues

https://github.com/canonical/checkbox/issues/1630

Debian version of glmark2 returns the correct string on Mali platforms whereas unity_support_test doesn't.

## Documentation

We are basically doing the same check as unity_support_test. The only difference is how we get the GL_RENDERER string.

## Tests

Original unit tests
Real hardware: x86 laptops and the Mali device in #1630 

https://certification.canonical.com/hardware/202312-33205/submission/432097/ (original mali device, checkbox snap with sideloading)
https://certification.canonical.com/hardware/202506-36824/submission/434449/ (checkbox deb with sideloading)
https://certification.canonical.com/hardware/202503-36447/submission/434882/ (one more snap test with sideloading)
https://certification.canonical.com/hardware/202506-36823/submission/434913/ (locally built snap, no sideloading, x86)


pieqq:
@tomli380576 thanks for the hard work on this. Could you 

- rebase your changes on the latest commit from `main`
- test them on the device where this problem was originally found and upload the submission to C3
- test them on another device in the lab and upload the submission to C3
- update this PR description with a link to these C3 subs

 (since this script is used in the warm/cold reboot tests, it would be good to have a full run to make sure nothing explodes during the stress test)

Thanks!

providers/base/bin/reboot_check_test.py
```diff
@@ -275,42 +274,133 @@ def has_display_connection(self) -> bool:
...
+            if RUNTIME_ROOT:
+                # the official way to specify the location of the data files
+                # is "--data-path path/to/data/files"
+                # but 16, 18, 20 doesn't have this option
+                # and the /usr/share/glmark2 is hard-coded inside glmark2
+                os.symlink(
+                    "{}/usr/share/glmark2".format(RUNTIME_ROOT),
+                    "/usr/share/glmark2",
+                    target_is_directory=True,
+                )
```
tomli380576, line 336: This is a bit of a hack that I'm a bit worried about since it symlinks a file inside the snap. Is this considered bad practice inside checkbox?

tomli380576: and the /usr/share/glmark2 path is a build time environment variable so we also can't change this by just specifying a different env

tomli380576:
Sorry about the last minute changes. I found a machine that was able to consistently lag a little bit between network.target and graphical.target, during which `wait_for_graphical_target` is invoked to wait for gnome to be ready. If glmark2 is called during that time period, then it will fail with an error about missing XDG_RUNTIME_DIR. 

The final few commits fixed this issue by dumping all the envs from gnome into the glmark2 process to guarantee valid XDG* env variables. (tbh I'm surprised we never had this problem with unity_support_test, maybe there's some magic going on with `XOpenDisplay`)

PR #1682, opened 2025-01-15
[checkbox-ce-oem] Enhance serial test to be more stable (Bugfix)

hanhsuan:
## Description

1. The server mode shouldn't be exited while getting the payload that couldn't be decoded.
2. The client mode should clean the buffer before starting the test to avoid receiving non testing payload.

## Tests

https://certification.canonical.com/hardware/202405-34046/submission/408239/


rickwu666666 (approved):
LGTM, Thanks for the patch.

LiaoU3 (approved):
Thank you for the patch!

hanhsuan:
@rickwu666666 mention the client mode couldn't work on his side, let's check it before merging.

rickwu666666 (approved):
I've tried the change on two different projects that have RS485. And works fine.

PR #1685, opened 2025-01-21
add manifest required condition for optee test cases (Bugfix)

stanley31huang:
add manifest condition for optee test cases

## Description
add manifest required condition for optee test cases (Bugfix), so we could have more clear description in the test output logs.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A


rickwu666666 (approved):
LGTM

PR #1686, opened 2025-01-21
Automate wake-on-LAN tests (New)

eugene-yujinwu:
Automate wake-on-LAN tests (New)

## Description
We would like to automate the wake-on-LAN test instead of running manual tests for it.
Please refer to wake-on-LAN-automatic-tests.md in the commit for more details

## Resolved issues
Make the Wake-on-LAN test from manual to automatic.

## Documentation
wake-on-LAN-automatic-tests.md

## Tests
"[wol_S3_auto] Both success":  https://certification.canonical.com/submissions/status/300209

"[wol_S3_auto] Cannot connect to the WOL server" : https://certification.canonical.com/submissions/status/300211

"[wol_S3_auto] One of the WOL disabled" : https://certification.canonical.com/submissions/status/300210

 "[wol_auto_test] One of NIC has wake-on-LAN issue" : https://certification.canonical.com/submissions/status/300212
 
 "[wake-on-Lan] test for one NIC": https://certification.canonical.com/hardware/202406-34088/submission/408713/


providers/base/units/ethernet/jobs.pxu
```diff
@@ -436,3 +436,26 @@ command: network_reconnect_resume_test.py -t 10 -d wired
...
+template-resource: device
+template-engine: jinja2
+template-filter: device.category == 'NETWORK' and device.mac != 'UNKNOWN'
+_summary: Wake on LAN (WOL) automatic test from S3 - {{ interface }}
+plugin: shell
+environ: SERVER_WAKE_ON_LAN WAKE_ON_LAN_DELAY WAKE_ON_LAN_RETRY PLAINBOX_SESSION_SHARE
+imports: from com.canonical.plainbox import manifest
+requires:
+ manifest.has_ethernet_adapter == 'True'
+ manifest.has_ethernet_wake_on_lan_support== 'True'
```
stanley31huang, line 453: I would suggest we have a manifest about the WoL server has been setup or not.

stanley31huang: one more issue is that we uses the RTC to wakeup the system in case the WoL is not working, shall we also add a RTC wakeup manifest as well?

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,155 @@
...
+    powertype = args.powertype
+    timestamp_file = args.timestamp_file
+    delay = args.delay
+    max_retries = args.retry
+
+    logging.info("Interface: {}".format(interface))
+    logging.info("PowerType: {}".format(powertype))
+
+    test_start_time = float(get_timestamp(timestamp_file))
+    actual_start_time = datetime.datetime.fromtimestamp(test_start_time)
```
stanley31huang, line 125: I would suggest we move this part (convert the float variable to timestamp) to the `get_timestamp` function.
as it is the function to read the timestamp from specific file.

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,155 @@
...
+    logging.debug(
+        "Test started at: {}({})".format(test_start_time, actual_start_time)
+    )
+
+    system_back_time = float(get_suspend_boot_time(powertype))
+    actual_back_time = datetime.datetime.fromtimestamp(system_back_time)
```
stanley31huang, line 131: I have some concerns for this part.
first, I would suggest we seperate the `get_suspend_boot_time` into two functions.
e.g.
```
    command = ["journalctl", "-b", "0", "--output=short-unix"]
    result = subprocess.check_output(
        command, shell=False, universal_newlines=True
    )
    logs = result.splitlines()
    get_up_timestamp = get_wakeup_timestamp if powertype == "s3" else get_first_boot_timestamp
    actual_back_time = get_up_timestamp(logs)
```
second, move the part that convert float variable into timestamp into get_wakeup_timestamp and get_first_boot_timestamp.
third, in current implementation, there's a risk that the scripts would crashed (`get_suspend_boot_time` would return None) when no wakupe message been found in get_suspend_boot_time.

providers/base/bin/wol_client.py
```diff
@@ -0,0 +1,282 @@
...
+    if resp.status_code != 200 or result_dict["result"] != "success":
+        raise SystemExit(
+            "get the wrong response: {}".format(result_dict["result"])
+        )
+
+    # bring up the system. The time should be delay*retry*2
+    bring_up_system("rtc", delay * retry * 2)
+    logging.debug(
+        "set the rtcwake time: {} seconds ".format(delay * retry * 2)
+    )
```
stanley31huang, line 272: the scripts would raise SystemExit if configure the RTC wake failed, but it's a bit weird to me that we don't test WoL because of RTC wake is not working.

providers/base/tests/test_wol_server.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+import unittest
+from unittest.mock import patch, MagicMock
+from wol_server import (
+    tasker_main,
+    send_wol_command,
+    is_pingable,
+    run_task,
+)
```
tomli380576, line 27: Since fastapi is imported at the top level inside wol_server, the patch needs to happen before the import. See `test_gnome_monitor.py` for an example.

providers/base/bin/wol_check.py
eugene-yujinwu: Rewrite it as the suggestion. Thanks!

providers/base/units/ethernet/jobs.pxu
eugene-yujinwu: A has_wake_on_lan_server manifest was added.
For the RTC wake up manifest check, after discussion, will keep this not change since for PC project, it should support RTC wake up. For other projects, it maybe need some more work on power management.

providers/base/tests/test_wol_server.py
eugene-yujinwu: Remove the wol_server.py from this PR, for the reason: 1) the wol_server.py doesn't called by anyplace in the checkbox. 2) The wol_server.py will imported fastapi, which doesn't support by python 3.5, and make the unit tests hard to be write. 
@tomli380576 Thanks for your review and insightful suggestion.

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,162 @@
...
+import re
+import argparse
+import logging
+import sys
+
+
+def get_timestamp(file):
+    with open(file, "r") as f:
+        saved_timestamp = float(f.read())
+        # logging.info("saved_timestamp: {}".format(saved_timestamp))
```
stanley31huang, line 31: do you still need to log time here? If not, please remove this line.

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,162 @@
...
+
+    for log in reversed(logs):
+        if r"suspend exit" in log:
+            logging.debug(log)
+            latest_system_back_time = extract_timestamp(log)
+            readable_back_time = datetime.datetime.fromtimestamp(
+                latest_system_back_time
+            )
+            logging.debug("System back time: {}".format(readable_back_time))
+            return float(latest_system_back_time)
```
stanley31huang, line 59: the `latest_system_back_time` is float already, so you could just return it.

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,162 @@
...
+    logs = result.splitlines()
+
+    # The first line of system boot up
+    log = logs[0]
+    latest_system_back_time = extract_timestamp(log)
+    readable_back_time = datetime.datetime.fromtimestamp(
+        latest_system_back_time
+    )
+    logging.debug("System back time: {}".format(readable_back_time))
+    return float(latest_system_back_time)
```
stanley31huang, line 79: same here, the variable type is float already.

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,162 @@
...
+    :param args: arguments from sys
+    :type args: sys.argv
+    """
+    parser = argparse.ArgumentParser(
+        description="Parse command line arguments."
+    )
+
+    parser.add_argument(
+        "--interface", required=True, help="The network interface to use."
+    )
```
stanley31huang, line 95: is this argument necessary for checking the system been wakeup? If not, I would suggest we remove this argument.

providers/base/bin/wol_client.py
```diff
@@ -0,0 +1,284 @@
...
+    req = urllib.request.Request(url, data=data_encoded, headers=headers)
+
+    attempts = 0
+    while attempts < retry:
+        try:
+            with urllib.request.urlopen(req) as response:
+                logging.info("in the urllib request.")
+                response_data = json.loads(response.read().decode("utf-8"))
+                logging.debug("Response: {}".format(response_data))
+                status_code = response.getcode()
```
stanley31huang, line 48: the `HTTPResponse.getcode` been deprecated since Python3.9, is this line still works on your test environment?
https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse.getcode

stanley31huang: Is it possible to return 4xx or 5xx error in WoL server? so we could just ensure the status code here.

providers/base/bin/wol_client.py
```diff
@@ -0,0 +1,284 @@
...
+        "retry_times": args.retry,
+        "wake_type": args.waketype,
+    }
+
+    send_request_to_wol_server(url, data=req, retry=3)
+
+    bring_up_system("rtc", delay * retry * 2)
+    logging.debug(
+        "set the rtcwake time: {} seconds ".format(delay * retry * 2)
+    )
```
stanley31huang, line 274: I would suggest we move this part into the bring_up_system function.

providers/base/bin/wol_check.py
```diff
@@ -0,0 +1,162 @@
...
+    logging.info("Interface: {}".format(interface))
+    logging.info("PowerType: {}".format(powertype))
+
+    test_start_time = get_timestamp(timestamp_file)
+
+    system_back_time = (
+        get_wakeup_timestamp()
+        if powertype == "s3"
+        else get_first_boot_timestamp()
+    )
```
stanley31huang, line 141: the system_back_time might be None when there's no suspend entry in the logs. Thus, we need to check the variable type here.
Otherwise, the scripts would crashed while doing the timestamp calculation.

eugene-yujinwu: Some return check was added.

stanley31huang (approved):
LGTM

eugene-yujinwu:
Thank you @stanley31huang for detailed and helpful reviews.

PR #1690, opened 2025-01-21
Add LXD VM GPU passthrough test (New)

pedro-avalos:
## Description

- Add LXD VM physical GPU passthrough test to GPGPU provider
  - This is similar to the existing LXD container test, but it uses a virtual machine instead of a container

## Resolved issues

- Resolves CHECKBOX-1719

## Tests

The test was already present (though misconfigured in the setup steps), so the unit tests are already present in Checkbox.

https://certification.canonical.com/submissions/status/301364


Hook25 (approved):
+1, the attached test log doesn't make it 100% clear this did actually pass, but we can do some cleanup afterwards I think. Good job

PR #1693, opened 2025-01-22
Remove deprecated libraries as flagged by pytest (Bugfix)

Hook25:
## Description

Checkbox doesn't even start on python3.13, additionally, some tests are now broken in CI because on noble we aren't somehow installing setuptools (so we are lacking pkg_resource that is now deprecated). This PR solves this by removing al usages of deprecated libraries that pytest pointed out.
- Some of these were soft deprecated (like pkg_resource)
- some others were soon to be removed (tar usage with no filter to be gone 3.14 ) 
- some are now gone (imghdr gone in 3.13)

## Resolved issues

Fixes: CHECKBOX-1720

## Documentation

Comments provided where weird try catches are needed to maintain backward compatibility

## Tests

Most of these are tests or tested already

### DIFF
--- a/checkbox-ng/checkbox_ng/test_certification.py
+++ b/checkbox-ng/checkbox_ng/test_certification.py
@@ -43,7 +43,7 @@ try:
     def resource_string(module, path):
         return files(module).joinpath(path).read_bytes()
 
-except (ModuleNotFoundError, ImportError):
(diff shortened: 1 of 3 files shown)


pieqq (approved):
Thanks for taking the time to clean things up!

Issue #1197, opened 2024-04-19
ethernet/hotplug-{{ interface }} fails without potential latency


scotthu27:
### Bug Description

The ethernet functions normally even though the test case fails repeatedly.

Possible reason seems to be the test case starts pinging on cable detection, ignoring the potential latency between cable plug-in and system network setup. See implementation of the test case:

    $ eth_hotplugging.py {{ interface }} && gateway_ping_test.py --interface {{ interface }}

This could be fixed by adding few seconds of latency in between. For example:

    $ eth_hotplugging.py {{ interface }} && sleep 5 && gateway_ping_test.py --interface {{ interface }}

The SKU could pass the modified test case.

### To Reproduce

1. run ethernet/hotplug-{{ interface }} test

### Environment

OS: jammy
checkbox type: debian
checkbox version: 4.0.0.dev194

### Relevant log output

```shell
[Checkbox job `com.canonical.certification::ethernet/hotplug-enp2s0` output]

stdout
------
Press enter and unplug the ethernet cable from the port enp2s0 of the System.
After 15 seconds plug it back in.
Checkbox session may be interrupted but it should come back up.
Waiting for cable to get disconnected.
...Cable unplugged!
Waiting for the cable to get connected.
..........Cable detected!
Failed to find a host to ping on interface enp2s0
FAIL: Unable to ping any host with the above interfaces

stderr
------
ERROR:root:Could not find def gateway info in /proc
WARNING: systemd-networkd is not running, output will be incomplete.

Failed to query link bit rates: Unit dbus-org.freedesktop.network1.service not found.
Failed to query link DHCP leases: Unit dbus-org.freedesktop.network1.service not found.
```

LiaoU3:
I just tested on ASL (202407-34231), `sleep 5` does help.

## Without sleep
```
root@ubuntu:/snap/checkbox24/current/providers/checkbox-provider-base/bin# ./eth_hotplugging.py enp0s30f4 && ./gateway_ping_test.py --interface enp0s30f4
Press enter and unplug the ethernet cable from the port enp0s30f4 of the System.
After 15 seconds plug it back in.
Checkbox session may be interrupted but it should come back up.

Waiting for cable to get disconnected.
...Cable unplugged!
Waiting for the cable to get connected.
.....Cable detected!
ERROR:root:Could not find def gateway info in /proc
Failed to find a host to ping on interface enp0s30f4
FAIL: Unable to ping any host with the above interfaces
```
## With sleep
```
[...]

Issue #1260, opened 2024-05-30
ethernet/hotplug-.* failed due to unable to ping the host


LiaoU3:
### Bug Description

[ethernet/hotplug-eth0](https://certification.canonical.com/hardware/202311-32317/submission/371615/test/112526/result/41212712/) failed, however, the function is well indeed.

submission: https://certification.canonical.com/hardware/202311-32317/submission/371615/

### To Reproduce

1. sudo snap install checkbox24 --beta
2. sudo snap install checkbox --channel 24.04/beta
3. checkbox.checkbox-cli
4. Select `client-cert-iot-server-24-04-manual`
5. Select `Ethernet Hotplut Test`

### Environment

 - OS: Server 24.04
 - Checkbox Type: Snap
 - Checkbox Version: 4.0.0-dev264
 - Hardware: KV260 Rev2 ([202311-32317](https://certification.canonical.com/hardware/202311-32317/submission/371615/))

### Relevant log output

```shell
Press enter and unplug the ethernet cable from the port eth0 of the System.
After 15 seconds plug it back in.
Checkbox session may be interrupted but it should come back up.
Waiting for cable to get disconnected.
....Cable unplugged!
Waiting for the cable to get connected.
......Cable detected!
ERROR:root:Could not find def gateway info in /proc
Failed to find a host to ping on interface eth0
FAIL: Unable to ping any host with the above interfaces
```

baconYao:
Same issue can be reproduced on UC22 as well

Failure rate: 100%

Baoshan
- G1200: https://certification.canonical.com/hardware/202307-31859/submission/375967/test-results/fail/?term=ethernet%2Fhotplug-eth0
- G700: https://certification.canonical.com/hardware/202308-31931/submission/375968/test-results/fail/?term=ethernet%2Fhotplug-eth0
- G350: https://certification.canonical.com/hardware/202309-32029/submission/375952/test-results/fail/?term=ethernet%2Fhotplug-eth0

Hook25:
Dumb question could this be because we try right after detecting the re-plugging connectivity? I think a few retries with some exponential backoff can't be bad in this situation while we just `detect connected && check connection`, which is not optimal. What do you think? Can you just run this command and see if running the second one a few seconds from the first fixes this?

LiaoU3:
This is duplicated to #1197

PR #1694, opened 2025-01-22
Eth hotplug check cable and routable before pinging (Bugfix)

seankingyang:
## Description

Some of the machines (most of the IOT devices) can't ping after plugging the eth cable.
Some of the machines (most of the PC devices) can ping after plugging the eth cable.

### In Desktop image, the networking renderer is Networkmanager
It will delay to turn down the eth interface and keep the ip (will not send req to DHCP to get the IP)  when unpluggin the cable.
- If the tester remove the cable and connect the cable immediatly, the test case can be pass (This is the work around method for the original script)
- If the tester remove the cable, make sure the NetworkMangager trun the eth interface down, and reconnect the cable back. In this situation, some of the device will get the ip slowly and this test case will fail. But it really can get the original ip and available to ping.

### In Server/UC image, the networking renderer is systemd-networkd
It will immediatly turn the interface down when unplugging the cable.
- If the tester remove the cable and reconnect the cable immedatly, this test case will still fail. But it really can get the original ip and available to ping.

### Change
Need to make sure the networking renderer actually not routable if remove the cable; it actually routable if reconnect the cable before ping.

### Flow
1, Check interface cable state (make sure this eth interface is existed)
2. Ask Tester to remove the eth cable
3. check cable is remove
4. check network renderer is Not routable
5. Ask Tester to reconnect the eth cable
6. check cable is connect
7. check network renderer is routable
8. Ping the gateway 

## Resolved issues

#1197 ethernet/hotplug-{{ interface }} fails without potential latency
#1260 ethernet/hotplug-.* failed due to unable to ping the host

## Tests

PC Desktop image with NetworkManager: https://certification.canonical.com/hardware/202411-35941/submission/409179/
IOT UC image with systemd-networkd: https://certification.canonical.com/hardware/202307-31859/submission/409181/

### DIFF
--- a/providers/base/bin/eth_hotplugging.py
+++ b/providers/base/bin/eth_hotplugging.py
@@ -7,7 +7,103 @@
 """Check if hotplugging works on an ethernet port."""
 
 import sys
+import os
 import time
+import glob
+import yaml
+import subprocess as sp
+
+from gateway_ping_test import perform_ping_test
+
+NETPLAN_CFG_PATHS = ("/etc/netplan", "/lib/netplan", "/run/netplan")
+
+
+def netplan_renderer():
+    """
+    Check the renderer used by netplan on the system if it is networkd or
+    NetworkManager.
+    This function looks for the renderer used in the yaml files located in the
+    NETPLAN_CFG_PATHS directories, and returns the first renderer found.
+    If the renderer is not found, it defaults to "networkd".
+    If the netplan file is not found, it defaults to "NetworkManager".
+    """
+    netplan_file_exist = False
+    for basedir in NETPLAN_CFG_PATHS:
+        if os.path.exists(basedir):
+            files = glob.glob(os.path.join(basedir, "*.yaml"))
+            for f in files:
+                netplan_file_exist = True
+                with open(f, "r") as file:
+                    data = yaml.safe_load(file)
+                    if "renderer" in data["network"]:
+                        return data["network"]["renderer"]
+    if netplan_file_exist:
+        return "networkd"
+    return "NetworkManager"
+
+
+def get_interface_info(interface, renderer):
+    """
+    Get the interface information (state and gateway) from the renderer.
+    """
+    if renderer == "networkd":
+        cmd = "networkctl status --no-pager --no-legend {}".format(interface)
+        key_map = {"State": "state", "Gateway": "gateway"}
+    elif renderer == "NetworkManager":
+        cmd = "nmcli device show {}".format(interface)
+        key_map = {"GENERAL.STATE": "state", "IP4.GATEWAY": "gateway"}
+    else:
+        raise ValueError("Unknown renderer: {}".format(renderer))
+
+    return _get_cmd_info(cmd, key_map, renderer)
+
+
+def _get_cmd_info(cmd, key_map, renderer):
+    info = {}
+    try:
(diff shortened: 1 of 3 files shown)


providers/base/bin/eth_hotplugging.py
```diff
@@ -1,13 +1,122 @@
...
+    info = get_interface_info(interface, renderer)
+    state = info.get("state", "")
+    if renderer == "networkd":
+        routable = "routable" in state
+    elif renderer == "NetworkManager":
+        routable = "connected" in state and "disconnected" not in state
+    return (routable, state)
+
+
+def wait_for_routable_state(
```
pieqq, line 107: You should check the [`retry` helper decorator](https://github.com/canonical/checkbox/blob/e5b7c6f57965a382fbb91f04d0a7f7d4db762d92/checkbox-support/checkbox_support/helpers/retry.py#L84) available in checkbox-support. (you can check how it's used [here](https://github.com/canonical/checkbox/blob/e5b7c6f57965a382fbb91f04d0a7f7d4db762d92/providers/base/bin/networking_http.py#L29) for instance). This decorator was written to prevent multiple implementations of a retry mechanism in our different test scripts.

(and we provide a `mock_retry` object for writing unit tests, see the [different examples available in the repo](https://github.com/search?q=repo%3Acanonical%2Fcheckbox%20mock_retry&type=code))

providers/base/bin/eth_hotplugging.py
```diff
@@ -17,6 +126,47 @@ def has_cable(iface):
         return carrier.read()[0] == "1"
 
 
+def wait_for_cable_state(iface, do_cable=True, max_wait=30):
```
pieqq, line 129: Same here, using the `retry` decorator would be better.

providers/base/bin/eth_hotplugging.py
```diff
@@ -1,13 +1,122 @@
...
+            if ":" not in line:
+                continue
+            key, val = line.strip().split(":", maxsplit=1)
+            key = key.strip()
+            val = val.strip()
+            if key in key_map:
+                info[key_map[key]] = val
+    except sp.CalledProcessError as e:
+        print("Error running {} command: {}".format(renderer, e))
+    return info
```
pieqq, line 89: Is there a case where you would want to return an empty dictionary and keep executing the script, even though the command returned something wrong?

Wouldn't it be better to just let the `CalledProcessError` exception bubble up?

pieqq (changes requested):
Thanks for undertaking this big change!

I noticed you used your own retry mechanism for different functions in the script. I recommend using the `retry` decorator available in checkbox-support (see comments inline below).

providers/base/bin/eth_hotplugging.py
seankingyang: Agree, it will be better to make the exception bubble up.

providers/base/bin/eth_hotplugging.py
seankingyang: Done

providers/base/bin/eth_hotplugging.py
seankingyang: OK, I finished these changes.

pieqq (approved):
LGTM. Thanks!

PR #1695, opened 2025-01-22
Fix gpu-burn test estimated duration (BugFix)

pedro-avalos:
## Description

Update the GPU Burn estimated duration to match the test execution time in seconds.


Hook25 (approved):
+1

PR #1699, opened 2025-01-23
Hidden manifests implementation (new)

Hook25:
## Description

We need a “hidden” manifest type to be able to include/exclude jobs runs in the lab without impact normal end user experience:

Hidden = not available for general consumption, not shown in the Manifest entries screen

hidden keyword raises questions, so need an additional hidden_reason field to explain why this manifest is not shown to the user (e.g. “No OBEX server testing required in the HW Cert lab”)

These manifest entries must be false (i.e. not set) for submissions made to issue a Certificate (Cert issuance process must include this check, this is a critical violation because it can lead to important things not being tested)

Hidden manifest IDs must start with _ (underscore). This is for easier reading/parsing/debugging, and must have a hidden-reason field to explain why they were introduced.

## Resolved issues

Fixes: CHECKBOX-1698

## Documentation

New field documented in reference

## Tests

Tested via metabox + unittest

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -639,7 +639,8 @@ class RemoteController(ReportsStage, MainLoopStage):
 
     def select_jobs(self, all_jobs):
         if self.launcher.get_value("test selection", "forced"):
-            self._save_manifest(interactive=False)
+            if self.launcher.manifest:
+                self._save_manifest(interactive=False)
         else:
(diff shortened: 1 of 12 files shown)


fernando79513 (commented):
Good job, Max!
The PR looks good to me. The hidden manifest implementation is what we've discussed previously and the renamings here and there look coherent.
Also, the metabox tests seem to cover the functionality properly.
I've left a small comment if you fancy to change it, but I leve it to you.

fernando79513 (approved):

PR #1702, opened 2025-01-29
Clean suspend time test (BugFix)

fernando79513:
## Description

We had several some tests with conflicting output files.
We have removed one unused test, and set coherent names for the attachment files.

Apart from the naming issues fixed in this PR, I think that all of these suspend tests could be simplified by:
 - Merge `suspend_single.log` and `suspend_single_times.log`. Since `suspend_single_times.log` already includes the contents of `suspend_single.log`, we could just have one log file that included all the information and check the part that we are interested in. It seems that we are already doing that in [some suspend tests](https://github.com/canonical/checkbox/blob/clean-suspend-time-test/providers/base/units/stress/suspend_cycles_reboot.pxu#L82)
 - Get rid of file checks inside the command:
  `[ -e "$PLAINBOX_SESSION_SHARE"/suspend_single_times.log ] && sleep_time_check.py "$PLAINBOX_SESSION_SHARE"/suspend_single_times.log`
  And just check if the file exists in the `sleep_time_check.py` script

We could tackle these in a follow-up PR or address them here if we consider that's the way to go.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1644

### DIFF
--- a/providers/base/units/suspend/suspend-graphics.pxu
+++ b/providers/base/units/suspend/suspend-graphics.pxu
@@ -302,10 +302,10 @@ user: root
 command:
  if type -P fwts >/dev/null; then
   echo "Calling fwts"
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
+1, we should really unify all those names

Issue #422, opened 2023-04-06
cpu/clocktest should not run on s390x


bladernr:
### Bug Description

The clocktest script doesn't work on Z due to issues with sched_setaffinity.  Having talked  with Frank before we don't think this is really necessary as a test case on IBM Z anyway so it's safe to just skip it on s390x arch systems.

### To Reproduce

Run cpu_clocktest on an LPAR or zVM guest on the Z 

### Environment

IBM Z running Lunar (has been seen on other releases too, test is inconsequential on Z and can be skipped)

### Relevant log output

```shell
sched_setaffinity: Invalid argument
Testing for clock jitter on 282 cpus
```

pieqq:
The job in question is in the base provider:

https://github.com/canonical/checkbox/blob/main/providers/base/units/cpu/jobs.pxu#L126-L135

It is currently executed for any architecture, but this can be tailored using the `requires` field of the [job unit](https://checkbox.readthedocs.io/en/latest/units/job.html) and the `cpuinfo` resource job, [similar to the `cpu/cstates` job](https://github.com/canonical/checkbox/blob/main/providers/base/units/cpu/jobs.pxu#L209) for instance.

PR #1704, opened 2025-01-31
Ensure cpu/clocktest does not run on the s390x architecture (BugFix)

pesader:
## Description

The cpu/clocktest test doesn't work on s390x due to issues with sched_setaffinity (as discussed in issue #422), so make sure this test doesn't run on this architecture. To do that, add a `requires` to the cpu/clocktest so it doesn't fulfill its requirements if the detected CPU architecture is "s390x".

## Resolved issues

- #422

## Tests

I haven't written any Python code for this contribution, so I don't think the test coverage has changed. Still, I ran `coverage` and attached its HTML coverage report below.

[htmlcov.zip](https://github.com/user-attachments/files/18612112/htmlcov.zip)

However, I have manually tested this change by running checkbox and making sure cpu/clocktest still runs on my x86 laptop and it did (the test description is "Tests the CPU for clock jitter"):

image

I don't have an IBM mainframe at hand 😅, so I tried changing "s390x" to "x86_64" to see if the test didn't run on my laptop and it didn't:

image

I make myself available for further changes and explanations. Thank you for reviewing this!

### DIFF
--- a/providers/base/units/cpu/jobs.pxu
+++ b/providers/base/units/cpu/jobs.pxu
@@ -126,6 +126,8 @@ _purpose:
 plugin: shell
 category_id: com.canonical.plainbox::cpu
 id: cpu/clocktest
+requires:
+ cpuinfo.platform not in ("s390x")
 flags: also-after-suspend
 estimated_duration: 300.0
 command: clocktest


Hook25 (approved):
+1, note that this works only if the platforms are uniform due to how negations work in resource expressions, either way, I think we can make this assumption

PR #1708, opened 2025-02-04
Fix clear old sessions for local (BugFix)

Hook25:
## Description

This fixes the `--clear-old-sessions` flag being overall ignored, also, it fixes the way that these flags are used for local only.

Lastly, it fixes a missing catch for both local and remote, that would make the agent crash when a selected job is deleted (from an update or manually) and a session is then resumed (as the error won't be InvalidJobError but CorruptedSessionError).

## Resolved issues

Fixes: CHECKBOX-1710

## Documentation

N/A

## Tests

New metabox test that stresses this scenario for local

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
+++ b/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
@@ -73,7 +73,7 @@ def handle_top_parser(args, ctx):
         logging.basicConfig(level=logging_level)
         set_all_loggers_level(logging.DEBUG)
         ctx.args.debug = True
-    if "--verbose" in sys.argv or "-v" in sys.argv:
+    elif "--verbose" in sys.argv or "-v" in sys.argv:
         logging_level = logging.INFO
         logging.basicConfig(level=logging_level)
         set_all_loggers_level(logging.INFO)
--- a/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
+++ b/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
@@ -22,14 +22,14 @@ Checkbox Launcher Interpreter Application
 
 import gettext
 import logging
-import os
-import subprocess
 import sys
+
+import checkbox_ng
 
 from plainbox.impl.jobcache import ResourceJobCache
 from plainbox.impl.session.assistant import SessionAssistant
 
-from checkbox_ng.config import load_configs
+from checkbox_ng.utils import set_all_loggers_level
 from checkbox_ng.launcher.subcommands import (
     Launcher,
     List,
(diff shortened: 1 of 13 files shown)


checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
```diff
@@ -62,6 +62,35 @@ def reset_sa(self):
...
+    if "--debug" in sys.argv:
+        logging_level = logging.DEBUG
+        logging.basicConfig(level=logging_level)
+        set_all_loggers_level(logging.DEBUG)
+        ctx.args.debug = True
+    if "--verbose" in sys.argv or "-v" in sys.argv:
+        logging_level = logging.INFO
+        logging.basicConfig(level=logging_level)
+        set_all_loggers_level(logging.INFO)
+        ctx.args.verbose = True
```
fernando79513, line 80: Minor:
If for some reason both `--verbose` and `--debug` flags are set in the command, the verbose takes priority since they are not exclusive. I think it should be the other way around.
I don't think this would ever be a problem, but maybe someone somewhere sets both at the same time unintentionally?

checkbox-ng/checkbox_ng/launcher/test_subcommands.py
```diff
@@ -632,14 +632,30 @@ def test__auto_resume_session_from_ctx_unknown_session(self):
     def test__auto_resume_session_autoresume(self):
         self_mock = MagicMock()
         resume_candidate_mock = MagicMock(id="session_to_resume")
+        # session id wasnt provided directly via the cli
```
fernando79513, line 635:
```suggestion
        # session id wasn't provided directly via the cli
```

checkbox-ng/checkbox_ng/launcher/test_subcommands.py
```diff
@@ -632,14 +632,30 @@ def test__auto_resume_session_from_ctx_unknown_session(self):
...
 
         self.assertTrue(
             Launcher._auto_resume_session(self_mock, [resume_candidate_mock])
         )
         self.assertTrue(self_mock._resume_session.called)
 
+    def test__auto_resume_session_no_autoresume_on_clear(self):
+        self_mock = MagicMock()
+        resume_candidate_mock = MagicMock(id="session_to_resume")
+        # session id wasnt provided directly via the cli
```
fernando79513, line 649:
```suggestion
        # session id wasn't provided directly via the cli
```

fernando79513 (commented):
LGTM!
I tried the `--clear-old-session` flag and seems to work fine.
Taking a look at the debug logs, they seem really verbose, I wonder if some of them should be traces instead. Anyways, it's not something to tackle here.

fernando79513 (approved):

fernando79513 (approved):

PR #1715, opened 2025-02-10
Added tests for IOMMU config (New)

fernando79513:

## Description

We want to check if the `CONFIG_INTEL_IOMMU_DEFAULT_ON` config flag is present in versions of the kernel >=6.8.0-20

Note: I've included the test into the submission-cert-automated test-plan. It's already being run in all of the devices and it includes also some info tests, so it made sense for me to include it here. Let me know if you think it should be in another test-plan
Note: I had to remove the config iterator (see commit: removed iterator for python 3.5). I don't like this too much, but the unit tests failed in python 3.5 if I read the file line by line.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1608

## Tests

Tested with:
classic image, ubuntu 24, kernel "6.8.0-52-generic"
```
-------------[ Check if the kernel is compiled with IOMMU support ]-------------
ID: com.canonical.certification::info/kernel-config-iommu-flag
Category: com.canonical.plainbox::info
Flag CONFIG_INTEL_IOMMU_DEFAULT_ON is present and set to 'y'.
```
core image, ubuntu 24, pc-kernel "6.11.0-17-generic"
https://certification.canonical.com/hardware/201805-26252/submission/410961/test-results/
```
-------------[ Check if the kernel is compiled with IOMMU support ]-------------
ID: com.canonical.certification::info/kernel-config-iommu-flag
Category: com.canonical.plainbox::info
Flag CONFIG_INTEL_IOMMU_DEFAULT_ON is present and set to 'y'.
```

core image, ubuntu 22, pc-kernel "5.15.0-132-generic"
https://certification.canonical.com/hardware/201802-26085/submission/410964/test-results/
```
-------------[ Check if the kernel is compiled with IOMMU support ]-------------
ID: com.canonical.certification::info/kernel-config-iommu-flag
Category: com.canonical.plainbox::info
Skipping: kernel version 5.15.0-132 is lower than 6.8.0-20.
```

core image, ubuntu 22, pc-kernel "5.15.0-132-generic"
https://certification.canonical.com/hardware/201802-26085/submission/410964/test-results/
```
-------------[ Check if the kernel is compiled with IOMMU support ]-------------
ID: com.canonical.certification::info/kernel-config-iommu-flag
Category: com.canonical.plainbox::info
Skipping: kernel version 5.15.0-132 is lower than 6.8.0-20.
```

core image, ubuntu 24, pi-kernel "6.8.0-1016-raspi" (ARM)
https://certification.canonical.com/hardware/202310-32203/submission/411715/test-results/
```
-------------[ Check if the kernel is compiled with IOMMU support ]-------------
ID: com.canonical.certification::info/kernel-config-iommu-flag
Category: com.canonical.plainbox::info
 - resource expression 'cpuinfo.platform in ("i386", "x86_64")' evaluates to false
```

### DIFF
--- a/providers/base/bin/kernel_config.py
+++ b/providers/base/bin/kernel_config.py
@@ -30,8 +30,8 @@ def get_kernel_config_path():
     kernel_version = os.uname().release
     if on_ubuntucore():
         kernel_snap = get_kernel_snap()
-        return f"/snap/{kernel_snap}/current/config-{kernel_version}"
-    return f"/boot/config-{kernel_version}"
+        return "/snap/{}/current/config-{}".format(kernel_snap, kernel_version)
+    return "/boot/config-{}".format(kernel_version)
 
 
 def get_configuration(output=None):
@@ -69,7 +69,7 @@ def check_flag(flag, min_version):
                 print("Flag {} is present and set to 'y'.".format(flag))
                 return
 
-    raise SystemExit("Flag {} not found in the config.".format(flag))
+    raise SystemExit("Flag {} not found in the kernel config.".format(flag))
 
 
 def parse_args():
--- a/providers/base/bin/kernel_config.py
+++ b/providers/base/bin/kernel_config.py
@@ -16,7 +16,6 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
-
 import argparse
 import os
 import shutil
@@ -55,7 +54,7 @@ def check_flag(flag, min_version):
         min_version
     ):
         print(
-            "Skipping flag check as kernel version"
+            "Skipping: kernel version"
             " {} is lower than {}.".format(kernel_version, min_version)
         )
         return
@@ -70,7 +69,7 @@ def check_flag(flag, min_version):
                 print("Flag {} is present and set to 'y'.".format(flag))
                 return
 
-    raise SystemExit("Flag {} not found in the kernel config.".format(flag))
+    raise SystemExit("Flag {} not found in the config.".format(flag))
 
 
 def parse_args():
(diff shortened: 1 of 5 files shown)


providers/base/bin/kernel_config.py
```diff
@@ -0,0 +1,115 @@
...
+    version_parts = os.uname().release.split("-")
+    kernel_version = "-".join(version_parts[:2])
+    if min_version and version.parse(kernel_version) < version.parse(
+        min_version
+    ):
+        print(
+            "Skipping: kernel version"
+            " {} is lower than {}.".format(kernel_version, min_version)
+        )
+        return
```
pieqq, line 60: Shouldn't Checkbox mechanism be used to skip the job instead of passing the job with a "skipping..." comment?

To do this, you can use the `uname` resource job that would contain the version number, for instance. Then you could use this as a requirement (`uname.release >= "6.8.0-20"`) to decide whether or not to skip the job.

providers/base/bin/kernel_config.py
```diff
@@ -0,0 +1,115 @@
...
+        # Check the header and ignore arm architecture configurations
+        lines = config.readlines()
+        header = lines[:4]
+        values = lines[4:]
+
+        for line in header:
+            if "Kernel Configuration" in line:
+                if "arm" in line:
+                    print("Skipping: arm architecture detected.")
+                    return
```
pieqq, line 73: Same here, you could use the `cpuinfo.platform` resource information to skip if not `x86_64`.

pieqq (changes requested):
Good call on creating an attachment to have a copy of the kernel config in each submission.

I've provided a few comments inline. I think Checkbox resource jobs can be used to skip the job instead of running it when not needed.

providers/base/bin/kernel_config.py
fernando79513: As discussed in Mattermost, I will rephrase this but keep the version check inside the script.

PR #1717, opened 2025-02-11
Fix NTFS SD card mount failure by adding sync and udevadm settle (BugFix)

Jefferyyen:
Fix an issue where Checkbox aborts during the SDHC storage test due to a race condition when mounting an NTFS-formatted SD card. The test fails with:

    ntfs-3g-mount: mount failed: Device or resource busy

because `mount_usb_storage()` executes before Ubuntu's auto-mount process completes.

Add `sync` to flush pending writes and `udevadm settle` to wait for udev events, ensuring the device is ready before mounting.

## Description
This PR enhances the storage mounting function in Checkbox by adding sync and udevadm settle. These additions ensure that the system has fully registered the device before attempting to mount it, preventing premature failures.

## Resolved issues
When running the **insertion + read/write + removal test**, if the SD card is formatted as **NTFS**, Checkbox may attempt to mount the device before Ubuntu's auto-mount process is complete. This results in an error log:

    ntfs-3g-mount: mount failed: Device or resource busy

which causes the test to abort prematurely.

By adding `sync` and `udevadm settle`, we ensure that the system completes the device registration before proceeding with the mounting process, preventing this issue.

## Tests

The fix was verified by running the following test procedure:

Test Steps:
1. Run the command:`checkbox-cli run com.canonical.certification::mediacard/sdhc-storage-manual`

2. When prompted, insert an **NTFS-formatted SD card**.
3. Observe the output logs:
```
--------- Testing insertion ---------  
INSERT NOW  
Timeout: 30 seconds  
```

4. Checkbox detects the card and attempts to mount it. Before applying this fix, the following error occurs:
```
------- Insertion test passed -------  
--------- Testing read/write --------  
DEBUG: generating a random file  
Mounting the USB storage  
sda1  
DEBUG: try to mount USB storage for testing  
ntfs-3g-mount: mount failed: Device or resource busy    🔴this one
ERROR: mount /dev/sda1 on /tmp/tmpXXXXXX failed.  
```

5. With the applied fix, `sync` and `udevadm settle` ensure that the mounting process does not fail due to a premature device state.


.github/workflows/deb_validator.yaml
```diff
@@ -59,8 +59,9 @@ jobs:
       # needed by providers that pull checkbox-support
       - name: Install PPA and dependencies (fix GPG Key Issue)
         run: |
-          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 753F8DAABFD945830D46A6BB968504F7952C9377
-          sudo add-apt-repository ppa:checkbox-dev/edge
+          sudo mkdir -p /etc/apt/keyrings
+          curl -fsSL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x753F8DAABFD945830D46A6BB968504F7952C9377" | sudo gpg --dearmor -o /etc/apt/keyrings/checkbox-dev.gpg
+          echo "deb [signed-by=/etc/apt/keyrings/checkbox-dev.gpg] https://ppa.launchpadcontent.net/checkbox-dev/edge/ubuntu $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/checkbox-dev.list
```
fernando79513, line 64: Why do you need this change? 
I think the issue is with your GPG keys not being properly verified.

checkbox-support/checkbox_support/scripts/tests/test_mount_usb_storage.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python3
```
fernando79513, line 1: I think this file is in the propper place, but I don't see a reason why it is separated from the `test_usb_read_write.py` script.
I would add the contents of that test to this one, and rename this one to `test_usb_read_write.py`

fernando79513 (commented):
Thank you very much for this PR!
It's really well documented and explained. I've just added a couple of comments.
The main issue is that I think you have to validate your gpg keys. That will fix the issue with the deb validator and with the merge of the PR.

.github/workflows/deb_validator.yaml
Jefferyyen: I encountered a timeout error with `sudo add-apt-repository ppa:checkbox-dev/edge` in CI/CD, so I used an alternative method to fetch the key to avoid blocking my commit.
Here is the log, did i get it wrong and need to reset/rebase this commit?

```
deb_validation (ubuntu-20.04, providers/sru)
failed 1 hour ago in 2m 16s

Run sudo add-apt-repository ppa:checkbox-dev/edge
 
To align our distribution channels, have transitioned to a new set of PPAs.
These new channels will replace our existing ones and provide a clearer, more
consistent structure for our releases.

------------------------------------------------------------------------------
Edge: ppa:checkbox-dev/edge: For testing recent developments not yet in beta.
------------------------------------------------------------------------------

To install Checkbox testing suite, please do the following:

sudo add-apt-repository ppa:checkbox-dev/edge
sudo apt-get update
sudo apt-get install canonical-certification-client

You can then run Checkbox by running:
- checkbox-cli (textual user interface)
 More info: https://launchpad.net/~checkbox-dev/+archive/ubuntu/edge
Error: retrieving gpg key timed out.
Error: Process completed with exit code 1.

```

checkbox-support/checkbox_support/scripts/tests/test_mount_usb_storage.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Define a side_effect to simulate different return values for various commands:
+        def call_side_effect(args, **kwargs):
+            if args[0] == "mount":
+                return 0  # Simulate mount success
+            else:
+                return 0  # Other commands also return success
+
+        mock_call.side_effect = call_side_effect
+
```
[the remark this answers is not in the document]
Jefferyyen: Thanks for pointing that out! I've updated the code as suggested.

checkbox-support/checkbox_support/scripts/tests/test_mount_usb_storage.py
Jefferyyen: I've committed the changes as suggested. I'm not sure if my understanding is completely correct, please let me know if anything needs further adjustment. Thank you.

fernando79513: I think I didn't make it super clear with my comment. The first part was fine (rename the test to `test_usb_read_write.py` under `checkbox_support/scripts/tests`
There was also another `test_usb_read_write.py` under  `checkbox_support/tests` that you have now merged. You could just remove this one, since it's already merged with the one in `checkbox_support/scripts/tests/test_usb_read_write.py`
This was an issue in our part because the test was not in the propper place to begin with, but thank you for the fix!

checkbox-support/checkbox_support/scripts/tests/test_usb_read_write.py
```diff
@@ -0,0 +1,210 @@
...
+            "checkbox_support.scripts.usb_read_write.FOLDER_TO_MOUNT",
+            "/mnt/usb",
+        ):
+            with self.assertRaises(SystemExit):
+                with mount_usb_storage("sda1"):
+                    # As soon as we enter the context, a non-zero return from mount should trigger sys.exit(1)
+                    pass
+
+        # Check that sys.exit was called correctly (with 1)
+        mock_exit.assert_called_once_with(1)
```
fernando79513, line 206: I think this test is more complex than it has to be. You don't have to mock the `sys.exit()` or the `join()` functions, it should be just okay mocking the subprocess. You could also check that the right message has been logged though.
```suggestion
    @patch("checkbox_support.scripts.usb_read_write.subprocess.call")
    @patch("logging.error")
    def test_mount_usb_storage_failure(self, mock_log, mock_call):
        """
        Test the failure scenario:
          - Simulate that the mount command returns 1 (failure).
          - The program should call sys.exit(1) and raise a SystemExit.
        """

        def call_side_effect(args, **kwargs):
            if args[0] == "mount":
                return 1  # Simulate mount failure
            else:
                return 0

        mock_call.side_effect = call_side_effect

        with patch(
            "checkbox_support.scripts.usb_read_write.FOLDER_TO_MOUNT",
            "/mnt/usb",
        ):
            with self.assertRaises(SystemExit) as context:
                with mount_usb_storage("sda1"):
                    # As soon as we enter the context, a non-zero return from
                    # mount should trigger sys.exit(1)
                    pass

        self.assertEqual(context.exception.code, 1)

        # Verify that the program logs the error message
        mock_log.assert_called_once_with(
            "mount /dev/sda1 on /mnt/usb failed."
        )
```

Jefferyyen:
Thank you for your suggestion. I've committed the changes to the PR again. Please don't hesitate to let me know if anything needs to change.

fernando79513:
As mentioned [here](https://github.com/canonical/checkbox/pull/1717#discussion_r1971190946), please try to remove the changes to the GitHub actions, these shouldn't be needed after the latest fixes, If they still don't work, try rebasing your branch with main again.

Jefferyyen:
Thanks for the feedback. I've removed the GitHub Actions changes and deleted the redundant test file. Please let me know if you have any further suggestions.

Issue #1716, opened 2025-02-11
Got `KeyError: 'hidden'` when going into the page to select manifest


LiaoU3:
### Bug Description

When running a test plan which requires selecting manifest (e.g. com.canonical.certification::client-cert-iot-server-22-04), it will go directly to  `KeyError: 'hidden'`. This bug will not happen on the testplans which don't require manifest.

Note that it happens on both `beta` and `edge` channel of checkbox core snap `checkbox24` (not front end snap `checkbox`), but not on `stable` of the host instead of the DUT.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. checkbox.checkbox-cli control {DUT_IP}
2. Select `com.canonical.certification::client-cert-iot-server-22-04`
3. Press **T** to start testing

### Expected Result

Go to the page to select manifest

### Actual Result

```terminal
vincent@vincent-XPS-9320:~$ checkbox.checkbox-cli control 10.102.182.225
$PROVIDERPATH is defined, so following provider sources are ignored ['/home/vincent/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop'] 
Connecting to 10.102.182.225:18871. Timeout: 600s
==========[ Bootstrap com.canonical.certification::bootloader (1/24) ]==========
===========[ Bootstrap com.canonical.certification::cpuinfo (2/24) ]============
============[ Bootstrap com.canonical.certification::device (3/24) ]============
==========[ Bootstrap com.canonical.certification::interface (4/24) ]===========
======[ Bootstrap com.canonical.certification::net_if_management (5/24) ]=======
=============[ Bootstrap com.canonical.certification::lsb (6/24) ]==============
=========[ Bootstrap com.canonical.certification::dmi_present (7/24) ]==========
=============[ Bootstrap com.canonical.certification::dmi (8/24) ]==============
==========[ Bootstrap com.canonical.certification::executable (9/24) ]==========
=======[ Bootstrap com.canonical.certification::model_assertion (10/24) ]=======
=====[ Bootstrap com.canonical.certification::wifi_interface_mode (11/24) ]=====
=[ Bootstrap com.canonical.certification::bluez-internal-rfcomm-tests (12/24) ]=
==[ Bootstrap com.canonical.certification::bluez-internal-hci-tests (13/24) ]===
===[ Bootstrap com.canonical.certification::bluez-internal-uc-tests (14/24) ]===
[...]

PR #1718, opened 2025-02-11
Backward compatible hidden check (bugfix)

Hook25:
## Description

This change was not supposed to be backward incompatible. This fixes it, default is False as before the hidden_manifests change, no manifest was hidden

## Resolved issues

Fixes: CHECKBOX-1735
Fixes: https://github.com/canonical/checkbox/issues/1716

## Documentation

N/A

## Tests

I installed checkbox stable, that doesnt have this feature, and connected running the provided test plan. It didn't crash as in the bug report


pieqq (approved):
+1, thanks for this quick patch!

PR #1730, opened 2025-02-17
Add DSS reboot testing (New)

motjuste:
> [!NOTE]
> This is a POC to get feedback on.  #1725 and its children need to be merged before this PR can be put into the final shape.

## Description

- Added three new test-plans.
  - First one reboots the machine
  - Second one performs tests on DSS that need to pass after reboot.
  - Third one uses the original test-plan and the two new ones above as nested parts, running first the original plan, then the reboot + "after reboot" tests.
- Added a new script / app to the `checkbox-dss` snap that now runs the test plan with reboot.
  - It explicitly requires IP to a remote that needs to be controlled.

- Missing:
  - update to README
  - bump snap version?
  - GH workflow using Testflinger multi-job to perform the tests.

## Resolved issues

- [CHECKBOX-1692](https://warthogs.atlassian.net/browse/CHECKBOX-1692)

## Documentation

No changes to Checkbox documentation.

## Tests

> [!NOTE]
> Work in Progress

[CHECKBOX-1692]: https://warthogs.atlassian.net/browse/CHECKBOX-1692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
+++ b/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
@@ -350,7 +350,9 @@ depends: dss/initialize
 _summary: Check that the DSS checks pass after reboot.  This job performs the reboot.
 estimated_duration: 5m
 user: root
-command: echo "rebooting"
+command:
+  echo "rebooting"
+  reboot
 
 id: dss/wait_after_reboot
 category_id: dss-regress
--- a/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
+++ b/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
@@ -340,3 +340,248 @@ depends: dss/create_tensorflow_cuda_notebook
 _summary: Check that the Tensorflow CUDA notebook can be removed
 estimated_duration: 1m
 command: check_dss.sh can_remove_notebook tensorflow-cuda
+
+# Reboot and re-test ##########################################################
+
+id: dss/reboot
+category_id: dss-regress
+flags: simple noreturn
+depends: dss/initialize
+_summary: Check that the DSS checks pass after reboot.  This job performs the reboot.
+estimated_duration: 5m
+user: root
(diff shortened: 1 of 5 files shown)


contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
```diff
@@ -340,3 +340,250 @@ depends: dss/create_tensorflow_cuda_notebook
...
+_summary: Check that PyTorch can use CPU in notebook created after reboot
+estimated_duration: 1m
+command: check_notebook.sh pytorch-cpu-ar can_run_python_script_in_pod verifying_pytorch_can_use_cpu
+
+id: dss/remove_pytorch_cpu_notebook_after_reboot
+category_id: dss-regress
+flags: simple
+imports: from com.canonical.certification import executable
+requires: executable.name == 'dss'
+depends: dss/create_pytorch_cpu_notebook_after_reboot
```
fernando79513, line 412: I think we should give a second thought to all of this `depends` on the tests, but maybe that will have to wait until you are back from your leave:
 - What happens if the "can_use_cpu" test fails? Then the notebook will not be removed, and in the following run, it will fail again.
 I think an "after" flag will make more sense in that case so the removal test is executed despite the check test failing.
 But that will not be the solution either, because in the rerun we will not remove the environment if the test passed the first time.
 I think the solution here is to first check if the notebook exists before creating it, try to remove it if exists and then create it, so we can have a "clean" state despite the results of previous runs

motjuste: This is a better idea.  I was unaware of the "after" flag (still learning, sorry).  I will pick this PR up after I am back from the leave.

motjuste:
Closing this PR as it has been stuck in Draft for too long.  We'll pick it up soon with a fresh mind and with all the updates from other PRs available to us.

PR #1732, opened 2025-02-17
Noble testing infra (infra)

Hook25:
## Description

We have to start testing checkbox for noble, so python 3.12 and metabox testing on noble. This updates all tox and metabox workflow

## Resolved issues

Fixes: CER-2857

## Documentation

N/A

## Tests

N/A

to verify this, beside the pipeline, check that all the packages are at the same version in the tox jobs

### DIFF
--- a/tools/release/tox.ini
+++ b/tools/release/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py310
+envlist = py312
 
 [testenv]
 commands =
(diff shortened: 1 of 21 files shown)


tools/release/tox.ini
pieqq, line 1: Is this modified because ubuntu-latest in GitHub Actions is now pointing to Noble?

pieqq (dismissed):
Thanks for taking this! One small question regarding the tox.ini in the release dir, otherwise LGTM

tools/release/tox.ini
Hook25: yup, that is already the case. This didnt bite us yet, but it will in the future

pieqq (approved):
+1

PR #1733, opened 2025-02-17
Only check changed paths (infra)

Hook25:
## Description

This makes the pipeline slightly less wasteful by only re-testing directories that did actually change.

## Resolved issues

N/A

## Documentation

N/A

## Tests

See this commit for the pipeline in action: https://github.com/canonical/checkbox/actions/runs/13372119135
See this commit for the pipeline not running (as it should): [27bc0df](https://github.com/canonical/checkbox/pull/1733/commits/27bc0dfd0a90253c1c440af6230676ccd31b8ec6)

### DIFF
--- a/.github/workflows/deb_validator.yaml
+++ b/.github/workflows/deb_validator.yaml
@@ -84,7 +84,9 @@ jobs:
           sudo apt-get update
           sudo apt-get install -qq -y dpkg-dev
       - name: Prepare installation
-        run: mv ${{ matrix.path }}/debian .
+        env:
+          MATRIX_PATH: ${{ matrix.path }}
+        run: mv "$MATRIX_PATH/debian" .
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/deb_validator.yaml
```diff
@@ -29,7 +29,37 @@ on:
...
+
+          paths=()
+          for path in \
+              checkbox-ng \
+              checkbox-support \
+              providers/*/; do
+            if changed $path; then
+              paths+=("\"$path\"");
+            fi;
+            echo $path
```
p-gentili, line 55: Wouldn't be better to rather print $path**s** at the end of the loop?

Hook25: (this was a debug print, I'll remove it)

p-gentili (approved):
+1

PR #1735, opened 2025-02-18
update SNAPD_TASK_TIMEOUT for checkbox-ce-oem snap (BugFix)

stanley31huang:
update SNAPD_TASK_TIMEOUT for checkbox-ce-oem snap

## Description
extend the timeout to 300 seconds for those platform with low performance

## Resolved issues

the snappy/snap-refresh related tests would failed due to the system took more than one minutes, that's because of the system is not powerful and/or the system needs to pull core snap as well, or the Lab network is not stable. And the purpose of this test is validate the system is able to install/refresh/revert a test snap. So we gonna extend the timeout for these tests.


zongminl (approved):
+1, thank you

Hook25:
Merging as it was approved

PR #1736, opened 2025-02-18
Remove keyboard mouse test program (bugfix)

pieqq:
This PR undoes what was done in #1582 .

This C program does not compile properly on 16.04, which is still a requirement for testing.

Moreover, we don't have the expertise to maintain it for the different archs and platform we provide Checkbox for.


Hook25 (approved):
thanks

Hook25 (approved):

Issue #1737, opened 2025-02-19
wifi_nmcli_test.py: not enough values to unpack (expected 2, got 1)


nancyc12:
### Bug Description

In the lab, when running the `com.canonical.certification::wireless/wireless_connection_open_bg_nm_wlan0` test on a device that fails to connect to an SSID, the function `wait for connected` fails with an error stating `not enough values to unpack (expected 2, got 1)` (see [Jenkins job](http://10.102.156.15:8080/job/cert-tillamook-pls63-uc20-network-manager-20-beta/26/testReport/junit/com.canonical/certification__wireless_wireless_connection_open_bg_nm_wlan0/com_canonical_certification__wireless_wireless_connection_open_bg_nm_wlan0/))

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Run the test with an SSID that doesn't exist.

### Expected Result

The test fails with the expected error message when the SSID is disconnected

### Actual Result

The test fails with a `not enough values to unpack (expected 2, got 1)` exception.

### Environment

- Checkbox Type: Snap
- Chekbox Version: checkbox20 4.4.0-dev6

### Relevant log output

```shell
===========================================
Attempt 1/5 (function 'wait_for_connected')
===========================================
+ nmcli -m tabular -t -f GENERAL.STATE,GENERAL.CONNECTION d show wlan0
30 (disconnected)


Attempt 1 failed:
not enough values to unpack (expected 2, got 1)

Waiting 1.20 seconds before retrying...
```

Hook25:
Hello, I agree that the output is ugly but it is doing exactly what it is supposed to, given that the test failed to connect, it is spitting out an error

PR #1738, opened 2025-02-19
Prettier output for the wifi_test (bugfix)

Hook25:
## Description

If the wifi test fails to connect it spits out a weird  `not enough values to unpack (expected 2, got 1)` instead of just reporting that the test failed to connect. This fixes that.

## Resolved issues

Fixes: CHECKBOX-1759
Fixes: https://github.com/canonical/checkbox/issues/1737

## Documentation

Explained why I used next and `*` in a comment above

## Tests

N/A


fernando79513 (approved):
Thanks for the fix, I saw the error the other day during wifi test review, but I forgot to fix it. Good job!
LGTM +1!

PR #1739, opened 2025-02-19
Use indexed option directive on unit reference documentation (Infra)

tang-mm:
## Description

The unit references currently use definition lists for explaining each field, but each element is not indexed so cannot be cross-referenced. This PR replaces those `<dl>` list and manually defined anchors with the [`:option:` directive in Sphinx](https://www.sphinx-doc.org/en/master/usage/domains/standard.html)

* Field names are defined with `.. option:: name`, and then are indexed during doc build 
  * the defined fields can be cross-referenced by using ``:option: `name` ``
* Scope can be defined with `.. program:: unit`, so that duplicated option names can be specified by adding the scope name before the field ``:option: `unit name` `` 
* Update CSS style for the generated `.sig` elements
* Update spelling check settings to ignore field names with .sig CSS class

Chore:
* Enabled `fail_on_warning` on readthedocs build to prevent broken links

## Resolved issues

CHECKBOX-1747

## Documentation

NA

## Tests

PR checks on Readthedocs
Preview: https://canonical-checkbox--1739.com.readthedocs.build/en/1739/reference/units/job.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -20,7 +20,7 @@ build:
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
    configuration: docs/conf.py
-
+   fail_on_warning: true   # Fail on all warnings to avoid broken references
 
 # Optionally declare the Python requirements required to build your docs
 python:
--- a/docs/.sphinx/_static/checkbox-doc.css
+++ b/docs/.sphinx/_static/checkbox-doc.css
@@ -1,3 +1,12 @@
+/* :option: directive for unit fields */
+.sig-name {
+    color: var(--color-code-foreground);
+}
+.sig:not(.sig-inline) {
+    background-color: var(--color-inline-code-background);
+}
+
+/* mermaid diagrams */
 .mermaid {
     display: flex;
     justify-content: center;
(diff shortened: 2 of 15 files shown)


tang-mm:
Preview: https://canonical-checkbox--1739.com.readthedocs.build/en/1739/reference/units/job.html

CSS style behaves differently than my local build, fix needed :(

Hook25 (approved):
lgtm, should we enable fail on warning on all doc sets?

tang-mm:
`fail_on_waring` is enabled by default in later versions of the Sphinx template. Checkbox is an early adopter so it was not here, but most other doc sets already have them :)

Hook25:
do double check on the others if you have some time, I'm pretty sure testflinger doesn't have it

tang-mm:
Nice reminder! The option was removed from Testflinger two weeks ago :thinking:

Hook25:
now I remember, I approved that PR. The problem with that option is that the build fails on readthedocs but not locally and I couldn't find a way to make it behave the same, which is pretty crappy. Can you please look into that as well? The behaviour should be the same

PR #1748, opened 2025-02-24
 Make sure OS resource job gets the proper version on devices running Ubuntu Core (bugfix)

pieqq:
## Description

hen installed on a Ubuntu Core system, the /etc/[lsb|os]-release files
accessed by Checkbox might be the ones provided by the core snap
installed. As a result, if the core snap does not match the version of
Ubuntu Core, it will result in wrong information provided. For instance,
if checkbox20 is installed on a device running UC22, the information
returned by the `lsb_resource` job will be "20.04" when it should be
"22.04".

To fix this, if run from a snap, the resource job script will now look
at the content from `/var/lib/snapd/hostfs/etc/os-release` to make sure it
retrieves the correct information.

Moreover, the script now uses data from the `os-release` file instead of
the `lsb-release` file, as the former is standardized on all
systemd-enabled systems.

## Resolved issues

[C3-1000](https://warthogs.atlassian.net/browse/C3-1000)

## Tests

New unit tests

[C3-1000]: https://warthogs.atlassian.net/browse/C3-1000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Hook25 (dismissed):
see if the below make sense

Hook25 (approved):

PR #1755, opened 2025-02-26
Improve daily builds reliability (infra)

Hook25:
## Description

We have decided to avoid relying too much on Launchpad for our daily builds because that makes us dependent on the build farm to release hotfixes more in the periods where the the build farm is busiest (as we tend to need hotfixes aroud point releases or major packages re-building). 
This:
- Moves all `remote-build` snap builds to the new cross-arch build action
- Gives sensible names to the workflows and the yamls (following our new and consolidated naming)
- Disables store pushes for manually triggered snap builds
- Removes pointless bash-ism to upload to the store (replaced with the action and if conditional)

Minor: Patch get_version.py to work on ubuntu 20, we need it to be able to build for xenial

## Resolved issues

Fixes: CHECKBOX-1763

## Documentation

Updated all references to the old workflows and their meaning

## Tests

I've ran the workflow (given that running it manually doesn't push to the store now!)
Native builds:
- https://github.com/canonical/checkbox/actions/runs/13547017328/job/37861069872
Cross-Builds:
- https://github.com/canonical/checkbox/actions/runs/13545880731/job/37857284234


pieqq (changes requested):
I think I spotted a typo in the docs, see below.

Also, I'm not sure if I understand correctly: "cross-builds" means the armhf version only, and native mean x86 and arm64?

Hook25:
cross-build as in cross-compile. In general, all the builds that we don't do natively (arm64 on arm64 or amd64 on amd64) we cross-build using qemu (armhf on amd64 for example)

.github/workflows/checkbox-daily-cross-builds.yaml
```diff
@@ -0,0 +1,158 @@
...
+            arch: arm64
+          - release: 22
+            arch: amd64
+          - release: 22
+            arch: arm64
+          - release: 24
+            arch: amd64
+          - release: 24
+            arch: arm64
+    runs-on: ${{ matrix.release == 16 && 'ubuntu-20.04' || 'ubuntu-latest' }}
```
pieqq, line 97: Can you add a comment of why UC16 requires 20.04 to be built? For our future selves :)

pieqq (approved):
Really nice stuff! I can't wait to see it in action for our snap releases :)

pieqq (approved):
+1

PR #1760, opened 2025-02-28
Use process tree to detect if the subprocess child is gone (bugfix)

Hook25:
## Description

This test is unreliable as there is a race condition between the process winding down and the actual pid being checked. Checking the process tree is more reliable, as children will be recorded in the multiprocessing tree (they inherit the names as well) so we are able to query and see if they are still alive.

## Resolved issues

Builds intermittendly failing on LP (deb)

## Documentation

N/A

## Tests

Tested locally


pieqq (approved):
LGTM

PR #1762, opened 2025-03-01
Add arm64 to GPGPU provider (New)

pedro-avalos:
## Description

- Add `arm64` support for NVIDIA GPGPU tests
  - The LXD VM passthrough test will fail for now, that is an upstream issue, not an issue with the test.
  - Most of the CUDA sample tests depend on a binary file that is only provided for `amd64`. Only `deviceQueryDrv` works for now.
- Increased LXD passthrough threshold to account for more systems
- Ensure `lxd` snap is installed for GPGPU tests

## Resolved issues

Resolves SERVCERT-1809

## Tests

[Submission on `hinyari`](https://certification.canonical.com/hardware/202409-35559/submission/414630/)

### DIFF
--- a/providers/gpgpu/bin/gpu_passthrough.py
+++ b/providers/gpgpu/bin/gpu_passthrough.py
@@ -57,7 +57,7 @@ This can be overwritten by the `--count` option or by setting the environment
 variable `LXD_GPU_RUNS`. With priority in that order.
 """
 
-GPU_THRESHOLD_SEC = 12.0
+GPU_THRESHOLD_SEC = 30
 """The threshold for the GPU test to pass.
 
 This can be overwritten by the `--threshold` option or by setting the
--- a/providers/gpgpu/bin/gpu_passthrough.py
+++ b/providers/gpgpu/bin/gpu_passthrough.py
@@ -57,7 +57,7 @@ This can be overwritten by the `--count` option or by setting the environment
 variable `LXD_GPU_RUNS`. With priority in that order.
 """
 
-GPU_THRESHOLD_SEC = 30
+GPU_THRESHOLD_SEC = 12.0
 """The threshold for the GPU test to pass.
 
 This can be overwritten by the `--threshold` option or by setting the
--- a/providers/gpgpu/debian/control
+++ b/providers/gpgpu/debian/control
@@ -16,7 +16,7 @@ Build-Depends: checkbox-ng,
 Standards-Version: 3.9.6
 
 Package: checkbox-provider-gpgpu
-Architecture: all
+Architecture: any-amd64 arm64
 Depends: checkbox-provider-base, fwts, ipmitool, ${plainbox:Depends}, ${misc:Depends}
 Suggests: ubuntu-drivers-common, ${plainbox:Suggests}
 X-Plainbox-Provider: yes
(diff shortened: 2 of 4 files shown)


providers/gpgpu/debian/control
```diff
@@ -16,15 +16,15 @@ Build-Depends: checkbox-ng,
 Standards-Version: 3.9.6
 
 Package: checkbox-provider-gpgpu
-Architecture: amd64
-Depends: checkbox-provider-base, ${plainbox:Depends}, ${misc:Depends}
+Architecture: all
```
Hook25, line 19: I don't really know but isn't there a way to say amd64,arm64?

providers/gpgpu/debian/control
```diff
@@ -16,15 +16,15 @@ Build-Depends: checkbox-ng,
...
+Depends: checkbox-provider-base, fwts, ipmitool, ${plainbox:Depends}, ${misc:Depends}
```
Hook25, line 20: Why are we adding these dependencies?

providers/gpgpu/bin/gpu_passthrough.py
```diff
@@ -57,7 +57,7 @@
 variable `LXD_GPU_RUNS`. With priority in that order.
 """
 
-GPU_THRESHOLD_SEC = 12.0
+GPU_THRESHOLD_SEC = 30
```
Hook25, line 60: Possibly followup material, but rather than increasing the global testiog expectation, this should probably read an envvar and default to 12, so that you can set it from the launcher!

Hook25 (approved):
lgtm, consider the comments below

providers/gpgpu/debian/control
pedro-avalos: When I set up hinyari and installed the gpgpu provider, these packages didn't get installed, and some of the jobs that the GPGPU launcher has require these packages

providers/gpgpu/bin/gpu_passthrough.py
pedro-avalos: Ah yeah, it already takes it in as an environment variable, so editing the launcher should work

providers/gpgpu/debian/control
pedro-avalos: I don't think so; the debian packaging guide didn't say you could add comma-separated values. It mentioned a wildcard syntax, but seemed quite complicated.

Hook25: (ty @nadzyah for [this](https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture))

Change it to `Architecture: any-amd64 arm64`

pedro-avalos: Done, that's one interesting syntax

providers/gpgpu/debian/control
Hook25: for ipmitool ok, I trust it is needed, but I dont get why fwts is needed, it is a depedency of base

pedro-avalos: I've removed these packages from the dependencies. They should be installed by the required base and certification-server providers (which are the providers that contain the tests that actually depend on these packages)

Hook25 (approved):

PR #1765, opened 2025-03-03
Remove workflow call check (infra)

Hook25:
## Description

This replaces the old condition (which doesn't work) with a new one, an input parameter. It is useful to have this workflow "triggerable" without pushing to the store so that is why I kept it.

Minor: Changed the order in the daily builds of the snap and deb build so that it is easier to read (snap first, debs then)


pieqq (approved):
LGTM

PR #1768, opened 2025-03-03
Fix slugify when str is empty/none (BugFix)

pedro-avalos:
## Description

Currently the `slugify` function assumes the `_string` argument is not empty/`None`. However, I saw this when trying to run some tests on `hinyari`. One of the device attributes was empty, but existed.

```
  File "/home/ubuntu/checkbox/providers/resource/bin/udev_resource.py", line 198, in <module>
    main()
  File "/home/ubuntu/checkbox/providers/resource/bin/udev_resource.py", line 194, in main
    dump_udev_db(udev)
  File "/home/ubuntu/checkbox/providers/resource/bin/udev_resource.py", line 76, in dump_udev_db
    value = getattr(device, attribute)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/checkbox_support/parsers/udevadm.py", line 933, in vendor_slug
    return slugify(self.vendor)
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/checkbox_support/helpers/slugify.py", line 33, in slugify
    if _string[0].isdigit():
```

## Tests

Added a unit test

### DIFF
--- a/checkbox-support/checkbox_support/helpers/slugify.py
+++ b/checkbox-support/checkbox_support/helpers/slugify.py
@@ -27,7 +27,7 @@ def slugify(_string):
     identifers.
     """
     if not _string:
-        return "_"
+        return _string
 
     valid_chars = frozenset(
         "_{}{}".format(string.ascii_letters, string.digits)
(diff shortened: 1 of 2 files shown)


Hook25:
Uhm this is a bit dangerous, `None` and `""` would expand to the same symbol that is not even `""`. The semantic here is not clear at all imo. What if we introduce a kwarg with a default value simmilar to `dict.get` for a default when the string is `None`? For `""` I would expect `""` to be the expansion, similarly to what str.capitalize does for example

Hook25 (approved):

PR #1769, opened 2025-03-04
Update security pipeline and permissions (infra)

Hook25:
## Description

Updating our security scanning tool zizmor to `1.4.1` I discovered a few new warnings that it finds. The following updates the tool and fixes all the warnings.

This may break a few pipelines, but we can fix them once we find them, the GH documentation about permissions is very bad and unclear about what each thing needs.

## Resolved issues

Fixes: CHECKBOX-1774

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,5 +1,6 @@
 name: Check formatting with Black
-
+permissions:
+  contents: read
 on:
   push:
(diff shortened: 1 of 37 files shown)


pieqq (approved):
This is pretty hard to test as is, let's land it and see what breaks, as you said.

PR #1770, opened 2025-03-04
Move to output parameter via expansion (infra)

Hook25:
## Description

The current implementation doesn't work because the upload artifact action doesn't support globs. This goes to the output paramter to the snap building action to get the same result that we wanted to get in the other case.

## Resolved issues

N/A

## Documentation

N/A

## Tests

See: https://github.com/canonical/checkbox/actions/runs/13658334143
See: https://github.com/canonical/checkbox/actions/runs/13658363158 (failures are due to the brownout)
Failing workflow: https://github.com/canonical/checkbox/actions/runs/13637347230/job/38119282521

### DIFF
--- a/.github/workflows/checkbox-daily-cross-builds.yaml
+++ b/.github/workflows/checkbox-daily-cross-builds.yaml
@@ -48,8 +48,9 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
 
-      - name: Build (retries on fail)
-        uses: Wandalen/wretry.action@master
+      - id: snap_build
+        name: Build (retries on fail)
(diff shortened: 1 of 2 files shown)


pedro-avalos (approved):
LGTM

pedro-avalos (changes requested):
Upload snap actions don't seem to be run in the linked examples, maybe this doesn't work? Or at least it's not fully tested. I see you're re-running one of them, though

pedro-avalos (approved):
After discussing, the snap publish action shouldn't be run for this testing. It should pass, if the upload artefact is anything to go by.

The failing/re-running workflows have now passed.

PR #1775, opened 2025-03-06
A test case for randomly generated input events (New)

GabrielChenCC:
## Description

We discovered this problem during last year's testing: https://bugs.launchpad.net/sutton/+bug/2045249 . We found this bug by manually coordinating the random use of the mouse and keyboard during screen rotation. Now I use this C code to implement the random automatic behavior of the mouse and keyboard. It can be used in conjunction with another PR https://github.com/canonical/checkbox/pull/1576 submitted by my colleague Hanhsuan to complete the random movement of the mouse and keyboard at different resolutions while rotating the screen.

Note: The keyboard and mouse random C code needs to execute via 'sudo'. We have also tried to use Python's uinput to avoid using sudo permissions, but by looking up the document https://python-evdev.readthedocs.io/en/latest/tutorial.html#injecting-input, Python cannot avoid sudo either, and C code is available for the rotation test case without introducing external dependence. So I choose C code, it is more suitable for the rotation test case.  

## Resolved issues

https://warthogs.atlassian.net/browse/OEMQA-5102 
https://bugs.launchpad.net/sutton/+bug/2045249

## Tests

$ providers/base/bin/mouse_keyboard.py
The demo file works as design

Run the test case mouse_keyboard.py without any errors


providers/base/bin/mouse_keyboard.py
```diff
@@ -0,0 +1,163 @@
...
+
+# Constants
+FREQUENCY_USEC = 100000  # Frequency of events in microseconds
+N_EPISODES = 81  # Number of events to generate
+WEIGHT_MOUSEMOVE = 10  # Weight of mouse movements
+WEIGHT_KEYPRESS = 1  # Weight of key presses
+WEIGHT_SUM = WEIGHT_MOUSEMOVE + WEIGHT_KEYPRESS  # Total weight
+
+MOVE_MAX = 100  # Maximum mouse movement distance
+MOVE_DELTA = 5  # Mouse movement step size
```
pieqq, line 142: These should go at the top of the script.

providers/base/bin/mouse_keyboard.py
pieqq, line 1: This script should be made executable (`chmod +x mouse_keyboard.py`).

providers/base/bin/mouse_keyboard.py
```diff
@@ -0,0 +1,163 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+"""mouse_key_random utility."""
+
+
+from evdev import UInput, ecodes as e
+import time
+import random
```
pieqq, line 26:
```suggestion
import time
import random
from evdev import UInput, ecodes as e
```

Importing stdlib stuff first.

providers/base/bin/mouse_keyboard.py
```diff
@@ -0,0 +1,163 @@
...
+    time.sleep(1)  # Give userspace time to read the remaining events
+    device.close()
+
+
+# Simulate a key press
+def key_press(device, key):
+    device.write(e.EV_KEY, key, 1)  # Press the key
+    device.syn()  # Synchronize the event
+    device.write(e.EV_KEY, key, 0)  # Release the key
+    device.syn()  # Synchronize the event
```
pieqq, line 96:
```suggestion
    device.write(e.EV_KEY, key, 1)  # Press the key
    device.write(e.EV_KEY, key, 0)  # Release the key
    device.syn()  # Synchronize the event
```

I don't think the first synchronize even is required, [the tutorial](https://python-evdev.readthedocs.io/en/latest/tutorial.html#injecting-input) shows one sync event is enough (and I've checked on my side).

pieqq (changes requested):
I made some comments in the code of the script.

Other than that, please also update the packaging for the snaps and the debian package recipes to pull `python3-evdev`. Hopefully this works on Xenial (I [just checked](https://packages.ubuntu.com/xenial/allpackages) and there is a `python3-evdev` package there, albeit a very old one, so fingers crossed!)

pieqq:
(side comment: I didn't really understand the connection between what's explained in the description about gnome-randr and this script, could you provide more information?)

GabrielChenCC:
> (side comment: I didn't really understand the connection between what's explained in the description about gnome-randr and this script, could you provide more information?)

Hi Pierre, I have updated the description and all story text. I guess it could fix your concern. Thanks.

GabrielChenCC:
> I checked and the `python3-evdev` package should be available everywhere we need it, so it's good. That said, you need to add the packaging information for Debian and Snaps (for all the series. See for instance [the snapcraft.yaml for series24](https://github.com/canonical/checkbox/blob/main/checkbox-core-snap/series24/snap/snapcraft.yaml); you would probably need to add this dependency in the [`stage-packages` section of the base provider](https://github.com/canonical/checkbox/blob/main/checkbox-core-snap/series24/snap/snapcraft.yaml#L256)).
> 
> I saw that in [7031c05](https://github.com/canonical/checkbox/commit/7031c059e23858cdde5c9df0cc9e9225a04739d1) you added this for Debian, but then you removed it in [b9b3606](https://github.com/canonical/checkbox/commit/b9b360639dc36dd474a078fe833c7d23f7f361c4). Not sure why?
I saw this dependency had been introduced by other test cases before, so It was not needed there.

> Other than that, I tried running the script with `$ ./mouse_keyboard.py` on my 24.04 laptop. Out of 5 runs, it failed twice with the following error:
> 
> ```
>   File "/home/pieq/dev/work/checkbox/providers/base/bin/./mouse_keyboard.py", line 159, in <module>
>     main()
>   File "/home/pieq/dev/work/checkbox/providers/base/bin/./mouse_keyboard.py", line 150, in main
>     rand_mouse_moves(device)  # Simulate mouse movement
>     ^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/pieq/dev/work/checkbox/providers/base/bin/./mouse_keyboard.py", line 134, in rand_mouse_moves
>     rest_x = x % steps
>              ~~^~~~~~~
> ZeroDivisionError: integer modulo by zero
> ```
> 
> Can you look into that?
Fixed in the latest commit.

pieqq (approved):
Thanks!

PR #1777, opened 2025-03-06
Restore netplan config even when exception is raised (bugfix)

### LINKED ISSUE #1674: if `generate_test_config` failed, `wifi_client_test_netplan.py` won't restore config file
### Bug Description

The [generate_test_config](https://github.com/canonical/checkbox/blob/21cd1c69544668af0aa528d9a308ec28829dd1bf/providers/base/bin/wifi_client_test_netplan.py#L150) will raise SystemExit while missing ssid. Then the [main](https://github.com/canonical/checkbox/blob/21cd1c69544668af0aa528d9a308ec28829dd1bf/providers/base/bin/wifi_client_test_netplan.py#L465) will stop without restoring config.

This will remove files under `/etc/netplan`, `/lib/netplan`, `/run/netplan` that defined in [here](https://github.com/canonical/checkbox/blob/a233067674b5c257bcc8b2075464fd8bf39f8a63/providers/base/bin/wifi_client_test_netplan.py#L50) then make the debsum check fail.

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

1. run wifi test without setting SSID

### Expected Result

It should restore all config files while executing the script failed.

### Actual Result

script won't restore config file if `generate_test_config` failed.

### Environment

- OS: ubuntu 24.04
- checkbox type: debian
- checkbox version: 4.3.0.dev36

### Relevant log output

```shell
Interface wlp0s20f3 using module iwlwifi
  Parameters:
    11n_disable: 0
    power_save: N
    swcrypto: 0
    power_level: 0
    amsdu_size: 0
    uapsd_disable: 3
    fw_restart: Y
    disable_11be: N
    led_mode: 0
    disable_11ax: N
    enable_ini: 16
    disable_11ac: N
    bt_coex_active: Y
    remove_when_gone: N
    nvm_file: (null)

Checking kernel ring buffer for iwlwifi messages:


## Backup any existing netplan configuration files

## Delete any existing netplan configuration files

## Generate a test netplan configuration
A SSID is required for the test
```

### Additional context

_No response_

pieqq:
## Description

Use context managers to handle original netplan config backup and restore, as well as the test config writing and removing.

Modify function `netplan_apply_config` to raise SystemExit if the
subprocess call to netplan failed for any reason.

- journal entries are still printed out to the log if there is any
SystemExit
- test config removal and original config restoration are run regardless
of what happens

## Resolved issues

Fix #1674

https://warthogs.atlassian.net/browse/CHECKBOX-1718

## Tests

unit tests updated

### DIFF
--- a/providers/base/bin/wifi_client_test_netplan.py
+++ b/providers/base/bin/wifi_client_test_netplan.py
@@ -218,7 +218,10 @@ def write_test_config(config):
 
 def delete_test_config():
     print_head("Delete the test file")
-    os.remove(NETPLAN_TEST_CFG)
+    try:
+        os.remove(NETPLAN_TEST_CFG)
+    except FileNotFoundError:
+        print("Test config {} not found, ignoring...".format(NETPLAN_TEST_CFG))
     print()
 
 
--- a/providers/base/bin/wifi_client_test_netplan.py
+++ b/providers/base/bin/wifi_client_test_netplan.py
@@ -460,6 +460,7 @@ def handle_original_np_config():
     try:
         yield
     finally:
+        delete_test_config()
         netplan_config_restore()
         netplan_apply_config()
 
--- a/providers/base/bin/wifi_client_test_netplan.py
+++ b/providers/base/bin/wifi_client_test_netplan.py
@@ -25,6 +25,8 @@ import sys
 import ipaddress
 import yaml
 
+from contextlib import contextmanager
+
 from gateway_ping_test import ping
 from checkbox_support.snap_utils.system import get_series
 
@@ -235,6 +237,7 @@ def netplan_apply_config():
     env.pop("PYTHONUSERBASE", None)
     retcode = sp.call(cmd, shell=True, env=env)
     if retcode != 0:
+        print_journal_entries(start_time, renderer)
         raise SystemExit("ERROR: failed netplan apply call")
     print()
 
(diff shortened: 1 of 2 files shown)


hanhsuan:
Tested on [202412-36142](https://certification.canonical.com/hardware/202412-36142/submission/415840/), the config of netplan is deleted if there is no checkbox.conf.

pieqq:
After adding the fix in commit 9711a32, the execution [seems to run correctly](https://certification.canonical.com/hardware/202412-36142/submission/415873/test-results/?term=_np_), that is, if SSID is omitted, the test will fail with an error message ("A SSID is required for the test"), but the config will be restored.

Hook25 (changes requested):
This doesn't look very good to me ngl. Catching BaseExceptions like SystemExit is in general not great. My best solution if I had to choose is to make the config a context manager, that would be super gorgeous 
```python
with generate_test_config(...) as config_data:
    ...
```

If you really don't want to do that, you can always use `atexit`, but I would advise you against that, it is not super clean as it introduces a global side effect to closing the application which is always a bit yucky if you can avoid it

pieqq:
I tested my latest changes on a device, and the tests fail without an SSID, but the config files are restored:

```
==============[ Running job 1 / 1. Estimated time left: 0:00:15 ]===============
----[ Connect to unencrypted 802.11be Wi-Fi network on wlp0s20f3 - netplan ]----
ID: com.canonical.certification::wireless/wireless_connection_open_be_np_wlp0s20f3
Category: com.canonical.plainbox::wireless
Interface wlp0s20f3 using module iwlwifi
(...)
## Backup any existing netplan configuration files
Clear backup location
Backing up from /etc/netplan
  /etc/netplan/90-NM-21dbe869-a81c-481b-9c30-5dd25bda5ad7.yaml
Backing up from /etc/netplan
  /etc/netplan/01-network-manager-all.yaml
Backing up from /lib/netplan
  /lib/netplan/00-network-manager-all.yaml

## Delete any existing netplan configuration files
  /etc/netplan/90-NM-21dbe869-a81c-481b-9c30-5dd25bda5ad7.yaml
  /etc/netplan/01-network-manager-all.yaml
  /lib/netplan/00-network-manager-all.yaml

## Generate a test netplan configuration
## Restore configuration files
Restoring:
  /etc/netplan/90-NM-21dbe869-a81c-481b-9c30-5dd25bda5ad7.yaml
  /etc/netplan/01-network-manager-all.yaml
  /lib/netplan/00-network-manager-all.yaml
+ netplan --debug apply

A SSID is required for the test
Outcome: job failed
```

```
$ ll /*/netplan/
/etc/netplan/:
total 24
drwxr-xr-x   2 root root  4096 Mar 26 17:10 ./
drwxr-xr-x 139 root root 12288 Mar 26 16:58 ../
-rw-r--r--   1 root root   104 Mar 26 17:10 01-network-manager-all.yaml
-rw-------   1 root root   727 Mar 26 17:10 90-NM-21dbe869-a81c-481b-9c30-5dd25bda5ad7.yaml

/lib/netplan/:
total 16
drwxr-xr-x  2 root root 4096 Mar 26 17:10 ./
drwxr-xr-x 97 root root 4096 Mar 26 16:58 ../
-rw-------  1 root root  104 Mar 26 17:10 00-network-manager-all.yaml
-rw-r--r--  1 root root  204 Aug 14  2024 PLACEHOLDER

```

Hook25 (approved):
+1, LGTM

Issue #1779, opened 2025-03-07
Manifest selection is set to "No" automatically


LiaoU3:
### Bug Description

When running a test, if we did not specify the manifest in advanced (e.g. running without a launcher with config variables), all the selections are set to `No` on the page of manifest.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. checkbox.checkbox-cli control {ANY_DUT_IP}
2. Go to the manifest selection page

### Expected Result

The selections should be empty

### Actual Result

The selections are filled `No` automatically.

### Environment

 - Checkbox Type: Snap
 - Checkbox Version: 4.3.0-dev71

### Relevant log output

```shell

```

PR #1782, opened 2025-03-07
Fix manifest always no (bugfix)

Hook25:
## Description

The default value in the selector unexpectedly changed due to `None` and `""` mixing up in the implementation. This changes the code so that `None` is always used for undefined variables in the browser/manifest repr getter.

## Resolved issues

Fixes: CHECKBOX-1786
Fixes: https://github.com/canonical/checkbox/issues/1779

## Documentation

N/A

## Tests

This improves the tests as well as it encodes the function expectations in the unit itself

### DIFF
--- a/checkbox-ng/plainbox/impl/session/assistant.py
+++ b/checkbox-ng/plainbox/impl/session/assistant.py
@@ -1380,14 +1380,14 @@ class SessionAssistant:
         for m in manifest_list:
             prompt = m.prompt() or m.default_prompt()
             value = config_manifest.get(m.id)
-            if m.is_hidden:
-                # set all hidden manifests not set in the launcher to default
-                value = value or m.default_value()
-            else:
-                # only load from the disk_manifest values of non-hidden manifests
-                value = value or disk_manifest.get(m.id, "")
-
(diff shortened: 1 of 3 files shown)


checkbox-ng/plainbox/impl/session/assistant.py
```diff
@@ -1380,14 +1380,14 @@ def didnt_run_yet(job):
...
-
-            if value:
+            if value is None:
+                if m.is_hidden:
+                    # set all hidden manifests not set in the launcher to default
+                    value = m.default_value()
+                else:
+                    # only load from the disk_manifest values of non-hidden manifests
+                    value = disk_manifest.get(m.id)
+            if value is not None:
```
pieqq, line 1390: Why not:

```suggestion
            else:
```

?

pieqq (approved):
Nice, thanks for that quick fix!

checkbox-ng/plainbox/impl/session/assistant.py
Hook25: because we modify value inside the if and it may be/become None

PR #1783, opened 2025-03-07
various - update platform constraints on some jobs to not run on either Power arch (ppc64le and ppc64el) (BugFix)

### LINKED ISSUE #879: Need to account for both ppc6le and ppc64el cpu platform names when setting constraints using cpu.platform
### Bug Description

On a recent Power 10 test, I discovered a test had been skipped because the cpu.platform constraint was not met.  This was incorrect as it SHOULD have matched and triggered the test to run.:

`job cannot be started: resource expression 'cpuinfo.platform in ("ppc64el")' evaluates to false`

Looking at the code this is used as a constraint a few times.  According to this: https://www.reddit.com/r/qemu_kvm/comments/b6d91v/ppc64el_vs_ppc64le/

they are the same, but it's a debianism to use ppc64el, while it appears IBM has gone with ppc64le on Power systems.  I think, in the past on older Power Hardware, ppc64le worked. So to be sure, lets just use both.

This raises the question, can constraints be substring matched?  in otherwords, instead of:

`cpu.platofrm in ["ppc64el", "ppc64le"] `

can we use 

`cpu.platform = "ppc64*"`

to match either of the two?

### To Reproduce

1. review this job failure to see the constraint is not met
    https://certification.canonical.com/hardware/202312-32378/submission/347345/test/76846/result/38196711/
2. review the platform info to see that the CPU platform is not the one we've used.
    https://certification.canonical.com/hardware/202312-32378/

### Environment

Power 10 LPAR is needed. will have to work with Patricia Domingues  in PE to test the fixes.

### Relevant log output

_No response_

### Additional context

_No response_

bladernr:
## Description
For jobs in checkbox-provider-base that are meant to run, or not run, on Power arch systesm, ensure that the constraint specifies both ppc64le and ppc64el to ensure the jobs don't run on known Power systems.

(had to add thius apparently)

## Resolved issues

Fixes SERVERCERT-253
Fixes #879 

## Documentation

NA - just adds consistency in power arch constraints

### DIFF
--- a/providers/base/units/cpu/jobs.pxu
+++ b/providers/base/units/cpu/jobs.pxu
@@ -5,7 +5,7 @@ estimated_duration: 150.0
 requires:
  executable.name == 'fwts'
  'userspace' in cpuinfo.governors
- cpuinfo.platform not in ("ppc64el", "s390x")
+ cpuinfo.platform not in ("ppc64el", "ppc64le", "s390x")
 user: root
 environ: PLAINBOX_SESSION_SHARE LD_LIBRARY_PATH SNAP
 command:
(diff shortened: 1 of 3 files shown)


Hook25 (approved):
+1, LGTM, you had to add the sru notice because some of these jobs are in the sru testplan!

PR #1788, opened 2025-03-11
Add new collectors and remove fetch-on-checkpoint mechanism (New)

Hook25:
## Description

It is often useful to analyze the logs after a session is done to see why something went south. In order to do this, this PR adds a journal and a dmesg collector to system information. 

This PR also moves the collection to the end of the session. I was about to create some complex machinery to run some collectors at the start of a run and some at the end, but I talked myself out of it. The collection was placed where it was placed because we thought we could use it during the run somehow. This idea went out of the window and all that is remaining now is the cost of running the collection at the start of the run, which is about 3s just to run inxi on my machine, for any session that is started, even if no submission is ever generated. I don't think the value we get (which is nothing right now) justifies this cost. If we will ever want to do something with that data during the run we can always revert (partially) this commit and tag the collectors/create functions to call them at finalize and first checkpoint, for now, I think we should do the easy thing and just collect at the end.

## Resolved issues

Fixes: CHECKBOX-1753

## Documentation

N/A

## Tests

Tried this locally, in general I don't expect this to cause major issues as collectors can safely fail

### DIFF
--- a/checkbox-ng/plainbox/impl/session/resume.py
+++ b/checkbox-ng/plainbox/impl/session/resume.py
@@ -1268,29 +1268,6 @@ class SessionResumeHelper7(MetaDataHelper7MixIn, SessionResumeHelper6):
     pass
 
 
-class SessionResumeHelper8(SessionResumeHelper7):
-    def _restore_SessionState_system_information(
-        self, session_state, session_repr
-    ):
-        _validate(session_repr, key="system_information", value_type=dict)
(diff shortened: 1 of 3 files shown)


pieqq (dismissed):

pieqq (approved):
Glad we caught the weird JSON formatting of journalctl for each line in the journal!

+1

pieqq (approved):

PR #1790, opened 2025-03-12
[checkbox-ce-oem] Add a test for splash screen (New)

vandah:
## Description
This PR adds a manual test for checking if a splash screen appears on boot. To do this, the device is first rebooted and the result is recorded upon resuming the Checkbox session. The rebooting method is the command from `power-management/warm-reboot`.

## Tests
- tested on a Raspberry Pi 4 Model B with UC22


vandah:
Please review, @alexclewontin.

PR #1794, opened 2025-03-14
Os release should always provide codename (bugfix)

Hook25:
## Description

This bugfix didn't take into consideration that ubuntu core doesn't have a codename. This patch provides backward compatibility by using the description (which is the pretty name of the os) as a codename (which is indeed proper for core).

## Resolved issues

Fixes: CHECKBOX-1793

## Documentation

Added a comment

## Tests

Added a test

### DIFF
--- a/providers/resource/bin/os_resource.py
+++ b/providers/resource/bin/os_resource.py
@@ -38,13 +38,18 @@ def get_release_info(release_file_content: str):
         "VERSION_CODENAME": "codename",
     }
     os_release = {}
-    for line in release_file_content.strip().splitlines():
-        if line:
-            (key, value) = line.split("=", 1)
-            if key in os_release_map:
-                k = os_release_map[key]
-                # Strip out quotes and newlines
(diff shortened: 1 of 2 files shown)


pieqq (approved):
Looks good to me. Thanks!

rickwu666666:
I've try this change with my UC24 system. And it worked for me. This PR is quite critical for the IoT QA who is test with Ubuntu Core. Since the current checkbox in beta and edge track will get the invalid submission that will block submission from upload to C3.
The invalid submission contact the distribution field as follow which is missing "codename"
```
 "distribution": {
        "description": "Ubuntu Core 24",
        "distributor_id": "Ubuntu Core",
        "release": "24"
    },
```
After sideload this chage, we can get something like this.
```
"distribution": {
        "codename": "Ubuntu Core 24",
        "description": "Ubuntu Core 24",
        "distributor_id": "Ubuntu Core",
        "release": "24"
    },
```

PR #1795, opened 2025-03-14
providers/base : add tests for Intel QAT

hector-cao:
QAT (Intel QuickAssist Technology) is an accelerator for crypto/compression operations. The hardward is available on recent Xeon processors.


bladernr:
I made a couple inline comments... I'm interested in including this in the server suite.  Mostly my questions are around the dependency on the detect job, which simply checks that someone ticked Y on a manifest entry (server suite doesn't use manifests, it uses resources).  

Would it be reasonable for the detect job to pass if either condition is there (e.g. the resource job returns a `pf: present` or whatever if the QAT PFs are detected by qatctl.py, OR if the manifest is true)?

and add that extra bit to the resource similar to other resource constraints in use? so... something like this for the detect job:

`requires: manifest.has_intel_qat == 'True' or pf.present == 'True'`

And keep the rest of it as-is with the dependency on the detect job?

hector-cao:
> I made a couple inline comments... I'm interested in including this in the server suite. Mostly my questions are around the dependency on the detect job, which simply checks that someone ticked Y on a manifest entry (server suite doesn't use manifests, it uses resources).
> 
> Would it be reasonable for the detect job to pass if either condition is there (e.g. the resource job returns a `pf: present` or whatever if the QAT PFs are detected by qatctl.py, OR if the manifest is true)?
> 
> and add that extra bit to the resource similar to other resource constraints in use? so... something like this for the detect job:
> 
> `requires: manifest.has_intel_qat == 'True' or pf.present == 'True'`
> 
> And keep the rest of it as-is with the dependency on the detect job?

Thanks @bladernr for your feedback !

Based on your comments, I re-designed the tests, please take a look and let me know things you would want to be improved.

providers/base/units/intel-qat/jobs.pxu
```diff
@@ -0,0 +1,254 @@
...
+  there is no VFIO files for VFs and when we load the vfio-pci
+  each VF will have it own character file in /dev/vfio/
+command:
+  rmmod vfio-pci || true
+  nb_vfio=$(qatctl.py status --devices {{ pf }} --vfio | wc -l)
+  [ "$nb_vfio" -le 0 ] || (echo "nb vfio devices should be <= 0" && exit 1)
+  # we have to pass the VF device ids
+  modprobe vfio-pci ids=8086:4941,8086:4943,8086:4945,8086:4947
+  nb_vfio=$(qatctl.py status --devices {{ pf }} --vfio | wc -l)
+  [ "$nb_vfio" -gt 0 ] || (echo "nb vfio devices should be > 0" && exit 1)
```
pieqq, line 110: Several suggestions (some apply to other jobs in this PR too):

1. Put these into a separate bash script in `bin/`, and start it with `set -e` so that it fails as soon as possible if anything goes wrong.
2. a `check` flag could be added to the `qatctl.py` script to avoid relying on additional bash commands (such as `[ "$nb_vfio" -le 0 ] || (echo "nb vfio devices should be <= 0" && exit 1)`)
3. maybe this test could be split in 2 (the second test would depend on the first): 
  - check that no VFIO files are present if the `vfio-pci` module is removed
  - check that VFIO are there when the module is reloaded
4. if something goes wrong before you reload the `vfio-pci` module, all the tests running after will fail, so you probably have to make sure the modules are reloaded regardless of the outcome.

providers/base/units/intel-qat/resource.pxu
```diff
@@ -0,0 +1,65 @@
+# Copyright 2025 Canonical Ltd.
+# All rights reserved.
+#
+# Written by:
+#   Hector Cao <hector.cao@canonical.com>
+
+unit: job
+id: qat_pf
```
pieqq, line 8: This resource job looks like it's doing exactly the same thing as `qat` below, except it doesn't add the `available` field. Consider removing this and rely on `qat` only.

pieqq (changes requested):
Sorry for the long time to provide feedback! Please check my comments to see if they make sense.

providers/base/units/intel-qat/jobs.pxu
```diff
@@ -0,0 +1,254 @@
...
+template-engine: jinja2
+template-unit: job
+id: intel-qat-qatlib/{{ available }}-cpa-sample-code-sym
+category_id: intel-qat
+flags: simple
+user: root
+_summary: Run CPA symmetric crypto tests
+requires:
+  package.name == 'qatlib-examples'
+  package.name == 'qatlib-service'
```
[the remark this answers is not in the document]
hector-cao: These packages are available in the archive (universe) from Ubuntu 24.04 onwards

providers/base/units/intel-qat/jobs.pxu
```diff
@@ -0,0 +1,254 @@
...
+command:
+  # TODO : for unknown reason, cpa_sample_code still runs when we cancel the
+  # checkbox session (ctrl-c)
+  # 1  : symmetric crypto tests only
+  # 2  : RSA crypto tests only
+  # 32 : Compression tests only
+  # switch all devices to crypto sym mode
+  printf "POLICY=0\nServicesEnabled=sym\n" | tee /etc/sysconfig/qat
+  systemctl restart qat
+  cpa_sample_code runTests=1
```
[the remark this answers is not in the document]
hector-cao: This tool is part of the `qatlib-examples` package

providers/base/units/intel-qat/jobs.pxu
```diff
@@ -0,0 +1,254 @@
...
+  if [ "${PFS}" -le 0 ]; then
+    echo "manifest.has_intel_qat is set to True but no device found !"
+    exit 1
+  fi
+
+unit: template
+template-resource: qat
+template-engine: jinja2
+template-unit: job
+id: intel-qat-common/{{ available }}-attach-devices
```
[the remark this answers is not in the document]
hector-cao: Thanks ! you are right but I tend to use jinja as the default every time I have to deal with template job, I do not see any strong reason to switch to a less powerful template engine, can we consider using it even if it is not mandatory right now ?

providers/base/units/intel-qat/resource.pxu
hector-cao: it is quite subtle but the 2 resources are not doing the samething, the `qat` resource only output the first QAT PF (physical device) (see the break in the control loop)

providers/base/units/intel-qat/jobs.pxu
pieqq: 1. It's up to you. However, adding `set -e` is required, otherwise you may run things after failed commands, which is probably not desired.
3. You're right, it's probably better to keep it coupled then.

hector-cao:
> @hector-cao can you check the tox error?
> 
> ```
> FAIL: test_qat_dev_manager_list_devices (test_qatctl.TestQatDevManager)
> ----------------------------------------------------------------------
>   File "/home/ubuntu/actions-runner/_work/_tool/Python/3.6.15/x64/lib/python3.6/unittest/mock.py", line 1183, in patched
>     return func(*args, **keywargs)
>   File "/home/ubuntu/actions-runner/_work/checkbox/checkbox/providers/base/tests/test_qatctl.py", line 388, in test_qat_dev_manager_list_devices
>     self.assertIn("00:00.0 4xxx", printed_lines)
> AssertionError: '00:00.0 4xxx' not found in ['args', 'args', 'args', 'args']
> ```

Sorry, I missed that, let me take a look now

hector-cao:
> @hector-cao are you still planning to own this?

Hello, I am struggling to have time to wrap this up, do you think you can help me on it ?

Issue #1792, opened 2025-03-13
CAN FD test cases require manifest


LiaoU3:
### Bug Description

The test cases
 - [socketcan/send_packet_local_fd_virtual](https://github.com/canonical/checkbox/blob/5baa4d0f5159ebf7fe0c14d15dd1057c270ad722/providers/base/units/socketcan/jobs.pxu#L59C5-L59C43)
 - [socketcan/send_packet_local_fd_{interface}](https://github.com/canonical/checkbox/blob/5baa4d0f5159ebf7fe0c14d15dd1057c270ad722/providers/base/units/socketcan/jobs.pxu#L132)
 - [socketcan/send_packet_remote_fd_{interface}](https://github.com/canonical/checkbox/blob/5baa4d0f5159ebf7fe0c14d15dd1057c270ad722/providers/base/units/socketcan/jobs.pxu#L214)

should add manifest like [has_socket_can_fd](https://github.com/canonical/checkbox/blob/5baa4d0f5159ebf7fe0c14d15dd1057c270ad722/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/socketcan/manifest.pxu#L2) to prevent running on fd-not-supported platforms

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Run socketcan tests on FD not supported platforms

### Expected Result

1. Should be skipped

### Actual Result

Still run

### Environment

The environment does not matter

### Relevant log output

```shell
---------------[ CAN device support test can0 (Raw, Remote, FD) ]---------------
ID: com.canonical.certification::socketcan/send_packet_remote_fd_can0
Category: SocketCAN interface tests
--------------------------------------------------------------------------------
+ ip link set can0 down
+ ip link set can0 type can bitrate 1000000 dbitrate 2000000 fd on
Error: bitrate error: 1.9%.

```

LiaoU3:
Assigned to myself, since I will propse a fix in the near future.

PR #1797, opened 2025-03-17
Add has_socket_can_fd for all can fd test cases (BugFix)

LiaoU3:
## Description
Due to the fact that FD mode is not supported on all platforms (e.g. [KD240](https://certification.canonical.com/hardware/202307-31904/)), and currently can fd test cases do not have manifest for it, we need to add a manifest for it. On top of that, we have `has_socket_can_fd` in `checkbox-ce-oem` already, so we could simply move them from checkbox-ce-oem to generic checkbox.

## Resolved issues
#1792 

## Documentation
N/A

## Tests
https://certification.canonical.com/hardware/202307-31904/submission/417423/test-results/


pieqq (approved):
+1, thanks!

Issue #1801, opened 2025-03-18
Precheck script complains if a USB2 device is not present, even if a USB3 device is present


rodwsmith:
### Bug Description

Changes to the server certification rules quite some time ago permit accepting submissions that have USB3 devices even if USB2 devices are not present. (Previously, both devices had to be present, except on systems with no USB3 support.) The canonical-certification-precheck script still enforces the old rules and should be updated.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Run canonical-certification-precheck on a system with a USB3 drive plugged in but no USB2 device.

### Expected Result

The output should not complain about the missing USB2 device.

### Actual Result

The summary complains that the USB testing has failed.

### Environment

Any system that supports both USB2 and USB3 devices

### Relevant log output

```shell

```

rodwsmith:
PR submitted, accepted, and merged; issue is now closed.

rodwsmith:
Close issue

PR #1802, opened 2025-03-18
Update canonical-certiciation-precheck for new USB device rules (bugfix)

rodwsmith:
Title: Updated canonical-certification-precheck script to accept system with USB3 disk but no USB2 disk (bugfix)

## Description

This PR adjusts the logic of the USB tests to flag a lack of a USB2 device only if the system also lacks a USB3 device. If both devices are present, the script still checks for suitable filesystems on both devices, since presumably the tester wants to test both devices in this case.

## Resolved issues

https://github.com/canonical/checkbox/issues/1801

## Documentation

No documentation changes are needed. The Server Self-Test Guide doesn't delve deep enough into this issue to need changing, 

## Tests

I deployed a node and tested with all combinations of USB2 and USB3 drives plugged in and removed; and with the drives correctly partitioned and not. The script passed the configurations in which the USB2 drive was removed when the USB3 drive was present and complained (correctly) when the USB3 drive was unplugged or when either drive was plugged in but had no valid partition.

### DIFF
--- a/providers/certification-server/tools/canonical-certification-precheck
+++ b/providers/certification-server/tools/canonical-certification-precheck
@@ -366,8 +366,8 @@ for disk in $(lsblk -r | grep disk | cut -f 1 -d " "); do
         fi
     fi
 done
-if [ -z "$usb2_disks" ]; then
-    echo " No USB2 disks detected."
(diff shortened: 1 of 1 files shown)


bladernr (approved):
LGTM...

PR #1809, opened 2025-03-19
Pin actions (infra)

Hook25:
## Description

As a response to the recent breach in one commonly used action we have decided to pin all of our actions to their revision, in order to be less exposed to this kind of vulnerability in the future

## Resolved issues

Fixes: CHECKBOX-1796

## Documentation

N/A

## Tests

N/A


p-gentili (approved):
All non-local actions are pinned, as expected. Dependabot next?

PR #1815, opened 2025-03-21
Push uc24 to the latest/edge channel (Infra)

LiaoU3:
## Description
Push uc24 to the latest/edge channel instead of classic24.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A

### DIFF
--- a/.github/workflows/checkbox-ce-oem-edge-builds.yml
+++ b/.github/workflows/checkbox-ce-oem-edge-builds.yml
@@ -82,12 +82,13 @@ jobs:
             do \
               echo "Uploading $snap..." ; \
               if [ ${{ matrix.type }} = 'classic' ]; then \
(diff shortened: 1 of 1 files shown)


Hook25:
Should this be done for the mainline checkbox?

Also: consider copy pasting the new workflow submit step, it is so much better in my opinion

Hook25 (approved):

LiaoU3:
> Should this be done for the mainline checkbox?

No, I dont think so. TBH, I should retrieve this PR, since it did not make any difference to our testing and other project checkbox.

>Also: consider copy pasting the new workflow submit step, it is so much better in my opinion

What do you mean by *copy pasting the new workflow submit step*?

Hook25:
well everything is devmode here, so the only thing that changes is that now you can use latest on core

PR #1816, opened 2025-03-21
Use the js action instead (infra)

Hook25:
## Description

I used the wrong commit as the js action one works, while the other seems to fail to propagate the output making our daily fail.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1814

## Documentation

N/A

## Tests

Failing tonight: https://github.com/canonical/checkbox/actions/runs/13976225083/job/39164709794
Built now and passing: https://github.com/canonical/checkbox/actions/runs/13994194106


p-gentili (approved):
For context: they have two branches `master` and `js_action`. Of  course we care about the latter, which is tagged differently.

Issue #1461, opened 2024-09-09
The iwlwifi_microcode_crash test only checks the journal log for the target kernel that is under test.


kevinyehk:
### Enhancement Proposal

The `iwlwifi_microcode_crash` checks all the journal logs on the DUT, but when the DUT is noprovision, it might cause false alarm, as the microcode carsh event could have occurred in previous kernel.  
Therefore, I think we should modify the test to only check the target kernel/snap that is under test.

anthonywong:
Can we stop reporting test failures like https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2077537 until this bug is fixed?

KaiChuan-Hsieh:
@kevinyehk Do you think clear old journal log before running the SRU test plan can avoid the situation?

kevinyehk:
@KaiChuan-Hsieh yes, clear old log in the beginning of entire test session should be OK.

boukeas:
I vacuumed the journal on `hp-elitebook-650-156-inch-g10-c31123`, which is a `noprovision` device, removing anything older than two days old, and reran the SRU test in which only the `iwlwifi_microcode_crash` test was [previously failing](https://certification.canonical.com/hardware/202301-31123/submission/393701/test-results/fail/). I can confirm [there were no failures](https://certification.canonical.com/hardware/202301-31123/submission/396673/).

kevinyehk:
if we check the kernel version at the first line in the log matches the one running now instead of cleaning all the journal log. maybe it's a better way? 
```
    bootidx=$(echo "$line" | cut -d " " -f1)
    kernel=$(journalctl -k -b "$bootidx" | grep -oP 'Linux version \d+\.\d+\.\d+-\d+-.*' | cut -d " " -f 3 )
    if [ "$kernel" = "$(uname -r)" ]; then
            echo "checking microcode error...."
            if journalctl -k -b "$bootidx" | grep -q "Microcode SW error detected"; then
                    echo "Boot $line, Microcode SW error detected"
                    exit 1
            fi
    else
            echo "skip...."
    fi
```

PR #1820, opened 2025-03-25
Check kernel version is the same when looking for iwlwifi microcode errors (bugfix)

pieqq:
## Description

As suggested by Kevin in #1461 , a check is added to make sure we are parsing journalctl logs from a boot that uses the same version of the kernel.

In addition, the script has been rewritten in Python to add unit tests.

## Resolved issues

#1461 

## Tests

- Unit tests
- Tested locally on 24.04.


providers/base/bin/check_iwlwifi_microcode_errors.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_boot_ids():
+    cmd = ["journalctl", "--list-boots", "--output", "json"]
+    boots = json.loads(subprocess.check_output(cmd, universal_newlines=True))
+    boot_ids = [boot["boot_id"] for boot in boots]
+    return boot_ids
+
+
+def get_kernel_version_from_journal(boot_id):
+    # Capture first line of journal (info about booted Linux kernel)
```
Hook25, line 16: Are you sure this is 100% the case? Every time the first line is the kernel version log line? Where was this documented?

providers/base/bin/check_iwlwifi_microcode_errors.py
```diff
@@ -0,0 +1,51 @@
...
+    cmd = ["journalctl", "-k", "-b", boot_id, "-n", "+1", "--output", "json"]
+    journalctl_log = json.loads(
+        subprocess.check_output(cmd, universal_newlines=True)
+    )
+    linux_line = journalctl_log["MESSAGE"]
+    # "Linux version 6.11.0-19-generic ..."
+    linux_version = linux_line.split()[2]
+    return linux_version
```
Hook25, line 24: This looks a bit dangerous, can you at least check that `Linux version` is in the string?

providers/base/bin/check_iwlwifi_microcode_errors.py
```diff
@@ -0,0 +1,51 @@
...
+    boot_ids = get_boot_ids()
+    for boot_id in boot_ids:
+        linux_version = get_kernel_version_from_journal(boot_id)
+        if linux_version == current_linux_version:
+            check_error(boot_id)
+    print("No microcode software errors detected.")
+
+
+if __name__ == "__main__":
+    current_linux_version = os.uname().release
```
Hook25, line 50: Why do you do it this way? does it work 100% of the times? You could just use the last boot id kernel's version (in the code above, go in reverse order, set it the first you encounter and then compare it backward!)

providers/base/tests/test_check_iwlwifi_microcode_errors.py
```diff
@@ -0,0 +1,82 @@
...
+            "(buildd@lcy02-amd64-014) (x86_64-linux-gnu-gcc-13 "
+            "(Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for "
+            "Ubuntu) 2.42) #19~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Feb 17 "
+            '11:51:52 UTC 2 (Ubuntu 6.11.0-19.19~24.04.1-generic 6.11.11)",'
+            '"_BOOT_ID":"1","__MONOTONIC_TIMESTAMP":"1","__CURSOR":"test",'
+            '"_TRANSPORT":"kernel","SYSLOG_IDENTIFIER":"kernel",'
+            '"__SEQNUM_ID":"1","_RUNTIME_SCOPE":"system",'
+            '"SYSLOG_FACILITY":"0","__REALTIME_TIMESTAMP":"1",'
+            '"_HOSTNAME":"test","__SEQNUM":"1"}\n'
+        )
```
Hook25, line 38: What about json.dumps the object? it is unreadable this way!

providers/base/bin/check_iwlwifi_microcode_errors.py
pieqq: implemented in https://github.com/canonical/checkbox/pull/1820/commits/a479de108d4129ee997768576219797b92b11a53

providers/base/bin/check_iwlwifi_microcode_errors.py
pieqq: implemented in https://github.com/canonical/checkbox/pull/1820/commits/a479de108d4129ee997768576219797b92b11a53

providers/base/tests/test_check_iwlwifi_microcode_errors.py
pieqq: implemented in https://github.com/canonical/checkbox/pull/1820/commits/a479de108d4129ee997768576219797b92b11a53

providers/base/bin/check_iwlwifi_microcode_errors.py
pieqq: Indeed, it's not documented so I should not rely on this. With https://github.com/canonical/checkbox/pull/1820/commits/a479de108d4129ee997768576219797b92b11a53 I now parse the whole log for that boot, and return the first linux version found.

Hook25 (approved):

PR #1828, opened 2025-03-28
fix otg serial and networking tests (Bugfix)

stanley31huang:
## Description
Fix OTG test scripts for following issues
- add retries to remove kernel modules
- initial serial server on RPYC server side
- wait 2 seconds for the networking initialization

## Tests
https://certification.canonical.com/hardware/202309-32027/submission/420194/

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/multiple_otg.py
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/multiple_otg.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse
+import copy
 import glob
 import logging
 import os
@@ -102,7 +103,8 @@ class OtgConfigFsOperatorBase:
             text=True,
             universal_newlines=True,
         )
-        return output.strip("\n").split(",") if output.strip("\n") else []
+        output = output.strip("\n")
+        return output.split(",") if output else []
 
     def enable_otg_module(self, modules):
         for module in modules:
@@ -111,9 +113,23 @@ class OtgConfigFsOperatorBase:
             )
 
     def disable_otg_related_modules(self, modules):
-        for module in modules:
-            subprocess.run(
-                "modprobe -r {}".format(module), shell=True, check=True
+        cp_modules = copy.deepcopy(modules)
+        i = 0
+        while i <= 3 and cp_modules:
+            for module in modules:
+                if module not in cp_modules:
+                    continue
+                cmd = "modprobe -r {}".format(module)
+                logging.info("Removing %s module", module)
+                logging.info("$ %s", cmd)
+                ret = subprocess.run(cmd, shell=True)
+                if ret.returncode == 0:
+                    cp_modules.remove(module)
+            i += 1
+
+        if cp_modules:
+            raise RuntimeError(
+                "failed to remove modules: {}".format(cp_modules)
             )
 
     def otg_setup(self):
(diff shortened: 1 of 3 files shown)


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/multiple_otg.py
```diff
@@ -348,7 +366,8 @@ def function_check_with_rpyc(self, rpyc_ip):
         test_func = getattr(_module, "perform_ping_test")
         ret = test_func([self._net_dev], "169.254.0.10")
         if ret != 0:
-            raise RuntimeError("Failed to ping DUT from RPYC server")
+            raise RuntimeError(
+                "Failed to ping RPYC server from DUT through OTG network")
```
stanley31huang, line 370:
```suggestion
                "Failed to ping RPYC server from DUT through OTG network"
            )
```

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/multiple_otg.py
```diff
@@ -111,9 +113,23 @@ def enable_otg_module(self, modules):
...
+                logging.info("Removing %s module", module)
+                logging.info("$ %s", cmd)
+                ret = subprocess.run(cmd, shell=True)
+                if ret.returncode == 0:
+                    cp_modules.remove(module)
+            i += 1
+
+        if cp_modules:
+            raise RuntimeError(
+                "failed to remove modules: {}".format(cp_modules)
```
baconYao, line 132: I assume modules is just a straightforward list of module names—is that right? If so, I refactored the code below to make it easier to maintain and more intuitive.

```suggestion
        if not modules:
            logging.info("No need to remove any OTG module since no candidate be provided")
            return

        retry_failed_set = set()

        #  Go through each module
        for module in modules:
            max_retry_count = 3
            while max_retry_count:
                cmd = "modprobe -r {}".format(module)
                logging.info("Removing %s module", module)
                logging.info("$ %s", cmd)
                try:
                    ret = subprocess.run(cmd, shell=True)
                    if ret.returncode == 0:
                        break
                    max_retry_count -= 1
                    if max_retry_count == 0:
                        retry_failed_set.add(module)
                except subprocess.CalledProcessError as e:
                    logging.warning("Command failed:", e) 

        if retry_failed_set:
            raise RuntimeError(
                "Failed to remove modules: {}".format(retry_failed_set)
```

baconYao (commented):
See my inline suggestion

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/multiple_otg.py
stanley31huang: yes, that's just a list of module name. But there're some dependencies between them, and you have to remove them in specific order. Otherwise, you won't be able to remove all of them.

let's said the list is `usb_f_rndis, g_ether, usb_f_ecm`, so the first module we would like to remove is `usb_f_rndis`, but we will get an error while we trying to remove it.
```
ubuntu@ubuntu:~$ lsmod  | grep libcom
libcomposite           77824  3 usb_f_rndis,g_ether,usb_f_ecm
ubuntu@ubuntu:~$ sudo modprobe -r usb_f_rndis
modprobe: FATAL: Module usb_f_rndis is in use.
```

So I use another mechansim to remove modules, I trying to remove all of modules at first round, and then do the same actions for rest of modules.

stanley31huang:
dropped this one as we would like to change the mechanisim.

PR #1832, opened 2025-04-01
Add dependabot configuration for gh actions (infra)

Hook25:
## Description

Now that actions are pinned, we need to also track the updates "manually" for new tags. For this, dependabot will help us proposing updates in prs

## Resolved issues

Fixes: CHECKBOX-1823

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


p-gentili (approved):
Title will not be compliant with the expected format, but *dependabot* doesn't support configuring the suffix. We can live with that, or if it becomes too annoying, we will have to reconsider the title validation.

But for now, LGTM!

PR #1838, opened 2025-04-01
Fix output to match the script (bugfix)

Hook25:
## Description

The image info collector is wrong, I don't recall where the oem replace comes from but it is wrong. Also, the canonical replace was slightly different, now it is replacing the prefix correctly.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1826
Internal discussion: https://chat.canonical.com/canonical/pl/in4u5zcqijrp7m35qui4isxuhw

## Documentation

N/A

## Tests

Added new tests

### DIFF
--- a/checkbox-ng/plainbox/vendor/image_info.py
+++ b/checkbox-ng/plainbox/vendor/image_info.py
@@ -47,8 +47,7 @@ def dcd_string_to_info(dcd_string):
     - url
     """
     # prefix, should always be present
-    dcd_string = dcd_string.replace("canonical-", "")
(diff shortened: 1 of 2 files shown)


KaiChuan-Hsieh:
It is PC oem image DCD string "canonical-oem-somerville-noble-oem-24.04b-proposed-20250326-211". There is prefix "canonical-oem-" before the project. It might be parsed wrongly, if we only replace "canonical-".

pieqq (approved):

PR #1842, opened 2025-04-04
Add extensive CUDA testing for more machines (New)

antlassagne:
The current CUDA-related tests in the GPGPU test plan are using a snap package that contains the CUDA runtimes. Thus it does not really test what the client will have on his machine, not does it allow to chose specific version or architecture. Right now it is only compatible with CUDA 12.8, amd64 desktops and arm server machines. This PR adds other tests that compile what is required at runtime. They assume that both the driver and CUDA are installed. They allow to choose the sample version, as well as specific sets of features.

- It's based on a shell script that download, build an runs the cuda samples. 
- The script takes an argument, a number between 0 and 8. That makes 9 test in one test plan, to split the tests.
- Some environment variables can be set in the manifest. They mainly allow to chose a specific CUDA version, but also to enable / disable specific CUDA features, or to skip some tests entirely for specific platforms.

This PR was tested on both a Desktop machine and a specific Nvidia machine (the target). The documentation does not mention any unit-testing of bash file, but I'm open to suggestions.

### DIFF
--- a/providers/gpgpu/bin/run_cuda_sample_set.py
+++ b/providers/gpgpu/bin/run_cuda_sample_set.py
@@ -1,0 +1,391 @@
+#!/usr/bin/env python3
+"""Script to build and run cuda samples, to test cuda features on nvidia gpus.
+
+Copyright (C) 2025 Canonical Ltd.
+
+Authors
+  Antone Lassagne <antone.lassagne@canonical.com>
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License version 3,
+as published by the Free Software Foundation.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import argparse
+import logging
+import os
+import shutil
+import subprocess
+
+from pathlib import Path
+
+
+def cleanup_temporary_files(orig_dir, test_set):
+    """Cleanup the files and folder that were created during the tests
+
+    Args:
+        orig_dir (Path): Path of the root folder
+        test_set (number): Index of the test set
+    """
+    test_set_dir = Path(orig_dir) / test_set
+    logging.info("Cleaning up %s", test_set_dir)
+    shutil.rmtree(str(test_set_dir), ignore_errors=True)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Run a set of CUDA tests with customizable configurations."
+    )
+
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        dest="log_level",
+        action="store_const",
+        default=logging.INFO,
+        const=logging.DEBUG,
+        help="Increase logging level",
+    )
+
+    parser.set_defaults(missing_files=[])
+
+    # sub parsers
+    subparsers = parser.add_subparsers(required=True)
+    intro_parser = subparsers.add_parser("introduction", help="Introduction")
+    intro_parser.set_defaults(test_set=0)
+
+    utilities_parser = subparsers.add_parser("utilities", help="Utilities")
+    utilities_parser.set_defaults(test_set=1)
+
+    concepts_parser = subparsers.add_parser(
+        "concepts", help="Concepts_and_Techniques"
+    )
+    concepts_parser.set_defaults(test_set=2)
+    concepts_parser.set_defaults(
+        missing_files=[
+            # list of tuple with src, dest, and extension ?
+            (
+                Path("Samples")
+                / "2_Concepts_and_Techniques"
+                / "EGLStream_CUDA_Interop",
+                Path("build")
+                / "Samples"
+                / "2_Concepts_and_Techniques"
+                / "EGLStream_CUDA_Interop"
+                / "bin",
+                ".yuv",
+            )
+        ]
+    )
+
+    features_parser = subparsers.add_parser("features", help="CUDA_Features")
+    features_parser.set_defaults(test_set=3)
+
+    libraries_parser = subparsers.add_parser(
+        "libraries", help="CUDA_Libraries"
+    )
+    libraries_parser.set_defaults(test_set=4)
+
+    domain_parser = subparsers.add_parser("domain", help="Domain_Specific")
+    domain_parser.set_defaults(test_set=5)
+
+    performance_parser = subparsers.add_parser(
+        "performance", help="Performance"
+    )
+    performance_parser.set_defaults(test_set=6)
+
+    libnvvm_parser = subparsers.add_parser("libnvvm", help="libNVVM")
+    libnvvm_parser.set_defaults(test_set=7)
+    libnvvm_parser.set_defaults(
+        missing_files=[
+            (
+                Path("Samples") / "7_libNVVM" / "ptxgen",
+                Path("build") / "Samples" / "7_libNVVM" / "ptxgen" / "bin",
+                ".ll",
+            )
+        ]
+    )
+
+    platform_parser = subparsers.add_parser(
+        "platform", help="Platform_Specific"
+    )
+    platform_parser.set_defaults(test_set=8)
+    platform_parser.set_defaults(
+        missing_files=[
+            (
+                Path("Samples")
+                / "8_Platform_Specific"
+                / "Tegra"
+                / "cudaNvSciBufMultiplanar",
+                Path("build")
+                / "Samples"
+                / "8_Platform_Specific"
+                / "Tegra"
+                / "cudaNvSciBufMultiplanar"
+                / "bin",
+                ".yuv",
+            )
+        ]
+    )
+
+    parser.add_argument(
+        "--cuda-samples-version",
+        default=os.getenv("CUDA_SAMPLES_VERSION", "12.8"),
+        help="CUDA samples version.",
+    )
+    parser.add_argument(
+        "--cuda-ignore-tensorcore",
+        default=os.getenv("CUDA_IGNORE_TENSORCORE", "0"),
+        choices=["0", "1"],
+        help="Ignore TensorCores if the machine does not have them.",
+    )
+    parser.add_argument(
+        "--cuda-multigpu",
+        default=os.getenv("CUDA_MULTIGPU", "0"),
+        choices=["0", "1"],
+        help="Enable if the machine has multiple NVIDIA GPUs.",
+    )
+    parser.add_argument(
+        "--no-clone",
+        action="store_true",
+        help="[DEBUG] Don't clone the repo",
+    )
+    parser.add_argument(
+        "--keep-cache",
+        action="store_true",
+        help="[DEBUG] Keep the cache",
+    )
+
+    parser.add_argument(
+        "--cuda-ignore-tests",
+        default=os.getenv("CUDA_IGNORE_TESTS", ""),
+        help="Space-separated list of tests to ignore.",
+    )
+
+    args = parser.parse_args()
+
+    args.cuda_ignore_tests = (
+        args.cuda_ignore_tests.strip().split(" ")
+        if args.cuda_ignore_tests and args.cuda_ignore_tests.strip()
+        else []
+    )
+
+    if args.cuda_ignore_tensorcore:
+        args.cuda_ignore_tests.extend(
+            ["dmmaTensorCoreGemm", "tf32TensorCoreGemm", "bf16TensorCoreGemm"]
+        )
+
+    if args.cuda_multigpu:
+        args.cuda_ignore_tests.extend(
+            [
+                "simpleP2P",
+                "simpleAttributesMPU",
+                "simpleCUFFT_MGPU",
+                "streamOrderedAllocationP2P",
+                "simpleCUFFT_2d_MGPU",
+                "conjugateGradientMultiDeviceCG",
+            ]
+        )
+
+    return args
+
+
+def remove_add_subdirectory_line(cmake_file, dir_name):
+
+    dir_name = os.path.basename(str(dir_name))
+    cmake_file_path = Path(cmake_file)
+
+    # Read the current content of the CMakeLists.txt
+    with cmake_file_path.open("r") as file:
+        lines = file.readlines()
+
+    # Write back to the CMakeLists.txt excluding lines with the specified
+    # add_subdirectory
+    with cmake_file_path.open("w") as file:
+        for line in lines:
+            if "add_subdirectory(" + dir_name + ")" not in line:
+                file.write(line)
+
+
+def copy_and_set_permissions(src, dst, file_extension):
+    Path(dst).mkdir(parents=True, exist_ok=True)
+
+    # Copy the files with the given extension
+    for filename in os.listdir(str(src)):
+        if filename.endswith(file_extension):
+            shutil.copy(os.path.join(src, filename), dst)
+
+            # Set file permissions to remove execute bit
+            file_path = os.path.join(dst, filename)
+            os.chmod(file_path, 0o644)
+
+
+def clone_and_build(orig_dir, test_set, cuda_samples_version):
+    """Function to clone the repository and build the correct subfolder
+
+    Args:
+        orig_dir (Path): Path of the root folder
+        test_set (number): index of the set (0 to 8)
+        cuda_samples_version (_type_): tag to clone (ex: 12.8)
+
+    """
+    test_set_dir = Path(orig_dir) / test_set
+    if test_set_dir.exists():
+        raise FileExistsError("Error: folder {} exists".format(test_set_dir))
+
+    logging.info(
+        "Cloning CUDA Samples v%s. Version can be set in the manifest.",
+        cuda_samples_version,
+    )
+    subprocess.run(
+        [
+            "git",
+            "clone",
+            "-b",
+            "v" + str(cuda_samples_version),
+            "--single-branch",
+            "https://github.com/NVIDIA/cuda-samples.git",
+            str(test_set_dir),
+        ],
+        check=True,
+    )
+
+    cmake_file = test_set_dir / "CMakeLists.txt"
+    cmake_file.write_text(
+        "{}\n{}".format(
+            'set(EXECUTABLE_OUTPUT_PATH "bin")', cmake_file.read_text()
+        )
+    )
+
+    # Remove unnecessary folders
+    samples_dir = test_set_dir / "Samples"
+    for folder in samples_dir.glob("[0-9]_*/"):
+        folder_number = folder.name.split("_")[0]
+        if folder_number != test_set:
+            shutil.rmtree(str(folder))
+            remove_add_subdirectory_line(
+                Path(samples_dir, "CMakeLists.txt"), folder
+            )
+        else:
+            logging.info("Keeping directory: %s", folder.name)
+
+    # Build the sample
+    build_dir = test_set_dir / "build"
+    build_dir.mkdir(parents=True, exist_ok=True)
+    cuda_path = os.getenv("CUDA_PATH", "/usr/local/cuda")
+    subprocess.run(
+        [
+            "cmake",
+            "-DCMAKE_CUDA_ARCHITECTURES=native",
+            "-DCMAKE_CUDA_COMPILER=" + cuda_path + "/bin/nvcc",
+            "-DCMAKE_LIBRARY_PATH=/usr/local/cuda/lib64/",
+            "-DCMAKE_INCLUDE_PATH=/usr/local/cuda/include",
+            str(test_set_dir),
+        ],
+        check=True,
+        cwd=str(build_dir),
+    )
+
+    subprocess.run(
+        ["make", "-j", str(os.cpu_count() - 1)],
+        cwd=str(build_dir),
+        check=True,
+    )
+
+
+# Function to run tests
+def run_tests(orig_dir, test_set, exclude_list):
+    """Run the test
+
+    Args:
+        orig_dir (Path): path of the root folder
+        test_set (_type_): index of the set (0 to 8)
+        exclude_list ([str]): list of test to skip
+
+    Returns:
+        (int, int): total tests, skipped tests
+    """
+    test_set_dir = Path(orig_dir) / test_set / "build" / "Samples"
+
+    executable_list = [
+        exe
+        for exe in test_set_dir.rglob("*/*/bin/*")
+        if os.access(str(exe), os.X_OK)
+    ]
+
+    skipped = 0
+    total = len(executable_list)
+
+    for index, exe in enumerate(executable_list, 0):
+        logging.info("Step %i of %i: %s", index, total, exe)
+
+        exe_name = exe.name
+        excluded = any(exe_name == pattern for pattern in exclude_list)
+
+        if excluded:
+            logging.info("Skipping %s", exe)
+            skipped += 1
+            continue
+
+        logging.info("Running: %s in %s", exe.name, os.path.dirname(str(exe)))
+        exe_args = "test.ll" if exe_name == "ptxgen" else None
+
+        proc = subprocess.run(
+            [str(exe), exe_args] if exe_args else [str(exe)],
+            check=True,
+            cwd=os.path.dirname(str(exe)),
+        )
+
+        code = proc.returncode if not isinstance(proc, bool) else 0
+
+        logging.error("Error code : " + str(code))
+
+    logging.info("All %i tests done,; %s skipped.", total, skipped)
+    return total, skipped
+
+
+def main():
+    args = parse_args()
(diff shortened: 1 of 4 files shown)


providers/gpgpu/units/cuda.pxu
```diff
@@ -0,0 +1,133 @@
...
+    CUDA_SAMPLES_VERSION
+    CUDA_IGNORE_TESTS
+    CUDA_IGNORE_TENSORCORE
+    CUDA_MULTIGPU
+plugin: shell
+estimated_duration: 4
+requires:
+    graphics_card.vendor == 'NVIDIA Corporation'
+_summary: NVIDIA CUDA 8 - platform specific
+command: run_cuda_samples_set.sh 8
```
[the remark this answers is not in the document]
antlassagne: Thank you, I did not know about this feature. Fixed and rebased.

providers/gpgpu/bin/run_cuda_samples_set.sh
pedro-avalos, line 1: I'd recommend turning this into a Python script so we can more easily run quality tests on your logic here

pedro-avalos (changes requested):
Raising from comment to request changes, please re-write the script to Python and add some unit tests.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
+#!/usr/bin/env python3
```
pedro-avalos, line 1: Please add a docstring for the string that includes a short description of the script and the Copyright notice

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+from pathlib import Path
+import atexit
+
+
+def should_clone():
+    return os.getenv("CLONE", "1") != "0"
+
+
+def should_keep_cache():
+    return os.getenv("KEEP_CACHE", "0") == "1"
```
pedro-avalos, line 16: Unless these are going to change during the script, why not just store them in global variables?

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def print_colored(color: str, text: str):
+    color_map = {
+        "green": "\033[32m",
+        "red": "\033[31m",
+    }
+    color_code = color_map.get(color, "")
+    reset_code = "\033[0m"
+    print(color_code + text + reset_code)
```
pedro-avalos, line 26: Don't bother adding color. Checkbox will override this. `stdout` is green and `stderr` is red. If you need to log things you should take a look at the [`logging`](https://docs.python.org/3/library/logging.html) module in Python. [`gpu_passthrough.py`](https://github.com/canonical/checkbox/blob/main/providers/gpgpu/bin/gpu_passthrough.py) should server as a simple example of how to do logging (and pass a `-v` flag to the script).

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import subprocess
+import shutil
+from pathlib import Path
+import atexit
```
pedro-avalos, line 8: Nitpick: Could you `isort` these imports?

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+    Set CUDA_IGNORE_TENSORCORE=1 if your machine does not have TensorCores.
+    Set CUDA_MULTIGPU=1 if your machine has multiple NVIDIA GPUs.
+    Set KEEP_CACHE=1 to keep the file from the session.
+    Then set CLONE=0 to reuse the files from a previous session.
+        """
+        print(help_text)
+        sys.exit(1)
+
+    else:
+        test_set = sys.argv[1]
```
pedro-avalos, line 318: Please take a look at [`argparse`](https://docs.python.org/3/library/argparse.html). For an example of how to have arguments that are optional and default to an environment variable you can take a look at [`gpu_passthrough.py`](https://github.com/canonical/checkbox/blob/main/providers/gpgpu/bin/gpu_passthrough.py). `argparse` also has the ability to limit choices to a list/set of choices.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def exit_trap(orig_dir, test_set):
+    if should_keep_cache():  # pragma: no cover
+        return
+
+    else:
+        test_set_dir = Path(orig_dir) / test_set
+        print("Cleaning up", test_set_dir)
+        shutil.rmtree(str(test_set_dir), ignore_errors=True)
```
pedro-avalos, line 36: I would strongly recommend against using `atexit` and exit traps. I would use `try`/`except` to handle errors and cleanup. If you do need a wider exception/cleanup handling perhaps you can rewrite it as a context manager. For an example you can take a look at how we wrap LXD in [`lxd_support.py`](https://github.com/canonical/checkbox/blob/main/checkbox-support/checkbox_support/lxd_support.py).

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def write_first_line_to_file(file_path: str, line: str):
+    with open(str(file_path), "r") as f:
+        lines = f.readlines()
+
+    lines.insert(0, line + "\n")  # you can replace zero with any line number.
+    with open(str(file_path), "w") as f:
+        f.writelines(lines)
```
pedro-avalos, line 45: Feel free to just use the `pathlib` functions instead of converting to and from `Path` and `str`.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+
+    else:
+        test_set = sys.argv[1]
+        orig_dir = Path.cwd()
+        cuda_samples_version = os.getenv("CUDA_SAMPLES_VERSION", "12.8")
+        cuda_ignore_tensorcore = (
+            os.getenv("CUDA_IGNORE_TENSORCORE", "0") == "1"
+        )
+        cuda_multigpu = os.getenv("CUDA_MULTIGPU", "0") == "1"
+        cuda_ignore_tests = os.getenv("CUDA_IGNORE_TESTS", "")
```
pedro-avalos, line 325: These should probably also be part of the `argparse` optional arguments that default to an environment variable / constant value.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+
+        print("Exclusion list:", exclude_list)
+
+        # Clone and build the repo
+        clone_and_build(orig_dir, test_set, cuda_samples_version)
+
+        # Run the tests
+        run_tests(orig_dir, test_set, exclude_list)
+        sys.exit(0)
```
pedro-avalos, line 338: Just returning without an error should be the same as `sys.exit(0)`.

pedro-avalos (changes requested):
Initial review. This Python script still smells of a bash script; I've left some feedback on how to improve its code style to match what we have in Checkbox.

Regarding the exit handling, I think perhaps a context manager would be the way to go with this script, take a look at the comment related to that and the `lxd_support.py` example.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: because I don't like global variables in general. But if you prefer, I'll do it.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: It's a lot more convenient when the program is ran outside from checkbox (which some features are exclusively for, CLONE, KEEP_CACHE). Once an issue has been detected from the checkbox suite, it's faster to debug with just the script.

providers/gpgpu/bin/run_cuda_sample_set.py
pedro-avalos: Ah `py35`, I forgot about that. Then no, migrate as much as possible, but we do want the tests to pass for `py35`

providers/gpgpu/bin/run_cuda_sample_set.py
pedro-avalos: @antlassagne fair on the global vars. One option is to use a `parse_args` function that returns a namespace. Set the `--clone` or `--keep-cache` options as optional with these as defaults, then any place just needs a `args` instance that they can refer to `args.clone` or `args.keep-cache` bools.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: This is done.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: This is done.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: This is done.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: This is done.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: I used try / except to keep it simple

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: This is done.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: This is done.

antlassagne:
@pedro-avalos I made the requested changes. However the CI still fails, and it seems that it is not because of my commits. 2 errors:
- `tox` job fails due to [some package not found during the setup](https://github.com/canonical/checkbox/actions/runs/14350803154/job/40229068629?pr=1842). Is `pkg-config` required?
- One of the deb validation fails [because it asks for confirmation before installing the dependencies](https://github.com/canonical/checkbox/actions/runs/14350803164/job/40229078208?pr=1842) (or so I suppose)

antlassagne:
> @antlassagne Could you rebase and retry? I think this was resolved by #1854

Same issue I'm afraid

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+                "simpleAttributesMPU",
+                "simpleCUFFT_MGPU",
+                "streamOrderedAllocationP2P",
+                "simpleCUFFT_2d_MGPU",
+                "conjugateGradientMultiDeviceCG",
+            ]
+        )
+    if cuda_ignore_tests:
+        exclude_list.extend(cuda_ignore_tests.split(" "))
+    return exclude_list
```
pedro-avalos, line 300: Hm, this shouldn't be necessary. If you follow my earlier comment to make the other arguments `store_true`, then you can do the following within the `parse_args` function:

```python
parser = argparse.ArgumentParser()
parser.add_argument("--exclude", action="append")
parser.add_argument("--cuda-ignore-tensorcore", action="store_true")
parser.add_argument("--cuda-multigpu", action="store_true")

args = parser.parse_args()

args.exclude = args.exclude or []
if args.cuda_ignore_tensorcore:
    args.exclude.extend(
        ["dmmaTensorCoreGemm", "tf32TensorCoreGemm", "bf16TensorCoreGemm"]
    )
if args.cuda_multigpu:
    args.exclude.extend(
        [
            "simpleP2P",
            "simpleAttributesMPU",
            "simpleCUFFT_MGPU",
            "streamOrderedAllocationP2P",
            "simpleCUFFT_2d_MGPU",
            "conjugateGradientMultiDeviceCG",
        ]
    )
```

Notice that I have changed the `--exclude-list` to an `--exclude` option that you can repeat multiple times. Sadly the action `extend` is a Python 3.8 or above feature.

If you still want to emulate the `extend` functionality, you can:

```python
parser = argparse.ArgumentParser()
parser.add_argument("--exclude")
parser.add_argument("--cuda-ignore-tensorcore", action="store_true")
parser.add_argument("--cuda-multigpu", action="store_true")

args = parser.parse_args()

args.exclude = args.exclude.strip().split(" ") if args.exclude.strip() else []
if args.cuda_ignore_tensorcore:
    args.exclude.extend(
        ["dmmaTensorCoreGemm", "tf32TensorCoreGemm", "bf16TensorCoreGemm"]
    )
if args.cuda_multigpu:
    args.exclude.extend(
        [
            "simpleP2P",
            "simpleAttributesMPU",
            "simpleCUFFT_MGPU",
            "streamOrderedAllocationP2P",
            "simpleCUFFT_2d_MGPU",
            "conjugateGradientMultiDeviceCG",
        ]
    )
```

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+            [str(exe), exe_args] if exe_args else [str(exe)],
+            check=True,
+            cwd=os.path.dirname(str(exe)),
+        )
+
+        code = proc.returncode if not isinstance(proc, bool) else 0
+
+        print("Error code : " + str(code))
+        # if code != 0:
+        #     sys.exit(1)
```
pedro-avalos, line 345: Why not combine all of these into a single for loop?

```python
for file in test_set_dir.rglob(...):
    if not os.access(file, os.X_OK):
        continue
    ...
```

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+    color_code = color_map.get(color, "")
+    reset_code = "\033[0m"
+    print(color_code + text + reset_code)
+
+
+def write_first_line_to_file(file_path: str, line: str):
+    with open(str(file_path), "r") as f:
+        lines = f.readlines()
+
+    lines.insert(0, line + "\n")  # you can replace zero with any line number.
```
pedro-avalos, line 120: It seems `pathlib.Path` has `write_text` and `read_text` since Python 3.5. You could simplify this to (if `file_path` is changed to a `Path` argument) the following:

```python
file_path.write_text("{}\n{}".format(line, file_path.read_text()))
```

Since this is now a one-liner, and I only see one reference to this function, then there is no need for the helper function.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+        dst_dir = os.path.join(
+            str(orig_dir),
+            str(test_set),
+            "build",
+            "Samples",
+            "7_libNVVM",
+            "ptxgen",
+            "bin",
+        )
+        copy_and_set_permissions(src_dir, dst_dir, ".ll")
```
pedro-avalos, line 275: See comment in `parse_args`, this can be generalized so you don't have to check for different `test_set` values.

pedro-avalos (changes requested):
General changes I want to see before moving forward:

- Docstrings
- Remove unnecessary comments (don't repeat yourself)
- Look at the argparse suggestions I made, should make the script CLI and code more readable/user-friendly
- Please get rid of colored output and replace it with proper logging
- Please fix the `try`/`except` blocks.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: I entirely removed the color, for the sake of simplicity. You are right that we should not need to look for colors in and endless log.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+        default=os.getenv("CLONE", "1"),
+        choices=["0", "1"],
+        help="[DEBUG] Clone - default is 1",
+    )
+    parser.add_argument(
+        "--keep_cache",
+        default=os.getenv("KEEP_CACHE", "0"),
+        choices=["0", "1"],
+        help="[DEBUG] Keep cache - default is 0",
+    )
```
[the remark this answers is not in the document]
antlassagne: Definitely. It's jsut some bad refactoring from me. I'll fix that with dashes only.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+            file_path = os.path.join(dst, filename)
+            os.chmod(file_path, 0o644)
+
+
+# Function to clone the repository and build the correct subfolder
+def clone_and_build(orig_dir, test_set, cuda_samples_version):
+    test_set_dir = Path(orig_dir) / test_set
+    if test_set_dir.exists():
+        raise FileExistsError(
+            "Error: folder " + str(test_set_dir) + " exists."
```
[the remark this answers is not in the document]
antlassagne: I'm doing logging indeed, but I can't use f" because python 3.5 is not compatible. Shall I keep these like that?

providers/gpgpu/units/test-plan.pxu
```diff
@@ -45,3 +45,23 @@ bootstrap_include:
     executable
     package
     snap
+
+id: gpgpu-cuda-features
+unit: test plan
+_name: Nvidia CUDA Feature testing
```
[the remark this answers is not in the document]
antlassagne: I'm not sure, so I'll change it

providers/gpgpu/bin/run_cuda_sample_set.py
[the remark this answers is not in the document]
pedro-avalos: Right, we can't use f-strings. For logging, you want to use, for example:

`logging.info("Error: folder %s exists.", test_set_dir)`

This is the recommended way to do logging, as it technically is more efficient (it avoids formatting when the string won't even be displayed because of the configured log level)

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+
+        result = 0
+        # Clone and build the repo
+        if args.clone != "0":
+            clone_and_build(
+                orig_dir, str(args.test_set), args.cuda_samples_version
+            )
+
+        run_tests(orig_dir, str(args.test_set), exclude_list)
+    except Exception:
```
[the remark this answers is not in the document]
antlassagne: I did the following:

- run_test generates [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError)
- `clone_and_build` gets his own try catch just for the sake of cleaning up in the end I made it a CalledProcessError, FileExistsError, OSError. Is OSError to large?
- The rest is not in the try/catch, since there is no need for cleanup.

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: That was to get the length of the list in the beginning. But I can also decrement. Will do.

providers/gpgpu/bin/run_cuda_sample_set.py
pedro-avalos: That's a good point, if you do the `extend` example, the space-separated env variable should work

providers/gpgpu/bin/run_cuda_sample_set.py
pedro-avalos: Hm, I guess you can get the list if you need to show the totals.

```python
executables = [exe for exe in test_set_dir.rglob(...) if os.access(exe, os.X_OK)]
total = len(executables)
```

providers/gpgpu/bin/run_cuda_sample_set.py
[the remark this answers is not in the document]
pedro-avalos: re `clone`, you may want to use `store_false` for that one and use `--no-clone`.

re manifest question: I'm not sure regarding the manifests, the only option I can think of at the script-level is to use the env vars as suggested.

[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+        help="[DEBUG] Keep cache - default is 0",
+    )
+
+    parser.add_argument(
+        "--cuda_ignore_tests",
+        default=os.getenv("CUDA_IGNORE_TESTS", ""),
+        help="Space-separated list of tests to ignore.",
+    )
+
+    # Parse the arguments
```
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+        if filename.endswith(file_extension):
+            shutil.copy(os.path.join(src, filename), dst)
+
+            # Set file permissions to remove execute bit
+            file_path = os.path.join(dst, filename)
+            os.chmod(file_path, 0o644)
+
+
+# Function to clone the repository and build the correct subfolder
+def clone_and_build(orig_dir, test_set, cuda_samples_version):
```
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/units/test-plan.pxu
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+            1 --> Utilities
+            2 --> Concepts_and_Techniques
+            3 --> CUDA_Features
+            4 --> CUDA_Libraries
+            5 --> Domain_Specific
+            6 --> Performance
+            7 --> libNVVM
+            8 --> Platform_Specific
+            """,
+    )
```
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,385 @@
...
+
+        run_tests(orig_dir, str(args.test_set), exclude_list)
+    except Exception:
+        print("Test failed", file=sys.stderr)
+        result = 1
+
+    if not args.keep_cache == "1":
+        exit_trap(orig_dir, str(args.test_set))
+
+    exit(result)
```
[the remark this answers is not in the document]
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: Pushed

providers/gpgpu/bin/run_cuda_sample_set.py
antlassagne: Pushed

pedro-avalos (changes requested):
This looks better. I just have some concerns and next steps:

1. I am still concerned with the use of `print` instead of the `logging` module.
2. I assume you've tested this locally as a script?
3. I want you to run your Test Plan on some machines in the lab, then submit the Checkbox results to C3 when promtped. Link those in this PR.
4. I see there is a cleanup step, can you do step 3 twice and Checkbox won't complain / run into any issues? I want to make sure this test can be re-run without any manual steps in between.

providers/gpgpu/bin/run_cuda_sample_set.py
```diff
@@ -0,0 +1,389 @@
...
+import shutil
+
+import subprocess
+import sys
+
+from pathlib import Path
+
+
+def cleanup_temporary_files(orig_dir, test_set):
+    """Cleanup the files and folder that were created duriong the tests
```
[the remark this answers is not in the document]
antlassagne: 1. You are right, and I was lazy, but it's now done.
2. I tested within checkbox to make sure the manifest inputs were working correctly.
3. This is it https://certification.canonical.com/hardware/202502-36337/submission/424080/
4. What is step 3? The cleanup can run several times in a row without errors. I added a test for that. There is a case where I get missing file, it is when I run the checkbox suite and spam ctrl-c during the run.

pedro-avalos (approved):
Thanks for all the changes, this should be okay now.

Note: The two failing tests in the submission are due to an upstream bug

PR #1854, opened 2025-04-09
Tag is wrong (infra)

Hook25:
## Description

The tag expands to the uninterpreted string and not an array, this replaces the string with a call to fromJSON as brackets don't seem to be supported in that syntax

## Resolved issues

Broken (not started) dailies from yesterday

## Documentation

N/A

## Tests

See: https://github.com/canonical/checkbox/actions/runs/14358219772/job/40252650316


pedro-avalos (approved):
Thank you, I should have known this was going to end up being the right solution...

PR #1858, opened 2025-04-10
Add remodel tests for Ubuntu Core 20 => 22 and 22 => 24 (New)

Meulengracht:
## Description

Introduce two new tests, one specifically for Ubuntu Core 20 that verifies the possibility of remodeling to Ubuntu Core 22, and one identical one for Ubuntu Core 22 that remodels to 24.

The tests are implemented with the assumption that the Ubuntu Core images used by certifications use the official models sourced from either cdimage (which use models from https://github.com/canonical/models) or they are built by certifications themself from the mentioned models.

There are currently a lot of limitations on how remodel works, which means that the models must match brand / model if done offline or the device must have a serial assertion (as offline remodels cannot be done from model to a new model, only upgrades and we do not have official models for that currently). 

If we end in a situation where the remodeling has to be done offline, we must discuss how we solve this for certification testing. Because if things don't line up currently, then we simply cannot test remodelling in certifications without either a new set of models specifically for this, or we have to lax some of the requirements for remodeling (i.e. implement/support some missing cases).

It's important to be aware of the fact that these test cause the device to reboot *multiple times*.

## Resolved issues

I don't know if there is a issue for covering this specific area of testing.

## Documentation

Depending on the test-plan they get added to, we may have to update the coverage in https://certification.canonical.com/docs/programmes/iot/ ?

## Tests

These tests have been tested in qemu for amd64 platform and on a Raspberry Pi4 for pi-arm64 platform, by installing my own custom checkbox frontend containing the two added tests, then run with the following commands:

```
checkbox-remodel.checkbox-cli run com.canonical.qa.remodel::remodel-20-to-22
```
and
```
checkbox-remodel.checkbox-cli run com.canonical.qa.remodel::remodel-22-to-24
```

The python script was tested before it was tested in a snapped checkbox context by running it from the users home directory with the parameters in the tests.

### DIFF
--- a/providers/base/bin/remodel_test.py
+++ b/providers/base/bin/remodel_test.py
@@ -5,8 +5,8 @@
 # Written by:
 #    Authors: Philip Meulengracht <philip.meulengracht@canonical.com>
 
+import argparse
 import os
-import sys
 import platform
 import subprocess
 from urllib.request import urlretrieve
@@ -20,93 +20,76 @@ def get_platform():
         return "pi-armhf"
     elif "x86_64" in plt:
         return "amd64"
-    raise Exception(f"unrecognized platform {plt}")
+    raise SystemExit(f"platform not supported for remodeling test: {plt}")
 
 
-def resolve_target_remodel():
-    runtime = os.getenv("CHECKBOX_RUNTIME", "/snap/checkbox/current")
-    if "checkbox20" in runtime:
-        return "uc22"
-    elif "checkbox22" in runtime:
-        return "uc24"
-    raise Exception(f"unsupported version for remodel: {runtime}")
-
-
-def resolve_model(uc_ver):
+# Currently images used in certifications are sourced from cdimage,
+# those images builds using the models supplied in canonical/models.
+# Make sure we use the same models that come from the same authority,
+# otherwise remodeling will fail.
+def download_model(uc_ver):
     base_uri = "https://raw.githubusercontent.com/canonical/models/"
     branch = "refs/heads/master/"
     model = f"ubuntu-core-{uc_ver}-{get_platform()}-dangerous.model"
-    print(f"resolving model for remodeling: {base_uri + branch + model}")
-    path, _ = urlretrieve(base_uri + branch + model, f"uc-{uc_ver}.model")
+    print(f"downloading model for remodeling: {base_uri + branch + model}")
+    path, _ = urlretrieve(base_uri + branch + model)
     return path
 
 
-def resolve_snaps(uc_ver):
+# downloads a snap to the tmp folder
+def download_snap(name, out, channel):
+    dir = os.getcwd()
+    os.chdir("/tmp")
     subprocess.run(
-        ["snap", "download", f"core{uc_ver}", f"--basename=core{uc_ver}"]
+        [
+            "snap",
+            "download",
+            name,
+            f"--channel=latest/{channel}",
+            f"--basename={out}",
+        ]
     )
-    # for the kernel snap use beta
+    os.chdir(dir)
+
+
+def download_snaps(uc_ver):
+    # use stable for remodel, we are not testing the snaps we are
+    # remodeling to, but rather the process works.
+    channel = "stable"
+    download_snap(f"core{uc_ver}", "base", channel)
     if "pi" in get_platform():
-        subprocess.run(
-            [
-                "snap",
-                "download",
-                "pi",
-                f"--channel={uc_ver}/edge",
-                "--basename=gadget",
-            ]
-        )
-        subprocess.run(
-            [
-                "snap",
-                "download",
-                "pi-kernel",
-                f"--channel={uc_ver}/beta",
-                "--basename=kernel",
-            ]
-        )
+        download_snap("pi", "gadget", f"--channel={uc_ver}/{channel}")
+        download_snap("pi-kernel", "kernel", f"--channel={uc_ver}/{channel}")
     else:
-        subprocess.run(
-            [
-                "snap",
-                "download",
-                "pc",
-                f"--channel={uc_ver}/edge",
-                "--basename=gadget",
-            ]
-        )
-        subprocess.run(
-            [
-                "snap",
-                "download",
-                "pc-kernel",
-                f"--channel={uc_ver}/beta",
-                "--basename=kernel",
-            ]
-        )
+        download_snap("pc", "gadget", f"--channel={uc_ver}/{channel}")
+        download_snap("pc-kernel", "kernel", f"--channel={uc_ver}/{channel}")
 
 
 def main():
     """Run remodel of an Ubuntu Core host."""
 
-    uc_ver = ""
-    if len(sys.argv) > 1:
-        uc_ver = sys.argv[1]
-    else:
-        uc_ver = resolve_target_remodel()
-
-    # resolve the model for the current platform
-    model_path = resolve_model(uc_ver)
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "target",
+        help="which verison of ubuntu-core that should be remodeled to",
+        choices=["22", "24"]
+    )
 
     # resolve the snaps for the remodel if offline has been requested
     # (currently offline was used for testing in certain scenarios during
     # test development) - for normal testing offline should not be needed
-    offline = False
-    if len(sys.argv) > 2 and sys.argv[2] == "offline":
-        offline = True
+    parser.add_argument(
+        "--offline",
+        help="whether the remodel should be offline",
+        action='store_true'
+    )
+    args = parser.parse_args()
 
-    if offline:
-        resolve_snaps(uc_ver)
+    # resolve the model for the current platform
+    model_path = download_model(args.target)
+
+    if args.offline:
+        download_snaps(args.target)
 
         # instantiate the offline remodel
         print("initiating offline device remodel")
@@ -117,17 +100,17 @@ def main():
                 "remodel",
                 "--offline",
                 "--snap",
-                f"core{uc_ver}.snap",
+                "/tmp/base.snap",
                 "--assertion",
-                f"core{uc_ver}.assert",
+                "/tmp/base.assert",
                 "--snap",
-                "gadget.snap",
+                "/tmp/gadget.snap",
                 "--assertion",
-                "gadget.assert",
+                "/tmp/gadget.assert",
                 "--snap",
-                "kernel.snap",
+                "/tmp/kernel.snap",
                 "--assertion",
-                "kernel.assert",
+                "/tmp/kernel.assert",
                 model_path,
             ]
         )
--- a/providers/base/bin/remodel_test.py
+++ b/providers/base/bin/remodel_test.py
@@ -81,7 +81,7 @@ def main():
     parser.add_argument(
         "--offline",
         help="whether the remodel should be offline",
-        action='store_true',
+        action="store_true",
     )
     args = parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


providers/base/bin/remodel_test.py
```diff
@@ -0,0 +1,141 @@
...
+def resolve_target_remodel():
+    runtime = os.getenv("CHECKBOX_RUNTIME", "/snap/checkbox/current")
+    if "checkbox20" in runtime:
+        return "uc22"
+    elif "checkbox22" in runtime:
+        return "uc24"
+    raise Exception(f"unsupported version for remodel: {runtime}")
+
+
+def resolve_model(uc_ver):
```
Hook25, line 35: Same issue with the name as below, this doesn't resolve much, it downloads model file from a repo. Also, is it ok to download this file from a "moving" target like main? Is it a reasonable thing or should this use at least the latest tag

providers/base/bin/remodel_test.py
```diff
@@ -0,0 +1,141 @@
...
+    print(f"resolving model for remodeling: {base_uri + branch + model}")
+    path, _ = urlretrieve(base_uri + branch + model, f"uc-{uc_ver}.model")
+    return path
+
+
+def resolve_snaps(uc_ver):
+    subprocess.run(
+        ["snap", "download", f"core{uc_ver}", f"--basename=core{uc_ver}"]
+    )
+    # for the kernel snap use beta
```
Hook25, line 48: Please document this better, why is this using both edge and beta? Is this significant or should we use beta for both? Why beta and not stable? Why aren't we using the same risk as the currently installed snap? What does `sudo snap remodel` do? Is it the same?

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
 plugin: manual
 category_id: ubuntucore
 
+id: ubuntucore/remodels
+category_id: ubuntucore
+unit: resource
```
Hook25, line 55: resource is not a unit, it is a plugin

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
 plugin: manual
 category_id: ubuntucore
 
+id: ubuntucore/remodels
```
Hook25, line 53: Please add a summary and a description

providers/base/units/ubuntucore/jobs.pxu
Hook25: remember `./manage.py validate`, it will catch all of these mistakes for you

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+category_id: ubuntucore
+unit: resource
+command:
+  echo source: 22
+  echo destination: 24
+  echo
+  echo source: 20
+  echo destination: 22
+
+id: ubuntucore/remodel-{{source}}-to-{{destination}}
```
Hook25, line 63: On templates: this can't work, if you dont specify a template engine (and you shouldn't) the default one is python string formatting, if you put a `{{ }}` around your field you get the opposite of what you want! Use a single one

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+category_id: ubuntucore
+_summary: Remodel Ubuntu Core {{source}} to {{destination}}
+_purpose:
+ Verify the system can remodel from ubuntu core {{source}} to {{destination}} under normal
+ circumstances. The requirement is that the official models from canonical
+ are used for the running Ubuntu Core host. Multiple reboots will happen
+ as a part of this test.
+unit: template
+template-resource: remodels
+template-filter: lsb.distributor_id == 'Ubuntu Core'
```
Hook25, line 73: This doesn't work, you can't `template-filter` a resource that is not your template resource (and you don't want to use a `template-filter` here either way!) What you want is a normal requires:
```suggestion
requires: os.distributor_id == 'Ubuntu Core'
```

Minor: use os, we are deprecating the lsb resource

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+id: ubuntucore/remodel-{{source}}-to-{{destination}}
+category_id: ubuntucore
+_summary: Remodel Ubuntu Core {{source}} to {{destination}}
+_purpose:
+ Verify the system can remodel from ubuntu core {{source}} to {{destination}} under normal
+ circumstances. The requirement is that the official models from canonical
+ are used for the running Ubuntu Core host. Multiple reboots will happen
+ as a part of this test.
+unit: template
+template-resource: remodels
```
Hook25, line 72: Wrong id
```suggestion
template-resource: ubuntucore/remodels
```

Hook25, line 72: Wrong id
```suggestion
template-resource: ubuntucore/remodels
```

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+template-filter: lsb.distributor_id == 'Ubuntu Core'
+template-unit: job
+plugin: shell
+user: root
+flags: noreturn
+estimated_duration: 300.0
+command:
+ snap model | grep "ubuntu-core-{{source}}"
```
Hook25, line 80: Note: checkbox doesn't set -e by default, so this line is a no-op, if it is supposed to be informative print the model line instead ( `| grep "^model"`)

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+_summary: Remodel checkpoint to verify system.
+_purpose:
+ Verify the current system status, if the system is up, then we exit cleanly.
+unit: job
+plugin: shell
+user: root
+flags: noreturn
+estimated_duration: 0.1
+command:
+  cat /proc/cmdline | grep snapd_recovery_mode=run
```
Hook25, line 95: Note this will most likely hang your session if I understand what you want it to do:
- If the system is up and it wont reboot anymore, this hangs the session because noreturn jobs will wait for a reboot forever
- If the system is not ready and will reboot again, I don't see the point of this as no output will we be stored in the submission (noreturn doesn't store the output of the job)

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+ circumstances. The requirement is that the official models from canonical
+ are used for the running Ubuntu Core host. Multiple reboots will happen
+ as a part of this test.
+unit: template
+template-resource: remodels
+template-filter: lsb.distributor_id == 'Ubuntu Core'
+template-unit: job
+plugin: shell
+user: root
+flags: noreturn
```
Hook25, line 77: This is a noreturn job, which is fine but just know that you won't get any output in the submission from it. Isn't there an option so that snapd waits for the user to trigger the reboot instead of just rebooting itself? that way we could prepare the remodel and in a subsequent job trigger the reboot manually

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+estimated_duration: 300.0
+command:
+ snap model | grep "ubuntu-core-{{source}}"
+ remodel_test.py {{destination}}
+ reboot
+
+id: ubuntucore/remodel-check-up
```
Hook25, line 84: This job doesn't depend on the previous one, so (could) be in a random spot in the test run, I assume this is not what you want. use `after: ` or `depends: `

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+ Verify the current system status, if the system is up, then we exit cleanly.
+unit: job
+plugin: shell
+user: root
+flags: noreturn
+estimated_duration: 0.1
+command:
+  cat /proc/cmdline | grep snapd_recovery_mode=run
+
+id: ubuntucore/remodel-{{source}}-to-{{destination}}-verify
```
Hook25, line 97: Same as all the feedback on the previous template (formatter, template filter etc.)

providers/base/units/ubuntucore/jobs.pxu
```diff
@@ -50,6 +50,63 @@ _verification:
...
+_summary: Verify remodel of Ubuntu Core {{source}} to {{destination}}
+unit: template
+template-resource: remodels
+template-filter: lsb.distributor_id == 'Ubuntu Core'
+template-unit: job
+plugin: shell
+user: root
+estimated_duration: 300.0
+command:
+ snap model | grep "ubuntu-core-{{destination}}"
```
Hook25, line 108: This is not a great idea because it hides the result of the command, consider duplicating the output of snap model somehow so that it is in the log, else debugging this will be a considerable issue

providers/base/units/ubuntucore/test-plan.pxu
```diff
@@ -24,3 +24,21 @@ include:
     ubuntucore/os-recovery-mode
     ubuntucore/os-fail-boot-(?!with-refresh-control).*
     ubuntucore/sshd
+
+id: ubuntucore-20-team
+unit: test plan
+_name: Automated tests from the Ubuntu Core Team for Ubuntu Core 20
+_description: Runs automated OS feature tests for Ubuntu Core 20 devices.
```
Hook25, line 31: you are using 2 resources in your test plan but they are not bootstrap included. Include them!
```suggestion
_description: Runs automated OS feature tests for Ubuntu Core 20 devices.
bootstrap_include:
  os
  ubuntucore/remodels
```

providers/base/units/ubuntucore/test-plan.pxu
Hook25, line 1: Remember that you can check which tests are gonna be run on a machine by running
```
checkbox-cli list-bootstrapped com.canonical.certification::ubuntucore-20-team
```

Hook25 (changes requested):
Nice, I would change a few of the following points (see review) but the two biggest one I didnt note everywhere are: 
- we package the base provider for all platforms, we validate it on all platforms so we can't use non-python 3.5 apis. This includes fstrings as our CI is screaming. You will have to change that.
- run doesn't check if the command didnt blow up, use check_call instead (else you will get a failing command followed by a bunch of also failing command, making the log unreadable)

providers/base/bin/remodel_test.py
```diff
@@ -0,0 +1,141 @@
...
+        return "pi-arm64"
+    elif "raspi" in plt:
+        return "pi-armhf"
+    elif "x86_64" in plt:
+        return "amd64"
+    raise Exception(f"unrecognized platform {plt}")
+
+
+def resolve_target_remodel():
+    runtime = os.getenv("CHECKBOX_RUNTIME", "/snap/checkbox/current")
```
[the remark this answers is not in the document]
Meulengracht: changed this

providers/base/units/ubuntucore/jobs.pxu
Meulengracht: since these are templates i cannot use after or depends, but we discussed this

providers/base/bin/remodel_test.py
Meulengracht: It's not really moving very much, and if there are changes  to the models in main then we want those. Updated naming and placement of model

providers/base/units/ubuntucore/jobs.pxu
Meulengracht: fixed!

Meulengracht: fixed!

PR #1870, opened 2025-04-17
add optee manifest to optee test cases (Bugfix)

stanley31huang:
## Description
add optee manifest to optee test cases, so we could skipped them when the system does not has OPTEE.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/optee/jobs.pxu
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/optee/jobs.pxu
@@ -25,6 +25,8 @@ _summary: Validate OPTEE tee-supplicant is activate
 plugin: shell
 user: root
 estimated_duration: 5.0
+requires: manifest.has_optee == 'True'
+imports: from com.canonical.plainbox import manifest
 depends: ce-oem-optee/device-node
 command:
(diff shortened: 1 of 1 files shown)


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/optee/jobs.pxu
```diff
@@ -52,6 +56,8 @@ estimated_duration: 1
 plugin: shell
 user: root
 category_id: optee
+requires: manifest.has_optee == 'True'
+imports: from com.canonical.plainbox import manifest
```
LiaoU3, line 60: Do we also need `depends: ce-oem-optee/device-node` for this job?

LiaoU3 (changes requested):
Please see my comment.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/optee/jobs.pxu
stanley31huang: nope, this test case is trying to capture the OPTEE OS version from journal logs. So it does not matter the optee node exists or not.

LiaoU3 (approved):
+1

PR #1872, opened 2025-04-17
Better beta promotion gate (infra)

Hook25:
## Description

The current beta promotion gate script fails sometimes even though it should pass. What it does is checking that the beta branch was bumped to a revision that created a daily build (green workflow). 

There are two situations when this is wrong:
- The beta build was frankenstained together locally and in CI, but it validated file (daily can't be green in this case)
- The launchpad workflow failed (because some package failed to build while others didnt, and today we are retrying the same revision, leading the one that passed yesterday to be rejected today due to the duplicate upload!). When this happens the debian build will always be red and edge will never be promoted for no good reason.

This updates the workflow to use the same tool we use on jenkins to check that all packages were published. There is still a tiny race condition but it is definetly more reliable this way

## Resolved issues

Fixes: CHECKBOX-1822

## Documentation

Added a comment explaining why things are done the way they are

## Tests

N/A


pieqq (dismissed):
Looks good as it is, I just left an inline suggestion.

pieqq (approved):

Issue #1868, opened 2025-04-16
The *gl_support* test fails with NV driver installed under Wayland


kevinyehk:
### Bug Description

As mentioned in the subject, this test fails when using Wayland and NV driver installed.
```
./unity_support_test -p
OpenGL vendor string:   NVIDIA Corporation
OpenGL renderer string: NVIDIA T1000/PCIe/SSE2
OpenGL version string:  4.6.0 NVIDIA 570.133.07

Not software rendered:    yes
Not blacklisted:          yes
GLX fbconfig:             yes
GLX texture from pixmap:  no
GL npot or rect textures: yes
GL vertex program:        yes
GL fragment program:      yes
GL vertex buffer object:  yes
GL framebuffer object:    yes
GL version is 1.4+:       yes

Unity 3D supported:       no
```
This tool hasn't been updated for years.
https://git.launchpad.net/nux

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

1. Deploy 24.10 on a desktop with Nvidia GPU
2. Install nvidia driver by `sudo apt install nvidia-driver-***`
3. Run checkbox test

### Expected Result

All features should be supported.

### Actual Result

GLX texture from pixmap:  no
Unity 3D supported:       no

### Environment

- OS ubuntu 24.10
- Checkbox Type Debian
- Checkbox Version 4.4.0.dev94

### Relevant log output

```shell

```

tomli380576:
This tool is really only intended for unity desktop which has been deprecated since ubuntu 16. I'm not sure if we have a newer equivalent to this yet but we can do some basic checks:
- hardware rendering
- opengl version > 3.0 for 16, 18, 20, >3.2 for 22+ according to [gtk](https://github.com/GNOME/gtk/blob/0c166eb8d8bd66b43495ca97bfa651457df9ecdb/gdk/gdkglcontext.c#L1144)

tomli380576:
The `GLX texture from pixmap` seems to be a x11 specific extension according to this: https://askubuntu.com/questions/50708/what-do-the-nos-mean-in-the-output-of-unity-support-test

PR #1880, opened 2025-04-21
Full glmark2 dependency (New)

tomli380576:
## Description

This PR adds the dependencies for resolving https://github.com/canonical/checkbox/issues/1868 and https://github.com/canonical/checkbox/pull/1677

It looks like a lot of dependencies but the bulk of the library is already in checkbox with the base glmark2 package.

```bash
# after installing just glmark2
❯ sudo apt install glmark2-es2 glmark2-es2-wayland glmark2-wayland
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'glmark2-es2-x11' instead of 'glmark2-es2'
The following NEW packages will be installed:
  glmark2-es2-wayland glmark2-es2-x11 glmark2-wayland
0 upgraded, 3 newly installed, 0 to remove and 1 not upgraded.
Need to get 693 kB of archives.
After this operation, 1,916 kB of additional disk space will be used.
```

## Documentation

The only thing I'm not sure about are the dependencies for ubuntu 16 and 18. I've checked that the packages exist on them and can be `apt install`'ed, but I'm not sure why it wasn't there before.

## Tests

Manually checked that the packages exist on 16 ~ 24


pieqq (approved):
Looks good to me. For the record, I checked in the [available packages for xenial](http://archive.ubuntu.com/ubuntu/dists/xenial/) and indeed, even `glmark2-es2-wayland` is there.

PR #1888, opened 2025-04-24
Document how to work on python3.5 (infra)

Hook25:
## Description

Iterating on Python3.5 is hard and confusing the first time one does it. I've created this new method which is pretty hands off, 1 command and you are in the environment to iterate freely.

Minor: while dogfooding this, I've discovered that ancient versions of painbox dont pull psutil because they use the setup.cfg file, which is wrong. This also fixes that.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1878

## Documentation

This is documentation

## Tests

Dogfooded the how-to, just run through it and test the environment by starting Checkbox / running some tests from the manage.py

### DIFF
--- a/checkbox-ng/setup.cfg
+++ b/checkbox-ng/setup.cfg
@@ -4,6 +4,7 @@ sign=true
 packages=find_namespace:
 include_package_data=true
 install_requires=
+  psutil
   packaging
   requests >= 1.0
   urwid >= 1.1.1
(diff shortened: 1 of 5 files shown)


fernando79513 (approved):
LGTM +1!
Just so you know, in case you want to include them, I had to install these additional packages to make the providers/base tests pass (I also had to install psutil, but that will be merged in this PR):
```
apt install libgl1 g++
pip install natsort opencv_python PyYAML libsvm
```
We could add that to the cloud init script, but maybe that will cause some issues with the versions in the future. I leave it up to you.

PR #1894, opened 2025-04-29
Change the user on failed service check (bugfix)

Hook25:
## Description

The failed_permission check fails to collect logs on *some* platforms because the normal user on those platforms doesn't have the permission to read the journal on *some* services. There is no reason not to run this as root.

See:
```
Found 1 failed units

Failed units:
  UNIT                      LOAD   ACTIVE SUB    DESCRIPTION
● plymouth-show-run.service loaded failed failed Display state on splash

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.
1 loaded units listed.

Logs for plymouth-show-run.service:
Hint: You are currently not seeing messages from other users and the system.
      Users in groups 'adm', 'systemd-journal' can see all messages.
      Pass -q to turn off this notice.
No journal files were opened due to insufficient permissions.
```

## Resolved issues

Fixes: CHECKBOX-1738

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/providers/base/units/power-management/jobs.pxu
+++ b/providers/base/units/power-management/jobs.pxu
@@ -204,7 +204,7 @@ requires:
   cpuinfo.other != 'emulated by qemu'
 user: root
 environ: RTC_DEVICE_FILE
-command: 
+command:
   if [[ -n "$RTC_DEVICE_FILE" ]]; then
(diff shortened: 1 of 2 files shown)


fernando79513 (approved):
LGTM+1!

PR #1898, opened 2025-04-30
Add manifest for fde tests (BugFix)

LiaoU3:
## Description
Add manifest for FDE tests, so that it could be skipped easily while running manual tests.

## Resolved issues
N/A

## Documentation
N/A

## Tests
N/A

### DIFF
--- a/providers/base/units/disk/encryption.pxu
+++ b/providers/base/units/disk/encryption.pxu
@@ -3,10 +3,12 @@ category_id: com.canonical.plainbox::disk
 plugin: shell
 template-engine: jinja2
 user: root
+imports: from com.canonical.plainbox import manifest
 requires:
   executable.name == 'lsblk'
   executable.name == 'dmsetup'
   executable.name == 'cryptsetup'
+  manifest.has_fde == 'True'
 _summary: Test that Full Disk Encryption is in use
 _purpose:
   Examine the system to detect if one of the standard full disk encryption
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
This looks ok to me but it makes me wonder if we should somehow just use 
```
(manifest.has_tpm == 'True' or manifest.has_tpm2 == 'True')
```

What's the difference here?

LiaoU3:
> This looks ok to me but it makes me wonder if we should somehow just use
> 
> ```
> (manifest.has_tpm == 'True' or manifest.has_tpm2 == 'True')
> ```
> 
> What's the difference here?
Like you said, of course TPM and secureboot are prerequisites for enabling FDE. However, a device has TPM supported does not really need to enable FDE. For instance, a laptop nowadays has TPM and secureboot supported, however, we currently do not enable FDE on them now.

pieqq:
Is it possible to detect if FDE is enabled for a given image? If so, we could use a resource job instead to determine whether or not to test FDE-related stuff.

My concern with using a manifest is that FDE is enabled by the image install/config, not by the hardware, so in the lab it's going to be more complicated to define whether or not FDE tests should be run (if we test a stock Ubuntu version, for instance, FDE wouldn't be enabled, so if the manifest says it is using FDE, we will get wrong results).

Hook25:
I'm pretty sure at least for CI this will be a bit of a mess either way. I'm pro using having this new manifest, but just know this will cause issues once we start enforcing manifests, given that if enabled, it breaks the image.

PR #1899, opened 2025-04-30
Run hotspot test with root (Bugfix)

### LINKED ISSUE #1873: Wi-Fi access point test case do not run as root
### Bug Description

For the other Wifi checkbox test (that use nmcli). They are running as root https://github.com/canonical/checkbox/commit/b130346409

The topic on where network manager require root permission in different use case is properly a different topic but to align on the test environment/setup. I would suggest we do the same for Wi-Fi access point test (wireless/nmcli_wifi_ap*)

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

I think the wireless/nmcli_wifi_ap* tests are failing in 24.04 Server/Desktop image which are NOT running under GUI session(e.g. GNOME)

### Expected Result

The test should pass. Or the test should not fail because of autorization issue on nmcli.

### Actual Result

Test fail due to permission issue in nmcli when it is not running under GNOME session.

[nmcli_wifi_ap.txt](https://github.com/user-attachments/files/19793156/nmcli_wifi_ap.txt)

### Environment

- Ubuntu Desktop/Server 24.04
- Qualcomm RB3 Gen 2 

### Relevant log output

```shell
=============================
Attempt 1/5 (function 'main')
=============================

======================================
Attempt 1/5 (function 'device_rescan')
======================================
## Calling a rescan
+ nmcli d wifi rescan
Error: org.freedesktop.NetworkManager.wifi.scan request failed: not authorized.
Attempt 1 failed:
Command '['nmcli', 'd', 'wifi', 'rescan']' returned non-zero exit status 1.

Waiting 4.98 seconds before retrying...
```

### Additional context

I believe this issue started to appear since 24.04 after some network manager changes. But since other wifi test already use root to run. So this only happen on wifi_ap test only and not all the platform run this test.

hkhonming:
## Description

nmcli require authorization to run under a non GUI session. Other wireless testing relying on nmcli is already running with root. So this commit only align the setup as the other wireless/nmcli based tests

## Resolved issues

This PR fix #1873 

## Tests

Running wireless/nmcli_wifi_ap_bg and wireless/nmcli_wifi_ap_a tests under non GUI session (SSH) no longer failed due to authorisation error

### DIFF
--- a/providers/base/units/wireless/nm-hotspot.pxu
+++ b/providers/base/units/wireless/nm-hotspot.pxu
@@ -5,6 +5,7 @@ template-engine: jinja2
 template-unit: job
 id: wireless/nmcli_wifi_ap_bg_{{ interface }}
 template-id: wireless/nmcli_wifi_ap_bg_interface
+user: root
 _summary: Create 802.11b/g Wi-Fi AP on {{ interface }} using NetworkManager
 _purpose: Create an 802.11b/g Wi-Fi access point on a specified interface using NetworkManager command-line tools.
 command:
@@ -30,6 +31,7 @@ template-engine: jinja2
 template-unit: job
 id: wireless/nmcli_wifi_ap_a_{{ interface }}
 template-id: wireless/nmcli_wifi_ap_a_interface
+user: root
 _summary: Create 802.11a Wi-Fi AP on {{ interface }} using NetworkManager
 command:
   net_driver_info.py "$NET_DRIVER_INFO"


Hook25 (approved):

Hook25:
Can you sign your commits please?

Hook25:
Ported and signed

hkhonming:
@Hook25 , thanks for porting and merging the change. I am out in the Product sprint.

I double check my commit. I did find my signature there.

`
hmhui@hmhui-Latitude-3420:~/checkbox$ git log --show-signature -2
commit 5280ab2685d53fb26b368142f244493ce7a11bc5 (HEAD -> 1873-use-root-for-hotspot-test, origin/1873-use-root-for-hotspot-test)
gpg: Signature made Wednesday, April 30, 2025 AM10:06:45 BST
gpg:                using RSA key 47040F5D32E017B5AA5DB6C451A4B38485D109C1
gpg:                issuer "honming.hui@canonical.com"
gpg: Good signature from "Hon Ming Hui <honming.hui@canonical.com>" [ultimate]
Author: Hon Ming Hui <honming.hui@canonical.com>
Date:   Tue Apr 29 18:01:21 2025 +0100

    Run hotspot test with root
    
    nmcli require authorization to run under a non GUI session. Other wireless
    testing relying on nmcli is already running with root. So this commit
    only align the setup as the other wireless/nmcli based tests

`

I wanna know what did I missed so that next time my PR will be in a good shape.

PR #1902, opened 2025-05-01
Added base pre-commit hooks to include whitespace and EOL checks.

leander-dsouza:
## Description

- I have added the following hooks as part of the pre-commit:
   * `end-of-file-fixer`
   * `mixed-line-ending`
   * `trailing-whitespace`

   This prevents empty whitespace and additional newlines from registering as a separate commit during development.
- Added the pre-commit to the workflow.
- For future updates, we can also include `black` to be part of the pre-commit.

## Resolved issues

* N/A

## Documentation

* This PR focuses mainly on formatting fixes, so probably for development, an additional `pre-commit` dependency could be introduced:

   ```bash
   sudo apt-get install pre-commit
   ```

## Tests

* The formatting verification can be done locally using:

   ```bash
   pre-commit run --all
   ```


Hook25:
Hello, thanks for this contribution but it is basically impossible for me to review it as it is right now. Github is clearly not too happy with 5+k lines contributions. Checked the changes locally and it seems to track what the PR title states.

More on the merit of pre-commit hooks, this is a deviation from how we usually work on this project, we don't use pre-commit hooks and we rely on quality pipelines on github. Checkbox is also opinionated on linters in providers that are run via `./manage.py test`. All this to say that this seems to break away and create yet another testing facility for little to no value in its current form (these are surely nice to have, but clearly not too major imo, beside, in their current form they touch vendorized code which is clearly sub-optimal). For all these reasons, I think this needs more discussion before moving forward. Feel free to reply here if you disagree.

leander-dsouza:
Thank you for the detailed insight, Massimiliano :)
I definitely now understand why this may be sub-optimal in the case of vendor code. My main idea in the first place was to prevent empty spaces from generating as a separate commit.

PR #1906, opened 2025-05-05
Kioxia storage detect (bugfix)

Hook25:
## Description

Currently this specific revision of a Kioxia USB key is recognized as a mediacard instead of a usb key. The reason is that the model id is `TransMemory`, which contains `SM` (case insensitive), which is part of the `CARD_READER_RE` categorization regex. This wrongly identifies the usb key as a mediacard. To resolve this I have removed the `SM` part of the regex that from my research seems to be present in [SmartMedia cards](https://en.wikipedia.org/wiki/SmartMedia) (please correct me if I'm wrong) wich are an ancient and dead standard we can safely drop support for.

Minor: This also adds some output to the detect job so that it easier to debug what is going wrong

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1229

## Documentation

N/A

## Tests

Added a stripped down udev and lsblk and a new test to verify that all works according to plan


providers/base/units/usb/usb.pxu
```diff
@@ -228,6 +228,7 @@ category_id: com.canonical.plainbox::usb
 plugin: shell
 _summary: Detect storage partitions on a device on the USB bus
 command:
+  echo "Searching udev for a partitioned device connected via usb"
```
pieqq, line 231:
```suggestion
  echo "Searching udev for a partitioned device connected via usb..."
```

I like adding ellipsis at the end of such a sentence to notice that something else is coming afterwards (in that case, some string about USB busses, or an error message)

pieqq (dismissed):
Thanks for this!

I added a suggestion, but it can land as is.

pieqq (approved):

PR #1907, opened 2025-05-06
Skip journal collection in `merge-submissions` (Bugfix)

### LINKED ISSUE #1905: `merge-submissions` subcommand was incorrectly including the journal files from the machine that executed the merge, not the DUT
### Bug Description

When using `checkbox-cli merge-submissions submission1.tar.xz submission2.tar.xz -o output.tar.xz`, checkbox would include a massive list of journal outputs in the final `submission.json` from the machine where `merge-submissions` was used, not from the DUTs nor the input tar files. This leads to performance hits with the journal collection step taking more than 30 seconds to finish for 2 to 3 input files. 

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Download these submissions
   - https://certification.canonical.com/hardware/202504-36675/submission/426853/
   - https://certification.canonical.com/hardware/202504-36675/submission/426512/
   - https://certification.canonical.com/hardware/202504-36675/submission/426291/
   Any submission works, these are the ones that took 30 seconds to merge in the description
2. `checkbox-cli merge-submissions` with the above submissions
3. In the submission.json's `["system_information"]["journalctl"]["outputs"]` array, there should be at least 1 entry that has the "_HOSTNAME" key. This key is the same as the machine that did the merge, not the DUT's, which indicates that all the journal collected here are not from the DUT. 

### Expected Result

The submission.json file doesn't include info that's not from the DUT

### Actual Result

`submission.json` contains journal from the machine where the `merge-submissions` command is used

### Environment

- OS: ubuntu desktop 24.04
- Checkbox type: snap
- Checkbox version: 4.5.0.dev12

### Relevant log output

```shell

```

### Additional context

I think an easy way around this is to just use a separate `checkbox.json` template that doesn't have this line for `merge-submissions` and `merge-reports`:
```
{%- set system_information = state.system_information -%}
```
and all the stuff in the template that references "system_information"

tomli380576:
## Description

This PR fixes #1905 by introducing an option in the tar exporter to use a json template that doesn't have the line `{%- set system_information = state.system_information -%}`

## Resolved issues

#1905 

Also added a few type annotations

## Documentation

The basic idea is to avoid accessing `state.system_information` altogether when using merge-submissions since it has a side effect of collecting journals of the machine that executed the merge. 

This PR implements this solution by adding another exporter unit `json-nosysinfo` and an option `OPTION_SKIP_SYSINFO_IN_JSON` in the `TARSessionStateExporter`. The merge-submissions subcommand would then pass in the option when calling
```py
exporter = self._create_exporter(
    "com.canonical.plainbox::tar", ["skip-sysinfo-in-json"]
)
```
and the tar exporter will use the json template that doesn't have state.system_information

The `checkbox-nosysinfo.json` might look like a lot of changes but it actually only has 2 lines of diffs compared to `checkbox.json`

## Tests

- Original unit tests
- Comparing the merged submission.json without the `system_information` key with current implementation and they produced the same result.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/merge_reports.py
+++ b/checkbox-ng/checkbox_ng/launcher/merge_reports.py
@@ -24,6 +24,7 @@ import json
 import os
 import tarfile
 from tempfile import TemporaryDirectory
+from typing import List, Any
 
 from plainbox.impl.ctrl import gen_rfc822_records_from_io_log
 from plainbox.impl.providers.special import get_exporters
@@ -55,12 +56,14 @@ class MergeReports:
             help="save combined test results to the specified FILE",
         )
 
-    def _parse_submission(self, submission, tmpdir, mode="list"):
+    def _parse_submission(
+        self, submission: str, tmpdir: TemporaryDirectory, mode="list"
+    ):
         try:
             with tarfile.open(submission) as tar:
                 tar.extractall(tmpdir.name)
                 with open(os.path.join(tmpdir.name, "submission.json")) as f:
-                    data = json.load(f)
+                    data = json.load(f)  # type: dict[str, Any]
             for result in data["results"]:
                 result["plugin"] = "shell"  # Required so default to shell
                 result["summary"] = result["name"]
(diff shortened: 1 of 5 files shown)


Hook25:
Ok, this is very complicated and duplicates the template, I would really like not to do this. Additionally, it doesn't solve the underlying issue, you loose the system information, which isn't great imo.

PR #1910, opened 2025-05-06
system_information from submission json (bugfix)

Hook25:
## Description

`system_information` was not taken into account in merge reports. This made it so, when creating the exported tar, Checkbox would re-calculate it instead of taking it from the submissions, slowing down the process significantly (as it has to do it once per submission) but most importantly, making it wrong, as the merge process is usually done on a random machine.

Note: This also makes the process about 20% faster, as it avoids unpacking and repacking the xz twice.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1887
Fixes: https://github.com/canonical/checkbox/issues/1905

## Documentation

Documented why the tar change was done
The rest of the change adheres to the code around it

## Tests

Submission.json now generates correctly without querying the `system_information` collector

Exported before and after the multi export change. With the same subs diff reports only one change (date now)
```
(venv)  z >  diff -r *
diff -r before/submission.html after/submission.html
6132c6132
<                 <p>This report was created using Checkbox 4.2.1.dev38+gc9201b347.d20250107 on 2025-05-06T13:03:08</p>
---
>                 <p>This report was created using Checkbox 4.2.1.dev38+gc9201b347.d20250107 on 2025-05-06T13:57:48</p>
```

### DIFF
--- a/checkbox-ng/plainbox/impl/session/system_information.py
+++ b/checkbox-ng/plainbox/impl/session/system_information.py
@@ -27,6 +27,13 @@ class CollectorOutputs(dict):
 
     @classmethod
     def from_dict(cls, dct):
+        """
+        This loads the class from a dct
+        Note: this ignores the version the CollectionOutputs was generated from
+              as we have no good way of using it for now
+        """
+        # ignore the version, this shouldn't matter as this is only used in
+        # merge-reports and merge-submissions
         _ = dct.pop("version")
         to_load = {
             name: CollectionOutput.from_dict(value)
--- a/checkbox-ng/plainbox/impl/session/system_information.py
+++ b/checkbox-ng/plainbox/impl/session/system_information.py
@@ -25,6 +25,15 @@ class CollectorOutputs(dict):
         to_dump["version"] = self.COLLECTOR_OUTPUTS_VERSION
         return json.dumps(to_dump, indent=4)
 
+    @classmethod
+    def from_dict(cls, dct):
+        _ = dct.pop("version")
+        to_load = {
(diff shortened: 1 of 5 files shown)


checkbox-ng/plainbox/impl/session/system_information.py
```diff
@@ -25,6 +25,15 @@ def to_json(self) -> str:
         to_dump["version"] = self.COLLECTOR_OUTPUTS_VERSION
         return json.dumps(to_dump, indent=4)
 
+    @classmethod
+    def from_dict(cls, dct):
```
pieqq, line 29: Can you document this a tiny bit? Especially why `version` gets discarded? It's not immediately clear you did this because of the merge-submissions feature (especially since it's part of `system_information`).

pieqq (changes requested):
I followed @tomli380576 's original steps and submissions, and it seems to work better now. I see a `localhost.localdomain` `_HOSTNAME` which is not the one my local machine is using, and I see some OEM kernel related messages in journalctl which look like the ones in the submissions.

I just have a little documentation request for the `from_dict` function.

checkbox-ng/plainbox/impl/session/system_information.py
Hook25: done

fernando79513 (approved):
LGTM+1!
87.5% of patch coverage can be easily rounded to 90%.
It's still a mess, but a faster and more accurate mess, XD.

fernando79513 (approved):

PR #1917, opened 2025-05-09
Run hotspot test as root (bugfix)

Hook25:
nmcli require authorization to run under a non GUI session. Other wireless testing relying on nmcli is already running with root. So this commit only align the setup as the other wireless/nmcli based tests

Ported from: https://github.com/canonical/checkbox/pull/1899


fernando79513 (approved):
LGTM+1!

PR #1918, opened 2025-05-09
Make dkms test run as root (bugfix)

Hook25:
## Description

As per sru review, there is no reason why this runs as normal user, on some system this fails due to a lack of permissions (not being added to a specific group.) Lets make this run as root.
## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
Lets hit theese jobs with one more root!!!
LGTM+1!

PR #1921, opened 2025-05-13
add a check if wakealarm is supported (fix)

aglinserer:
## Description

This avoids a failing test and skips it instead if the rtc does not support an automatic wakeup.
The `suspend_advanced_auto` job is failing even though it cannot be run when the wakealarm is not supported.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


providers/base/units/suspend/suspend.pxu
```diff
@@ -251,7 +251,7 @@ category_id: com.canonical.plainbox::suspend
 id: suspend/suspend_advanced_auto
 requires:
     sleep.mem == 'supported'
-    rtc.state == 'supported'
+    rtc.state == 'supported' and rtc.wakealarm == 'supported'
```
pieqq, line 254:
```suggestion
    rtc.state == 'supported'
    rtc.wakealarm == 'supported'
```

You can just place the new requirement on another line and it acts as an `and` (see [doc](https://canonical-checkbox.readthedocs-hosted.com/en/stable/reference/units/job.html#cmdoption-job-arg-requires))

pieqq (changes requested):
Thanks for your contribution!

I left a comment inline.

Can you describe the use case you have and the issue you're encountering? This job is very sensitive because it's used in almost every test plan we maintain, so modifying it is always a bit risky.

aglinserer:
> Can you describe the use case you have and the issue you're encountering? This job is very sensitive because it's used in almost every test plan we maintain, so modifying it is always a bit risky.

Currently, I am running Checkbox on hardware that does not support the wakealarm functionality. My thought is that if the interface or functionality is not supported, failing this test - based solely on rtc.state == supported and mem.sleep == supported - is incorrect. I would assume that checking for wakealarm functionality should be a requirement for this since it depends on this.

After giving it some more thought you could also interpret suspendability as a basic requirement. Then failing this would be the correct choice. This thought only occured after creating this pull request :)

PR #1925, opened 2025-05-20
Rename genio brightness so that it doesn't clash (bugfix)

Hook25:
## Description

This removes the warning, but most importantly, it fixes the manual test in genio, because else the wrong script is used (as only the first one is succesfully symlinked)

## Resolved issues

N/A

## Documentation

N/A

## Tests

Tests are still working locally.


fernando79513 (approved):
Thanks Max!!
Ez +1

PR #1928, opened 2025-05-26
Exclude v4l2-compliance to not break existing certs (BugFix)

tomli380576:
## Description

This PR is a short workaround for v4l2-compliance tests that was breaking on older ubuntu versions. It moves the v4l2-compliance tests to it's own test plan and is not included anywhere for now.

## Documentation

v4l2-compliance tests now lives in its own test plan `camera-v4l2-compliance`

### DIFF
--- a/providers/certification-client/units/client-cert-desktop-24-04.pxu
+++ b/providers/certification-client/units/client-cert-desktop-24-04.pxu
@@ -90,6 +90,7 @@ nested_part:
     audio-cert-automated
     bluetooth-cert-automated
     camera-cert-automated
+    camera-v4l2-compliance
     thunderbolt-cert-automated
     monitor-gpu-cert-automated
     graphics-gpu-cert-automated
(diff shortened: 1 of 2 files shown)


providers/certification-client/units/client-cert-desktop-24-04.pxu
```diff
@@ -90,6 +90,7 @@ nested_part:
     audio-cert-automated
     bluetooth-cert-automated
     camera-cert-automated
+    camera-v4l2-compliance
```
fernando79513, line 93: I think we should wait until we discuss this on Friday to see which test-plans should include these non-blocker tests, but we can merge already the rest of the PR, so it does not mess up the SRU tests.

fernando79513 (commented):
I would wait until we discuss where this should be included to include it in the `client-cert-desktop-24-04` test-plan, but I think we can merge the rest of the PR

providers/certification-client/units/client-cert-desktop-24-04.pxu
tomli380576: Sounds good, I'll remove this until we discuss this on Friday

PR #1935, opened 2025-05-27
Add before field (Breaking)

fernando79513:
## Description

This PR adds a new field called "before" to the jobs. This field allows a new type of dependency for the jobs.
The way the dependency manager sorts the jobs has not changed. Instead, the way the before flag works is by adding the current job as an after dependency of the jobs listed in the before field.

These references are added to the job object as a new attribute and included to the `after_deps` set when retrieving the dependency set.

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-1919

## Documentation

The "before" flag has also been added to the corresponding section in the checkbox documentation.

## Tests

Unit tests for this feature have been added In the PR

### DIFF
--- a/docs/reference/units/job.rst
+++ b/docs/reference/units/job.rst
@@ -148,10 +148,10 @@ Following fields may be used by the job unit:
     In the case of the before field, if the job was not previously included in
     the test plan, it won't be added and the dependency will be ignored.
 
-    If a job has another in the before field and uses with the 
-    ``also-after-suspend`` flag, both of the job and the generated  sibling
-    will include the same before dependency. This means that the referenced
-    job will be moved to the end of the test plan, after the
+    When a job lists another job in its ``before`` field and has the
+    ``also-after-suspend`` flag, both the original job and its generated sibling
+    share that identical ``before`` dependency. As a result, the referenced job
+    is pushed to the end of the test plan, positioned after the
     ``also-after-suspend`` sibling.
 
 .. option:: salvages
--- a/docs/tutorial/using-checkbox/advanced-commands.rst
+++ b/docs/tutorial/using-checkbox/advanced-commands.rst
@@ -128,9 +128,9 @@ fields available from the jobs. To see what fields are available, run:
 .. note::
 
     There is currently a bug with this command. It does not actually list all
-    the available fields, but the fields that are used in all the jobs in the
-    scope. In most cases, this will be all the fields available, but a new field
-    that is not used in any job will not be listed.
+    the available fields. It shows only the fields that are used throughout the
+    jobs that are in scope. In most cases, this will be all the fields
+    available, but a new field that is not used in any job will not be listed.
 
 To create a table listing each job id and their summary, run:
 


checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -133,12 +135,31 @@ def get_dependency_set(self, job, job_list=None):
         )
         return result
 
+    def add_before_deps(self, job, job_map):
+        """
+        Add a all "before" references declared in a job to the corresponding
```
pieqq, line 140:
```suggestion
        Add all "before" references declared in a job to the corresponding
```

checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -133,12 +135,31 @@ def get_dependency_set(self, job, job_list=None):
...
+        jobs as an "after" dependency in the before_references set.
+
+        If a job (B) has a "before" field, we add this job as an "after"
+        dependency to the job (A).
```
pieqq, line 144: This is not very clear, since job A and job B are not defined.

You could show a very simple job definition, and how this mechanism turns

```
id:job-a

id: job-b
before: job-a
```
into

```
id:job-a
after: job-b

id: job-b
```

checkbox-ng/plainbox/impl/unit/job.py
```diff
@@ -682,11 +687,40 @@ def visit_Text_node(visitor, node: Text):
...
+        Compute and return a set of after dependencies.
+
+        If this job (A) is referenced in the "before" field of another job (B),
+        then the job B will be added to the set of dependencies of A. This way
+        the dependency graph only works with "after" dependencies:
+
+            id: A          id: A
+                       ->  after: B
+            id: B      ->
+            before: A      id: B
```
pieqq, line 706: Ha, this is where job A and B are defined :) (see my other comment above)

docs/reference/units/job.rst
```diff
@@ -139,6 +139,14 @@ Following fields may be used by the job unit:
     run (regardless of the outcome). Multiple job names, separated by spaces,
     can be specified.
 
+.. option:: before
+
+    (optional). If specified, the job will only run before all the listed jobs
+    have run (regardless of the outcome). Multiple job names, separated by
+    spaces, can be specified.
+    The way the before flag works is by adding the current job as an after
```
pieqq, line 147:
```suggestion
    The way the before flag works is by adding the current job as an ``after``
```

docs/reference/units/job.rst
```diff
@@ -139,6 +139,14 @@ Following fields may be used by the job unit:
     run (regardless of the outcome). Multiple job names, separated by spaces,
     can be specified.
 
+.. option:: before
```
pieqq, line 142: There should probably be some kind of warning when using this field alongside the `also-after-suspend` flag, as this field will also be present in the generated sibling, which means any job listed in the `before` field will be pushed much later in the test plan, which might not be the desired behavior!

pieqq (approved):
I tested it locally by butchering jobs in the `client-cert-desktop-24-04-automated` test plan and seeing what was happening, and it works as expected in my quick test.

If possible, add more warnings in the documentation about this field.

metabox/metabox/scenarios/basic/ordering.py
```diff
@@ -0,0 +1,89 @@
...
+from metabox.core.actions import (
+    AssertPrinted,
+    Start,
+)
+from metabox.core.scenario import Scenario
+from metabox.core.utils import tag
+
+
+@tag("ordering")
+class OrderingDepends(Scenario):
```
Hook25, line 28: Uhm, why is this remote and local, but uses run, so it is just local?

metabox/metabox/scenarios/basic/ordering.py
```diff
@@ -0,0 +1,89 @@
...
+    steps = [
+        Start("run 2021.com.canonical.certification::ordering_depends"),
```
Hook25, line 30: Why do you use run instead of a launcher?

metabox/metabox/scenarios/basic/ordering.py
```diff
@@ -0,0 +1,89 @@
...
+        AssertPrinted(
```
Hook25, line 31: why do you do this this way? `AssertPrinted` and `Expect` are ordered matches
```
AssertPrinted('a')
AssertPrinted('b')
```
means that `a` must be after `b`!

metabox/metabox/metabox-provider/units/ordering.pxu
Hook25, line 1: oh man, the ids here could use some improvement, at least put a summary to state what each thing is supposed to test

docs/tutorial/using-checkbox/advanced-commands.rst
```diff
@@ -123,6 +125,13 @@ fields available from the jobs. To see what fields are available, run:
     The underscore before some of the fields names simply means the content
     of this field can be translated into another language.
 
+.. note::
+
+    This command does not actually list all the available fields, but the
+    fields that are used in all the jobs in the scope. In most cases, this
+    will be all the fields available, but a new field that is not used in
+    any job will not be listed.
```
Hook25, line 133: This is false, its the abstract tree viewer or whatever it is called that is bugged and should be fixed, this is not an intended feature, its a bug!

docs/reference/units/job.rst
Hook25, line 147: Implementation detail, not documentation worthy imo, also wording is a bit wonky (wdym regardless of outcome, you are before the other test, how would you know the outcome :laughing:). Please do also include the fact that it doesn't pull the other job in the run if its not present

checkbox-ng/plainbox/impl/unit/job.py
```diff
@@ -939,6 +974,23 @@ class fields(SymbolDef):
...
+                CorrectFieldValueValidator(
+                    lambda value, unit: (
+                        unit.get_before_dependencies() is not None
+                    )
+                ),
+                UnitReferenceValidator(
+                    lambda unit: unit.get_before_dependencies(),
+                    constraints=[
+                        ReferenceConstraint(
+                            lambda referrer, referee: referee.unit == "job",
```
Hook25, line 988: Uhm, I'm a bit rusty on this, does it mean this wouldn't work for templated jobs? if so, I dont know if this 100% what we need

checkbox-ng/plainbox/impl/ctrl.py
```diff
@@ -133,12 +135,31 @@ def get_dependency_set(self, job, job_list=None):
...
+        """
+        before_deps = job.get_before_dependencies()
+        for dep_id in before_deps:
+            # Check if the dep_id is a valid job
+            if dep_id not in job_map:
```
Hook25, line 149: My main point: I think this should not be the case! If the job id is not in the job map, we are fine

How often is this piece of code ran? Whenever adopting a new result? Else this may not work for templated jobs

Hook25 (changes requested):
See below, please consider not failing on missing job

metabox/metabox/metabox-provider/units/ordering.pxu
fernando79513: Yes, I will add some explanation to the metabox test.

metabox/metabox/scenarios/basic/ordering.py
fernando79513: Oh, that's what you would expect. But the assert printed looks for a string printed in the whole output, not just the last line:
https://github.com/canonical/checkbox/blob/db66b4535f52b89821ba1c48ffef5e90de8fc2bd/metabox/metabox/core/scenario.py#L135
I knew you would not like the regex trickery, but the approach you proposed does take into acount the order in which they are printed.

metabox/metabox/scenarios/basic/ordering.py
fernando79513: That's me not having pushed these changes, I used the launcher for `local + remote` runs

checkbox-ng/plainbox/impl/unit/job.py
fernando79513: I've just tested that you can include templated jobs both in the before field and include templated jobs as a dependency:
```
unit: template
id: testC{id}
before: testE

unit: template
id: testD{id}
before: testE

id: testE

id: ordering_resource
plugin: resource
command:
    echo 'id: 1'
    echo
    echo 'id: 2'

id: template_order
unit: test plan
include:
 testE
 testD.*
 testC.*

 ☑ : testC2
 ☑ : testD2
 ☑ : testC1
 ☑ : testD1
 ☑ : testE
```

checkbox-ng/plainbox/impl/depmgr.py
```diff
@@ -320,93 +336,217 @@ def __init__(self, job_list):
...
+        # Solve again for order dependencies, using the pulled jobs as the
+        # new visit list
+        self._clear_state_map()
+        for job in self._pull_solution:
+            self._visit(job)
+
+        # Perform a sanity check to ensure that no jobs have been added or
+        # removed from the solution.
+        if len(self._order_solution) != len(self._pull_solution):
+            raise ValueError(
```
pieqq, line 397: I don't know how this can happen, but I know the output would be better if it contained the diff of `self._order_solution` and `self._pull_solution`, because otherwise debugging will be difficult.

checkbox-ng/plainbox/impl/depmgr.py
```diff
@@ -320,93 +336,217 @@ def __init__(self, job_list):
...
+    def _visit(self, job, trail=None, pull=False):
         """
         Internal method of DependencySolver.
 
         Called each time a node is visited. Nodes already seen in _visited are
         skipped. Attempts to enumerate all dependencies (both direct and
         resource) and resolve them. Missing jobs cause DependencyMissingError
         to be raised. Calls _visit recursively on all dependencies.
+        If the pull flag is set to True, the dependencies are pulled form the
+        search_map, if not, not.
```
pieqq, line 418:
```suggestion
        Dependencies are pulled form the search_map only if the pull flag is
        set to True.
```

pieqq (changes requested):
two little suggestions

docs/tutorial/using-checkbox/advanced-commands.rst
fernando79513: Is this solved by your new PR, or should I keep this note??

Hook25: didn't fix it yet, if you do have a sec, please do so. For now you could leave this comment, even though it is not relevant to this pr. For now add the before in there!

pieqq (approved):
I left some inline comments regarding the documentation. It would be good if you took a look at it, but otherwise I think the feature works as expected.

I've done the following test, since we have an issue with debsum being run after a wireless test that modifies netplan config:

1. Moved the `misc-client-cert-automated` nested part in the `client-cert-desktop-24-04-automated` test plan to the very end. This nested part includes `miscellanea/debsums`.
1. Ran `checkbox-cli list-bootstrapped "com.canonical.certification::client-cert-desktop-24-04-automated" > /var/tmp/misc_np_at_the_end`
1. Looked at the result, and indeed the jobs from that nested part are all towards the end, including the debsum job.
1. Edited the `miscellanea/debsums` job to add `before: wireless/detect`
1. Ran `checkbox-cli list-bootstrapped "com.canonical.certification::client-cert-desktop-24-04-automated" > /var/tmp/misc_np_at_the_end_before_field_added`

Compared the two, and sure enough, all the jobs from the `misc-client-cert-automated` nested part are still at the end of the list, **except** `miscellanea/debsums` which sits just before `wireless/detect`. (this should fix our problem, so it means the feature works great! :)

PR #1942, opened 2025-06-02
Run wifi AP tests as root (ce-oem)

weizhenwu:
## Description
Update the wifi AP tests to run as root.
since 24.04, Network Manager requires permission to perform network configuration.

Reference: https://bugs.launchpad.net/carmel/+bug/2103387/comments/3

## Resolved issues
N/A

## Documentation
N/A

## Tests
Test submission on carmel devices: 
https://certification.canonical.com/hardware/202502-36339/submission/428352/


hanhsuan (approved):
LGTM. Would it be possible to change this to some kind of web service in the future ? ssh into host device might not be safe.

PR #1959, opened 2025-06-13
Make run_watcher print every line it reads from journal (BugFix)

p-gentili:
## Description

The error you get from any storage watcher job is quite useless, it's just a timeout error:

```
    <...>
    raise TimeoutError(""Task unable to finish in {}s"".format(timeout_s))
TimeoutError: Task unable to finish in 30s
```

This change makes it print every line it reads from journal, making it easier to understand what was going on in case of failure. I don't expect this log to get big enough in 30s to be a concern in terms of size.

## Resolved issues

N/A

## Documentation

N/A

## Tests

- Added unit test and updated the rest of them

### DIFF
--- a/checkbox-support/checkbox_support/scripts/run_watcher.py
+++ b/checkbox-support/checkbox_support/scripts/run_watcher.py
@@ -75,7 +75,6 @@ class StorageInterface(ABC):
     these methods
     """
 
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
Well done, +1

PR #1960, opened 2025-06-13
Update Readme by removing broken image (BugFix)

baconYao:
## Description

Update Readme file due to a non-existent image file.

## Resolved issues

截圖 2025-06-13 下午12 41 06

## Documentation

N/A

## Tests

N/A


Hook25 (approved):
thanks <3

Hook25:
bypassing the signature check because the party is trustworthy and the change is self contained in a readme file. 

Thanks again man :)

PR #1964, opened 2025-06-18
Allow non-exact matching for ids in cli subcommands (New)

Hook25:
## Description

It is often bothersome, and rarely useful, to include the namespace for all ids in the checkbox-cli subcommands.

This changes them all so that any id provided will be searched both as a `namespace::id` and a `?::id` input. This doesn't affect the user that provides the namespace, but makes it way more accessible for all users to use these tools.

Whenever this behaviour is not desirable, for whatever reason, `--exact` can be provided to all changed subcommands to have the old behaviour back. 

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1951

## Documentation

Documents the new `--exact` option

## Tests

Tested new functions added (most)
Launched manually the commands with the new and old behaviour, ie.
```
checkbox-cli show sru
checkbox-cli show com.canonical.certification::sru
checkbox-cli show com.canonical.certification::sru --exact
```
should all yield the same result

### DIFF
--- a/checkbox-ng/plainbox/impl/highlevel.py
+++ b/checkbox-ng/plainbox/impl/highlevel.py
@@ -23,6 +23,7 @@
 """
 
 from collections import OrderedDict
+import operator
 import logging
 
 from plainbox.impl.session.storage import WellKnownDirsHelper
--- a/checkbox-ng/plainbox/impl/highlevel.py
+++ b/checkbox-ng/plainbox/impl/highlevel.py
@@ -119,7 +119,7 @@ class PlainBoxObject:
 
     def find_children_by_name(self, names: list, exact=False):
         to_explore = [self]
-        name_founds = {name: [] for name in names}
+        name_matches = {name: [] for name in names}
 
         def match_f(a, b):
             # a is namespace::id, b may not contain namespace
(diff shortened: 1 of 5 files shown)


checkbox-ng/plainbox/impl/highlevel.py
```diff
@@ -116,6 +117,26 @@ def attrs(self):
...
+            match_f = operator.eq
+
+        while to_explore:
+            node = to_explore.pop()
+            matching = [name for name in names if match_f(node.name, name)]
+            for match in matching:
+                name_founds[match].append(node)
+            to_explore += node.children
+
+        return name_founds
```
fernando79513, line 138: Weird name, I'm not sure if it's correct English, I'd go for `name_matches`
```suggestion
        name_matches = {name: [] for name in names}

        def match_f(a, b):
            # a is namespace::id, b may not contain namespace
            return a == b or a.split("::", 1)[-1] == b

        if exact:
            match_f = operator.eq

        while to_explore:
            node = to_explore.pop()
            matching = [name for name in names if match_f(node.name, name)]
            for match in matching:
                name_matches[match].append(node)
            to_explore += node.children

        return name_matches
```

fernando79513 (commented):
I've just left two small comments. Can't wait to have this in place. I've just had to write twice the `com.canonical.certification` this morning.

pieqq (dismissed):

PR #1973, opened 2025-06-23
Ensure debsums job is run before wireless tests

pieqq:
## Description

Because of [a bug in netplan][1], the `miscellanea/debsums` will fail if it's run after any of the wireless tests.

Checkbox now supports the [before field for job units][2].

Since `wireless/detect` is a dependency for all of the wireless tests, using it ensures the `miscellanea/debsums` job will be called before any wireless tests happen.

[1]: https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2078759
[2]: https://github.com/canonical/checkbox/pull/1935

## Resolved issues

Fix https://warthogs.atlassian.net/browse/CHECKBOX-1916

### DIFF
--- a/providers/base/units/miscellanea/jobs.pxu
+++ b/providers/base/units/miscellanea/jobs.pxu
@@ -497,6 +497,7 @@ estimated_duration: 60
 id: miscellanea/debsums
 user: root
 requires: executable.name == 'debsums'
+before: wireless/detect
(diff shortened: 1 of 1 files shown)


pieqq:
This PR is not required anymore since Antoine provided another PR that fixes the issue: #1933

PR #1983, opened 2025-06-30
Use json encoding to avoid spamming remote API (breaking)

Hook25:
## Description

Checkbox Remote is very, very slow when used over a slow connection. This is mostly due to a reason: When returning a mutable objects over an rpyc function, any further query on that object is a rpyc call. This is most evident in the following situations:
- Test plan selection (iterating over the testplan list takes ~600 (number of testplans) rpyc calls
- Test selection (modifying the desired list takes one call per test that will be selected, so deselecting just one test could lead to hundreds of rpyc calls)
- Calculating the return code of a test run (This iterated over the whole job_state_map, leading to ~10k rpyc calls as undesired jobs are present when no reboots are done)
- Exporting (Additional Configurations that were created client side lead to hundreds of rpyc remote calls when used afterward)

Additionally (this is a bit of a bonus), every test on a medium to high latency will always pay 0.5s to "wait" for the test to be completed + the latency. This adds a backoff mechanism that starts from 0 and goes to .5s

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1956
Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1955

## Documentation

Documented this in the class docstring

## Tests

Added unit tests for /most/ of the api changed.

To try this change and simulate this issue locally modify `checkbox-ng/plainbox/vendor/rpyc/core/protocol.py` add to the serve function of the Connection class the following
```python
    def serve(self, timeout=1, wait_for_lock=True):  # serving
        import time
        time.sleep(.1)
```

This will simulate a 100ms latency (less than what it is Europe to Taiwan). Add the following test units to the metabox provider:
```
id: tons_of_jobs
_summary: Generates a lot of ids to fill the job_state_map
plugin: resource
shell: python
command:
  for i in range(100):
    print(f"id: stress_{i}")
    print()

unit: template
template-resource: tons_of_jobs
template-unit: job
id: {id}
template-id: template_tons_of_jobs
command:
  echo "Test {id}"
_summary: Test that stresses rpyc calls
flags: simple

unit: test plan
id: stress_remote_tps
_name: Stress test of the checkbox remote-api
bootstrap_include:
  tons_of_jobs
include:
  template_tons_of_jobs
```

A test run at 100ms latency of this testplan takes about 35minutes on main, 5m on this branch

Additionally, I've used this very ugly decorator to log all the calls made to the server to be able to track them down
```python
if os.getuid() == 0:
    log = open("/tmp/root_agent.log", "w+")
else:
    log = open("/tmp/usr_control.log", "w+")


def _log_call(f):

    # return f

    def _f(self, *args, **kwargs):
        global total
        total += 1
        import json

        log.write(json.dumps([str(x) for x in [total, *args, kwargs]]))
        log.write("\n")
        print(total, *args, kwargs)
        to_r = f(self, *args, **kwargs)
        if (
            not any(isinstance(to_r, x) for x in [str, int, float, bool])
        ) and to_r is not None:
            # breakpoint()
            ...
        return to_r

    return _f
```
Decorate all _handle_[...] functions in the Connection class to use this.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/stages.py
+++ b/checkbox-ng/checkbox_ng/launcher/stages.py
@@ -454,6 +454,11 @@ class ReportsStage(CheckboxUiStage):
 
     def _prepare_stock_report(self, report):
         try:
+            # This function is called by both remote and local. Local sa
+            # doesn't have this API so it will fallback to the "normal"
+            # Configuration type. In remote we get the agent Configuration type
+            # because the object is passed and mostly used by the agent, so
+            # creating this object as a netref saves a lot of rpyc calls
             ConfigurationType = self.sa.configuration_type()
         except AttributeError:
             ConfigurationType = Configuration
--- a/checkbox-ng/checkbox_ng/launcher/stages.py
+++ b/checkbox-ng/checkbox_ng/launcher/stages.py
@@ -453,9 +453,13 @@ class ReportsStage(CheckboxUiStage):
         self._export_fn = export_fn
 
     def _prepare_stock_report(self, report):
+        try:
+            ConfigurationType = self.sa.configuration_type()
+        except AttributeError:
+            ConfigurationType = Configuration
         new_origin = "stock_reports"
         if report == "text":
-            additional_config = Configuration.from_text(
+            additional_config = ConfigurationType.from_text(
                 textwrap.dedent(
                     """
                     [exporter:text]
@@ -473,7 +477,7 @@ class ReportsStage(CheckboxUiStage):
             )
             self.sa.config.update_from_another(additional_config, new_origin)
         elif report == "certification":
-            additional_config = Configuration.from_text(
+            additional_config = ConfigurationType.from_text(
                 textwrap.dedent(
                     """
                     [exporter:tar]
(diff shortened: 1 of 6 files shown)


checkbox-ng/checkbox_ng/launcher/stages.py
```diff
@@ -453,9 +453,13 @@ def _override_exporting(self, export_fn):
         self._export_fn = export_fn
 
     def _prepare_stock_report(self, report):
+        try:
+            ConfigurationType = self.sa.configuration_type()
+        except AttributeError:
+            ConfigurationType = Configuration
```
pieqq, line 459: Add some comments about what this does and why this needs to catch this AttributeError exception.

checkbox-ng/plainbox/impl/session/remote_assistant.py
```diff
@@ -141,9 +143,18 @@ def outcome(self):
...
+    Code in this class runs in the agent. Returning mutable types or receiving
+    mutable types as parameter from any of these functions creates an implicit
+    remote API (as any function/attribute used on the returned value will
+    result in a remote API call) and should therefore be avoided.
+    Favour creating a JSON API version of the function that returns the same
+    object but JSON encoded.
+    """
 
-    REMOTE_API_VERSION = 13
+    REMOTE_API_VERSION = 15
```
pieqq, line 157:
```suggestion
    REMOTE_API_VERSION = 14
```

pieqq (changes requested):
First feedback, will test it locally later.

checkbox-ng/plainbox/impl/session/remote_assistant.py
Hook25: done

pieqq (approved):

Issue #1978, opened 2025-06-26
miscellanea/secure_boot_mode test is being skipped


rodwsmith:
### Bug Description

The `miscellanea/secure_boot_mode` test is being skipped. The logged message is:

`job cannot be started: resource expression "manifest.has_secure_boot == 'True'" evaluates to false`

A quick scan of the code suggests that the `has_secure_boot` variable is defined in `providers/base/units/miscellanea/manifest.pxu`, but I don't see where it's being set.

It's unclear if the `has_secure_boot` variable is meant to indicate whether the machine supports Secure Boot or if it's meant to indicate that Secure Boot is active. If the latter, then this should definitely _not_ be used in this case. If the former, then it's more reasonable; however, the `miscellanea/secure_boot_mode` test is intended, by design, to fail if the machine does not support Secure Boot, since all UEFI machines should support Secure Boot at this time; if they don't, then that's a design flaw or a bug in the firmware. As the `miscellanea/secure_boot_mode` test is a non-blocking test, a failure for this reason would not gate certification, but discussions should take place between Canonical and the vendor in this case.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Launch Checkbox with the command `test-firmware`, or any other launcher that runs the `miscellanea/secure_boot_mode` test, on a system with UEFI firmware.
2. Observe the output

### Expected Result

The `miscellanea/secure_boot_mode` test should run. Whether it passes or fails depends on the firmware and whether Secure Boot is active, but the test should run.

### Actual Result

The test always fails to run.

### Environment

Reported by a partner; verified with the following:

- Ubuntu 22.04.5
- Checkbox 4.6.0~dev48~ubuntu22.04.1
- Submission: https://certification.canonical.com/hardware/201508-19150/submission/435242/

### Relevant log output

```shell
`job cannot be started: resource expression "manifest.has_secure_boot == 'True'" evaluates to false`
```

pieqq:
`has_secure_boot` is a manifest. When using Checkbox interactively, a screen (like the one at the beginning of [this tutorial](https://canonical-checkbox.readthedocs-hosted.com/en/stable/tutorial/using-checkbox/manifest.html)) is presented to the tester to help them choose the value for each of the required manifest entries for the test plan they are running.

[...]

PR #1984, opened 2025-06-30
Run Secure Boot test on all UEFI servers. (Bugfix)

rodwsmith:
Title: The miscellanea/secure_boot_mode test should always run on UEFI-based servers (Bugfix)

## Description

This PR adds a `[manifest]` section to the Server Certification `/etc/xdg/canonical-certification.conf` configuration file and sets the `has_secure_boot` variable to `True`. This causes the `miscellanea/secure_boot_mode` test to run on all UEFI servers. (Other conditions prevent it from running on non-UEFI servers.)

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/1978

## Documentation

## Tests

I installed the modified file on servers running both UEFI and non-UEFI firmware. The test ran on the former but not on the latter, both as in the past.


rodwsmith:
After internal discussions, we've decided on another way to do this, so I'm closing this PR.

Issue #1720, opened 2025-02-12
camera/roundtrip-qrcode doesn't find libmmal.so library


omar-selo:
### Bug Description

`camera/roundtrip-qrcode` job fails because of `OSError: libmmal.so: cannot open shared object file: No such file or directory`.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

Rerun http://10.102.156.15:8080/job/cert-rpi3aplus-arm64-core22-beta/

### Expected Result

The test should've passed.

### Actual Result

The test errors out because it can't find a particular library.

### Environment

See this submission https://certification.canonical.com/hardware/201907-27199/submission/411144/

### Relevant log output

```shell
Testing device name: vchiq

Input string: NlGGPCJAtc
Generating QR code...
Displaying on screen
Capture image of screen
Traceback (most recent call last):
  File "/tmp/nest-sb4_fxrb.16427a450e537741fc47ed5c55f4b9235c8cabe69831215cfcd65732683b2462/roundtrip_qr.py", line 128, in <module>
    main()
  File "/tmp/nest-sb4_fxrb.16427a450e537741fc47ed5c55f4b9235c8cabe69831215cfcd65732683b2462/roundtrip_qr.py", line 113, in main
    file = capture_rpi(name)
  File "/tmp/nest-sb4_fxrb.16427a450e537741fc47ed5c55f4b9235c8cabe69831215cfcd65732683b2462/roundtrip_qr.py", line 34, in capture_rpi
    import picamera
  File "/snap/checkbox/12338/checkbox-runtime/lib/python3.10/site-packages/picamera/__init__.py", line 72, in <module>
    from picamera.exc import (
  File "/snap/checkbox/12338/checkbox-runtime/lib/python3.10/site-packages/picamera/exc.py", line 41, in <module>
    import picamera.mmal as mmal
  File "/snap/checkbox/12338/checkbox-runtime/lib/python3.10/site-packages/picamera/mmal.py", line 49, in <module>
    _lib = ct.CDLL('libmmal.so')
  File "/snap/checkbox/12338/checkbox-runtime/usr/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libmmal.so: cannot open shared object file: No such file or directory
```

boukeas:
Perhaps this issue (and the responses) can be of use: https://github.com/waveform80/picamera/issues/716

PR #1992, opened 2025-07-02
Skip camera/roundtrip-qrcode test on arm64 (bugfix)

Hook25:
## Description

This test relies on libmmal that doesn't exist for arm64. It should therefore be skipped in this context (else it will fail, see linked issue)

Minor: this also replaces lsb with os given that we have soft deprecated lsb

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1739
Fixes: https://github.com/canonical/checkbox/issues/1720

## Documentation

Documented in a jinja comment (I'm sorry, but its already a jinja template so...) why this is done

## Tests

Tested it in the lab on rpi3 both:
- [amd64 (to see that it is skipped)](https://certification.canonical.com/hardware/202311-32356/submission/435955/)
- [armhf (to see that it still runs)](https://certification.canonical.com/hardware/202311-32356/submission/435966/)

### DIFF
--- a/providers/base/units/camera/jobs.pxu
+++ b/providers/base/units/camera/jobs.pxu
@@ -165,7 +165,7 @@ flags: also-after-suspend
 _summary: Webcam BRISQUE score for {product_slug}
 estimated_duration: 20s
 depends: camera/detect
-requires: lsb.release >= '22.04'
+requires: os.release >= '22.04'
 command:
   camera_quality_test.py -d {name} -o "$PLAINBOX_SESSION_SHARE"
 _purpose:
@@ -184,7 +184,7 @@ flags: also-after-suspend
 _summary: Attach the image used for the BRISQUE score for {product_slug}
 estimated_duration: 1s
 after: camera/camera-quality_{name}
-requires: lsb.release >= '22.04'
+requires: os.release >= '22.04'
 command:
   [ -f "$PLAINBOX_SESSION_SHARE"/quality_image_{name}.jpg ] &&
   cat "$PLAINBOX_SESSION_SHARE"/quality_image_{name}.jpg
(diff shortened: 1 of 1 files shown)


providers/base/units/camera/jobs.pxu
```diff
@@ -247,10 +247,16 @@ depends:
   camera/detect
   {% endif -%}
 requires:
+  {#
+    If the device that generated this test is MMAL, check that we are on armhf
+    (libmmal doesn't exist for amd64 or arm64).
+    See: https://github.com/waveform80/picamera/issues/716#issuecomment-1063878114
+  #}
+  (device.name == '{{name}}' and device.category == 'MMAL' and dpkg.architecture == 'armhf') or (device.name == '{{ name}}' and device.category == 'CAPTURE')
```
fernando79513, line 255: Why do we have different requirements for the [`camera/multiple-resolution-images-rpi`](https://github.com/canonical/checkbox/blob/main/providers/base/units/camera/jobs.pxu#L219-L224) job and the `camera/roundtrip-qrcode` job?
```
id: camera/multiple-resolution-images-rpi-attachment_{name}
...
requires: cpuinfo.platform == 'armv7l'
```
I think both jobs should be run in the same scenarios.

If anything, the `camera/roundtrip-qrcode` job should be more restrictive, since it uses more functionalities apart from the `picamera` module and requires a screen.

This mismatch with the requirements was on the code before, but I think it's a nice opportunity to change it.

fernando79513 (changes requested):
Just a small comment regarding the `requires` field for `camera/multiple-resolution-images-rpi` and `camera/roundtrip-qrcode` jobs.

providers/base/units/camera/jobs.pxu
Hook25: Because that job is expanded only for MMAL devices, whle the one I just modified it expanded both for capture and mmal devices. Indeed my requires means, require that if the job was expanded for a mmal device, that we are also on armhf. That requires says, expand this on mmal devices but run it only on armhf. They are the same!

fernando79513: You are right, I missed the `MMAL` restriction in the template filter.

PR #2001, opened 2025-07-04
Make networking/info_device template depend on ethernet/detect (bugfix)

pieqq:
networking/info_device__index___interface template should only be run if ethernet/detect passed.

ethernet/detect requires the `has_ethernet_adapter` manifest, so if this manifest entry is set to false, both tests would be skipped.

Before this patch, the following problem could happen:

- on laptop without an onboard Ethernet adapter, testing teams sometimes plug an USB Ethernet adapter to make it easier to hook the device up in our testing lab (to be able to remote access it, etc.)
- when doing so, Checkbox will detect this adapter as if it was an onboard adapter
- testing teams rightfully set the `has_ethernet_adapter` manifest entry to true, yet the networking/info_device__index___interface job is still run for that device.

This can create unnecessary problems down the line to review test reports in order to issue a certificate.

## Resolved issues

Prevent unwanted failures in certificate requests (see [this for example](https://bugs.launchpad.net/certify-planning/+bug/2103972/comments/2))

### DIFF
--- a/providers/base/units/networking/jobs.pxu
+++ b/providers/base/units/networking/jobs.pxu
@@ -19,6 +19,7 @@ template-id: networking/info_device__index___interface
 flags: also-after-suspend
 _summary: Network Information of device {__index__} ({interface})
 estimated_duration: 1.0
+depends: ethernet/detect
 requires: executable.name == 'ethtool'
 command:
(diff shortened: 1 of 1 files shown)


Hook25 (approved):
Thanks, makes sense

PR #2017, opened 2025-07-14
Copy all packages and catch those that couldn't be copied (infra)

Hook25:
## Description

The strategy of trying to infer what packages are to be copied has two disadvantages:
- Doens't copy non-checkbox packages
- Doens't work if we pass validation the day the series is disabled (race condition, we may build the package and fail once we try to copy it)

This leads to partial beta releases, like the one we just had. This fixes the issue by simply skipping the copy actions that fails with the error that the target build is outdated.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1977

## Documentation

N/A

## Tests

Updated the tests and fixed the current release with this fix

### DIFF
--- a/tools/release/lp_update_recipe.py
+++ b/tools/release/lp_update_recipe.py
@@ -29,19 +29,44 @@ import sys
 import textwrap
 import argparse
 
-from utils import get_source_build_recipe
+from utils import get_source_build_recipe, get_launchpad_client
 
-# this is hardcoded because supporting a new release requires a bit of manual
-# effort (copying support packages to the new series and doing the first build
-# dance of building packages in dependency order)
-active_series = [
-    "https://api.launchpad.net/devel/ubuntu/bionic",
-    "https://api.launchpad.net/devel/ubuntu/focal",
-    "https://api.launchpad.net/devel/ubuntu/jammy",
-    "https://api.launchpad.net/devel/ubuntu/noble",
-    "https://api.launchpad.net/devel/ubuntu/plucky",
-    "https://api.launchpad.net/devel/ubuntu/questing",
-]
+
+def get_active_series():
+    # this is hardcoded because supporting a new release requires a bit of manual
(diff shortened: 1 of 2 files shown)


tools/release/lp_update_recipe.py
```diff
@@ -31,6 +31,18 @@
...
+# dance of building packages in dependency order)
+active_series = [
+    "https://api.launchpad.net/devel/ubuntu/bionic",
+    "https://api.launchpad.net/devel/ubuntu/focal",
+    "https://api.launchpad.net/devel/ubuntu/jammy",
+    "https://api.launchpad.net/devel/ubuntu/noble",
+    "https://api.launchpad.net/devel/ubuntu/plucky",
+    "https://api.launchpad.net/devel/ubuntu/questing",
+]
+
```
fernando79513, line 45: I don't think we can forget to add a new series here, but we may forget to remove deprecated ones.
Should we check if a series is deprecated before updating the recipe?

fernando79513 (changes requested):
I guessed there could be a race condition, though I didn't guess the exact reason, XD.

I like this approach better. There is just one small comment. Since now we are copying all the packages (not only the checkbox ones), there could be a chance that we remove a package from the source PPA, and it will remain in the destination PPA, forever.

I don't think this is a big issue, but it would be great to include a comparison of the packages in the source PPA and the destination PPA for debugging purposes, as we have the output of the packages we have successfully copied.

Hook25:
I've updated the code to also check if version are outdate and remove them from the desired list. AFAIK you can't remove packages from a PPA, so I'm not sure your comment applies. Still, we've never done something like that, I don't think it applies here

fernando79513 (approved):
LGTM+1!

Hook25:
ngl, I'm unsure about what that button does and how to get to it regularly (when copies didn't fail).

PR #2018, opened 2025-07-15
BugFix for docker/save-and-load remove image before try to load it again (BugFix)

locnnil:
## Description

- The test has an inconsistency, it saves an image get from docker hub on a tar file and immediately tries to load it back, which doesn't make sense because this operation will be cached/ignored.

- This adds a removal of the image before trying to load the tar file.

### DIFF
--- a/providers/docker/units/docker.pxu
+++ b/providers/docker/units/docker.pxu
@@ -244,6 +244,7 @@ command:
  set -e
  docker pull ubuntu
  docker save ubuntu > {root_dir}/ubuntu.tar
+ docker rmi ubuntu
  docker load < {root_dir}/ubuntu.tar
  docker images -a | grep ubuntu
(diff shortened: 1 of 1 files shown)


Hook25:
Hello, thanks for this, please provide at least a submission to prove this works

locnnil:
@Hook25 At first when I tried to run with my change it didn't work.

Then I revert it, but still, it doesn't work.

After investigating, I saw that the IO redirection for the save and load command weren't working inside Checkbox. Outside it works fine. 

So I changed the method to use to save and load the tarball, and now the tests passes.

Here is the requested submission:
https://certification.canonical.com/hardware/202203-30003/submission/442561/

Hook25:
Interesting, it is really weird that it doesn't work inside of Checkbox that command, it looks good to me. Either way, thanks for fixing it.

PR #2020, opened 2025-07-15
Allow agent to run as non-root for debugging (infra)

Hook25:
## Description

This makes debugging locally much easier because running the agent as root while iterating locally is quite bothersome and has quite a few undesired side effects.

## Resolved issues

N/A

## Documentation

N/A

I don't know if/where we would want to document this, we don't really have checkbox developer documentation

## Tests

Start a non-priviledged container, enable the Checkbox venv
```
$ checkbox-cli run-agent
Checkbox agent must be run by root!
```

Now try with the envvar exported:
```
$ ALLOW_CHECKBOX_AGENT_NONROOT=1 checkbox-cli run-agent
WARNING:agent:Forcing the agent to allow running as non-root
WARNING:agent:Note: this is a debugging tool, THE AGENT DOESN'T ACTUALLY WORK AS NON-ROOT
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,9 +75,11 @@ you have to activate the `checkbox-cli run-agent` on the Machine under test:
 
 > Note: Keep in mind that run-agent has to be run as root and needs the
 > virtual env, you may have to re-enable/activate it after a `sudo -s`
-> you can circumvent this constraint if you are developing for Checkbox by
-> exporting ALLOW_CHECKBOX_AGENT_NONROOT to `true`. This only makes sense if
-> you are debugging Checkbox internals. The agent will not work as normal user.
+>
+> You can circumvent this constraint if you are developing for Checkbox by
+> setting the `ALLOW_CHECKBOX_AGENT_NONROOT` environment variable to `true`.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -75,6 +75,9 @@ you have to activate the `checkbox-cli run-agent` on the Machine under test:
 
 > Note: Keep in mind that run-agent has to be run as root and needs the
 > virtual env, you may have to re-enable/activate it after a `sudo -s`
+> you can circumvent this constraint if you are developing for Checkbox by
+> exporting ALLOW_CHECKBOX_AGENT_NONROOT to `true`. This only makes sense if
+> you are debugging Checkbox internals. The agent will not work as normal user.
```
pieqq, line 80:
```suggestion
> virtual env, you may have to re-enable/activate it after a `sudo -s`.
> 
> You can circumvent this constraint if you are developing for Checkbox by
> setting the `ALLOW_CHECKBOX_AGENT_NONROOT` environment variable to `true`.
> This only makes sense if you are debugging Checkbox internals.
> The agent will not work as normal user.
```

pieqq (changes requested):

pieqq (approved):

Issue #1885, opened 2025-04-23
No `libOSSlib.so` while running audio/alsa-loopback-automated


LiaoU3:
### Bug Description

When running tests `audio/alsa-loopback-automated`, it shows the following error.

After investigation, `libOSSlib.so` is provided by `liboss4-salsa2`, which is not installed by default on Noble

1) We need to update the test to install the required package
OR
2) Update the test to not relay on optional packages

References:
- https://packages.ubuntu.com/noble/amd64/liboss4-salsa2/filelist
- https://launchpad.net/ubuntu/+source/alsa-lib

Related LP bug: https://bugs.launchpad.net/kobuk/+bug/2083895

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

1. Run job: audio/alsa-loopback-automated

### Expected Result

Pass.

### Actual Result

alsa_test: error while loading shared libraries: libOSSlib.so: cannot open shared object file: No such file or directory

### Environment

 - OS: Generic Server 24.04
 - Checkbox Type: Snap
 - Checkbox Version: 4.4.0-dev111

### Relevant log output

```shell
alsa_test: error while loading shared libraries: libOSSlib.so: cannot open shared object file: No such file or directory
```

stanley31huang:
I could reproduce this issue on my iMX6ULL (armhf) running server image as well, and I observed this issue could be only reproduced on checkbox shell environment.

### test passed while running `alsa_test` outside of checkbox shell environment.
```
ubuntu@ubuntu:~$ /snap/checkbox24/current/providers/checkbox-provider-base/bin/alsa_test loopback -d 5
Dominant frequency: 439.8
Deviation: 0.200012
```

LiaoU3:
I am not able to reproduce the issue.

```
vincent@vincent-XPS-9320:~$ amixer 
Simple mixer control 'Master',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 65536
  Mono:
  Front Left: Playback 0 [0%] [off]
  Front Right: Playback 0 [0%] [off]
Simple mixer control 'Capture',0
  Capabilities: cvolume cswitch cswitch-joined
  Capture channels: Front Left - Front Right
  Limits: Capture 0 - 65536
  Front Left: Capture 65536 [100%] [on]
  Front Right: Capture 65536 [100%] [on]
vincent@vincent-XPS-9320:~$ sudo checkbox.shell 
[...]

PR #2023, opened 2025-07-17
Stage "libasound2t64" instead of the virtual package "libasound2" (BugFix)

LiaoU3:
## Description
`libasound2` is not a virtual package before noble (e.g. [jammy](https://packages.ubuntu.com/jammy/libasound2)), but it is in [noble](https://packages.ubuntu.com/noble/libasound2). Moreover, it seems like in the build for [arm64](https://github.com/canonical/checkbox/actions/runs/16202677370/job/45745393365) (currently it only happened on arm64 and if you search for "asound", you will find that it installed both packages on arm), snapcraft tried to package both `libasound2t64` and `liboss4-salsa-asound2`, which should no happened (I will create a bug for snapcraft). Since these 2 packages conflict each other, and actually what we want is `libasound2t64`.

Overall, I not only to try to workaround the snapcraft issue, but also think it is better to point out the package we want.

```
$ sudo apt install libasound2
[sudo] password for vincent: 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package libasound2 is a virtual package provided by:
  libasound2t64 1.2.11-1ubuntu0.1 (= 1.2.11-1ubuntu0.1)
  liboss4-salsa-asound2 4.2-build2020-1ubuntu3
You should explicitly select one to install.

E: Package 'libasound2' has no installation candidate
```

## Resolved issues
#1885 

## Documentation
N/A

## Tests
I build a checkbox24 with replacing `libasound2` with `libasound2t64`
### Before
```
ubuntu@ubuntu:~$ checkbox.shell 
checkbox runtime shell, type 'exit' to quit the session
(checkbox-shell) ubuntu@ubuntu:~$ amixer 
amixer: error while loading shared libraries: libOSSlib.so: cannot open shared object file: No such file or directory
```
### After
```
ubuntu@ubuntu:~$ checkbox.shell 
checkbox runtime shell, type 'exit' to quit the session
(checkbox-shell) ubuntu@ubuntu:~$ amixer 
ALSA lib conf.c:4579:(snd_config_update_r) Cannot access file /snap/checkbox/15760/usr/share/alsa/alsa.conf
ALSA lib conf.c:4579:(snd_config_update_r) Cannot access file /snap/checkbox/15760/usr/share/alsa/pcm/default.conf
Simple mixer control 'Master',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 65536
  Mono:
  Front Left: Playback 65536 [100%] [on]
  Front Right: Playback 65536 [100%] [on]
Simple mixer control 'Capture',0
  Capabilities: cvolume cswitch cswitch-joined
  Capture channels: Front Left - Front Right
  Limits: Capture 0 - 65536
  Front Left: Capture 65536 [100%] [on]
  Front Right: Capture 65536 [100%] [on]
```

### DIFF
--- a/checkbox-core-snap/series24/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series24/snap/snapcraft.yaml
@@ -287,7 +287,7 @@ parts:
       - iw
       - jq
       - kmod
-      - libasound2
+      - libasound2t64
       - libcap2-bin
       - libfdt1
       - libsvm3


LiaoU3:
I create https://github.com/canonical/snapcraft/issues/5621 to track the snapcraft issue I mentioned above.

PR #2026, opened 2025-07-17
Print default value of launcher via subcommand (new)

Hook25:
## Description

When using Checkbox one is often left wondering what the default value of a configuration is. Additionally, in order to start writing a launcher, it is often tricky to new users to do so without continuosly referencing the documentation. 

This adds a new subcommand that "bootstraps" a launcher for the user with all keys set to their default value (so one can modify it and use it right away), and comments that explain what each configuration does.

## Documentation

Documented both via help and added it to the useful subcommands

## Tests

Added tests and tested it locally. To try this run
```
checkbox-cli config defaults /tmp/abc.conf
```
To check that what was produced was valid:
```
# this I've preserved for backward compatibility
checkbox-cli check-config /tmp/abc.conf
# this is new from this pr, does the same thing
checkbox-cli config check /tmp/abc.conf
```

### DIFF
--- a/docs/tutorial/using-checkbox/advanced-commands.rst
+++ b/docs/tutorial/using-checkbox/advanced-commands.rst
@@ -312,6 +312,57 @@ You can also run a whole test plan using the ``run`` command:
 This will run the Checkbox Base Tutorial test plan, executing all the jobs in
 it and providing a text summary of the test run.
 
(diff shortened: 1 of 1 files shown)


pieqq (approved):
Thanks, very neat!

PR #2029, opened 2025-07-17
Add checkbox-gfx Vulkan tests (New)

mckees:
## Description
This provider contains tests for the graphics stack. Testing currently checks Vulkan by running Khronos's VK-GL-CTS, a conformance test suite. These checks ensure that our graphics stacks are conforming to Khronos's expected functionality for Vulkan.

## Resolved issues

N/A

## Documentation

See README

## Tests

Tests have been run on Tiger Lake, Alder Lake, and Raptor Lake


contrib/checkbox-gfx/.github/workflows/build-vulkan-on-pr-arm.yaml
```diff
@@ -0,0 +1,31 @@
+name: "Build Vulkan - ARM"
```
fernando79513, line 1: Why do you want to build bulkan on each PR?
It seems that you are building always the same version.

Also, does it make sense to include Vulcan directly in the checkbox-provider-gfx snap, as we are doing with [opencv](https://github.com/canonical/checkbox/blob/main/checkbox-core-snap/series22/snap/snapcraft.yaml#L445) in checkbox?

Or maybe this package is too big to be included in the snap. Maybe @Hook25 knows a good way to store these binaries? Maybe in a PPA? Just the build folder seems to be 2.6GB,

contrib/checkbox-gfx/bin/setup-env.sh
```diff
@@ -0,0 +1,19 @@
...
+sudo chown -R $USER $WORKING_DIR
+
+# Get vendor
+if [[ $(lscpu | grep "GenuineIntel") ]]; then
+    export VENDOR=Intel
+elif [[ $(lscpu | grep "AuthenticAMD") ]]; then
+    export VENDOR=AMD
+elif [[ $(lscpu | grep "Qualcomm") ]]; then
+    export VENDOR=Qualcomm
+fi
```
fernando79513, line 18: Where is this `VENDOR` variable used?

contrib/checkbox-gfx/checkbox-provider-gfx/units/vulkan/jobs.pxu
```diff
@@ -0,0 +1,294 @@
+id: vkconformancetests
```
fernando79513, line 1: Check: https://github.com/canonical/checkbox/pull/2011#discussion_r2228842379

contrib/checkbox-gfx/snap/snapcraft.yaml
```diff
@@ -0,0 +1,74 @@
...
+    command-chain: [bin/wrapper_local]
+    command: bin/configure
+  test-vulkan:
+    command-chain: [bin/wrapper_local]
+    command: bin/test-vulkan
+  remote-slave:
+    command-chain: [bin/wrapper_local]
+    command: bin/checkbox-cli-wrapper slave
+    daemon: simple
+    restart-condition: always
```
fernando79513, line 36: Check https://github.com/canonical/checkbox/pull/2011#discussion_r2228592762

fernando79513 (changes requested):
Thanks for the PR and sorry for taking so long.
I've left some comments in the review, please ping me if you need further explanation.

contrib/checkbox-gfx/.github/workflows/build-vulkan-on-pr-arm.yaml
mckees: To the first question, this workflow only runs when a change is made to `bin/install-vulkan` due to the `paths` section. This was just to check that changes to the install files don't break the installer for both x86 and ARM (I broke the ARM build on several occasions before creating this check)

To the second point, I'm realizing that this might be a misleading name. `install-vulkan` actually builds and installs the Vulkan conformance test suite. I've gone more minimalist on naming so that people using it will just install and run by the name of the component they're wanting to test rather than something that sounds like a variant of vulkan like `install-vulkan-cts`. But I'm seeing that this could also cause some confusion making it seem like Vulkan itself is being installed, so I'm thinking something like `install-vulkan-tests` if that works better for you. 

I've also been tempted to package this CTS since it's such a large build each time you want to run the tests. Just haven't had the cycles yet

contrib/checkbox-gfx/bin/setup-env.sh
mckees: I may have jumped the gun leaving that in for this PR, but it gets used here:
https://github.com/canonical/checkbox-gfx/blob/f4c01a772c369bacd693fe9dad26b2938eab0b38/bin/install-lvl-zero#L11

And I think it will get used more, especially as we start testing on non-intel hardware and need a way to filter that out. If checkbox has some mechanism for that, we can switch to it. This could be deleted and make a comeback in Python scripts once we get to the point of needing to detect feature support based on each generation's stated feature set (I'm currently doing this for the media side of the testing)

contrib/checkbox-gfx/.github/workflows/build-vulkan-on-pr-arm.yaml
LiaoU3: I am just wondering, does it work if you don't put `.github/workflow` under the root of the git repo?

LiaoU3 (commented):
Just one minor comment.

contrib/checkbox-gfx/.github/workflows/build-vulkan-on-pr-arm.yaml
mckees: Great question. I don’t think so, but I wasn’t sure how the checkbox team wanted to handle it, so I left it in place for the initial review stages

fernando79513 (changes requested):
Thank you a lot for the changes!

I've added a few more comments. I'll discuss with Max tomorrow about how to handle the Vulkan build, and I think we can merge this soon.

fernando79513 (changes requested):
I talked with Max, and it should be okay building Vulkan and adding it as an artifact of the gh workflows as long as we don't do this too often.

I've just added a few more comments, but I think this should be good to land after that

contrib/checkbox-gfx/.github/workflows/build-vulkan-on-pr-arm.yaml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  push:
+    branches:
+      - test_gh_runner
```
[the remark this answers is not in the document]
mckees: Because the manual triggers had use a different job, I have used this as a way to verify that the workflow is working before merging it. I don't mind removing it

contrib/checkbox-gfx/checkbox-provider-gfx/units/vulkan/jobs.pxu
```diff
@@ -0,0 +1,283 @@
...
+category_id: mesa
+flags: simple
+user: root
+_summary: Run the api tests from VK-GL-CTS
+environ:
+  # necessary for local mode
+  XDG_SESSION_TYPE
+  XDG_RUNTIME_DIR
+estimated_duration: 8h
+command: run_vulkan_cts.py --test_file /usr/local/checkbox-gfx/VK-GL-CTS/external/vulkancts/mustpass/main/vk-default/api.txt
```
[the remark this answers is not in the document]
mckees: Agreed that it would make things look cleaner. Added in the latest PR

contrib/checkbox-gfx/README.md
```diff
@@ -0,0 +1,87 @@
...
+sudo snap install --dangerous --classic ./checkbox-gfx_1.0_<arch>.snap
+```
+
+Make sure that the provider service is running and active:
+
+```shell
+systemctl status snap.checkbox-gfx.run-agent.service
+```
+
+# Install dependencies
```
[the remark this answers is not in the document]
mckees: I think that would be nice for those on arm64 to have that option. With the checkbox repo, I'm guessing I can't exactly guide people to search through all completed actions, but my attempt to quickly automate this ends up becoming a script that will need to live somewhere. Once I get to this point, I'm going to have to either offer the user download options for their configuration or download artifacts for all configurations:
```
curl $(curl -s "https://api.github.com/repos/canonical/checkbox-gfx/actions/workflows/build-installers.yaml/runs?status=success&per_page=1" | grep -Po '"artifacts_url":\s*"\K[^"]+')
```
Should be a quick Python script if we want it, and I could use it in the installer script by default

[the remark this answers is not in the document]
fernando79513: I would add in the README the process to use the artifact built by the GH runers, since we are building it there. I don't see the reason to have the build workflow if we don't encourage to use it.
You can do it in a follow up PR if you want.

fernando79513 (approved):
LGTM +1!

PR #2031, opened 2025-07-18
Docs: ignore link check on GNU license (Infra)

tang-mm:
## Description

Update Sphinx configuration to ignore link check on the GNU license URL, which caused rate limiting:
```
-rate limited-   https://www.gnu.org/licenses/gpl-3.0.html | sleeping...
```

## Resolved issues

NA

## Documentation
NA

## Tests

Link check on PR succeeds.


pieqq (approved):
Thanks for the quick fix!

Issue #2027, opened 2025-07-17
iperf3 fails to start


rodwsmith:
### Bug Description

A partner has reported a bug involving the `network.py` script: If the formatting of the NUMA node list from `lscpu` does not match expectations, the script crashes with a Python traceback. A problematic example output resembles the following:

```
NUMA:
  NUMA node(s): 34
  NUMA node0 CPU(s): 0-71
  NUMA node1 CPU(s): 72-143
  NUMA node2 CPU(s):
  NUMA node3 CPU(s):
  NUMA node4 CPU(s):
  NUMA node5 CPU(s):
  NUMA node6 CPU(s):
```

(And so on up to node33.)

The problem is the empty list of NUMA nodes, which the script doesn't handle correctly.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Run the `network.py` script (directly or via a Checkbox launcher) on an affected system.
2. Watch it fail.

### Expected Result

The script should run and assign the threads to the appropriate NUMA nodes.

### Actual Result

The script crashes with a Python traceback.

### Environment

- An aarch64-based server
- 6.8.0.63-generic kernel
- 

### Relevant log output

```shell

```

PR #2028, opened 2025-07-17
Fix network.py crash on unexpected NUMA node listing data (BugFix)

rodwsmith:
Title: Fix network.py crash on unexpected NUMA node listing data (BugFix)

## Description

The `extract_core_list()` function in `network.py` loops through the NUMA node list data presented by `lscpu`, looking for lines of the form:

`NUMA node0 CPU(s):      0-15`

When one of these lines is empty, the code crashed with a Python traceback. This bug fix skips lines with empty CPU lists.

In addition, one minor formatting change (removing a space from the `cpu_list = line[colon + 1 :]` line) was made to fix a flake8 complaint.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/2027

## Documentation

No documentation changes are needed.

## Tests

The partner who encountered this bug provided the fix and tested it to their satisfaction. I've tested it on three other systems that have not exhibited the bug, and have seen no problems on these systems, so it seems to have no effect on these systems.

PR #2035, opened 2025-07-18
Fix network.py crash on unexpected NUMA node data (BugFix)

rodwsmith:
Title: Fix network.py crash on unexpected NUMA node listing data (attempt 2) (BugFix)

## Description

The `extract_core_list()` function in `network.py` loops through the NUMA node list data presented by `lscpu`, looking for lines of the form:

`NUMA node0 CPU(s):      0-15`

When one of these lines is empty, the code crashed with a Python traceback. This bug fix skips lines with empty CPU lists.

This PR replaces #2028, which ran into permissions problems that I and Pedro were unable to resolve.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/2027

## Documentation

No documentation changes are needed.

## Tests

The partner who encountered this bug provided the fix and tested it to their satisfaction. I've tested it on three other systems that have not exhibited the bug, and have seen no problems on these systems, so it seems to have no effect on these systems.

### DIFF
--- a/providers/base/bin/network.py
+++ b/providers/base/bin/network.py
@@ -213,6 +213,11 @@ class IPerfPerformanceTest(object):
         cpu_list = line[colon + 1 :]
         core_list = []
         for core_range in cpu_list.split(","):
+            # Skip it if the CPU list for the NUMA node is empty....
+            core_range = core_range.strip()
+            if not core_range:
+                logging.error("Empty core range in line: %s", line)
+                continue
             # core_range should be of the form "a-b" or "a"
             range_list = core_range.split("-")
             if len(range_list) > 1:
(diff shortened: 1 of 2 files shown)


pedro-avalos (approved):
LGTM

providers/base/bin/network.py
```diff
@@ -213,6 +213,11 @@ def extract_core_list(self, line):
         cpu_list = line[colon + 1 :]
         core_list = []
         for core_range in cpu_list.split(","):
+            # Skip it if the CPU list for the NUMA node is empty....
```
mreed8855, line 216: +1

mreed8855 (commented):
+1

PR #2038, opened 2025-07-22
Removed references to checkbox_support.contrib and graphics_stress_test (BugFix)

fernando79513:
## Description

`graphics_stress_test.py` was using importing the `checkbox-support.contrib` module, which does not exist anymore. We couldn't identify if this test was run anywhere, and in case it is still run somewhere, we should refactor it so that it does not import a non-existent module.

## Resolved issues

N/A

## Documentation

Removed from `checkbox-support/debian/control` Description

## Tests

N/A


Hook25 (approved):
Thanks

Issue #2024, opened 2025-07-17
audio/alsa-playback always failed with ALSA lib conf errors


weizhenwu:
### Bug Description

While running the `audio/alsa-playback` test, ALSA lib conf error is reported:

```
ALSA lib conf.c:1245:(parse_value) default is not a string
ALSA lib conf.c:2012:(_snd_config_load_with_include) _toplevel_:350:0:Invalid argument
ALSA lib conf.c:4639:(snd_config_update_r) /snap/checkbox24/current/usr/share/alsa/alsa.conf may be old or corrupted: consider to remove or fix it
Failed to change volume: Failed to attach HCTL to a Mixer
ALSA lib conf.c:1245:(parse_value) default is not a string
ALSA lib conf.c:2012:(_snd_config_load_with_include) _toplevel_:350:0:Invalid argument
ALSA lib conf.c:4639:(snd_config_update_r) /snap/checkbox24/current/usr/share/alsa/alsa.conf may be old or corrupted: consider to remove or fix it
terminate called after throwing an instance of 'Alsa::AlsaError'
  what():  Failed to open device: default. Invalid argument
```

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

1. Lunch checkbox-cli with project checkbox
2. Run the job `audio/alsa-playback`

### Expected Result

The test can run and get a pass/fail result

### Actual Result

The test failed with the error in the description

### Environment

- OS: ubuntu server 24.04
- checkbox type: snap
- checkbox version:
```
ubuntu@ubuntu:~$ snap list
Name                     Version             Rev    Tracking       Publisher            Notes
checkbox                 6.0.0-dev18         15760  24.04/beta     ce-certification-qa  classic
checkbox-carmel-classic  0.1dev              84     latest/edge    ce-certification-qa  classic
checkbox-ce-oem          1.0-noble           2269   24.04/edge     ce-certification-qa  classic,held
checkbox24               5.0.0-dev17         857    latest/beta    ce-certification-qa  held

```
- DUT: 202403-33871

### Additional context
checkbox submission: https://certification.canonical.com/hardware/202403-33871/submission/438468/test/68908/result/50508638/

_No response_

LiaoU3:
By the way, to prove that `default.conf` is not applicable as `ALSA_CONFIG_PATH`, I did an experiment like this. As you can see, this file is not working.
```
[...]

PR #2049, opened 2025-07-28
Remove pcm/default.conf from $ALSA_CONFIG_PATH and make playback functional (Bugfix)

LiaoU3:
## Description
As my investigation in the [comment](https://github.com/canonical/checkbox/issues/2024#issuecomment-3125604213), it is shown that `pcm/default.conf` is not applicable in `$ALSA_CONFIG_PATH`. Hence, we need to remove them.

As you can see in the comment, I also find some issue with playback for alsa_test. Hence, I also did some changes on it.

## Resolved issues
audio/alsa-playback always failed with ALSA lib conf errors #2024

## Documentation
N/A

## Tests
submission: https://certification.canonical.com/hardware/202412-36092/submission/441064/

## Future Action Items
According to the code [here](https://github.com/alsa-project/alsa-lib/blob/ee5a58f48e141aeb28df182247c3f789f4160d02/src/conf.c#L4515), it seems like if we pass several alsa.conf in`$ALSA_CONFIG_PATH`, it will do updating instead of replacing, which means:
1. If there is new key in the second alsa.conf, it will be added.
2. If there is different value but same key in the second alsa.conf, it will be updated.
3. If there is no the same key in the second alsa.conf, it will keep it there

Accordingly, there may be an unexpected behavior if there is difference between `/usr/share/alsa/pcm/default.conf` and `$RUNTIME/usr/share/alsa/pcm/default.conf`. I have not come up with a better solution now, but the following approach may be considerable.
1. Use only `/usr/share/alsa/pcm/default.conf`
2. Do not set a value for `$ALSA_CONFIG_PATH`, then it will use the default one, wich is `/usr/share/alsa/pcm/default.conf`

However, the above solutions may not work in core image, because there is no `alsa-utils` in snap. Anyway, I will leave this to be discussed in the future.

### DIFF
--- a/providers/base/src/alsa_test.cpp
+++ b/providers/base/src/alsa_test.cpp
@@ -354,6 +354,7 @@ int playback_test(float duration, int sampling_rate, const char* capture_pcm, co
     auto player = Alsa::Pcm<storage_type>();
     player.set_params(sampling_rate);
     player.sine(440, duration, 0.5f);
+    player.drain();
     return 0;
 }
 


LiaoU3:
After fixing #2024 , another issue popped up ... I think this may due to the fact that we are running alsa test as root user as [here](https://github.com/canonical/checkbox/blob/6c4101532e46d1cc4fad7168c1099ee00273be6d/providers/base/units/audio/jobs.pxu#L842), which should not be expected.
```
-----------------------------[ Running job 3 / 3 ]------------------------------
-------------------------------[ Playback works ]-------------------------------
ID: com.canonical.certification::audio/alsa-playback
Category: Audio tests
Purpose:

Check if sound is played through ALSA on the default output

Steps:

1. Make sure speakers or headphones are connected to the device
2. Commence the test

Pick an action
    => press ENTER to continue
  c => add a comment
  s => skip this job
  q => save the session and quit
[csq]: 
Failed to change volume: Failed to attach HCTL to a Mixer
terminate called after throwing an instance of 'Alsa::AlsaError'
  what():  Failed to open device: default. Host is down
```

LiaoU3:
After fixing the user issue, another issue popped up! (OMG... How many bugs are there...)
The playback finish so fast that people cannot here it.
```
ubuntu@ubuntu:/snap/checkbox24/current/providers/checkbox-provider-base/bin$ time ./alsa_test playback -d 5 -v
Using format: int16 and sampling rate: 48000
PCM opened. Name: default PCM handle: 0xaaaad72f5880 PCM mode: playback
got rate: 48000
period_time: 21333
state: PREPARED
no. of channels: 2
state: RUNNING
Closing PCM 0xaaaad72f5880

real	0m0.169s
user	0m0.086s
sys	0m0.036s
```
Another interesing thing is that, only when I set up to 30 seconds, it can play around 8 seconds
```
ubuntu@ubuntu:/snap/checkbox24/current/providers/checkbox-provider-base/bin$ time ./alsa_test playback -d 10 -v
Using format: int16 and sampling rate: 48000
PCM opened. Name: default PCM handle: 0xaaab115d9e00 PCM mode: playback
got rate: 48000
period_time: 21333
state: PREPARED
no. of channels: 2
state: RUNNING
Closing PCM 0xaaab115d9e00

real	0m0.173s
user	0m0.082s
sys	0m0.042s
ubuntu@ubuntu:/snap/checkbox24/current/providers/checkbox-provider-base/bin$ time ./alsa_test playback -d 20 -v
Using format: int16 and sampling rate: 48000
PCM opened. Name: default PCM handle: 0xaaab13db0c50 PCM mode: playback
got rate: 48000
period_time: 21333
state: PREPARED
no. of channels: 2
state: RUNNING
Closing PCM 0xaaab13db0c50

real	0m0.175s
user	0m0.095s
sys	0m0.025s
ubuntu@ubuntu:/snap/checkbox24/current/providers/checkbox-provider-base/bin$ time ./alsa_test playback -d 30 -v
Using format: int16 and sampling rate: 48000
PCM opened. Name: default PCM handle: 0xaaaac2570830 PCM mode: playback
got rate: 48000
period_time: 21333
state: PREPARED
no. of channels: 2
state: RUNNING
Closing PCM 0xaaaac2570830

real	0m8.281s
user	0m0.308s
sys	0m0.174s
```

I will investigate the alsa_test.cpp. Hope that I could fix this and there is not extra bug pop up...

LiaoU3:
I used the checkbox24 and checkbox for arm and amd from this [build](https://github.com/canonical/checkbox/actions/runs/16612406458), and it worked well on amd64.
However, for arm64, it only worked well on [local](https://certification.canonical.com/hardware/202412-36092/submission/441064/) run, instead, it failed on remote run with the error as below.
```
Failed to change volume: Failed to attach HCTL to a Mixer
terminate called after throwing an instance of 'Alsa::AlsaError'
  what():  Failed to open device: default. Host is down
```

I modified the command field and sideloaded for testing.
```
command:
 whoami
 echo $XDG_RUNTIME_DIR
 amixer
 alsa_test playback -d 5
 ```
The local result is
```
ubuntu
/run/user/1000
Simple mixer control 'Master',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 65536
  Mono:
  Front Left: Playback 49152 [75%] [on]
  Front Right: Playback 49152 [75%] [on]
Simple mixer control 'Capture',0
  Capabilities: cvolume cswitch cswitch-joined
  Capture channels: Front Left - Front Right
  Limits: Capture 0 - 65536
  Front Left: Capture 49152 [75%] [on]
  Front Right: Capture 49152 [75%] [on]
```
The remote result is
```
ubuntu

amixer: Mixer attach default error: Host is down
Failed to change volume: Failed to attach HCTL to a Mixer
terminate called after throwing an instance of 'Alsa::AlsaError'
  what():  Failed to open device: default. Host is down
```
Overall, it shows that checkbox somehow did not pass the environment variable `$XDG_RUNTIME_DIR` correctly. Hence, this should be a checkbox issue.

LiaoU3:
I created #2058 to track the `$XDG_RUNTIME_DIR` issue in remote session. I will leave it to the checkbox team.

PR #2057, opened 2025-07-29
Remote assistant refactoring (new)

Hook25:
## Description

Currently:
- The remote api relies on an obscure dict to re-connect to the agent depending on its state which is basically not documented or tested + weird partial usage that is unnecessary to pass arguments to functions
- bootstrapping (with or without a ui) uses duplicated code in two functions
- bootstrappping relies on a pointlessly duplicated api
- bootstrapping uses a flag in self to suppress resource jobs output (sigh)
- run_jobs is not used (and can't be used) to run any job, only normal jobs (post bootstrap)
- remote api stores states as string, not an enum
- there is a class local decorator in the remote api (which doesn't use the class and its args inferred wrong by the lsp)
- the decorator also raises an AssertionError for a runtime error with an obscure error message
- state transitions are done by setting a private variable and not logged
- bootstrap transition is not done when starting the bootstrap process but when running the first bootstrap job
- `whats_up` is poorly documented
- similarly to the other pr, resume session is called resume session but doesn't resume the session!

This refactor fixes all the above issues

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1983

## Documentation

Added documentation for many APIs

## Tests

Metabox tests, nothing should change about the behaviour

### DIFF
--- a/.github/workflows/tox-checkbox.yaml
+++ b/.github/workflows/tox-checkbox.yaml
@@ -93,6 +93,7 @@ jobs:
   tox_test_checkbox_component:
     needs: get_path_matrix
     strategy:
+      fail-fast: false
       matrix:
         path_flag: ${{ fromJson(needs.get_path_matrix.outputs.path_flag) }}
         python: ["3.5", "3.6", "3.8", "3.10", "3.12"]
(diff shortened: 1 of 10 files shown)


Hook25:
The session behaviour must be exactly the same after this PR. The code I've removed was a legacy resume mechanism that was even more brittle than the one we have right now.

fernando79513 (approved):
Yep, that was what I thought, but I wanted to double check it.
I think this is good to go then.
Great work!

Issue #1628, opened 2024-11-29
suspend-time-check not reporting any time


omar-selo:
### Bug Description

Suspend time check is failing on some machines with no reported times at all. Look at [this](https://certification.canonical.com/hardware/202212-30938/submission/403415/test-results/fail/) example.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

Run the job http://10.102.156.15:8080/job/cert-oem-sru-jammy-desktop-hp-elite-tower-880-g9-desktop-pc-c30938

### Environment

Look at the submission https://certification.canonical.com/hardware/202212-30938/submission/403415/

### Relevant log output

```shell
Warning: Time to sleep was only reported 0 times out of 1 iterations!

Warning: Time to resume was only reported 0 times out of 1 iterations!

Average time to sleep: N/A

ERROR: One or more sleep times was not reported correctly:
could not convert string to float: 'N/A'
Average time to resume: N/A

ERROR: One or more resume times was not reported correctly:
could not convert string to float: 'N/A'
```

fernando79513:
This seems to be an issue with the `suspend/suspend_advanced_auto` for some reason fwts is not reporting the sleep time.
The same error has been found in a set of devices:
 - [hp-z6-g5-a-workstation-desktop-pc-c34219](https://certification.canonical.com/hardware/202407-34219/submission/410223/)
 - [hp-z6-g5-workstation-desktop-pc-c35313](https://certification.canonical.com/hardware/202408-35313/submission/410227/)
 - [hp-z6-g5-workstation-desktop-pc-c35315](https://certification.canonical.com/hardware/202408-35315/submission/410230/)
 - [hp-z8-fury-g5-workstation-desktop-pc-c35886](https://certification.canonical.com/hardware/202410-35886/submission/410229/)
 - [hp-z8-g5-workstation-desktop-pc-c35572](https://certification.canonical.com/hardware/202409-35572/submission/410235/)
 - [hp-z8-g5-workstation-desktop-pc-c35573](https://certification.canonical.com/hardware/202409-35573/submission/410233/)
 - [hp-z8-g5-workstation-desktop-pc-c35574](https://certification.canonical.com/hardware/202409-35574/submission/410231/)

```
Calling fwts
==================== Test Results ====================

1 iterations total.
Warning: Time to sleep was only reported 0 times out of 1 iterations!
Warning: Time to resume was only reported 0 times out of 1 iterations!

Average time to sleep: N/A
[...]

PR #2062, opened 2025-07-31
Update FWTS to V25.07.00 for series18~24 snaps (infra)

pieqq:
This new version contains a fix that might help with the problem described in #1628.


Hook25:
Verified the snaps still build here:
- https://github.com/canonical/checkbox/actions/runs/16651143089
- https://github.com/canonical/checkbox/actions/runs/16651149119

PR #2076, opened 2025-08-12
Keepalive rpyc connection (bugfix)

Hook25:
## Description

Whenever the agent fails to close the connection (currently happening on one device, that fails to do so during the watchdog reset test), the session may hang forever. This wasn't an issue when we had a global attribute access timeout but now that we removed it, it is blocking at least one execution. 

This modifies the rpyc connection so that it also implements keepalive. This way the keepalive messages are also used to probe the connection and detect if it was closed. 

Additionally, this should also fix a very rare bug. When asking for interactions, the connection is not kept alive, so it may be unexpectedly closed by an intermidiary.

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1995 

## Documentation

N/A

## Tests

Tested on the machine, it now 100% works, before it worked very rarely.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -291,7 +291,7 @@ class RemoteController(ReportsStage, MainLoopStage):
                     keep_running = self._handle_interrupt()
                     if not keep_running:
                         break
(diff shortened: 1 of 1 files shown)


fernando79513 (approved):
LGTM+1!

PR #2083, opened 2025-08-20
Setup include (Breaking)

Hook25:
## Description

Setup include must be landed in trances because it is a large feature. This task includes:

- Partial Setup job support
- No dependencies
- No manifest field
- Inclusion mechanism
- Resume (local + remote)

## Resolved issues

Fixes: CHECKBOX-1996

## Documentation

WIP

## Tests

New metabox scenarios

### DIFF
--- a/checkbox-ng/plainbox/impl/session/test_remote_assistant.py
+++ b/checkbox-ng/plainbox/impl/session/test_remote_assistant.py
@@ -95,7 +95,7 @@ class RemoteAssistantTests(TestCase):
         self.assertEqual(self_mock.state, RemoteSessionStates.SetupCompleted)
         self.assertTrue(self_mock._sa.finish_setup.called)
 
-    def test_whats_up_seting_up(self):
+    def test_whats_up_setting_up(self):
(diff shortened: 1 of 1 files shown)


fernando79513 (changes requested):
Really good work here!
I couldn't find any big issue or flaw in the "setup" logic, but I added a few comments.
This feature is a huge one, but I think it will be really helpful in the future.
 
I also have one question. Could "setup jobs" have other "setup jobs" as a dependency?
How does it work when there are any "failed_setups"? Are all the "setup jobs" run again, or just the ones on the "failed_setups" list? 

Please also check metabox tests. It seems that there are some of them failing when running them "in local" after all of these changes:
https://github.com/canonical/checkbox/actions/runs/17234967164/job/48897625049
 - SetupIncludeResumeAutomated
 - ConfigLoadedAlsoAfterResume
 - ResumeAfterFinishPreserveOutputLocal
 - ResumeDontCrashIncompatibleSession
I run them on my laptop with the same results

pieqq (approved):

PR #2117, opened 2025-09-18
wifi test: fix checking of ping results (#2116) (BugFix)

### LINKED ISSUE #2116: In wifi_nmcli_test::perform_ping_test ping can return more than "count" packets due to the use of deadline mode
### Bug Description

The ping abstraction in gateway_ping_test use the ping deadline mode as a safeguard. This mode has quite an impact on the behavior of ping, and can thus give results that will surprise any test using it.
A test failed recently on one of our device due to the following ping result:
```
## Perform a ping test
Got gateway address: 192.168.144.1
Validated gateway address: 192.168.144.1
PING 192.168.144.1 (192.168.144.1) from 192.168.154.210 wlP1p1s0: 56(84) bytes of data.
64 bytes from 192.168.144.1: icmp_seq=1 ttl=64 time=60.6 ms
64 bytes from 192.168.144.1: icmp_seq=2 ttl=64 time=3.34 ms
64 bytes from 192.168.144.1: icmp_seq=3 ttl=64 time=318 ms
64 bytes from 192.168.144.1: icmp_seq=4 ttl=64 time=1058 ms
64 bytes from 192.168.144.1: icmp_seq=5 ttl=64 time=1066 ms
64 bytes from 192.168.144.1: icmp_seq=6 ttl=64 time=64.3 ms

--- 192.168.144.1 ping statistics ---
6 packets transmitted, 6 received, 0% packet loss, time 5021ms
rtt min/avg/max/mdev = 3.336/428.377/1066.461/458.995 ms, pipe 2

Ping result: {'transmitted': 6, 'received': 6, 'pct_loss': 0}
Connection test failed
```

The test has the following code:
```
def perform_ping_test(interface, renderer):
    target = get_gateway(interface, renderer)

    if target:
        count = 5
        result = ping(target, interface, count, 10)
        print("Ping result: {}".format(result))
        if result["received"] == count:
            return True

    return False
```

As you can see, ping transmitted more than `count` packets, because some answers took more than 1s to arrive.
The test is thus wrong, it should check for `result["received"] >= count` or `result["received"] == result["transmitted"]` to consider the case where a network has a latency spike.
At least two tests have this issue in checkbox:
- providers/base/bin/wifi_nmcli_test::perform_ping_test
- providers/base/bin/wifi_client_test_netplan::perform_ping_test

The perform_ping_test in gateway_ping_test is correct, so I think this issue is merely a mis-understanding of this subtle ping edge case and has not been noticed by other users of gateway_ping_test::ping.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Launch wifi_nmcli_test or wifi_client_test_netplan perform_ping_test with a latency of more than 1s for the fifth packet

### Expected Result

The job passes because the gateway is reachable.

### Actual Result

The job fails because it has more packets than expected.

### Environment

Ubuntu 24.04

### Relevant log output

```shell

```

### Additional context

_No response_

pinaraf:
## Description

Fix the tests relying on gateway_ping_test::ping to properly handle network latency above 1s, thus ping having more than `count` packets

## Resolved issues

Fixes #2116

### DIFF
--- a/providers/base/bin/wifi_client_test_netplan.py
+++ b/providers/base/bin/wifi_client_test_netplan.py
@@ -354,7 +354,7 @@ def perform_ping_test(interface, renderer):
         count = 5
         result = ping(target, interface, count, 10)
         print("Ping result: {}".format(result))
-        if result["received"] == count:
+        if result["received"] == result["transmitted"]:
             return True
 
     return False
--- a/providers/base/bin/wifi_nmcli_test.py
+++ b/providers/base/bin/wifi_nmcli_test.py
@@ -196,7 +196,7 @@ def perform_ping_test(interface):
     if target:
         count = 5
         result = ping(target, interface, count, 10)
-        if result["received"] != count:
+        if result["received"] != result["transmitted"]:
             raise ValueError(
                 "{} packets expected but only {} received".format(
                     count, result["received"]
--- a/providers/base/tests/test_wifi_client_test_netplan.py
+++ b/providers/base/tests/test_wifi_client_test_netplan.py
@@ -780,6 +780,33 @@ class WifiClientTestNetplanTests(TestCase):
         self.assertIn("Got gateway address: ", captured_output.getvalue())
 
 
+@patch("wifi_client_test_netplan.ping")
+@patch("wifi_client_test_netplan.get_gateway")
+@patch("wifi_client_test_netplan.sp.check_output")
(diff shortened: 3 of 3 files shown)


Hook25 (approved):
Thanks, please next time do fill out the full description template, it makes our job a little bit easier

PR #2118, opened 2025-09-18
Add reboot testing for DSS (New)

motjuste:
## Description

> [!NOTE]
> This PR is best merged / reviewed after #2112 

This PR adds testing DSS across machine reboot in the contributed Checkbox-DSS provider.  Most additions are adding sibling jobs to existing jobs and expanding the test plan to order their execution.

The test plan has been expanded so that the following jobs run:

1. All existing jobs except purging DSS.
2. New long-living notebooks are created and verified, but not removed.
3. Machine is rebooted.
4. Wait for the cluster to come back up and settle down.
5. The long-living notebooks are restarted and verified again, then removed.
6. All existing jobs are executed again, including purging DSS.

The tests now ensure that notebooks created in DSS are usable across reboots, irrespective of the Kubernetes cluster underneath.

### Waiting for cluster after reboot

We need to wait for the cluster to come back up after reboot.  For this, we first just wait for the cluster to be accessible using `kubectl`, and then, depending on which GPU plugins are expected to have been installed (based on which GPUs are available), we also await the respective Kubernetes daemonsets to finish restarting.  Due to some potential race conditions while the cluster is still restarting, we pad the checks with `sleep` calls in a few places.

### Restart necessary for long-living notebooks

Furthermore, the long-living notebooks that were left running before the reboot need to be restarted using `dss stop` and `dss start` commands.  This is another way to handle the "race condition" during the cluster settling down.  In this case, it happens so that the notebook pods get recreated before the NVIDIA GPU operator is properly back up ... i.e. the notebook pods get recreated without access to the NVIDIA GPU.  The behaviour of the notebooks is correct ... they do not require NVIDIA GPU to run, but can use the GPU if the cluster can.  Simply restarting the pods after the NVIDIA GPU operator is deployed is enough.

## Resolved issues

[CHECKBOX-1897](https://warthogs.atlassian.net/browse/CHECKBOX-1897).

## Documentation

No changes to the Checkbox's documentation.  There's no change in usage of this provider either.

## Tests

- Full GH workflow run on Canonical K8s: https://github.com/canonical/checkbox/actions/runs/18306830633
- Full GH workflow run on Microk8s: https://github.com/canonical/checkbox/actions/runs/18306843874

[CHECKBOX-1897]: https://warthogs.atlassian.net/browse/CHECKBOX-1897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
+++ b/contrib/checkbox-dss-validation/checkbox-provider-dss/units/jobs.pxu
@@ -84,9 +84,9 @@ imports:
   from com.canonical.certification import graphics_card
 requires:
   executable.name == 'dss'
-  graphics_card.vendor == 'NVIDIA Corporation'
+  graphics_card.vendor == 'Intel Corporation'
 depends: dss/status_mlflow
-_summary: Check that DSS status reports NVIDIA GPU acceleration is enabled
+_summary: Check that DSS status reports Intel GPU acceleration is enabled
 estimated_duration: 5s
 command:
   set -eo pipefail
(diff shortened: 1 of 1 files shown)


fernando79513 (approved):
Thanks a lot for the changes!

LGTM +1!

PR #2007, opened 2025-07-07
Add ability to customize VM specs in LXD VM GPU Passthrough test (Bugfix)

MitchellAugustin:
## Description

Some modern GPUs have BAR sizes that exceed the default QEMU pci64-hole-size allowance, or require additional host memory / CPU allocations to work properly in guest VMs when passed through.

This series enables users of the gpgpu provider to apply custom environment options to override these defaults, as-needed.

Users who do not specify these optional environment variables will see no change in behavior.

## Resolved issues

Options required for passthrough to work correctly with large-BAR GPUs can now be applied by the user

## Documentation

Since the actual functionality of these custom options can only be tested on specific platforms with GPUs that meet these criteria, and since the default case with no custom options already has a sufficient unit test, I have not adjusted the unit tests in this PR.

## Tests

This PR was verified on a DGX B200 (buzzsaw) with the following environment variables:
```
[environment]
QEMU_OPTS=" -global q35-pcihost.pci-hole64-size=8192G"
VM_RAM_MB=8192
VM_CPUS=8
```

This section was appended to the default ./test-gpgpu launcher. I then verified that the VM and container GPU passthrough tests both passed as expected when run with these parameters on that platform.

To verify that there were no regressions for other platforms, I loaded my branch on a platform with a consumer-grade GPU (luma), then ran the unmodified ./test-gpgpu and confirmed that all tests passed (as they had when this platform was certified originally) and no custom options were applied.

PR #2136, opened 2025-10-01
Update GPU passthrough tests (BugFix)

pedro-avalos:
## Description

- Replaces `mixbench` with `gpu-burn`
  - `mixbench` is not building for newer versions of CUDA or newer architectures
  - Since the test is just for NVIDIA GPUs at the moment, we can use `gpu-burn`, which we already use to test the GPU without passthrough
- Runs test once instead of re-running and taking average time
- Added missing `environ` section for LXDVM passthrough job (see #2007)

## Resolved issues

- ~~Related to CHECKBOX-2020~~

## Tests

- Updated unit tests
- Run checkbox on lab device and add submission
  - VM tests passing: https://testflinger.canonical.com/jobs/9d102692-c872-4f2e-8686-ac55eb85b90c (container tests failing due to wrong host driver installed)
  - All tests pass: https://testflinger.canonical.com/jobs/b7dbd054-ba55-48dd-a832-34b575396920

### DIFF
--- a/checkbox-support/checkbox_support/lxd_support.py
+++ b/checkbox-support/checkbox_support/lxd_support.py
@@ -212,9 +212,14 @@ class LXD:
         self.run("lxc restart {}".format(self.name))
 
     @retry(10, 10)
-    def wait_until_running(self):
+    def wait_until_running(self, allow_degraded: bool = False):
         """Waits for the instance to be up and running."""
-        self.run("systemctl is-system-running --wait", on_guest=True)
+        try:
+            self.run("systemctl is-system-running --wait", on_guest=True)
+        except subprocess.CalledProcessError as e:
+            if allow_degraded and e.stdout == "degraded":
+                return
+            raise
 
     def add_device(
         self,
--- a/checkbox-support/checkbox_support/lxd_support.py
+++ b/checkbox-support/checkbox_support/lxd_support.py
@@ -217,7 +217,7 @@ class LXD:
         try:
             self.run("systemctl is-system-running --wait", on_guest=True)
         except subprocess.CalledProcessError as e:
-            if allow_degraded and e.stdout == "degraded":
+            if allow_degraded and e.stdout.strip() == "degraded":
                 return
             raise
 
--- a/checkbox-support/checkbox_support/tests/test_lxd_support.py
+++ b/checkbox-support/checkbox_support/tests/test_lxd_support.py
@@ -285,6 +285,21 @@ class TestLXD(TestCase):
         self_mock = MagicMock()
         LXD.wait_until_running(self_mock)
 
+    def test_wait_until_running_degraded(self, logging_mock):
+        self_mock = MagicMock()
+        self_mock.run = MagicMock(
+            side_effect=subprocess.CalledProcessError(1, "", "degraded")
+        )
+        with self.assertRaises(subprocess.CalledProcessError):
+            LXD.wait_until_running(self_mock, allow_degraded=False)
+
+    def test_wait_until_running_allow_degraded(self, logging_mock):
+        self_mock = MagicMock()
+        self_mock.run = MagicMock(
+            side_effect=subprocess.CalledProcessError(1, "", "degraded")
+        )
+        LXD.wait_until_running(self_mock, allow_degraded=True)
+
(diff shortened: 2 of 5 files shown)


MitchellAugustin:
I would not expect this to cause any issues. @k-s-0 what do you think, any concern on your end?

k-s-0:
@MitchellAugustin thanks for pinging me on this, I wasn't aware of it :)

I see no issues with using gpu-burn given there were issues with building mixbench. It seems to have a default runtime of 10s which is fine. Running once instead of re-running and averaging runtime seems fine, too.

What is the purpose / value of allowing `degraded` state LXC containers?

I'm not confident this will resolve the linked [CHECKBOX-2020](https://warthogs.atlassian.net/browse/CHECKBOX-2020) bug, which I believe is specific to NVSwitch / fabricmanager systems, and the device this was tested on (doubletusk) doesn't appear to be such a device.

In the meantime, I'll add context to [CHECKBOX-2020](https://warthogs.atlassian.net/browse/CHECKBOX-2020) and test on Hidon (DGX-H100) and report back.

[CHECKBOX-2020]: https://warthogs.atlassian.net/browse/CHECKBOX-2020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHECKBOX-2020]: https://warthogs.atlassian.net/browse/CHECKBOX-2020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bladernr:
running gpu-burn for a few seconds on this is really just a quick sniff to make sure you can actually use the GPUs from within the VM, it's not really meant for anything more than that.  We could use any workload that actually puts things on the GPU.  gpu-burn was initially chosen mostly because it is an already well established way to exercise the GPUs in checkbox, so the only new thing to add  was just running it from within the container.

It might be better at some point to make this more real-world, but for now it feels like that much is sufficient for the purpose of the test.

k-s-0:
I added context about the original bug to CHECKBOX-2020 ([link](https://warthogs.atlassian.net/browse/CHECKBOX-2020?focusedCommentId=1070146)) :+1:

k-s-0:
Confirmed the hang still occurs on NVLink / Fabric Manager systems (tested on Hidon) with this branch

```
STDOUT: Device gpu added to testbed-20260420174120
DEBUG:checkbox_support.lxd_support: STDOUT: Device gpu added to testbed-20260420174120
Command: lxc start testbed-20260420174120
DEBUG:checkbox_support.lxd_support:Command: lxc start testbed-20260420174120
(no further output)
```

```
[84381.800600] VFIO - User Level meta-driver version: 0.3
[84381.828732] NVRM: Attempting to remove device 0000:1b:00.0 with non-zero usage count!
(no further output)
```

I think it's still worth merging to solve the mixbench issue, though, but CHECKBOX-2020 should be kept open :+1:

k-s-0:
Hi - this PR has become relevant for us as I'm seeing flakiness during 26.04 certification on at least one platform so far (DGX A100) due to Checkbox treating LXC reporting a container state as `degraded` as a failure condition (leading to 2-4 out of 8 VM GPU passthrough tests failing). Once `degraded` is not considered a failing state in Checkbox, if the VM GPU passthrough tests can actually pass in a `degraded` state failure (@pedro-avalos can you confirm if this was tested?), then this should be valuable to merge.

Per my earlier comment, this still won't resolve [CHECKBOX-2020](https://warthogs.atlassian.net/browse/CHECKBOX-2020), but the `degraded` container state tolerance may resolve a different issue.

[CHECKBOX-2020]: https://warthogs.atlassian.net/browse/CHECKBOX-2020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Hook25 (approved):

PR #2143, opened 2025-10-07
Add virtualization-kvm as part of requires of job verify_lxd_vm (New)

rickwu666666:
Add virtualization-kvm as part of requires of job virtualization/verify_lxd_vm

This can prevent the job execute on the system that not support KVM instead of execute the job and get fail in later stage

## Tests
Sideload result
The machine not support kvm
https://certification.canonical.com/hardware/202503-36377/submission/454484/test-results/

The machine support kvm
https://certification.canonical.com/hardware/202010-28347/submission/454489/test-results/

### DIFF
--- a/providers/base/units/virtualization/jobs.pxu
+++ b/providers/base/units/virtualization/jobs.pxu
@@ -6,6 +6,7 @@ estimated_duration: 60.0
 requires:
  executable.name == 'lxc'
  package.name == 'lxd-installer' or snap.name == 'lxd'
+ virtualization.kvm == 'supported'
 command: virtualization.py --debug lxdvm
 _purpose:
  Verifies that an LXD Virtual Machine can be created and launched


bladernr:
This will remove a red flag for systems that do support KVM but it's disabled in BIOS.  I have concern that simply skipping something that's expected to run and pass will lead to legitimate failures not being caught.

For example, when run on a system that supports virtualization:
```
$ kvm-ok
INFO: /dev/kvm exists
KVM acceleration can be used
```

and on the same system with virtualization "accidentally" disabled in BIOS:
```
$ kvm-ok
INFO: Your CPU does not support KVM extensions
INFO: For more detailed results, you should run this as root
HINT:   sudo /usr/sbin/kvm-ok
```

In the latter case, the test would simply be skipped saying "echo "kvm: unsupported" which is incorrect for the hardware... 

Perhaps a better way would be to not skip this based on something easily changed by misconfiguration, and rather depend on a manifest entry somewhere in the virtualization chain that says "yes, we expect virtualization to be supported"

PR #2144, opened 2025-10-07
Add checkbox-gfx crucible tests (New)

mckees:
# Description
This provider contains tests for the graphics stack. This runs Mesa's Crucible testing, which is a component of their CI testing.

# Resolved issues
N/A

# Documentation
See README

# Tests
Tests have been run on Tiger Lake, Alder Lake, and Raptor Lake


contrib/checkbox-gfx/checkbox-provider-gfx/bin/run_crucible_test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
fernando79513, line 1: This test is fairly simple, but it's already creating a temporary file and removing it.
I'd suggest creating a python script for it, so you can just use "subprocess.check_output"  (with `cwd` and `env`) to get the result of the tests and look for the failure string.
Again, the reason for that is that bash scripts are very hard to debug and we can't run unit tests on them.

fernando79513 (changes requested):
Good job here!
I've just left a small coment regarding the `run_crucible_test` scirpt

contrib/checkbox-gfx/checkbox-provider-gfx/bin/run_crucible_test.py
```diff
@@ -0,0 +1,112 @@
...
+        passed = SUCCESS_MARKER in completed.stdout
+        return passed
+
+    finally:
+        # cleanup
+        try:
+            if out_path.exists():
+                out_path.unlink()
+        except Exception:
+            pass
```
fernando79513: If you just read from the output, you don't need to clean up after

contrib/checkbox-gfx/checkbox-provider-gfx/bin/run_crucible_test.py
```diff
@@ -0,0 +1,112 @@
...
+    if use_fork:
+        cmd.append("--fork")
+    cmd.append(test_string)
+
+    env = os.environ.copy()
+    env["XDG_RUNTIME_DIR"] = xdg_runtime_dir
+
+    completed: Optional[subprocess.CompletedProcess[str]] = None
+    try:
+        completed = subprocess.run(
```
fernando79513: You are mostly interested in the test output, so you could use [subprocess.check_output()](https://docs.python.org/3/library/subprocess.html#subprocess.check_output)

fernando79513 (changes requested):
Thanks for the changes!
Please take a look at the comments. I think the python script could be simplified substantialy for our purposes.

fernando79513 (approved):
LGTM +1!

Issue #1211, opened 2024-04-26
RTC operation not permitted causes suspend/suspend_advanced_auto failed sometime


baconYao:
### Bug Description

I found the suspend failed sometimes because of "rtcwake: /dev/rtc0: unable to find device: Operation not permitted".

This problem was observed before, See Checkbox issue: https://github.com/canonical/checkbox/issues/857

At that moment, This problem could reproduced easily on G700 UC22 image. And there was a PR to fix this issue: https://github.com/canonical/checkbox/pull/979

However, I can still run into this problem sometime, not 100% on G1200-evk

[Failure Rate]
60% (3/5)

[Test Submissions]
- Pass:
  - https://certification.canonical.com/hardware/202307-31864/submission/366212/test-results/?term=suspend%2Fsuspend_advanced_auto
  - suspend only: https://certification.canonical.com/hardware/202307-31864/submission/366210/test-results/pass/
- Fail:
  - https://certification.canonical.com/hardware/202307-31864/submission/366214/test-results/fail/
  - https://certification.canonical.com/hardware/202307-31864/submission/366206/test-results/fail/
  - https://certification.canonical.com/hardware/202307-31859/submission/366215/test-results/fail/?term=suspend

### To Reproduce

1. Install checkbox22 via command `$ sudo snap install checkbox22 --beta`
2. Install checkbox via command `$ sudo snap install checkbox --channel="uc22/beta" --devmode`
3. Install bluez via command `$ sudo snap install bluez`
4. Connect checkbox and bluez via command `$sudo snap connect checkbox:bluez bluez:service`
5. Reboot DUT and make sure there's no connection such as SSH, serial console to it.
6. Start testing via Checkbox Control on Host $ checkbox.checkbox-cli control <IP address of DUT>
7. Choose IoT Client Certification for 22.04 classic images (Automated Tests) test plan (`com.canonical.certification::client-cert-iot-server-22-04-automated`)
8. Choose `bluetooth/detect-output` and `suspend/suspend_advanced_auto jobs`
9. Execute testing

### Environment

Image: Ubuntucore
Checkbox: Snap
Version:
```
ceqa@ubuntu:~$ snap list
Name Version Rev Tracking Publisher Notes
bluez 5.64-5 368 22/stable canonical✓ -
checkbox-baoshan 0.5dev-jammy 128 latest/edge ce-certification-qa devmode
checkbox-ce-oem 1.0-jammy 209 latest/edge ce-certification-qa devmode
[...]

PR #2155, opened 2025-10-10
Systemd-unit based runner (new)

Hook25:
## Description

This introduces a new experimental systemd-based runner. This new runner adresses in the short term the rfkill permission denied situation but the objective is to make it, in the long run, the default runner for checkbox strict frontends.

For now this new runner is disabled by default and clearly marked as experimental when enabled.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/1211
Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-2003

## Documentation

New runner clearly documented in the docstring

## Tests

New function is unit tested.

Tested on RPI4 on core24, This can be tested in any environment that has plz-run installed. Currently, only snap packaging is supported. 

To test this use the following launcher:
```
#!/usr/bin/env checkbox-cli
[launcher]
stock_reports = text
[test plan]
unit = com.canonical.certification::smoke
forced=True
[test selection]
forced=True
[features]
systemd_based_job_runner=True
```

The result must be unchanged, the runner should be completely transparent to the user.

Snap builds:
- https://github.com/canonical/checkbox/actions/runs/18403310406
- https://github.com/canonical/checkbox/actions/runs/18403306343


checkbox-ng/plainbox/impl/session/assistant.py
```diff
@@ -1640,8 +1640,16 @@ def run_job(
...
+                    self._metadata.FLAG_FEATURE_SYSTEMD_BASED_JOB_RUNNER
+                    in self._metadata.flags
+                )
                 result = self._runner.run_job(
-                    job, job_state, self._config.environment, ui
+                    job,
+                    job_state,
+                    self._config.environment,
+                    ui,
+                    systemd_unit=systemd_unit,
```
pieqq: Same as above.

checkbox-ng/plainbox/impl/execution.py
```diff
@@ -713,3 +729,53 @@ def get_execution_command(
         cmd += ["aa-exec", "-p", "unconfined"]
     cmd += [job.shell, "-c", job.command]
     return cmd
+
+
+def get_execution_command_systemd_unit(
+    job, environ, session_id, nest_dir, target_user, extra_env=None
+):
+    """
+    Generates a command that if executed runs a Checkbox command section
```
pieqq: This docstring header is exactly the same as the one for `get_execution_command_subshell`. Maybe


```suggestion
    Generates a command that if executed runs a Checkbox command section as a systemd unit
```

or something like that (not sure how to keep it under 79 chars while differentiating it from the other one though)

pieqq (dismissed):
Thanks for the fix and the additional logging information!

Issue #1823, opened 2025-03-26
Disk Tests MISSING_PARAM_1


mreed8855:
### Bug Description

On some of the newer systems such as Granite Rapids we are seeing disk test fail due to missing resources.  

"Cached read timing benchmark of MISSING_PARAM_1 using hdparm"
Template 'com.canonical.certification::benchmarks/disk/hdparm-cache-read_name' generated an invalid unit. Field 'command' has value 'hdparm -T /dev/{name}' but the template-resource didn't generate any value for 'name'

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

Run certify-ubuntu-server or test-disk to recreate

### Expected Result

The disk test run properly with the disk names .

### Actual Result

 disk/disk_stress_ng_MISSING_PARAM_1
 disk/read_performance_MISSING_PARAM_1

### Environment

- ubuntu 24.04

### Relevant log output

```shell

```

pieqq:
This is a recent addition to Checkbox. If a template job fails to generate jobs, this will be reported as you've seen. `benchmarks/disk/hdparm-cache-read_name` depends on the `device` resource job to work. Can you provide the output of the `device` job on that specific device?

mreed8855:
I have included one of the recent test runs that is showing this error.

https://certification.canonical.com/certificates/2503-16520/

mreed8855:
path: /devices/pci0000:d5/0000:d5:02.0/0000:d6:00.0/0000:d7:02.0/0000:da:00.0
bus: pci
driver: nvme
product_id: 20
vendor_id: 7695
subproduct_id: 8776
subvendor_id: 4136
vendor: KIOXIA Corporation
vendor_slug: KIOXIA_Corporation

path: /devices/pci0000:d5/0000:d5:02.0/0000:d6:00.0/0000:d7:02.0/0000:da:00.0/nvme/nvme0
name: nvme0
bus: nvme
category: OTHER
driver: nvme
product_id: 49200
vendor_id: 4096
subproduct_id: 9013
subvendor_id: 4136
product: PEX890xx PCIe Gen 5 Switch
vendor: KIOXIA Corporation
product_slug: PEX890xx_PCIe_Gen_5_Switch
vendor_slug: KIOXIA_Corporation

path: /devices/pci0000:d5/0000:d5:02.0/0000:d6:00.0/0000:d7:02.0/0000:da:00.0/nvme
bus: nvme
category: DISK
driver: nvme
product_id: 20
[...]

PR #2160, opened 2025-10-14
Fix missing name field for NVMe devices without DEVNAME (bugfix)

mreed8855:
NVMe controller devices that don't have a DEVNAME were 
being categorized as DISK but had no name field, causing 
template parameter substitution to fail with MISSING_PARAM_1 
errors.

This fix first attempts to extract nvmeXnY pattern from DEVPATH and  
falls back to extracting nvmeX and appending n1 if the former fails.

This should resolve MISSING_PARAM errors in disk tests for systems
with various NVMe device path structures.

## Resolved issues
https://github.com/canonical/checkbox/issues/1823
https://warthogs.atlassian.net/browse/ARGOS-1477

### DIFF
--- a/checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
+++ b/checkbox-support/checkbox_support/parsers/tests/test_udevadm.py
@@ -1205,12 +1205,6 @@ class TestUdevadmParser(TestCase, UdevadmDataMixIn):
 
     def test_DISK_SAMSUNG_KIOXIA(self):
         """
-        Test system with both standard NVMe disks and virtual NVMe disks.
-        This system has:
-        - 2 standard Samsung NVMe disks (nvme0n1, nvme1n1)
-        - 8 virtual Kioxia NVMe disks with controller notation (nvme2c2n1,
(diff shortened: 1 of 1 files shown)


pieqq (changes requested):
Looks good!

The added unit tests cover the different things that could happen. Could you add one more test that uses real udevadm data? We usually do that to have more regression data.

You can use the one from [this submission](https://certification.canonical.com/hardware/202504-36718/submission/427801/) (it's in the archive, in `attachment_files/com.canonical.certification__udev_attachment`), place it in `checkbox-support/checkbox_support/parsers/tests/udevadm_data` as a `.txt` file, and use the `parse` function like [here](https://github.com/canonical/checkbox/blob/dcb241925f8716a9ec2c306555976491b626097a/checkbox-support/checkbox_support/parsers/tests/test_udevadm.py#L1076) to extract the data, then run some checks on it, namely, that there are 10 disks identified and that their name is matching the expectations (`nvme0n1` and `nvme1n1` for the Samsung disks, and `nvme2c2n1`, `nvme3c3n1`, etc. for the Kioxia disks).

Thanks!

Issue #2159, opened 2025-10-14
Checkbox service does not start on Questing


bladernr:
### Bug Description

Discovered while trying to do 25.10 validation for the IBM Z project on two Z instances (one zVM and one LPAR). In both cases, I discovered that the checkbox-ng service is not starting.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

Install checkbox from edge or beta on Questing (I'm using server).
CHeck for the checkbox-ng service.

### Expected Result

service is running

### Actual Result

service is not running due to failures.

### Environment

25.10 server
IBM Z15 s390x LPAR and VM

### Relevant log output

```shell
ubuntu@hwe0008:~$ systemctl start checkbox-ng.service
==== AUTHENTICATING FOR org.freedesktop.systemd1.manage-units ====
Authentication is required to start 'checkbox-ng.service'.
Authenticating as: ubuntu
Password: 
==== AUTHENTICATION COMPLETE ====
ubuntu@hwe0008:~$ ps axf |grep checkbox-ng
   1711 pts/0    S+     0:00              \_ grep --color=auto checkbox-ng
ubuntu@hwe0008:~$ systemctl status checkbox-ng
× checkbox-ng.service - Checkbox Remote Service
     Loaded: loaded (/usr/lib/systemd/system/checkbox-ng.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2025-10-14 15:22:15 UTC; 11s ago
   Duration: 341ms
 Invocation: 36aafa2ca9fa4e7d8011e566a4f8b356
    Process: 1706 ExecStart=/usr/bin/checkbox-cli run-agent (code=exited, status=1/FAILURE)
   Main PID: 1706 (code=exited, status=1/FAILURE)

Oct 14 15:22:15 hwe0008 systemd[1]: checkbox-ng.service: Scheduled restart job, restart counter is at 5.
Oct 14 15:22:15 hwe0008 systemd[1]: checkbox-ng.service: Start request repeated too quickly.
Oct 14 15:22:15 hwe0008 systemd[1]: checkbox-ng.service: Failed with result 'exit-code'.
Oct 14 15:22:15 hwe0008 systemd[1]: Failed to start checkbox-ng.service - Checkbox Remote Service.
```

PR #2161, opened 2025-10-15
Use --non-interactive instead of -A (bugfix)

Hook25:
## Description

Checkbox on Questing is unable to start the service and always prompts for password even if passwordless sudo is configured. The reason is that `sudo-rs` doesn't support `-A` (askpass) option yet. Checkbox is using this option both to detect that we are able to run `sudo` when root (crashing the service, as checkbox thinks something is wrong with sudo itself) and detect if the password is required in local (interpreting the failure due to `-A` is not supported as passwordless sudo is not supported). 

This uses another option `--non-interactive` in combination with `--reset-timestamp` to have the same result as before, namely the command will fail if user interaction is required, which is what we want in this situation.

This also improves the error message for the agent, as before it was very difficoult to know exactly why the process was crashing, now we log the sudo output

See `-A` not supported: https://github.com/trifectatechfoundation/sudo-rs/issues/1249#issuecomment-3274142007

## Resolved issues

Fixes: CHECKBOX-2037
Fixes: https://github.com/canonical/checkbox/issues/2159

## Documentation

Added a comment to explain the logic behind the command and the reason for the switch in case someone will wonder in the future why we don't use `-A`

## Tests

To test this do the following:
```
lxc launch ubuntu:questing questing
lxc shell questing
root@questing: add-apt-repository ppa:checkbox-dev/edge
root@questing: apt update
root@questing: apt install checkbox-ng checkbox-provider-resource checkbox-provider-base
root@questing: sudo -u ubuntu bash
ubuntu@questing: sudo --non-interactive --reset-timestamp true; echo $?
0 
# this confirms that passwordless sudo is enabled! Try it on your machine, it should fail
# if you aren't a cowboy
 # Now try to run the smoke testplan, you will be prompted for a password 
ubuntu@questing: checkbox-cli 
# /smoke <enter>
Enter sudo password: <- this is actually not sudo, this is checkbox!
```
Verify that also the agent is broken
```
root@questing: checkbox-cli run-agent
BOOM
```
Now patch the installation with the sudo_broker from this PR:
```
root@questing: curl https://raw.githubusercontent.com/canonical/checkbox/refs/heads/passwordless_sudo_rs/checkbox-ng/plainbox/impl/secure/sudo_broker.py >  /usr/lib/python3/dist-packages/plainbox/impl/secure/sudo_broker.py
```
Rerun the above, everything should work as intended.

### DIFF
--- a/checkbox-ng/plainbox/impl/secure/sudo_broker.py
+++ b/checkbox-ng/plainbox/impl/secure/sudo_broker.py
@@ -33,6 +33,7 @@ import sys
 from plainbox.i18n import gettext as _
 from subprocess import (
     check_output,
+    check_call,
     CalledProcessError,
     STDOUT,
     DEVNULL,
--- a/checkbox-ng/plainbox/impl/secure/test_sudo_broker.py
+++ b/checkbox-ng/plainbox/impl/secure/test_sudo_broker.py
@@ -17,7 +17,10 @@
 from subprocess import CalledProcessError
 from unittest import TestCase, mock
 
-from plainbox.impl.secure.sudo_broker import is_passwordless_sudo
+from plainbox.impl.secure.sudo_broker import (
+    is_passwordless_sudo,
+    validate_pass,
+)
 
 
 class IsPasswordlessSudoTests(TestCase):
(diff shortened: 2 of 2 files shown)


fernando79513 (changes requested):
Everything looks fine, I've just added a question regarding if we should go back to askpass once the feature is implemented.

PR #2166, opened 2025-10-20
fix: add more mock time values in test_prime_offload_tester to handle (Bugfix)

seankingyang:
## Description
add more mock time values in test_prime_offload_tester to handle

The check_offload() method makes 5 time.time() calls in the failure scenario:
Call #1: deadline = time.time() + timeout → returns 0, sets deadline to 1
Call #2: while time.time() < deadline: → returns 0, check 0 < 1 = True, enters loop
Call #3: while time.time() < deadline: → returns 1, check 1 < 1 = False, exits loop
Call #4: self.logger.info("Checking fail:") → LogRecord creation calls time.time() → returns 2   (ref: https://docs.python.org/3/library/logging.html#logrecord-attributes)
Call #5: self.logger.info(" Couldn't find process") → LogRecord creation calls time.time() → returns 3

ref: `python3 -c "import logging, inspect; print(inspect.getsource(logging.LogRecord.__init__))"` , https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L283

## Resolved issues
PR #2162
https://github.com/canonical/checkbox/actions/runs/18641071113/job/53139932346?pr=2162

## Tests
cherry pick this commit 
https://github.com/canonical/checkbox/actions/runs/18641579136/job/53141487716?pr=2162

### DIFF
--- a/providers/base/tests/test_prime_offload_tester.py
+++ b/providers/base/tests/test_prime_offload_tester.py
@@ -325,7 +325,7 @@ class CheckOffloadTests(unittest.TestCase):
         cmd = ["echo"]
         # get_clients return string that doesn't include cmd
         mock_client.return_value = ""
-        mock_time.side_effect = [0, 0, 1, 2]
+        mock_time.side_effect = [0, 0, 1, 2, 3, 4]
         po = PrimeOffloader()
         self.assertEqual(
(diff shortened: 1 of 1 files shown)


hanhsuan (approved):
LGTM, Thanks for finding this bug in the unit test.

PR #2170, opened 2025-10-21
Fix net_driver_info crash when dumping parameters (Bugfix)

pinaraf:
## Description

When the net_driver_info script encounters a parameter that is not valid utf-8 text, it crashes and doesn't produce more information.
Some drivers (for instance rtl8822ce) have such parameters and thus shortens available information when there is an issue to investigate.

### DIFF
--- a/providers/base/bin/net_driver_info.py
+++ b/providers/base/bin/net_driver_info.py
@@ -42,7 +42,7 @@ for interface, driver in driver_list:
             if path.is_file():
                 # Path.read_text is new in python 3.5 but we want to support
                 # trusty as well, which uses python 3.4 by default.
-                with open(str(path), "r") as f:
+                with open(str(path), "r", errors="backslashreplace") as f:
                     print("    {}: {}".format(path.name, f.read().strip()))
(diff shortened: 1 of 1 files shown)


providers/base/bin/net_driver_info.py
```diff
@@ -42,8 +42,11 @@
             if path.is_file():
                 # Path.read_text is new in python 3.5 but we want to support
                 # trusty as well, which uses python 3.4 by default.
-                with open(str(path), "r") as f:
-                    print("    {}: {}".format(path.name, f.read().strip()))
+                try:
+                    with open(str(path), "r") as f:
+                        print("    {}: {}".format(path.name, f.read().strip()))
+                except:
```
Hook25: Please don't catch like this, flake8 is sad. Catch the exact exception if possible or at least `Exception`. Print what you catch! Else the error is gone from view

Hook25 (changes requested):
Please consider the following.

Feel free to not add unit tests for this change, disregard codecov

PR #2197, opened 2025-11-12
Add a test case to Verify the test audio I/O after suspend with previously occupied devices (new)

GabrielChenCC:
## Description
Validate the audio could input and output after suspending while audio devices are occupied before suspending
According to  https://warthogs.atlassian.net/browse/SOMERVILLE-1356,  OEMQA needs a way to test this kind of condition.

## Resolved issues
https://warthogs.atlassian.net/browse/SOMERVILLE-1356

## Tests
Please refer to the test result https://certification.canonical.com/hardware/202509-37925/submission/458650/
This case should only execute after suspend, so it needs to depend on suspend/suspend_advanced_auto. @hanhsuan gave me the suggestion, it was very helpful in completing this case. thanks.


hanhsuan (approved):
Tested in my side. LGTM.

PR #2206, opened 2025-11-19
Move boot status check to the beginning of reboot checks (BugFix)

tomli380576:
## Description

This PR moves the wait_for_graphic_target polling function to the beginning of main() to ensure a more consistent behavior since now it will always wait for the system to finish booting first instead of only when there's a desktop.

## Resolved issues

This "always wait for the system to finish booting" behavior already exists, implicitly, in the current implementation because `-g` flag is always specified in the test plan. However it was only called at the final graphics check which can lead to inconsistent results for the other tests like the failed services test being run after network.target but before graphical.target etc. Moving this polling step to the beginning can make sure no tests are run and no baseline data used for comparison is collected until the system has finished booting.

## Documentation

This PR also switches from `systemd-analyze critical-chain` to `systemctl is-system-running` to reduce the likelihood of timeouts. Based on https://unix.stackexchange.com/questions/460324/is-there-a-way-to-wait-for-boot-to-complete we don't have to call the slow `systemd-analyze` if we just need to know whether the system finished booting. `systemctl is-system-running` exists on ubuntu 16 as well so it's safe to call.

## Tests

Original unit tests

C3 submissions:

DUT that boots fast: https://certification.canonical.com/hardware/202511-38075/submission/460132/
- Failures are expected in this submission because they are fwts errors

DUT that boots slow: https://certification.canonical.com/hardware/202407-34222/submission/460133/
- Failed tests are also real failures not the script itself

### DIFF
--- a/providers/base/bin/reboot_check_test.py
+++ b/providers/base/bin/reboot_check_test.py
@@ -640,7 +640,7 @@ def main() -> int:
 
     args = create_parser().parse_args()
 
-    print("Waiting for boot to finish...", flush=True)
+    print("Waiting for boot to finish...")
     if poll_systemctl_is_system_running(args.boot_ready_timeout):
         print("[ OK ] System finished booting!")
     else:


Hook25 (approved):
the more you know. Thanks!

PR #2208, opened 2025-11-19
New documentation for groups feature (Infra)

fernando79513:
## Description

This PR includes:
 - A new "how-to" guide for using groups
 - A new section under the test jobs tutorial
 - The job reference page includes the "group" field  

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-2012

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -44,6 +44,7 @@ testplan
 TServer
 unary
 validator
+vendorize
 venv
 vfork
 wodim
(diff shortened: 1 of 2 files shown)


docs/how-to/using-groups.rst
```diff
@@ -0,0 +1,231 @@
...
+    after: device_test_read
+    command: echo "Removing device"
+
+    id: device_test_plan
+    name: device_test_plan
+    unit: test plan
+    _summary: Step by step device test
+    include:
+      device_.*
+      gather_device_info
```
pieqq: This is interesting, because although you put `gather_device_info` after the `device_*` jobs, it ends up being executed before the whole group. It would be good to add a little paragraph to explain this, something like

> Although `gather_device_info` is placed after the `device_*` jobs, it will be executed by Checkbox before them because `device_test_write` depends on it and because it's part of the `device_test` group like all the other `device_*` jobs.

pieqq (approved):
Thanks for the update!

The link check thing seems to be borked at the moment because of snapcraft.io, so hopefully it will pass at some point...

PR #2215, opened 2025-11-24
providers/base: add fscrypt test (New)

fhost:
New: Add a small fscrypt test into the base provider, in its own category.

## Description

The new fscrypt test generates an ext4 disk image, mount it and setup fscrypt.
Then it creates a file in the disk image mountpoint, locks it and verifies that the file is not accessible anymore. Finally, it unlocks it and check that the file is accessible and its content is correct.

## Tests

`checkbox-cli`, select `automated fscrypt test`, run the test plan. Tested on my laptop and on my project device (qualcomm arm64 device). Testing with fscrypt already setup and not setup.


hawesy-can (commented):
LGTM, but I'm not an expert in this area.

providers/base/bin/fscrypt_test.py
```diff
@@ -0,0 +1,141 @@
...
+                    "--source=raw_key",
+                    "--name=test_key",
+                    f"--key={str(key_file)}",
+                    str(test_dir),
+                ]
+            )
+
+            # Write a file inside
+            with test_file.open("w") as f:
+                f.write("test\n")
```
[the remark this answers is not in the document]
fhost: Fixed!

providers/base/bin/fscrypt_test.py
```diff
@@ -0,0 +1,141 @@
...
+                check=True,
+            )
+
+            # Confirm fscrypt is enabled
+            output = subprocess.check_output(
+                ["fscrypt", "status"],
+                text=True,
+            )
+            found = False
+            pattern = re.compile(rf"^{re.escape(str(mnt))}.*supported *Yes$")
```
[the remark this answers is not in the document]
fhost: I don't understand what's wrong there. I do reproduce on my side.

providers/base/bin/fscrypt_test.py
```diff
@@ -0,0 +1,144 @@
...
+
+            # Encrypt directory
+            subprocess.check_call(
+                [
+                    "fscrypt",
+                    "encrypt",
+                    "--quiet",
+                    "--source=raw_key",
+                    "--name=test_key",
+                    f"--key={str(key_file)}",
```
pieqq: f-strings are not supported in python 3.5 which we still support, so please use `.format` instead.

providers/base/bin/fscrypt_test.py
```diff
@@ -0,0 +1,144 @@
...
+            print("File correctly inaccessible when locked")
+
+            # Unlock the directory
+            subprocess.check_call(
+                ["fscrypt", "unlock", f"--key={str(key_file)}", str(test_dir)]
+            )
+
+            with test_file.open("r") as f:
+                content = f.read().strip()
+            if content != test_content:
```
pieqq: For easier debugging or review, please print the expected content and the actual content before raising `SystemExit`.

providers/base/bin/fscrypt_test.py
```diff
@@ -0,0 +1,144 @@
...
+                raise SystemExit("File contents not correct after unlock")
+            print("File is accessible and content is correct after unlock")
+        finally:
+            subprocess.check_call(["umount", str(mnt)])
+            if not fscrypt_setup and fscrypt_config.exists():
```
pieqq: I'm not sure to understand this line, given that at the top of the script, you set

```
fscrypt_setup = fscrypt_config.exists()
```

What's the expectation here?

pieqq (changes requested):
Sorry for the long silence. I've added a few comments, see if they make sense.

providers/base/bin/fscrypt_test.py
fhost: Depending if there was an error during the test, the configuration might not have been created. I could just remove the test (the exists() one) and do a "rm -f" if you prefer.

pieqq (changes requested):
One last tiny change request, and then it should be good.

pieqq (approved):
Thanks! Let's see if CI passes for older builds.

PR #2222, opened 2025-12-03
Migrate pc-sanity to new script for gpu drivers (New)

Artur-at-work:
This change adds python script to replace gpu-driver-checker.sh script. It also adds some unit tests.
This test case is the candidate to move to providers/base after running it for several cycles in pc-sanity

## Resolved issues
Closes: https://warthogs.atlassian.net/browse/OEX86-892

## Tests
Run old and new scripts on DUT with and without NV dgpu.
Verify outputs and return errors are the same.


cyruslien (approved):
LGTM.

PR #2242, opened 2025-12-12
Minimum resolution depends on a desktop session being available (BugFix)

p-gentili:
## Description

The `*_minimum_resolution_*` test case is not usually included in server test plans because it relies on an available desktop session. Yet, the job doesn't depend on it, making it fail in case it's being run on a server image.

This PR just adds a new dependency using an already existing resource job.

## Resolved issues

N/A

## Documentation

N/A

## Tests

1. Run it locally from source
3. Run it sideloading the provider on remote AND updating the agent


Hook25 (approved):
thanks

PR #2247, opened 2025-12-15
Systemd based runner default on core (bugfix)

Hook25:
## Description

The current subshell runner will be broken on core (in some scenarios) starting 6Jan with the release of snapd beta to stable. This comes in the form of a cgroup tracking error when calling a snap from a snap like the following:
```
/system.slice/snap.checkbox-riverside.remote-slave.service is not a snap cgroup for tag  snap.bluez.hcitool 
```

For this reason we must make the new systemd-based runner default on core. Note that the option to use the old runner will remain via the feature flag as the value from the user will always be respected. This only changes the default value from `False` (never enabled by default) to `None` (sometimes enabled by default). 

This PR also updates the build repo for `plz-run` to my Github fork (pending move under the Canonical org) as we need to be able to update this tool while the maintainer is OOO. Note that this is necessary as I have[ landed a bugfix there](https://github.com/Hook25/plz-run/pull/1).

Finally this pr also adds a small bugfix (there is a typo in the PAM module to use, to double check simply ls `/etc/pam.d`, there is no `systemd-login`, its `system-login`!)

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-2142

## Documentation

N/A

## Tests

Tested this locally to verify the bugfix fixes the regressions reported by QA.
Snap builds:
- https://github.com/canonical/checkbox/actions/runs/20227521842
- https://github.com/canonical/checkbox/actions/runs/20227530435

### DIFF
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -780,7 +780,7 @@ def get_execution_command_systemd_unit(
         target_user,
     ]
     if target_user != "root":
-        cmd += ["-pam", "systemd-login"]
+        cmd += ["-pam", "system-login"]
     env = get_differential_execution_environment(
         job, environ, session_id, nest_dir, extra_env
     )


fernando79513 (approved):
LGTM +1!

PR #2265, opened 2025-12-26
Fix gst discoverer variable (Bugfix)

rickwu666666:
## Description
We found that variable `GST_DISCOVERER` has been used in camera test for the same purpose. Therefore. we made the change to use the same variable in `gst_utilts.py`


stanley31huang (approved):
Let's have this workaround first, but we need to think about how to align this kind of variables later.

PR #2266, opened 2025-12-26
Fix wrong jobs that included in iot classic plan (Bugfix)

rickwu666666:
## Description
disk/encryption/detect and miscellanea/secure_boot_mode should be automated jobs
Also the secure_boot_mode.* will include the job that depended on gadget snap which only for ubuntu core.


pieqq (approved):
LGTM. Thanks!

PR #2276, opened 2026-01-13
Add printer test (New)

rickwu666666:
## Description
Adding the test for testing printer since we have a project which is a POS system with printer connected.

## Tests
Sideload result without include manifest.
https://certification.canonical.com/hardware/202511-38103/submission/467449/test-results/

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/printer_test.py
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/printer_test.py
@@ -148,11 +148,7 @@ def main():
 
         # 4. Monitor status
         try:
-            run_with_timeout(
-                monitor_job,
-                30,
-                printer_name,
-                job_id)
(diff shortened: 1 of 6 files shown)


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/printer_test.py
```diff
@@ -0,0 +1,168 @@
...
+    keyword = args.keyword
+    printer_name = "Printer_Tester"
+
+    # 1. Find URI
+    uri = find_printer_uri(keyword)
+    if not uri:
+        logging.error(
+            "Could not find a printer with keyword '{}'.".format(keyword)
+        )
+        raise SystemExit(1)
```
tomli380576: Optional: directly exit with the error message:

```py
raise SystemExit("Could not find a printer with keyword '{}'.".format(keyword))
```

this will still return 1

stanley31huang (approved):
LGTM

PR #2282, opened 2026-01-14
Add barcode scanner test (New)

rickwu666666:
## Description
Add barcode scanner test.
- automated test to check if the expected device been listed.
- manual test to check if the scanned barcode value are expected.

## Tests
https://certification.canonical.com/hardware/202511-38103/submission/467699/test-results/

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/barcode_scanner_test.py
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/barcode_scanner_test.py
@@ -85,7 +85,7 @@ def find_device_by_name(name_substring: str):
     return None
 
 
-def listen_and_decode(device: evdev.InputDevice) -> None:
+def listen_and_decode(device: evdev.InputDevice):
     """Listens for events with a timeout and decodes keystrokes."""
 
     # Grab device to prevent it from typing into other windows


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/barcode_scanner_test.py
```diff
@@ -0,0 +1,192 @@
...
+                                return True
+                            else:
+                                barcode_buffer += char
+
+    # Clean up
+    device.ungrab()
+
+
+if __name__ == "__main__":
+    logging.basicConfig(
```
hanhsuan: I cannot find the guide about there should not be added too much logical in the `if __name__ == "__main__":` for checkbox test case script. Do you remember where is it?

rickwu666666: I'm sorry but I don't recall. However, that make sense. I've push the fix.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/barcode_scanner_test.py
```diff
@@ -0,0 +1,197 @@
...
+        elapsed = time.time() - start_time
+        remaining = timeout_sec - elapsed
+
+        if remaining <= 0:
+            logging.error("Time limit reached.")
+            break
+
+        # select.select waits for data on the file descriptor
+        # This is efficient (sleeps until interrupt) vs a busy loop
+        r, w, x = select.select([device.fd], [], [], remaining)
```
hanhsuan: If you don't need `w` and `x`, you should use `_` instead of them.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/barcode_scanner_test.py
```diff
@@ -0,0 +1,197 @@
...
+        action="store_true",
+        help="Only check if the device is present",
+    )
+    args = parser.parse_args()
+
+    # 1. Find the device
+    target_device = find_device_by_name(args.name)
+
+    if not target_device:
+        raise SystemExit(1)
```
hanhsuan: Do you mind to print the exit reason instead of `1`

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/barcode_scanner_test.py
```diff
@@ -0,0 +1,186 @@
...
+                        else:
+                            char = char.upper()
+
+                    if char == "\n":
+                        logging.info(
+                            "[SUCCESS] Barcode Detected: {}".format(
+                                barcode_buffer
+                            )
+                        )
+                        return True
```
hanhsuan: What is the reason you have to use `return` instead of `break` ?

hanhsuan (changes requested):
Do you mind to add date type for the function input such as
```python
def find_device_by_name(name_substring: str):
```

PR #2284, opened 2026-01-14
Move "Not Supported Tests" in their own section in the HTML report (bugfix)

pieqq:
## Description

1. Move "Not Supported Tests" into a separate section of the HTML report. This section only shows up if there are "not-supported" jobs, and contains a small explanation paragraph. This helps reviewing test runs, since often, readers don't care about jobs that are skipped because the feature to be tested is not present in the device.
2. Remove "not-supported" counts from the pie chart: reviewers are more interested into results that target the DUT (pass/fail/skip), not tests that have not been executed because said DUT cannot test them in the first place.

The same was done for the multi-page HTML template which is used when launching the `checkbox-cli merge-reports` command.
## Resolved issues

Fix https://warthogs.atlassian.net/browse/CHECKBOX-651

## Tests

These HTML reports show the outcome when running locally with the smoke test plan, and when using the `checkbox-cli merge-reports` command with 2 submissions (this uses the multi-page template):

[html-reports.zip](https://github.com/user-attachments/files/24669947/html-reports.zip)

I've also run this on checkbox16 and it worked as expected.

Run locally using different test plans (manual, automated), manually skipping tests. Here is the screenshot of a test run on 24.04, with all the fields expanded (they're all collapsed by default):

Screenshot 2026-01-14 at 14-14-30 System Testing Report

### DIFF
--- a/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
+++ b/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
@@ -86,18 +86,24 @@
                                 datasets: [{
                                     data: [
                                         {%- for outcome, total in state.get_test_outcome_stats()|dictsort %}
+                                        {%- if outcome != 'not-supported' %}
                                         {{ total }},
+                                        {%- endif %}
                                         {%- endfor %}
                                     ],
                                     backgroundColor: [
                                         {%- for outcome, total in state.get_test_outcome_stats()|dictsort %}
+                                        {%- if outcome != 'not-supported' %}
                                         "{{ OUTCOME_METADATA_MAP[outcome].color_hex }}",
+                                        {%- endif %}
                                         {%- endfor %}
                                     ],
                                 }],
                                 labels: [
                                     {%- for outcome, total in state.get_test_outcome_stats()|dictsort %}
+                                    {%- if outcome != 'not-supported' %}
                                     "{{ OUTCOME_METADATA_MAP[outcome].tr_label }}",
+                                    {%- endif %}
                                     {%- endfor %}
                                 ]
                             },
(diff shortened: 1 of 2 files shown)


checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
```diff
@@ -198,8 +209,51 @@ <h3>{{ cat_name }}<span class="ui-li-count {{ OUTCOME_METADATA_MAP[category_outc
...
+                    {%- endif %}
+                    {%- endfor %}
+                    <p></p>
+                    {%- set has_any_not_supported = [] %}
+                    {%- for job_id, job_state in job_state_map|dictsort if job_state.result.outcome == 'not-supported' and job_state.job.plugin not in ("resource", "attachment") %}
+                        {%- if has_any_not_supported.append(1) %}{% endif %}
+                    {%- endfor %}
+                    {%- if has_any_not_supported|length > 0 %}
+                    <h2>Not Supported Tests</h2>
+                    <p>The following tests are part of the test plan but were not run because they don't match the device's available features (for instance, wireless tests are skipped on a device without a WiFi chip).</p>
```
Hook25: I suggest a slight tweak here because reading it `for instance` tricked me into thinking `in this instance`, which, although its my language parser failing, I suppose it is less clear:
```suggestion
                    <p>The following tests are part of the test plan but were not run because they don't match the device's available features (for example, wireless tests are skipped on a device without a WiFi chip).</p>
```

Hook25 (dismissed):
+1 LGTM, consider the comment below, or not

pieqq:
@Hook25 I realized I missed the multi-page template, so I added it here (and tested it). I also tested the HTML template on checkbox16 snap and it works.

Hook25 (approved):

PR #2300, opened 2026-01-22
Add credit card reader test (New)

rickwu666666:
## Description
Add test for credit card reader. We develop the test script according to the Verifone's credit card reader. However, we check if the device echo ACK `\x06` back after sending out enquiry string `\x05` which is a quite common method to check if serial device's connection established.

## Tests
sideload result
https://certification.canonical.com/hardware/202511-38101/submission/469103/test-results/


stanley31huang (approved):
LGTM

Issue #2094, opened 2025-08-27
Disk stress tests use /tmp, which sometimes isn't on the disk!


rodwsmith:
### Bug Description

Recent versions of Ubuntu sometimes place /tmp on a virtual (in-memory) filesystem; however, the disk stress tests in Checkbox (disk/disk_stress_ng_{device}) use this directory in a now-pointless attempt to stress-test the root (/) disk hardware. This problem does _not_ affect disks other than the root (/) disk.

In my initial experiments, it looks like MAAS-deployed nodes do not use a virtual /tmp, but machines deployed in other ways do.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Deploy a node
2. Do "df -h /tmp" to verify where /tmp is mounted
3. Run test-storage or some other test plan that runs the disk stress tests
4. Examine the results; the disk stress results for the disk holding the root (/) filesystem include a line at the top like "Using test directory: '//tmp/stress-ng-586d98c1-5cbb-11f0-b144-107c61539e49'", indicating use of /tmp

### Expected Result

The test should be run somewhere on the intended disk.

### Actual Result

The test sometimes runs in a virtual filesystem, which is pointless. 

### Environment

- OS: Ubuntu 24.10 or later
- It's unclear to me what triggers use of a virtual /tmp filesystem

### Relevant log output

```shell

```

PR #2302, opened 2026-01-23
Change location of stress-ng disk tests to avoid running on a tempfs (BugFix)

rodwsmith:
Change location of stress-ng disk tests to avoid running on a tempfs (BugFix)

## Description

By default, disk stress tests had been performed in a subdirectory of `/tmp` on the main disk or a mounted secondary disk (e.g., `/tmp/stress-ng-{uuid}`, where `{uuid}` is a UUID generated on a per-run basis, or `/mnt/sdb/tmp/stress-ng-{uuid}`); however, beginning with Ubuntu 24.10, `/tmp` is sometimes a tempfs, which means the test is not testing actual disk hardware. 

This PR moves the location of the disk test directory to a subdirectory of the disk's root directory (e.g., `/stress-ng-{uuid}` or `/mnt/sdb/stress-ng-{uuid}`), thus ensuring that actual disk hardware is tested.

Pre-existing code cleans up the test directory, so it should not clutter the filesystem after the test is run.

In addition to the trivial one-line change to the `disk_support.py` script, this PR required an entirely new unit test for that script. I used Copilot to create an initial unit test, then tweaked it so that it actually worked. (It got a few things wrong.)

## Resolved issues

Jira: https://warthogs.atlassian.net/browse/SERVCERT-1912
GitHub issue: https://github.com/canonical/checkbox/issues/2094

## Documentation

## Tests

I tested by running the test script (`stress_ng_test.py`) both directly and via the Checkbox `test-storage` launcher. (This PR changes the support file, `disk_support.py`, which is where the test location is actually set.) I observed the temporary directory being created under `/tmp` with the original script and under root (`/`) with the changed version, by monitoring those locations manually in another session. My main test system had two disks, to ensure that both the root (`/`) disk and other disk(s) would work correctly. (The problem is only with the main root disk, really, but the fix affects both disks. The change to secondary disks is unimportant to the test's functionality.) The test also reports where its temporary files are being created, as in:

```
Using test directory: '//stress-ng-b79cdd89-f7fe-11f0-81e4-00133b126f99'
```

A sample submission using the revised test is here:

https://certification.canonical.com/hardware/201409-15506/submission/469447/

### DIFF
--- a/checkbox-support/checkbox_support/disk_support.py
+++ b/checkbox-support/checkbox_support/disk_support.py
@@ -229,7 +229,7 @@ class Disk:
                         full_device, self.mount_point
                     )
                 )
-            self.test_dir = "{}/tmp/stress-ng-{}".format(
+            self.test_dir = "{}/stress-ng-{}".format(
                 self.mount_point, uuid.uuid1()
             )
             os.makedirs(self.test_dir, exist_ok=True)
--- a/checkbox-support/checkbox_support/tests/test_disk_support.py
+++ b/checkbox-support/checkbox_support/tests/test_disk_support.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+# encoding: UTF-8
+# Copyright (c) 2026 Canonical Ltd.
+#
+# Authors:
+#     Rod Smith <rod.smith@canonical.com> using CoPilot
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


pedro-avalos (approved):
Thanks for adding unit tests!

The change makes sense, the tests are passing, and the submission looks okay to me

PR #2303, opened 2026-01-23
Use AmbientCapabilities on non-core16 systems (bugfix)

Hook25:
## Description

fs caps dont work on newer bases because all mounted fs are no setuid, which also means no setcap, so the current implementation doesnt work. This also add setpriv to the dependencies, which is already staged on core!=18, but still putting it there so that if it gets removed we have it

## Resolved issues

N/A

## Documentation

Decisions documented in comments

## Tests

Tested manually on all bases and coreXX using image-garden, to replicate testing use the snaps built in this PR here:
- https://github.com/canonical/checkbox/actions/runs/21290922107
- https://github.com/canonical/checkbox/actions/runs/21290926322

### DIFF
--- a/checkbox-core-snap/series18/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series18/snap/snapcraft.yaml
@@ -198,6 +198,7 @@ parts:
       - zlib1g-dev
       - build-essential
     stage-packages:
+      - setpriv # used to downgrade AmbientCapabilities of systemd-based runner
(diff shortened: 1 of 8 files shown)


fernando79513 (changes requested):
It's Friday afternoon, so I hope some of the comments make sense here

fernando79513 (approved):
This is much clear now, thanks a lot for the refactoring.
LGTM +1!

PR #2316, opened 2026-02-02
Translator from pxu to yaml (New)

Hook25:
## Description

With this PR we introduce a new migration tool to translate PXUs to YAML. This already translates the unit with the new schema, which will be formalized (and bugfixed probably) in a followup PR. 

Support for comments is best effort. All inline comments should be supported. Header comments are supported. Comments on the top level (unindented commets) are not supported. Untranslatable comments yield a warning.

Jinja support is best effort. Jinja blocks the translation of fields, all fields that contain jinja template are translated as is. Given that jinja can technically modify the semantics (keys) of pxus, the translator, even dumping fields as they are, may break units. All jinja fields yield a warning.

The following fields are unchanged, lines with `.` are reduced to `""` as that is the semantic of them:
- description
- description
- summary
- purpose
- steps
- prompt
- hidden-reason
- verification
- template-summary

The following fields are now objects, no longer raw json (although json is an option, yaml > json)
 - data
 - siblings

The following fields are now a list:
 - environ
 - flags
 - after
 - salvages
 - depends
 - before
 - setup_include
 - include
 - bootstrap_include
 - nested_part
 - exclude
 - mandatory_include
 - requires
 - template-imports
 - imports
 - template-filter
 - certification_status_overrides
 - options

Inline overrides (which are effectively annotations in inclusions) are translated as such:

pxu:
```
include:
    job-baz certification-status=blocker
```

yaml:
```yaml
include:
  - job-baz:
      certification-status: blocker
```

## Resolved issues

Fixes: CHECKBOX-1393

## Documentation

Documentation of the new schema is not part of this PR. All decisions takes are documented in line in the code.

## Tests

Translator is tested

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
+++ b/checkbox-ng/checkbox_ng/launcher/checkbox_cli.py
@@ -47,6 +47,8 @@ from checkbox_ng.launcher.merge_submissions import MergeSubmissions
 from checkbox_ng.launcher.controller import RemoteController
 from checkbox_ng.launcher.agent import RemoteAgent
 
+from checkbox_ng.launcher.translator import Translator
+
 _ = gettext.gettext
 
 _logger = logging.getLogger("checkbox-cli")
@@ -109,6 +111,7 @@ def main():
         "tp-export": TestPlanExport,
         "run-agent": RemoteAgent,
         "control": RemoteController,
+        "translate": Translator,
     }
     deprecated_commands = {
         "slave": "run-agent",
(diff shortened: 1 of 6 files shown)


fernando79513 (changes requested):
Great job Max!!
I can't wait to start using yaml instead of the horrible PXUs.
I've added a few comments, but nothing big.
I tried to translate all the units in checkbox and I only encountered some comments and Jinja errors.

fernando79513 (changes requested):

fernando79513 (approved):
LGTM +1!

PR #2317, opened 2026-02-04
Detect dummy/virtual sinks with checkbox-support-pipewire-utils (New)

tomli380576:
## Description

This PR adds a new subcommand to `checkbox-support-pipewire-utils` that checks whether the default audio/video device is a real device. 

```
usage: Pipewire validator default_device_is_real [-h] -d {audio-source,audio-sink,video-source}

options:
  -d, --device {audio-source,audio-sink,video-source}
                        Device type, audio source/sink or video source.
```

Example:

```
# Real device

$ checkbox-support-pipewire-utils default_device_is_real -d audio-sink
OK! 'Lunar Lake-M HD Audio Controller Speaker' (id=62) is a real device

$ echo $?
0

# dummy output 

$ checkbox-support-pipewire-utils default_device_is_real -d audio-sink
Default audio sink 'Dummy Output' (id=33) is a virtual device

$ echo $?
250

# virtual node from applications

$ checkbox-support-pipewire-utils default_device_is_real -d audio-sink
Default audio sink 'Easy Effects Sink' (id=91) is a virtual device

$ echo $?
250
```

Running this command before the record/playback step can prevent manual test cases from passing when the sources/sinks aren't real devices like dummy sink or a pipewire node created by another application.

## Resolved issues

The current `audio/alsa_record_playback_internal`, `audio/playback_headphones`, `bluetooth/audio-a2dp`, `bluetooth/audio_record_playback` (and many others) are allowed to pass by default when the output device is a Dummy Sink or a virtual pipewire node as long as pipewire reports the device as a source/sink. This check can prevent these tests from passing when the device is not real. 

## Documentation

The basic idea is we ask `wpctl` for the ID of the current default device, then feed that ID into `pw-dump`. Inside the json returned by pw-dump there's an optional field called `node.virtual`. If it's explicitly set to true, then this node is definitely associated with a virtual/dummy device thus shouldn't be tested. Otherwise the device is likely real (or at least pipewire thinks it's real).

Some caveats:

1. The term "default" is a bit misleading. Technically we are checking if the "currently selected" device is real, but "default" is pipewire's terminology.
2. V4L2 loopback devices are considered real
3. "Dummy Output" devices will appear both as sinks and sources
4. There is no `video-sink`


checkbox-support/checkbox_support/scripts/pipewire_utils.py
```diff
@@ -548,6 +551,108 @@ def compare_wpctl_status(self, status_1: str, status_2: str):
...
+                "'pw-dump {}' did not return a list with >= 1 element, got {}".format(
+                    default_device_id, pw_dump_out
+                )
+            )
+
+        # sometimes pw-dump returns extra elements in pw_dump_json
+        # even if we specify the exact ID
+        real = None  # type: dict[str, t.Any] | None
+        for elem in pw_dump_json:
+            # with suppress(Exception):
```
fernando79513: Is this comment intended?
```suggestion
```

fernando79513 (commented):
Thanks a lot for the PR, and thanks also for splitting that into two different ones. I think it's more manageable this way. 

I added a couple of small comments, but they should be pretty straightforward.

I'm not sure if it makes sense to do it on this PR, but when we introduce the jobs, we should also check if the camera source is detected correctly on devices with a MIPI camera, because they usually handle virtual devices differently.

fernando79513 (approved):
LGTM +1!

PR #2319, opened 2026-02-04
Move boot tests to the end of the iot automated test plan (BugFix)

weizhenwu:
## Description
A cold boot, warm boot, and watchdog test will restart the system and might clean up the issues during the test run. Also, some devices (e.g., USB, cameras, thermal_zone) are loaded dynamically on each boot, so a restart may change device indices.

For these reasons, we would like to keep all tests within the same boot by moving the power-automated and watchdog-automated tests to the end of the test plan.

## Tests
sideload and [list-bootstrapped](https://pastebin.canonical.com/p/tzfTsgRmsc/) to confirm that coldboot, warmboot, and watchdog will be in the very end of the test plan.


Hook25 (approved):
I would suggest, if you want a test to run before another, to use `before`, else you are always praying the order doesn't get changed by the dependencies

PR #2322, opened 2026-02-06
gnome monitor doesn't return the *active* resolution (BugFix)

p-gentili:
## Description

Due to a simple regression in this module, the function `get_current_resolutions` is returning the last available resolution instead of the currently active.

## Resolved issues

Part of https://warthogs.atlassian.net/browse/RTW-577

## Documentation

N/A

## Tests

Updated the test cases.


Hook25 (approved):
Any chance that no mode is marked as current?

PR #2325, opened 2026-02-09
Support extra envvars per snap (New)

Hook25:
## Description

Some libraries are not the default locations and, without ldconfig support, which we don't have in snaps, we don't have an easy way to point to them. 
The old approach was to add them to the wrapper common, but this leads to:
- code duplication, all frontends have a copy of this file
- forced inheritance, Checkbox itself and all jobs of all providers see this variables and are influenced by them
- broken w/ the new interface. Now that custom frontends no longer have an independent agent,  this mechanism to define these variables doesn't work anymore. 

This introduces a new mechanism, in the form of a file in the root of the snap with a very simple sintax and comments support. This file is supposed to be created during snap creation and is in the form of:
```
# comments
VARIABLE_NAME+=value
```

## Resolved issues

Fixes: CHECKBOX-2172

## Documentation

Added a HowTo that expains how to use the new packaging feature

## Tests

Tested on a rebuilt version of Checkbox24. Use this command to run the test
```
snap download --edge checkbox24
unsquashfs checkbox24*.snap
cd squashfs-root
snap try --devmode .
checkbox24.checkbox run fwts
# ^^^ this fails
# lets patch it
```
Add this file under squashfs-root
```
#fwts is in /lib/fwts instead of /lib
LD_LIBRARY_PATH+=/lib/fwts
# lapack and blas are pulled by opencv and install libraries in a directory
LD_LIBRARY_PATH+=/usr/lib/x86_64-linux-gnu/lapack
LD_LIBRARY_PATH+=/usr/lib/x86_64-linux-gnu/blas
GI_TYPELIB_PATH+=/usr/lib/girepository-1.0
GI_TYPELIB_PATH+=/usr/lib/x86_64-linux-gnu/girepository-1.0
```
Now patch the snap and run the test(assuming `../checkbox` points to checkbox repo at this branch
```
cp -rT ../checkbox/checkbox-ng/checkbox_ng/ `find . -name "checkbox_ng"`; cp -rT ../checkbox/checkbox-ng/plainbox/ `find . -name "plainbox"`; checkbox24.checkbox run fwts
```

### DIFF
--- a/checkbox-ng/plainbox/impl/secure/providers/test_v1.py
+++ b/checkbox-ng/plainbox/impl/secure/providers/test_v1.py
@@ -1104,6 +1104,7 @@ class Provider1Tests(TestCase):
                 },
             ),
         ]
+        # Note the order, custom frontend variables have precedence
         self.assertEqual(
             Provider1.extra_snap_environment.func(provider),
             {
--- a/checkbox-ng/plainbox/impl/secure/providers/test_v1.py
+++ b/checkbox-ng/plainbox/impl/secure/providers/test_v1.py
@@ -1062,7 +1062,7 @@ class Provider1Tests(TestCase):
             path = Path("/frontend_root")
             extra_env = Provider1._parse_extra_environment_file(path)
 
-        self.assertEqual(list(extra_env), ["LD_LIBRARY_PATH", "PATH"])
+        self.assertEqual(set(extra_env), {"LD_LIBRARY_PATH", "PATH"})
         self.assertEqual(
             extra_env["PATH"], ["/frontend_root/extra/path/location"]
         )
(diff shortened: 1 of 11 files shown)


checkbox-ng/plainbox/impl/secure/providers/test_v1.py
```diff
@@ -1039,6 +1042,84 @@ def test_custom_frontend_root_value_error(self):
         with self.assertRaises(ValueError):
             self.provider.custom_frontend_root()
 
+    def test__parse_extra_environment_file(self):
```
fernando79513: Check this test or the next one, because we didn't catch the error with the path.

fernando79513 (changes requested):
Good job here. I just have a couple of questions/suggestions, but it looks good in general (appart from the path bug)

checkbox-ng/plainbox/impl/secure/providers/test_v1.py
Hook25: It didn't catch the issue because of bad assumptions, on  one side I assumed the path is the root, while on the other it was the file name of the extra env path. I've fixed the name and the construction

fernando79513 (approved):
LGTM +1!
Good job, thanks!
👍

PR #2328, opened 2026-02-10
Build static nsenter for core16 (bugfix)

Hook25:
## Description

We actually do use it if the test is launched as root. Uh oh

## Resolved issues

Fixes: CHECKBOX-2174

## Documentation

Removed misleading comment

## Tests

```
image-garden allocate ubuntu-core-16.x86_64
# download and scp the snap from the CI
# run test plan with agent/local
```

Build: https://github.com/canonical/checkbox/actions/runs/21832537172/job/62994651953

Note: failing runs are not due to this patch, its GH being weird yd. Those snaps aren't touched by this pr

### DIFF
--- a/checkbox-core-snap/series16/snap/snapcraft.yaml
+++ b/checkbox-core-snap/series16/snap/snapcraft.yaml
@@ -101,6 +101,34 @@ parts:
     source: https://github.com/Hook25/plz-run.git
     source-type: git
 ################################################################################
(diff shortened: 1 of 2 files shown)


Hook25:
Tested running smoke on image garden core16

fernando79513 (approved):
Damn, you are right, I missed it during review. 
We are only using the dangerous_nsenter when the user is not "root"

LGTM +1

PR #2332, opened 2026-02-12
Fix fstrim test to not run when fstrim is not supported (BugFix)

### LINKED ISSUE #216: LP1969663: fstrim fails when not supported
This issue was migrated from https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1969663

## Summary
| Status 	| Created on 	| Heat 	| Importance 	| Security related 	|
|--------	|------------	|------	|------------	|------------------	|
| New	| 2022-04-20 18:42:34	| 6	| Undecided	| False  	|

## Description
Seen during a Focal SRU run:
https://certification.canonical.com/hardware/202110-29522/submission/259880/test/90591/result/25139129/

INFO:root:Disk device is /dev/sdb
INFO:root:Testing partition /dev/sdb1
INFO:root:/dev/sdb1 is already mounted at /data-sdb
INFO:root:mount point is /data-sdb
fstrim: /data-sdb: the discard operation is not supported
INFO:root:

If the operation is not supported, should the test just pass with a note that it's passing because the operation is not supported?

## Attachments
No attachments

Tags:
['servcert-256']

rodwsmith:
Title: Fixed fstrim resource job to not run the disk/fstrim_{name} test when it's not supported (BugFix)

## Description

The disk/fstrim_{name} tests were running even on disks (mainly spinning disks) that do not support TRIM, leading to failed tests when those tests should be skipped instead. This submission changes that behavior, mainly by altering a resource job:

* There actually was a resource job to detect if a disk was a spinning disk, which would not support fstrim; however, it was broken because it was looking at a file in the `sys` directory that no longer existed. Also, a test for whether a disk is a spinning disk isn't quite adequate, because some SSDs don't support fstrim (if they're in a RAID array, for instance).
* This pre-existing resource job is renamed and re-worked so that it works to more directly and correctly detect TRIM support, not simply whether a disk is a spinning disk. The basic mechanism is the same; it looks for TRIM support in the `sys` directory tree, but in different files. Re-naming the job required modifying the user of the resource job (see below).
* These changes required changes to the unit test.
* The user of this resource job, in `providers/base/units/disk/jobs.pxu`, is updated to use the new name and syntax. Also, it had been broken because of a lack of parentheses around the `requires` test; this has been fixed.

## Resolved issues

Jira: https://warthogs.atlassian.net/browse/SERVCERT-256
GitHub: https://github.com/canonical/checkbox/issues/216
Launchpad: https://bugs.launchpad.net/plainbox-provider-checkbox/+bug/1969663 (defunct; including for completeness)

## Documentation

As noted, changes to the unit tests were required, and are included in this PR.

## Tests

I've tested on a number of systems, and they all now perform as expected, skipping disks that do not support fstrim. For instance:

* https://certification.canonical.com/hardware/201409-15506/submission/472493/ -- A test of teela, one of my personal systems with an SSD and a spinning disk;
* https://certification.canonical.com/hardware/202409-35406/submission/472498/ -- A test of amontons, a server in TOR-3 with one NVMe and four spinning disks

In all cases, disks that lack fstrim support are now skipped, while those that do support it are tested.

### DIFF
--- a/providers/base/units/disk/jobs.pxu
+++ b/providers/base/units/disk/jobs.pxu
@@ -103,7 +103,7 @@ template-id: disk/fstrim_name
 estimated_duration: 1.0
 user: root
 requires:
- block_device.rotation == 'no' and block_device.name == '{name}'
+ (block_device.fstrim == 'True' and block_device.name == '{name}')
 _summary: Filesystem TRIM check for {product_slug}
 _purpose: Take the path of the storage device and test its TRIM capabilities
 command: fstrim_test.py --device-file {name}
--- a/providers/resource/bin/block_device_resource.py
+++ b/providers/resource/bin/block_device_resource.py
@@ -57,19 +57,20 @@ def usb_support(name, version):
     return "unsupported"
 
 
-def device_rotation(name):
+def device_fstrim(name):
     """
-    Check the device queue/rotational parameter to determine if it's a spinning
-    device or a non-spinning device, which indicates it's an SSD.
+    Check the device hw maximum discard bytes value. If it's 0, fstrim is
+    not supported; if it's anything else, fstrim should work. Assume a
+    missing file is equivalent to a 0 value (no support).
     """
-    path = "/sys/block/{0}/device/block/{0}/queue/rotational".format(name)
+    path = "/sys/block/{}/queue/discard_max_bytes".format(name)
     if not os.path.exists(path):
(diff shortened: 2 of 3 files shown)


pedro-avalos (approved):
Changes look good, tests are passing, and the submissions look good to me. Thanks for these changes!

PR #2344, opened 2026-02-18
Drop only caps that were given to the process (bugfix)

Hook25:
## Description

This should be (almost) the same if the pam login doens't give the process more ambient caps than usual (which could happen). This was done because on core<20, `-all` is not supported

## Resolved issues

Fixes: 

## Documentation

N/A

## Tests

Tested on problematic machine via patching snap and running canary

Tested on all versions of ubuntu core using image garden + spread running canary


fernando79513 (approved):
LGTM +1!

PR #2352, opened 2026-02-23
preserve-cwd only when preserve-cwd is provided (bugfix)

Hook25:
## Description

Snaps services cwd is calculated by snapd and depends on the user that starts the service among other factors. Using the service cwd as the job cwd present one challenge, if the job is a user job it may not have access to that agent cwd (that may be in /root/snap/...). Additionally I believe that preserving cwd always in snaps was something done in the past because `/tmp` is private for snaps and it was a quick workaround, which also drops the leftovers check.

A solution that fixes both issues is using `/var/tmp` isntead of `/tmp`. We already use this quite a lot (for the nest), so until it becomes private as well, we can rely on it being globally accessible. 

## Resolved issues

Fixes: CHECKBOX-2181

## Documentation

N/A
I don't think this needs additional documentation as this restores the "normal" behaviour one expects from checkbox, it was weird we wouldn't create a temp cwd for snaps making all jobs effectively preserve-cwd

## Tests

Tested on image-garden.spread running the docker test (which, as it calls a snap it may be impacted)

### DIFF
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -426,14 +426,17 @@ class UnifiedRunner(IJobRunner):
         :returns:
             Pathname of the new directory
         """
-        if "preserve-cwd" in job.get_flag_set() or os.getenv("SNAP"):
+        if "preserve-cwd" in job.get_flag_set():
             yield os.getcwd()
             return
         # Create a nest for all the private executables needed for execution
         prefix = "cwd-"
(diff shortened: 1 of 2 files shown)


fernando79513 (approved):
Thanks for the comment in the description, it's very explanatory and will help us in case we have to come back to this in the future.
LGTM +1!

fernando79513 (approved):
Thanks for the comment in the description, it's very explanatory and will help us in case we have to come back to this in the future.
LGTM +1!

PR #2359, opened 2026-03-02
NPU test plan for known failures (New)

vandah:
## Description

This PR is adding a test plan that includes the known failures of NPU tests. With this PR there are three test plans, one without known failures, one with only known failures and one with both passing tests and known failures.

We want to run the known failures as part of the integration tests of the driver. Initially I thought we'd able to easily add them in a launcher but I found out that the tests still need to be a part of a test plan to do it.

## Tests

A run of the new test plan including both the regular tests and the known failures on a Meteor Lake laptop: [submission_2026-03-02T17.38.28.230364.zip](https://github.com/user-attachments/files/25691144/submission_2026-03-02T17.38.28.230364.zip)

### DIFF
--- a/providers/base/units/npu/test-plan.pxu
+++ b/providers/base/units/npu/test-plan.pxu
@@ -6,3 +6,18 @@ include:
     umd/.*
 bootstrap_include:
     com.canonical.certification::npu_gtest_resource
+
+id: npu-only-known-failures
+unit: test plan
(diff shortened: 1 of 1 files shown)


fernando79513 (dismissed):
LGTM +1!

fernando79513 (approved):
LGTM +1!

PR #2376, opened 2026-03-10
add manifest for ce-oem-gpio/loopback test (Bugfix)

stanley31huang:
## Description
add manifest for ce-oem-gpio/loopback test


seankingyang (approved):
LGTM~

PR #2386, opened 2026-03-13
Remove all (un)translatable warnings (bugfix)

Hook25:
## Description

This is the first step in hard deprecating translation, a feature that is overengineered and not used anymore. If we need them in the future, I hope we can implement them better without polluting the whole codebase + all the tests

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
LGTM+1!

Thanks for taking care of this. I don't think we are using it and it's crazy that just excluding the validation leads to 427 lines removed... Gj!

PR #2392, opened 2026-03-16
Metabox QOL (bugfix)

Hook25:
## Description

This is a bunch of QOL patches for Metabox that come out of innovation time. 
- pkg_resource: very deprecated and not quite what we need
- log to file: quite a recurrent need, see the commit message. In general this always creates a log file and dumps the log there as well, this way one can't loose the errors in the scrollback
- dispose (instead of do-not-dispose), we all have an alias, it should just be default
- dont-reprovision-existing instead of use-existing. Reprovisioning the recycled container is now default and has a name that makes sense.

## Resolved issues

Fix: CHECKBOX-2192

## Documentation

N/A

## Tests

Tested running metabox because metabox doesn't have unit tests sigh

### DIFF
--- a/metabox/metabox/core/lxd_provider.py
+++ b/metabox/metabox/core/lxd_provider.py
@@ -27,10 +27,8 @@ LXD machines are containers that can run metabox scenarios in them.
 
 import json
 import os
-import sys
 import time
 import yaml
-import subprocess
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


metabox/metabox/core/lxd_provider.py
fernando79513: Not used anymore
```suggestion
```

PR #2401, opened 2026-03-18
MANIFEST.in checking CI (infra)

Hook25:
## Description

MANIFEST.in inclusions are a recurrent rake we step on. For this reason, this adds CI that checks if a file that was added to `checkbox-support` or `checkbox-ng` is correctly included or excluded from the manifest.

The reason why we only see this issue in debian builds or snapcraft remote builds is because if `.git` is shipped to the build env, `setuptools-scm` automatically uses it to know what to include. In these two environments `.git` is not present, so we fallback to MANIFEST and we fail to build. 

Additionally this forces us to take into consideration all additions from this lens. This also caught another problem, we weren't shipping the LICENSE for some vendorized dependencies

## Resolved issues

Fixes: CER-3481

## Documentation

N/A

## Tests

This should fail first PR push, rebase on main should pass after [this](https://github.com/canonical/checkbox/pull/2400) lands

### DIFF
--- a/.github/workflows/verify-manifest.yaml
+++ b/.github/workflows/verify-manifest.yaml
@@ -1,0 +1,81 @@
+name: Verify MANIFEST.in is up to date
+permissions:
+  contents: read
+on:
+  push:
+    paths:
+      - checkbox-ng/**
+      - checkbox-support/**
+      - .github/workflows/verify-manifest.yaml
+  pull_request:
+    paths:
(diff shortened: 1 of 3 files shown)


fernando79513 (approved):
I read it and everything looks fine.
Most importantly, the CI failed when it should fail and passed when it should pass.

LGTM +1!

PR #2409, opened 2026-03-20
Add screen-private-ppa test to misc-client-cert-automated (New)

bnbhat:
## Description
Add miscellanea/screen-private-ppa test to  misc-client-cert-automated test plan.
This test checks for any private PPAs in the system and fails if found.

As suggest by @stanley31huang here: https://code.launchpad.net/~carmel-team/carmel/+git/checkbox-provider-carmel/+merge/495859

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/PECA-1400

## Tests

- Run this branch on any device with Private PPA entry in the source.list*

### DIFF
--- a/providers/base/units/miscellanea/test-plan.pxu
+++ b/providers/base/units/miscellanea/test-plan.pxu
@@ -41,6 +41,7 @@ include:
     miscellanea/ubuntu-desktop-recommends          certification-status=blocker
     miscellanea/ubuntu-desktop-minimal-recommends  certification-status=blocker
     miscellanea/grub_file_check                    certification-status=blocker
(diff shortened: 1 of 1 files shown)


kevinyehk (commented):
I think this change will affect Cert’s SRU testing, because some of our tests need to enable a private PPA, such as ESM kernel tests and real-time kernel tests. In addition, some PE projects, including MTK and Nvidia, also require a private PPA for certain tests.

Hook25:
I'm closing this as I believe that this is not desirable. Feel free to reopen this if you have a counter arguement. The only thing that is not in question is that this can't be in the cert client/sru testplans as it will fail in SRU/SUV every time

PR #2410, opened 2026-03-20
Add bios metadata to submission json (new)

pieqq:
## Description

Update device-bios script with a new `bios` subcommand that extracts data from `/sys/class/dmi/id/bios_*` files and returns them as JSON. Sysfs is accessed directly (instead of using dmidecode, for instance) since these files are user-readable (while dmidecode requires root access).

The collector version is bumped to v5.

## Resolved issues

Fix CHECKBOX-2191

## Tests

Tested locally:

```
$ device-info bios
{
    "date": "09/18/2025",
    "release": "5.35",
    "vendor": "American Megatrends International, LLC.",
    "version": "F8d"
}
```

### DIFF
--- a/checkbox-ng/checkbox_ng/support/device_info.py
+++ b/checkbox-ng/checkbox_ng/support/device_info.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
 import argparse
-import glob
 import json
-import os
+from pathlib import Path
 import platform
 import subprocess
+import sys
 
 from checkbox_ng.support.release_info import get_release_info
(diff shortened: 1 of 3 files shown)


Hook25 (dismissed):
LGTM but consider the following, I don't like that the keys /may/ change. It does, on core16 only these are present: (both core16 and uc16)
```
ubuntu@ubuntu-16-04:~$ cat /sys/class/dmi/id/bios_
bios_date     bios_vendor   bios_version  
```

Hook25 (approved):

Issue #1961, opened 2025-06-16
504 Gateway Time-out while uploading submission, but submission uploaded


hanhsuan:
### Bug Description

Currently, we are facing an issue while uploading the submission to C3 randomly. The condition is that Checkbox shows the submission is failed to upload with `504 Gateway Time-out`, but that submission did be uploaded.

![Image](https://github.com/user-attachments/assets/93239ffe-6ea6-4805-804e-72d12bd10c82)

This [example](https://certification.canonical.com/hardware/202501-36241/) shows that uploaded multiple times due to this issue.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Finish the test
2. Upload the submission

### Expected Result

If checkbox shows failed, then it should not upload the submission.

### Actual Result

The submission has been uploaded to C3, but checkbox shows upload failed.

### Environment

- OS: 24.04
- Checkbox type: debian
- Checkbox version: 4.6.0~dev23~ubuntu24.04.1

### Relevant log output

```shell

```

GabrielChenCC:
I encountered the same problem when submitting on June 13. Please refer to the sample: https://certification.canonical.com/hardware/202505-36737/

https://certification.canonical.com/hardware/202505-36737/submission/432906/
https://certification.canonical.com/hardware/202505-36737/submission/432907/
https://certification.canonical.com/hardware/202505-36737/submission/432908/

Both 432906 and 432907 failed to upload. I initially thought it might be a network problem, but when I successfully submitted the third time and checked the results on C3, I found that all three submissions were successfully submitted.

hanhsuan:
Encounter this [one](https://certification.canonical.com/hardware/202503-36450/submission/434934/) today.

hanhsuan:
Encounter this [one](https://certification.canonical.com/hardware/202501-36239/submission/438466/) today.

Hook25:
Closing this due to the issue being now solved.

PR #2417, opened 2026-03-23
Avoid collecting 80k lines of logs in checkbox collector (bugfix)

Hook25:
## Description

When exporting on devices with really low memory and no swap, Checkbox will be killed by the OOM killer (both local and remote) if there is a lot of logs or the logs weren't purged before the test run. 

## Resolved issues

Fixes: CHECKBOX-2153
May help with: https://github.com/canonical/checkbox/issues/1961

## Documentation

This adds an explanation in the docstring, given that the definition of low memory is arbitrary, I've increased it to 2Gb which is still quite low and also includes all devices with slightly more than 1Gb

## Tests

Create the following yaml somewhere:
```yaml
name: low_mem
description: "Low memory profile"
config:
  limits.cpu: "1"
  limits.memory: 512MB
  limits.memory.swap: "false"
project: default
```
Test via the following
```shell
# Import the profile
lxc profile create low_mem < low_mem.yaml 
# Create a machine with this low memory 
lxc launch ubuntu:noble low-mem-noble --profile low_mem --profile default
apt update && apt upgrade && apt install python3 python3-venv python3-pip
# Install checkbox + all
git clone https://github.com/canonical/checkbox.git
python3 -m venv venv
. venv/bin/activate
cd checkbox/checkbox-ng
pip install -e .
cd ../checkbox-support
pip install -e .
cd ../providers
for f in */manage.py; do ./$f develop; done
cd ../metabox/metabox/metabox-provider
./manage.py develop
# Now lets create a lot of data in the journal to collect
for i in {0..80000..1}; do echo some | systemd-cat; done
```
Launch checkbox-cli. Before this patch, it will be killed while creating the tar, after this patch, it will terminate without a problem

### DIFF
--- a/checkbox-ng/plainbox/impl/__init__.py
+++ b/checkbox-ng/plainbox/impl/__init__.py
@@ -210,3 +210,17 @@ def get_plainbox_dir():
     Return the root directory of the plainbox package.
     """
     return os.path.dirname(getabsfile(plainbox))
+
+
+def low_memory_device():
+    """
+    Returns if the system we are runing on has very low RAM installed
+
+    This is used to disable or tune down some subsystems to avoid filling
+    the RAM with Checkbox itself and crashing the system.
+    """
+    mem_bytes = os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES")
+    mem_mib = mem_bytes / (1024.0**2)
+    if mem_mib < 1024 * 2:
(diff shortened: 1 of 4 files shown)


pieqq (dismissed):
One little typo and a suggestion for the docstring, otherwise it looks like a good idea!

pieqq (approved):

PR #2425, opened 2026-03-25
add a manifest for cpufreq scaling test (Bugfix)

stanley31huang:
## Description
add a manifest for cpufreq scaling test, as this feature is not supported on new i.MX9 SoC

#### test result when cpu policy is available
https://certification.canonical.com/hardware/202605-38721/submission/496556/
#### test result when cpu policy is unavailable
```
ubuntu@juju-8ab9e7-7:~/carmel_launcher$ checkbox-cli control 10.102.181.187
Connecting to 10.102.181.187:18871. Timeout: 600s
Agent is using sideloaded providers
=========[ Bootstrap com.canonical.contrib::cpufreq_policy_list (1/1) ]=========
Outcome: job passed
There were no tests to select from!
```

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
@@ -493,6 +493,14 @@ class CPUScalingHandler:
 
 class CPUScalingTest:
     """A class for CPU scaling test operations."""
+    test_description = ""
+    _registry = {}
+
+    def __init_subclass__(cls, *, key=None, **kwargs):
+        """Register subclasses in the registry."""
+        super().__init_subclass__(**kwargs)
+        if key:
+            CPUScalingTest._registry[key] = cls
 
     def __init__(self, policy=0):
         """
@@ -503,6 +511,29 @@ class CPUScalingTest:
         """
         self.policy = policy
         self.handler = CPUScalingHandler(policy=self.policy)
+
+    @classmethod
+    def create(cls, governor, policy=0):
+        """
+        Create a test instance by governor name.
+
+        Args:
+            governor (str): The name of the governor test to create.
+            policy (int): The CPU policy number to be used (default is 0).
+
+        Returns:
+            CPUScalingTest: An instance of the appropriate test class.
+
+        Raises:
+            ValueError: If the governor is not supported.
+        """
+        test_class = cls._registry.get(governor.lower())
+        if test_class is None:
+            raise ValueError(
+                f"Governor '{governor}' not supported. "
+                f"Available: {', '.join(sorted(cls._registry.keys()))}"
+            )
+        return test_class(policy=policy)
 
     def print_policy_info(self):
         """
@@ -676,6 +707,29 @@ class CPUScalingTest:
                 sys.exit("Governor '{}' not supported".format(governor))
         return success
 
+    def test_governor(self) -> bool:
+        """
+        Run the CPU Scaling Governor Test.
+
+        This function is a placeholder for running the CPU scaling governor
+        test. It should be implemented in the subclasses for specific
+        governors.
+
+        Returns:
+            bool: True if the test passes, False otherwise.
+        """
+        raise NotImplementedError("This method should be implemented in subclass.")
+
+
+class UserspaceCPUScalingTest(CPUScalingTest, key="userspace"):
+    """
+    CPU scaling test operations specific to the userspace governor.
+    """
+    description = """
+    This job sets the governor to "userspace" and verifies the frequency 
+    when setting it to maximum and minimum.
+    """
+
     def test_userspace(self) -> bool:
         """
         Run the Userspace Governor Test.
@@ -696,7 +750,17 @@ class CPUScalingTest:
             self.handler.min_freq,
         )
 
-    def test_performance(self) -> bool:
+
+class PerformanceCPUScalingTest(CPUScalingTest, key="performance"):
+    """
+    CPU scaling test operations specific to the performance and powersave
+    governors.
+    """
+    description = """
+    This job sets the governor to "performance" and verifies whether 
+    the frequency is maximum.
+    """
+    def test_governor(self) -> bool:
         """
         Run the Performance Governor Test.
 
@@ -710,7 +774,17 @@ class CPUScalingTest:
         governor = "performance"
         return self.test_frequency_influence(governor)
 
-    def test_powersave(self) -> bool:
+
+class PowersaveCPUScalingTest(CPUScalingTest, key="powersave"):
+    """
+    CPU scaling test operations specific to the powersave governor.
+    """
+    description = """
+    This job sets the governor to "powersave" and verifies whether 
+    the frequency is minimum.
+    """
+
+    def test_governor(self) -> bool:
         """
         Run the Powersave Governor Test.
 
@@ -724,7 +798,17 @@ class CPUScalingTest:
         governor = "powersave"
         return self.test_frequency_influence(governor)
 
-    def test_ondemand(self) -> bool:
+
+class OndemandCPUScalingTest(CPUScalingTest, key="ondemand"):
+    """
+    CPU scaling test operations specific to the ondemand governor.
+    """
+    description = """
+    This job sets the governor to "ondemand" and verifies whether the 
+    frequency will be maximum after stressing CPUs and settling down after
+    sleeping for a few seconds.
+    """
+    def test_governor(self) -> bool:
         """
         Run the Ondemand Governor Test.
 
@@ -738,7 +822,18 @@ class CPUScalingTest:
         governor = "ondemand"
         return self.test_frequency_influence(governor)
 
-    def test_conservative(self) -> bool:
+
+class ConservativeCPUScalingTest(CPUScalingTest, key="conservative"):
+    """
+    CPU scaling test operations specific to the conservative governor.
+    """
+    description = """
+    This job sets the governor to "conservative" and verifies whether the 
+    frequency will be maximum after stressing CPUs and settling down after
+    sleeping for a few seconds.
+    """
+
+    def test_governor(self) -> bool:
         """
         Run the Conservative Governor Test.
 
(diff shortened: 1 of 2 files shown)


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/cpu/jobs.pxu
```diff
@@ -14,6 +14,8 @@ plugin: shell
 user: root
 category_id: com.canonical.plainbox::cpu
 flags: also-after-suspend
+requires: manifest.has_cpufreq_scaling_supported == 'True'
```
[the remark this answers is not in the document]
stanley31huang: thanks for the suggestion, but we decied to remove this manifest.

stanley31huang:
this PR will change the CPU governor test as following scenario
1. In the resource job, it will print out all CPU policy with avaiable scaling governor
2. the corresponding CPU governor tests will be run based on governor listed in available scaling governor.
3. Remove cpu driver detect test as it's not supported for all platform.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -169,7 +169,7 @@ def __init__(self, policy=0):
         self.max_freq = self.get_max_frequency()
         self.governors = self.get_supported_governors()
         self.original_governor = self.get_governor()
-        self.affected_cpus = self.get_affected_cpus()
+        self.affected_cpus = self.get_affected_cpus().split()
```
tomli380576, line 214: optional: can we put the split() call in `get_affected_cpus`? or update the return type to be `str`

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -692,7 +719,34 @@ def test_frequency_influence(self, governor, target_freq=None) -> bool:
...
+        This function is a placeholder for running the CPU scaling governor
+        test. It should be implemented in the subclasses for specific
+        governors.
+
+        Returns:
+            bool: True if the test passes, False otherwise.
+        """
+        raise NotImplementedError(
+            "This method should be implemented in subclass."
+        )
```
tomli380576: I think the more idiomatic way to do this is with an abstract base class:

```diff
-class CPUScalingTest
+class CPUScalingTest(abc.ABC)
```

then at the method definition, decorate it:
```py
@abc.abstractmethod
def test_governor(self) -> bool:
    # no need to write anything in the method body
    pass
```

this can also prevent the base class from being constructed altogether.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -485,6 +500,19 @@ def set_frequency(self, frequency) -> bool:
...
+
+    @classmethod
+    def register(cls, key):
+        """Register test classes by governor name."""
+
+        def decorator(test_cls):
+            cls._registry[key] = test_cls
+            return test_cls
+
+        return decorator
```
tomli380576, line 523: optional: i think it's less convoluted to just use a dict in main and put the class names in there:

```py
# inside main

TEST_CLASSES = {
    "powersave": PowersaveCPUScalingTest
    "schedutil": SchedutilCPUScalingTest
    # ...other ones
}

if args.governor not in TEST_CLASSES:
    raise SystemExit("Unavailable governor: {}".format(args.governor))

# governor available, do the rest of main
for governor, test_class in TEST_CLASSES.items():
    # do stuff

```

this can also eliminate the `.create()` method and allow `main()` to use normal class constructors

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -485,6 +500,19 @@ def set_frequency(self, frequency) -> bool:
 class CPUScalingTest:
     """A class for CPU scaling test operations."""
 
+    test_description = ""
```
tomli380576: in the child classes this was changed to `description`, did you mean to use 2 different names?

If not, a more robust way to enforce the existence of `test_description` is to use an abstract property

```py
# requires the parent class to be abstract

class CPUScalingTest(abc.ABC):
    @property
    @abc.abstractmethod
    def test_description() -> str:
        pass

class UserspaceCPUScalingTest(CPUScalingTest):
    @property
    def test_description() -> str:
        return "some test description"

# access it like a normal property

t = UserspaceCPUScalingTest()
print(t.test_description)
```

If the child class doesn't implement `test_description`, python will raise this:

```
TypeError: Can't instantiate abstract class UserspaceCPUScalingTest with abstract methods test_description
```

because not implementing the entire ABC's interface makes the child class implicitly abstract, thus cannot be constructed.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -254,13 +255,28 @@ def print_policies_list(self) -> bool:
...
+
+        available_governor_tests = CPUScalingTest._registry.keys()
+
         for policy in self.cpu_policies:
-            print("policy: {}".format(policy))
-            print("scaling_driver: {}".format(self.get_scaling_driver(policy)))
-            print("cpb: {}".format(self.get_cpb(policy)))
-            print()
-        return True
+            self.policy = policy
```
tomli380576: is there a reason why we need to keep overwriting `self.policy`?

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -495,6 +523,30 @@ def __init__(self, policy=0):
...
+            ValueError: If the governor is not supported.
+        """
+        test_class = cls._registry.get(governor.lower())
+        if test_class is None:
+            raise ValueError(
+                "Governor '{}' not supported. Available: {}".format(
+                    governor, ", ".join(sorted(cls._registry.keys()))
+                )
+            )
+        return test_class(policy=policy)
```
tomli380576: see the comment on 514

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -835,18 +947,24 @@ def main():
         handler.print_policies_list()
         sys.exit(0)
 
-    test = CPUScalingTest(policy=args.policy)
-    if args.driver_detect:
-        sys.exit(0) if test.test_driver_detect() else sys.exit(1)
+    if not args.governor:
+        parser.print_help()
+        sys.exit(1)
```
tomli380576, line 986: let's use `required=True` to enforce this:

```py
parser.add_argument(
      "--governor",
      dest="governor",
      required=True,
      help="Run Specific Governor Test",
  )
```

you can also eliminate the test for governor existence here with `choices=("powersave", "performance", "schedutil", "userspace")`

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -835,18 +947,24 @@ def main():
...
 
     try:
+        test = CPUScalingTest.create(args.governor, policy=int(args.policy))
+        logging.info(test.description)
         test.print_policy_info()
         if args.governor not in handler.governors:
-            probe_governor_module(args.governor)
-        if not getattr(test, "test_{}".format(args.governor))():
+            logging.error(
+                "Governor '%s' is not supported by CPU policy%s",
```
tomli380576, line 993: use `.format()`

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
```diff
@@ -835,18 +947,24 @@ def main():
...
+                args.governor,
+                args.policy,
+            )
+        if not test.test_governor():
             sys.exit(1)
-    except AttributeError:
-        logging.error("Given governor is not supported")
+    except ValueError as err:
+        logging.error(str(err))
```
tomli380576: optional: depending on if you want to see the literal `ValueError` string, you could use `repr`

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
tomli380576: oh wait i didn't see the `--policy-resource` one, ignore this

tomli380576 (changes requested):
the comments are mostly refactoring suggestions, lmk if i misunderstood anything

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
stanley31huang: I think it's okay to use logging likes this 
https://docs.python.org/3/library/logging.html#logging.info

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
stanley31huang: fixed

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
stanley31huang: fixed

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
stanley31huang: fixed

tomli380576 (approved):
lgtm we will worry about refactoring later

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/cpufreq_governors.py
stanley31huang: let's do the refactor in another PR.active-high

PR #2429, opened 2026-03-26
Implement context manager to handle codec test file download an deletion (BugFix)

rickwu666666:
Implement context manager to handle codec test file download and deletion

`VIDEO_CODEC_TESTING_DATA` is no longer mandatory. If not provided, files are stored in a `video` directory within the user's home folder. If provided, the files in that path are used as test data, and no new files are downloaded during testing if the valid file under the path.

## Tests
Sideload result
https://pastebin.canonical.com/p/Jz5mtwvwCy/


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/gst_transform_resize.py
```diff
@@ -229,24 +222,30 @@ def build_pipeline(self) -> str:
...
-        golden_reference_file=p.psnr_reference_file,
-        artifact_file=p.artifact_file,
-    )
-    delete_file(file_path=p.artifact_file)
+    with manage_test_file_by_params(
+        args.width_from, args.height_from, args.framerate, args.encoder_plugin
+    ):
+        with manage_test_file_by_params(
+            args.width_to, args.height_to, args.framerate, args.encoder_plugin
+        ):
```
baconYao, line 230: Why do we handle artifact_file here? It should be created, verified and delete via the following process instead of downloading from CodeCraft repo, right?

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/tox.ini
```diff
@@ -40,6 +40,7 @@ deps =
     pyparsing == 2.0.3
     distro == 1.0.1
     PyYAML == 3.11
+    opencv-python
```
baconYao, line 43: Why we need this package in our snap?

baconYao (commented):
It's a good way to tackle test files, but I have some questions.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/tox.ini
rickwu666666: That is for pass tox test and not pack into checkbox-ce-oem snap.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/gst_transform_resize.py
rickwu666666: I thought that was mandatory for testing resize.
The first context manager will download the original video size, which is with "width_from" and "height_from". And it is for the build_pipeline funciton used.
```
def build_pipeline(self) -> str:
        """
        Build the GStreamer commands based on the platform and codec.
        Returns:
            str: A GStreamer command based on the platform and
            codec.
        """
        pipeline = (
            "{} filesrc location={} ! decodebin ! v4l2convert ! "
            "video/x-raw,width={},height={} ! {} ! {} ! mp4mux ! filesink"
            " location={}"
        ).format(
            GST_LAUNCH_BIN,
            self._golden_sample,
            self._width_to,
            self._height_to,
            self._codec,
            self._codec_parser_map.get(self._codec),
            self.artifact_file,
        )
        return pipeline
```
And the second context manger will download the expect target after resized, which is with "width_to" and "height_to". And this is for psnr reference file used.
```
    def psnr_reference_file(self) -> str:
        """
        A golden reference which has been transformed in advance. It's used to
        be the compared reference file for PSNR.
        """
        golden_reference = get_test_file_path_by_params(
            self._width_to, self._height_to, self._framerate, self._codec
        )
        if not os.path.exists(golden_reference):
            raise SystemExit(
                "Error: Golden PSNR reference '{}' doesn't exist".format(
                    golden_reference
                )
            )

        return golden_reference
```

baconYao (approved):
After checking with @rickwu666666, there's no concern from me. LGTM

baconYao (approved):
LGTM

rickwu666666:
@copilot resolve the merge conflicts in this pull request

baconYao (approved):
LGTM

PR #2430, opened 2026-03-26
Remove all jinja conditionals from requires (bugfix)

Hook25:
## Description

This PR does 2 things:
1. Removes all jinja conditionals from the requires fields
2. Documents how to do it

Minor: also documents the new envvars we have introduced for this (and more!) purpose

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-1391

## Documentation

This also has the documentation

## Tests

This is basic boolean logic but given I too got super cofused while trying to encode it I wrote this script to double check my reasoning!

```python
from itertools import product


def jinja_version(a, b, c):
    if a:
        if b:
            return c
        return False
    return c


def non_jinja_version(a, b, c):
    return (not a or b) and c


inputs = product([True, False], [True, False], [True, False])

for a, b, c in inputs:
    print(a, b, c)
    assert jinja_version(a, b, c) == non_jinja_version(a, b, c)
```

### DIFF
--- a/docs/how-to/index.rst
+++ b/docs/how-to/index.rst
@@ -16,3 +16,4 @@ These how-to guides cover key operations and processes in Checkbox.
    using-groups
    ancient-python
    custom-frontend-envvars
+   migrate-away-jinja2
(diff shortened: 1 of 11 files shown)


pieqq (changes requested):
Left a few suggestions below.

pieqq (approved):

PR #2433, opened 2026-03-30
pre-commit: initialize (Infra)

Saviq:
## Description

Rather than rely on people remembering to run linters/formatters, and maintain a separate config in Actions, use https://pre-commit.com hooks to standardize.

`pre-commit install` locally to validate every commit.

If you configure https://github.com/apps/pre-commit-ci-lite to have access to this repo, it will even autofix PRs.

`pre-commit run --all-files` is quite unhappy at the moment, but that's out of scope for this PR.

## Resolved issues

N/A

## Documentation

N/A

## Tests

N/A


.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c  # frozen: v6.0.0
+  hooks:
+  - id: trailing-whitespace
+  - id: end-of-file-fixer
+  - id: check-yaml
+    args: [--allow-multiple-documents]
+  - id: check-added-large-files
+- repo: https://github.com/psf/black
```
tomli380576, line 11: there's a mirror that's _allegedly_ 2x as fast: https://github.com/psf/black-pre-commit-mirror 
```yaml
  - repo: https://github.com/psf/black-pre-commit-mirror
    rev: # version num here
    hooks:
      - id: black
```

Hook25:
Sorry for coming to this late. We don't want pre-commit hooks in Checkbox at all, we've discussed this internally multiple times but in our opinion don't fit with the way we work. The decision is not final as many others, but for now I'd say no. Thanks for working on this any way

PR #2436, opened 2026-03-30
Force using fork multiprocessing (bugfix)

Hook25:
## Description

On Python3.14, they've changed the default handler for Process, now defaulting to forkserver. This is good for a multitude of reasons but breaks our decorator because forkserver needs all arguments (including the target) to be picklable. This of course is not the case for decorators. It is not quite clear to me what is the principled solution to this, but one solution is to force the "fork" usage, which is problematic only in threaded applications, which our tests are for the most part not.

## Resolved issues

Partially fixes: https://warthogs.atlassian.net/browse/CHECKBOX-2201

## Documentation

Adds a comment to explain why we are doing the context thing

## Tests

Tests were already in place, they now pass on python3.14

### DIFF
--- a/checkbox-support/checkbox_support/helpers/timeout.py
+++ b/checkbox-support/checkbox_support/helpers/timeout.py
@@ -30,10 +30,17 @@ import traceback
 import subprocess
 
 from queue import Empty
-from functools import partial
+from functools import partial, wraps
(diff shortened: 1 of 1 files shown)


p-gentili (approved):
TIL

PR #2442, opened 2026-04-01
Simplify audio sink iteration in docking station tests (New)

tomli380576:
(the +6000 lines are the test data not actual code xD)
## Description

The original test case `dock/*-monitor-multi-head-audio-playback` was a little cumbersome because the user has to wait until every single audio sink has been tested before re-testing any of the earlier ones. This PR modifies the flow to add automatic audio sink switching and shows a short menu to allow the user to quickly switch between audio sinks without leaving checkbox.

Also added some type annotations.

## Resolved issues

QoL improvement only.

## Documentation

For automatic sink switching, the basic idea here is to find all the node IDs associated with an "available" audio sink, use `wpctl set-default <id>` to change it, then run the speaker test. Pipewire doesn't seem to provide a convenient way to query all available sinks, so this script iterates through all the sink nodes, finds the parent device, then see if the sink node's `card.profile.device` matches an `EnumRoute` in the device. If there's a match and that EnumRoute is available, then we run the test. 

1 caveat of this script is that if a pair of headphones is plugged into the built-in 3.5mm port of a DUT that has a built-in speaker, the speaker becomes hidden and can't be detected by this script. This behavior is:
- consistent with `wpctl status`, which also doesn't list the speaker when headphones are connected
- inconsistent with GNOME, which lists the built-in speaker AND the headphones

It's not a critical problem since we never test like this, but might be worth noting. Headphone ports on the dock will be correctly detected.

### What the test data contains

They are essentially `pw-dump` results from a laptop, that has built-in speakers and HDMI, connected to an external display that also has built-in speakers. 

## Tests

C3 Submission:https://certification.canonical.com/hardware/202411-36032/submission/482965/


checkbox-support/checkbox_support/scripts/pipewire_utils.py
```diff
@@ -395,46 +398,182 @@ def go_through_ports(self, cmd, mode):
...
-                        "yes",
-                        "unknown",
-                    ]:
-                        while chosen != "yes":
-                            self.logger.info(
-                                "Please select [{}] for "
-                                "testing (if selected, "
-                                "please enter 'yes')".format(p["description"])
-                            )
+                assert type(ports) is list
```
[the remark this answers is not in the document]
fernando79513: You changed the other asserts but not this one. Any reason to do that?

checkbox-support/checkbox_support/scripts/pipewire_utils.py
```diff
@@ -395,46 +398,213 @@ def go_through_ports(self, cmd, mode):
...
+                            "[ ERR ] Only {} audio sinks were tested, ".format(
+                                len(tested_ids)
                             )
+                            + "but expected {}".format(N)
+                        )
+
+                idx = int(choice)
+                subprocess.check_call(
+                    ["wpctl", "set-default", str(audio_sink_ids[idx][0])]
+                )
```
fernando79513, line 521: If this fails with `subprocess.CalledProcessError` it is not caught.
This could happen if you disconnect a device after discovery (I tried it locally). 

I think we could add an exception for that:

```suggestion
                subprocess.check_call(
                    ["wpctl", "set-default", str(audio_sink_ids[idx][0])]
                )
            except subprocess.CalledProcessError as e:
                print(
                    "ERROR: Failed to set default audio sink: {}".format(e),
                    file=sys.stderr,
                )
                continue
```

checkbox-support/checkbox_support/scripts/pipewire_utils.py
```diff
@@ -395,46 +398,213 @@ def go_through_ports(self, cmd, mode):
...
+                    )
+
+                print(
+                    "Choose an audio sink to test [0-{}],".format(N - 1),
+                    "or type 'q' to quit: ",
+                    end="",
+                    flush=True,
+                )
+                choice = input()
+
```
fernando79513, line 489: This is another QoL idea, I don't know if it makes sense. 
If you forgot to plug a connection that you want to test (e.g. Usb headphones) you can just connect it and re-discover all the audio sinks

```suggestion
            audio_sink_ids = list(self._find_available_audio_sinks().items())
            N = len(audio_sink_ids)
            try:
                for i, (node_id, node_description) in enumerate(
                    audio_sink_ids
                ):
                    print(
                        "({}) - '{}' {}".format(
                            i,
                            node_description,
                            ("- Tested" if node_id in tested_ids else ""),
                        )
                    )

                print(
                    "Choose an audio sink to test [0-{}],".format(N - 1),
                    "type 'r' to reload, or type 'q' to quit: ",
                    end="",
                    flush=True,
                )
                choice = input()

                if choice == "r":
                    continue

```

fernando79513 (changes requested):
GJ with this PR, and thanks for taking the time to improve the tooling.

This is a lengthy one, but mostly bc of the tests.

I added a couple of comments, nothing big, and I tested it locally.
In general, I think it's okay, but review the comments (also the one on the copilot PR regarding the use of assert)

checkbox-support/checkbox_support/scripts/pipewire_utils.py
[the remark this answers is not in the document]
tomli380576: oops i forgot

checkbox-support/checkbox_support/scripts/pipewire_utils.py
tomli380576: Thanks this is a really good idea!

fernando79513 (approved):
Very good work!

Thanks a lot for the effort here.

LGTM +1!

PR #2445, opened 2026-04-01
Completely deprecate the usage of gnome-randr package (BugFix)

p-gentili:
## Description

We have a library that replaces completely gnome-randr, yet we still have some usage of it here and there.

1. Complete the migration
2. Cleanup scripts
3. Stop building the gnome-randr package

## Resolved issues

N/A

## Documentation

N/A

## Tests

Please @hanhsuan help :)


Hook25 (approved):
thanks

PR #2449, opened 2026-04-03
providers/base: fix race condition in turn_down_nm_connections (BugFix)

eugene-yujinwu:
## Problem

There is a race condition in `turn_down_nm_connections()` in `wifi_nmcli_test.py`.

The function calls `_get_nm_wireless_connections()` to list active connections, then runs `nmcli c down <uuid>` for each activated one. However, between the query and the `nmcli c down` call, the connection may have already become inactive (e.g., due to a network event or prior disconnect).

In this case `nmcli` returns exit code **10** ("not an active connection"), which caused `sp.check_call()` to raise `CalledProcessError` and crash the test — even though the connection being down is exactly the desired outcome.

**Observed failure:**
```
Turn down connection Canonical
+ nmcli c down cf0b6f04-db09-46f5-a8bb-e703f97c003a
Error: 'cf0b6f04-db09-46f5-a8bb-e703f97c003a' is not an active connection.
subprocess.CalledProcessError: Command returned non-zero exit status 10.
```
https://warthogs.atlassian.net/browse/SUTTON-4809
https://certification.canonical.com/hardware/202603-38441/submission/482952/test/84491/result/57704127/
https://certification.canonical.com/hardware/202602-38419/submission/481505/test/221519/result/57390085/

## Fix

Replace `sp.check_call()` with `sp.call()` and explicitly handle the return code:
- Exit code `0`: success
- Exit code `10`: connection already inactive — goal achieved, treat as success
- Any other non-zero code: still raises `CalledProcessError`

## Tests

- Updated existing unit tests to mock `sp.call` instead of `sp.check_call`
- Added new test `test_turn_down_connection_already_inactive` to verify exit code 10 is handled gracefully
- https://certification.canonical.com/hardware/202603-38440/submission/482955/
- https://certification.canonical.com/hardware/202603-38440/submission/482963/
- https://certification.canonical.com/hardware/202603-38440/submission/483237/
- https://certification.canonical.com/hardware/202603-38442/submission/483250/

Re-run the test with the new way which use statce check other than ignore exit 10:
[CE-QA-PC]: Automatic (BugFix) #2449_use_state_check test for Paris-P-SIT-S15
https://certification.canonical.com/hardware/202603-38440/submission/483385/

### DIFF
--- a/providers/base/bin/wifi_nmcli_test.py
+++ b/providers/base/bin/wifi_nmcli_test.py
@@ -100,7 +100,10 @@ def turn_down_nm_connections():
         print("Turn down connection", name)
         cmd = "nmcli c down {}".format(uuid)
         print_cmd(cmd)
-        sp.check_call(shlex.split(cmd))
+        ret = sp.call(shlex.split(cmd))
+        # exit code 10: connection is not active — already down, that's fine
+        if ret not in (0, 10):
+            raise sp.CalledProcessError(ret, cmd)
         print("{} {} is down now".format(name, uuid))
     print()
 
--- a/providers/base/bin/wifi_nmcli_test.py
+++ b/providers/base/bin/wifi_nmcli_test.py
@@ -112,15 +112,7 @@ def turn_down_nm_connections():
             continue
         cmd = "nmcli c down {}".format(uuid)
         print_cmd(cmd)
-        try:
-            sp.check_call(shlex.split(cmd))
-        except sp.CalledProcessError as e:
-            # The connection may have gone inactive in the narrow window
-            # between the state pre-check and this call. Exit code 10
-            # ("connection/device/AP does not exist") means it is already
-            # down — treat as success.
-            if e.returncode != 10:
-                raise
+        sp.check_call(shlex.split(cmd))
         print("{} {} is down now".format(name, uuid))
     print()
 
--- a/providers/base/bin/wifi_nmcli_test.py
+++ b/providers/base/bin/wifi_nmcli_test.py
@@ -112,7 +112,15 @@ def turn_down_nm_connections():
             continue
         cmd = "nmcli c down {}".format(uuid)
         print_cmd(cmd)
-        sp.check_call(shlex.split(cmd))
+        try:
+            sp.check_call(shlex.split(cmd))
+        except sp.CalledProcessError as e:
+            # The connection may have gone inactive in the narrow window
+            # between the state pre-check and this call. Exit code 10
+            # ("connection/device/AP does not exist") means it is already
+            # down — treat as success.
+            if e.returncode != 10:
+                raise
         print("{} {} is down now".format(name, uuid))
     print()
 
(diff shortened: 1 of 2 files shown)


hanhsuan:
In [this doc](https://networkmanager.dev/docs/api/latest/nmcli.html) the exit code `10` means `Connection, device, or access point does not exist.` Do you mind to provide more information about your finding for the exit code `10` ?

providers/base/bin/wifi_nmcli_test.py
```diff
@@ -100,7 +100,10 @@ def turn_down_nm_connections():
         print("Turn down connection", name)
         cmd = "nmcli c down {}".format(uuid)
         print_cmd(cmd)
-        sp.check_call(shlex.split(cmd))
+        ret = sp.call(shlex.split(cmd))
+        # exit code 10: connection is not active — already down, that's fine
+        if ret not in (0, 10):
+            raise sp.CalledProcessError(ret, cmd)
```
tomli380576: optional: I think you can use try-catch here to avoid manually constructing the error
```py
try:
    sp.check_call(["nmcli", "connection", "down", uuid])
except sp.CalledProcessError as e:
    if e.returncode == 10:
        print("WARN: connection {} doesn't exist, ignoring".format(uuid))
    else:
        raise e
```

tomli380576 (changes requested):
See Hanhsuan's comment

providers/base/bin/wifi_nmcli_test.py
eugene-yujinwu: Thanks for the suggestion! We've actually taken it a step further in the latest commit. Rather than catching the error after the fact, we now pre-check the specific connection's state via UUID before issuing the down command:
```
nmcli -t -f GENERAL.STATE c show <uuid>
```
If the connection is no longer `activated`, we skip it. This removes the need to interpret exit code 10 entirely, and `check_call` reverts to its default behavior — any real error still raises as expected.

eugene-yujinwu:
Thanks for raising this — you're right that exit code 10 can mean multiple things ("connection/device/AP does not exist"), making it risky to ignore unconditionally.

Instead of relying on exit code interpretation, the latest commit takes a different approach: before calling `nmcli c down`, we now query the specific connection's state by UUID:
```
nmcli -t -f GENERAL.STATE c show <uuid>
```
If the state is no longer `activated`, we skip the down command entirely with a warning. This avoids the exit code 10 discussion altogether — `check_call` is back to its standard behavior and any unexpected error will still propagate normally.

Hook25:
@eugene-yujinwu two things:
1. please verify the AI output before asking for a review
2. please don't ask for a review on a draft PR, I've spent 20m checking if `nmcli -t -f GENERAL.STATE c show` works on xenial and you changed the implementation again. Use the draft status if you aren't ready
3. writing a description to a PR is the most important part of the work. I don't want to see AI generated description please. At least make it conform to the template if you really want to do it this way..

eugene-yujinwu:
> @eugene-yujinwu two things:
> 
> 1. please verify the AI output before asking for a review
> 2. please don't ask for a review on a draft PR, I've spent 20m checking if `nmcli -t -f GENERAL.STATE c show` works on xenial and you changed the implementation again. Use the draft status if you aren't ready
> 3. writing a description to a PR is the most important part of the work. I don't want to see AI generated description please. At least make it conform to the template if you really want to do it this way..

Sorry， I forgot to put it in draft PR as I still modify it.
Yes, I will double check these descriptions and code before push it. 
Sorry again.

PR #2450, opened 2026-04-03
Provide more information in the HTML reports (New)

pieqq:
## Description

Add the following to the default HTML report (used by Checkbox when
exporting to HTML and when using the `merge-submissions` subcommand) as well as the multi-page HTML report (used by the `merge-reports` subcommand):

- If `system_info_json` job has not been run, use some data from the
`system_information` collector instead (distribution, kernel, memory)
- Add more metadata from `system_information` collectors, if available
(these are collapsed tables by default and can be shown by clicking on
their title):
    - Image information
    - BIOS
    - Machine Manifest
    - Snaps
    - Debian Packages
- Add headers for each columns in the Tests Results
- Add execution times (in seconds) for each Test Result

A bug has been fixed in merge_reports.py to properly expose `system_information` to the templates.

The formatting of the HTML templates has been modified a bit.

Sample HTML report: [submission_2026-04-13T06.28.05.338628.zip](https://github.com/user-attachments/files/26667362/submission_2026-04-13T06.28.05.338628.zip)

Sample HTML report on UC16: [submission.html](https://github.com/user-attachments/files/26741504/submission.html) (it's run in a VM with a test plan that ends up not running any jobs, hence the lack of executed jobs).

When available, new sections appear (collapsed by default):

image

Each section can be clicked on to reveal the data:

image

Test Results now include headers as well as execution time:

image

## Resolved issues

https://warthogs.atlassian.net/browse/CHECKBOX-2024

## Tests

- Ran checkbox-cli to generate an HTML report, and checked that all the data was there
- Used submissions created with previous versions of Checkbox to test the `merge-reports` and `merge-submissions` subcommands, both work as expected.

### DIFF
--- a/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
+++ b/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
@@ -2,6 +2,7 @@
 {%- set state = manager.default_device_context.state -%}
 {%- set resource_map = state.resource_map -%}
 {%- set job_state_map = state.job_state_map -%}
+{%- set system_information = state.system_information -%}
 {%- set category_map = state.category_map_lite -%}
 {%- set category_outcome_map = state.category_outcome_map -%}
 {%- set resource_global_outcome = state.resource_global_outcome -%}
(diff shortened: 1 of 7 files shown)


checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
Hook25, line 1: I have two points on this:
1. system information is a better mechanism (imo) to gather informations, why are we using the old one and only falling back to it?
2. Why are we checking if the collectors were executed? Imo if a collector is removed but is used here, it should blow up so that we notice.

Hook25 (changes requested):
Really nice job, I dont really have any objections here but one: Can you provide a submission on core16 just to be sure?

Another minor consideration, you added so many cool and useful sys infos but I can't see them here, why not add to the report also (at least) uname and distribution?

checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
pieqq: > system information is a better mechanism (imo) to gather informations, why are we using the old one and only falling back to it?

The old one is a big script that extracts a bunch of things that are not present in our `system_information` blob yet, such as human-readable information about the Bluetooth module or the Webcam present on the system. On recent laptop submissions, it looks like this:

Image

If we replace this with the information available from `system_information`, we would be down to

- system (Ubuntu info, kernel info)
- BIOS info
- Memory info

and that's all.

I can nevertheless obliterate this if you think it's OK.


> Why are we checking if the collectors were executed? Imo if a collector is removed but is used here, it should blow up so that we notice.

I would have to re-test this, but when I implemented it, stuff like image_info run a command that's only present in classic desktop images, so it crashes on UC or server images. I'll revisit this.

Hook25: ok, makes sense, leave it as is.

Second point, I mean that you can just check if they pass, why do you check if they are present?

pieqq: For the second point, it's because these HTML templates are used in the merge-submissions and merge-reports commands that might work on older submissions that don't have these fields available.

Hook25 (approved):
Thanks

PR #2452, opened 2026-04-08
No more requests unixsocket (bugfix)

Hook25:
## Description

The current state of requests unixsocket is unclear. The main project was unmaintained and forked, the package in the archive is not quite clear if it works with new requests and it periodically causes issues like the one currently blocking edge validation. Additionally, we don't need this dependency for the only job it was pulled to do, as there is a builtin way to get the same result without any dependency.

## Resolved issues

Fixes: CHECKBOX-2222

## Documentation

N/A

## Tests

Ran the module as root on ubuntu 16.04 and it works out of the box as it doesn't have any dependency.


p-gentili (dismissed):
Makes sense, thanks!

p-gentili (approved):
Fly metabox, fly

PR #2457, opened 2026-04-10
Move custom frontend and load all for all jobs (bugfix)

Hook25:
## Description

Previous implementation only loaded envvars per provider, effectively binding them to the providers they target. This was a mistake, as that means that providers can't rely on each other and resources like executables are basically useless with this new architecture. 

This adresses this flaw by making all custom frontend variables apply to all jobs, as they should.

## Resolved issues

N/A

## Documentation

N/A

## Tests

Added/moved tests to the new module. 

Tested this locally by building checkbox24 and a custom-frontend provider. Snap trying both. Creating a new executable (touch a; chmod +x a) in the custom frontend `/usr/bin`, adding the following job to Checkbox to a random pxu and triggering it, it works.
```
id: test_path
flags: simple
requires: executable.name == 'a'
command:
  echo "it works :)"
_summary: Test that path includes all custom frontends
```

Command:
```
checkbox24.checkbox run test_path
```


p-gentili:
Was looking at the documentation, https://github.com/canonical/checkbox/blob/main/docs/how-to/custom-frontend-envvars.rst. You might want to rephrase "accessed globally by all tests in a frontend" with "accessed globally by all tests in a frontend regardless of which provider the test comes from". Or simple "accessed globally by all tests."

Testing now...

checkbox-ng/plainbox/impl/secure/providers/custom_frontend.py
```diff
@@ -0,0 +1,186 @@
...
+        "usr/lib/{}/site-packages".format(python_name),
+        "usr/lib/{}/lib-dynload".format(python_name),
+        "usr/lib/python3/dist-packages",
+        "usr/local/lib/{}/dist-packages".format(python_name),
+    ]
+    return [
+        path
+        for custom_frontend_root in custom_frontend_roots()
+        for path in paths_to_custom_frontend_path(custom_frontend_root, paths)
+    ]
```
p-gentili, line 139: I might be wrong, but it looks like the order in here is the opposite of the one you get from `extra_snap_environment`.

checkbox-ng/plainbox/impl/secure/providers/custom_frontend.py
```diff
@@ -0,0 +1,186 @@
...
+    CUSTOM_FRONTEND_LOCATION = Path(os.getenv("SNAP")) / "custom_frontends"
+else:
+    CUSTOM_FRONTEND_LOCATION = None
+
+
+def custom_frontend_roots() -> list:
+    if not CUSTOM_FRONTEND_LOCATION:
+        return []
+    try:
+        return list(filter(Path.is_dir, CUSTOM_FRONTEND_LOCATION.iterdir()))
```
p-gentili: speaking of order, it seems like [iterdir](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir)'s is arbitrary: do we care?

p-gentili (dismissed):
Tested and confirmed the behavior before and after this patch. Thanks a lot!

---

A comment below, interesting or not? I'm not sure anymore, but the orders of path/pythonpath seems to be opposite.

checkbox-ng/plainbox/impl/secure/providers/custom_frontend.py
Hook25: made it sorted

checkbox-ng/plainbox/impl/secure/providers/custom_frontend.py
Hook25: as far as I can see, the logic matches, they should be in the same order no? maybe I didnt understand this point.

p-gentili (approved):
Tested again - still working. Thanks for your patience.

PR #2461, opened 2026-04-14
Add Crucible (Mesa CTS) Testing (New)

mckees:
## Description

Crucible is a test suite for Mesa conformance.

Some notes in this PR: 
* This integration presented more issues around GPU selection than previous tests. We ran into the issue of prime-select preferences being ignored and forking threads crashing the test machine. We also found some places to improve the previous Vulkan testing because the vulkaninfo checks run in core tests before were not in the snap environment. 
* We also added some previously-requested refactoring work to pull common functions into a host_utils.py script

## Resolved issues

Mesa is already tested by the Vulkan CTS testing, but Crucible is the testing developed specifically for Mesa, so it is a great tool to make sure that it is integrated properly from the perspective of the upstream developers.

## Documentation

https://gitlab.freedesktop.org/mesa/crucible

## Tests

* Ran the classic and core test suites for Crucible
* Ran the classic and core test suites for Vulkan CTS (shared elements between them make a regression more likely)

### DIFF
--- a/checkbox-support/checkbox_support/helpers/host_utils.py
+++ b/checkbox-support/checkbox_support/helpers/host_utils.py
@@ -136,7 +136,9 @@ def _vendor_prefixes_from_vulkaninfo(output):
         for vid, prefixes in _PCI_VENDOR_ICD_PREFIXES.items():
             if vid in stripped:
                 return prefixes
-    raise VulkanDetectionError("no known GPU vendor found in vulkaninfo output")
+    raise VulkanDetectionError(
+        "no known GPU vendor found in vulkaninfo output"
+    )
 
 
 def active_vendor_prefixes():
@@ -146,8 +148,8 @@ def active_vendor_prefixes():
     then falls back to vulkaninfo via plz-run.
     Raises VulkanDetectionError if no method identifies the vendor.
     """
-    # prime-select is only present on NVIDIA hybrid systems; absence or
-    # unrecognised output (e.g. on-demand) is normal — fall through to vulkaninfo.
+    # prime-select is only present on NVIDIA hybrid systems;
+    # absence or unrecognised output is normal — fall through.
     try:
         vendor = prime_selected_vendor()
         return _PRIME_VENDOR_ICD_PREFIXES[vendor]
--- a/checkbox-support/checkbox_support/helpers/host_utils.py
+++ b/checkbox-support/checkbox_support/helpers/host_utils.py
@@ -38,7 +38,6 @@ def get_arch_triple():
 
 def find_plz_run():
     """Return the path to plz-run from PATH.
-
     Raises VulkanDetectionError if plz-run is not found.
     """
     path = shutil.which("plz-run")
@@ -47,6 +46,7 @@ def find_plz_run():
     return path
 
 
+# Loading these will hang
 _VIRTUAL_ICD_PREFIXES = {"gfxstream", "virtio"}
 
 # Maps prime-select vendor name to ICD filename prefixes.
--- a/checkbox-support/checkbox_support/helpers/host_utils.py
+++ b/checkbox-support/checkbox_support/helpers/host_utils.py
@@ -18,11 +18,14 @@
 
 """Shared utilities for host Vulkan test helpers."""
 
-import json
 import os
 import shutil
 import subprocess
 import sysconfig
+
+
+class VulkanDetectionError(Exception):
+    """Raised when a GPU/Vulkan detection step fails."""
 
 
 def get_arch_triple():
@@ -34,11 +37,17 @@ def get_arch_triple():
 
 
 def find_plz_run():
-    """Return the path to plz-run from PATH."""
-    return shutil.which("plz-run")
+    """Return the path to plz-run from PATH.
+
+    Raises VulkanDetectionError if plz-run is not found.
+    """
+    path = shutil.which("plz-run")
+    if path is None:
+        raise VulkanDetectionError("plz-run not found in PATH")
+    return path
 
 
-_VIRTUAL_ICD_LIBS = {"libvulkan_gfxstream.so", "libvulkan_virtio.so"}
+_VIRTUAL_ICD_PREFIXES = {"gfxstream", "virtio"}
 
 # Maps prime-select vendor name to ICD filename prefixes.
 # Used to filter the ICD list to the PRIME-selected GPU on multi-GPU systems.
@@ -48,8 +57,8 @@ _PRIME_VENDOR_ICD_PREFIXES = {
     "amd": ("radeon", "amd"),
 }
 
-# Maps PCI vendor ID from /sys/class/drm to ICD filename prefixes.
-_DRM_VENDOR_ICD_PREFIXES = {
+# Maps PCI vendor ID to ICD filename prefixes.
+_PCI_VENDOR_ICD_PREFIXES = {
     "0x8086": ("intel",),
     "0x1002": ("radeon", "amd"),
     "0x10de": ("nvidia",),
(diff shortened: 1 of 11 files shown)


providers/base/bin/crucible_host.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+# This file is part of Checkbox.
+#
+# Copyright 2025 Canonical Ltd.
```
Hook25: Minor: lol
```suggestion
# Copyright 2026 Canonical Ltd.
```

providers/base/bin/crucible_host.py
```diff
@@ -0,0 +1,120 @@
...
+  run-test ARGS...  Run a crucible test using host Vulkan libraries,
+                    forwarding all remaining arguments to crucible run.
+"""
+
+import logging
+import os
+import subprocess
+import sys
+
+from host_utils import (  # noqa: F401
```
Hook25: Why noqa?
```suggestion
from host_utils import (
```

mckees:
At time of writing, the Vulkan Classic test plan will fail because this job is still running (meaning I cannot promote the snap with the needed changes):
https://github.com/canonical/vulkan-cts-snap/actions/runs/24387359316

This one has been a bit trickier to integrate than the past two, so I'm definitely looking for some feedback where folks see a need for additional polishing. I've been trying to leave our options open for non-amd64 hardware, but the [crucible snap](https://github.com/canonical/crucible-snap) does not have a build for any other arch at the moment.

providers/base/bin/crucible_host.py
```diff
@@ -0,0 +1,120 @@
...
+    logging.error(
+        "Install libvulkan1 or equivalent before running host crucible tests"
+    )
+    return 1
+
+
+def cmd_run_test(test_args):
+    snap = "/snap/crucible/current"
+    # NODEVICE_SELECT disables VK_LAYER_MESA_device_select — the layer
+    # requires GLIBC_ABI_GNU2_TLS, which the snap's core24 glibc lacks.
```
Hook25, line 76: Is this a TODO to remove this on core26?

checkbox-support/checkbox_support/helpers/host_utils.py
Hook25, line 1: if you want to import this, it must move to the checkbox_support module. Probably under helpers.

mckees:
The vulkan snap has been updated, so the vulkan classic test plan will succeed.

providers/base/bin/host_utils.py
```diff
@@ -0,0 +1,224 @@
...
+                ["/usr/bin/prime-select", "query"],
+                stderr=subprocess.DEVNULL,
+                universal_newlines=True,
+            )
+            .strip()
+            .lower()
+        )
+        return output if output in _PRIME_VENDOR_ICD_PREFIXES else None
+    except (FileNotFoundError, OSError, subprocess.CalledProcessError):
+        return None
```
Hook25: No, don't hide exceptions, rise them. You can group them all into a single one if you want.

providers/base/bin/host_utils.py
```diff
@@ -0,0 +1,224 @@
...
+        output = (
+            subprocess.check_output(
+                ["/usr/bin/prime-select", "query"],
+                stderr=subprocess.DEVNULL,
+                universal_newlines=True,
+            )
+            .strip()
+            .lower()
+        )
+        return output if output in _PRIME_VENDOR_ICD_PREFIXES else None
```
Hook25: Why would this return None, if you call this function you expect this to return something, then raise if you find nothing to return

providers/base/bin/host_utils.py
```diff
@@ -0,0 +1,224 @@
...
+                "LD_LIBRARY_PATH={}".format(ld_library_path),
+                "--",
+                "/usr/bin/vulkaninfo",
+                "--summary",
+            ],
+            universal_newlines=True,
+            stderr=subprocess.STDOUT,
+        )
+    except subprocess.CalledProcessError:
+        return None
```
Hook25: Why

checkbox-support/checkbox_support/helpers/host_utils.py
Hook25, line 1: Ok I don't want to write the same comment on the whole file. In python we don't return None as a special value to say error. Error = Exception. I don't like exceptions but that is how the language works.

providers/base/bin/host_utils.py
```diff
@@ -0,0 +1,224 @@
...
+                    vid = f.read().strip().lower()
+                prefixes = _DRM_VENDOR_ICD_PREFIXES.get(vid)
+                if prefixes is not None:
+                    return prefixes
+                break
+            except OSError:
+                continue
+    except OSError:
+        pass
+    return None
```
Hook25: ? I'm sure we can do better than this, 2 nested os errors? this is really hard to test/reason about

providers/base/bin/host_utils.py
```diff
@@ -0,0 +1,224 @@
...
+                )
+                if lib in _VIRTUAL_ICD_LIBS:
+                    continue
+                if vendor_prefixes and not any(
+                    name.startswith(p) for p in vendor_prefixes
+                ):
+                    continue
+                result.append(path)
+            except (OSError, ValueError):
+                result.append(path)
```
Hook25: Why would we add the path erroring to result? At least comment on this action if you have to do it

providers/base/bin/host_utils.py
```diff
@@ -0,0 +1,224 @@
...
+    except OSError:
+        pass
```
Hook25: Same as above, are we sure we need two nested os errors? Remember `suppress` exists and makes your life easier AND raising exceptions is actually idiomatic if you want to say somthing went wrong

providers/base/units/graphics/vulkan-cts.pxu
```diff
@@ -18,15 +18,13 @@ plugin: resource
...
-requires:
-  executable.name == 'vulkaninfo'
 _summary: Checks for Vulkan-capable GPU
 environ:
   XDG_SESSION_TYPE
   XDG_RUNTIME_DIR
 estimated_duration: 1s
 command:
- if vulkaninfo --summary 2>/dev/null | grep -q "PHYSICAL_DEVICE_TYPE_.*_GPU"; then
+ if env -u DISPLAY vulkan-cts.vulkaninfo --summary 2>&1 | grep "PHYSICAL_DEVICE_TYPE_.*_GPU"; then
```
Hook25, line 29: Comment why you env -u that variable. Also, weird to do this in bash considering we did everything else in python

Hook25 (changes requested):
Consider the following

providers/base/bin/crucible_host.py
mckees: Claude sneaked it in when I asked it to reformat some flake8/black stuff, and I didn't notice, sorry

providers/base/bin/crucible_host.py
mckees: Ideally, yes. The device selection via VK_LAYER_MESA_device_select fails if the snap has a different glibc version than the host, which I suspect could continue to be an issue in the future. For instance, if we need to test on 26.10 with core26, the same issue should come back

checkbox-support/checkbox_support/helpers/host_utils.py
mckees: Done

providers/base/bin/host_utils.py
mckees: Done, converted everything to raise VulkanDetectionError

providers/base/bin/host_utils.py
mckees: This one specifically didn't raise the error if you find nothing because some devices will not have or need prime-select, so we don't want to error out if we need to fall down to the typical Vulkan GPU selection. I've changed it to raise the error as requested, so that's why you will see the next function down the line catching VulkanDetectionError and passing instead of noting the None return and moving on to letting Vulkan handle GPU selection

providers/base/bin/host_utils.py
mckees: Changed

providers/base/bin/host_utils.py
mckees: Refactored to remove this pattern

providers/base/bin/host_utils.py
mckees: A holdover from an earlier experiment when the correct path was not being added in the try block. Thanks for catching, I've changed it

providers/base/bin/host_utils.py
mckees: Changed

providers/base/units/graphics/vulkan-cts.pxu
mckees: Added a comment to explain that this is for a more headless experience

Hook25 (approved):
+1, thanks

PR #2463, opened 2026-04-14
Avoid re-connecting every Ctrl+C (bugfix)

Hook25:
## Description

This is a weird behaviour because regardless of what the user picks from the ctrl+c menu, we will always re-connect from scratch and sometimes mess up the session

image

## Resolved issues

N/A (Ctrl+C didnt work, with this patch it does)

## Documentation

N/A

## Tests

Tested locally, before this patch, Ctrl+C always reconnects, so it basically never works.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -286,6 +286,7 @@ class RemoteController(ReportsStage, MainLoopStage):
         while True:
             try:
                 if interrupted:
+                    # handling ctrl+c, assumes we are already connected
                     _logger.info("controller: Session interrupted")
                     interrupted = False  # we are handling the interruption ATM
                     # next line can raise exception due to connection being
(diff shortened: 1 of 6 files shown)


fernando79513:
Please add a more detailed description of what the behavior was before this PR, and how this solves the issue.
I tried to run smoke tests with the previous version with:
```
checkbox.checkbox-cli control 10.164.173.184 smoke_launcher 
```
And I was able to continue the execution, disconnect from the agent, and start a new one.

fernando79513 (changes requested):
Thanks for taking care of this.

I think this PR could benefit greatly from some metabox tests. I don't know how difficult it would be to add them, but I think they can help us avoid these kinds of issues in the future.

fernando79513:
I run the metabox tests manually yesterday and they are failing:
https://github.com/canonical/checkbox/actions/runs/24726686275/job/72329566224

fernando79513 (approved):
LGTM +1. Approving it to see the metabox tests run.

PR #2465, opened 2026-04-15
VRR capable monitor detection (New)

tomli380576:
## Description

This PR implements a script that can detect whether the DUT is connected to any VRR capable monitors. This is supposed to act as the resource job for `graphics/variable-refresh-rate-manual` when `has_vrr` is set to true. See #2464 for more context. **#2464 and this PR can be merged in any order, they don't reference each other yet.** There will be 1 last PR after this one is merged to put everything together.

## Resolved issues

#2464 as it is right now doesn't actually check whether the DUT is connected to a VRR capable monitor. 

## Documentation

The basic idea is to mimic [what `modetest -connectors` does](https://github.com/grate-driver/libdrm/blob/master/tests/modetest/modetest.c) (this command comes from libdrm-tests). `modetest` will print these lines for a VRR capable monitor:

```
103 vrr_capable:
    flags: immutable range
    values: 0 1
    value: 1
```
If VRR is impossible, this property is either 0 or missing altogether. This script uses the same libdrm API calls to find this value.

[The exact lines we are mimicking](https://github.com/grate-driver/libdrm/blob/3e3c53ed85c30331a6c0ba2af349223654caa5dd/tests/modetest/modetest.c#L450-L457)

## Tests

Manually checked the output of this script against GNOME's output in gnome settings.

### DIFF
--- a/providers/base/bin/detect_vrr.py
+++ b/providers/base/bin/detect_vrr.py
@@ -127,15 +127,16 @@ def get_vrr_capable_monitors(dri_card: Path) -> bool:
                 if not prop_ptr:
                     continue
 
-                prop_name = prop_ptr.contents.name.decode()
-                if prop_name == "vrr_capable":
-                    val = conn.prop_values[j]
-                    if val == 1:
-                        print(
-                            "Connection ID",
-                            conn_id,
-                            "is VRR capable",
-                        )
+                prop = prop_ptr.contents
+                if prop.name.decode() == "vrr_capable":
+                    # if we have the right name,
+                    # that means we are on the right index
+                    # use j to index into the value array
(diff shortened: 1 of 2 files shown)


baconYao:
Interesting topic!
One quick question, have you tried to pack the libdrm-test deb package into Checkbox, maybe in the checkbox-support?

tomli380576:
> Interesting topic! One quick question, have you tried to pack the libdrm-test deb package into Checkbox, maybe in the checkbox-support?

Nope, because this script is supposed to achieve the same thing without including `libdrm-test`. You can try this on any device by  replacing `slugify` with `str()` and run `python3 detect_vrr.py`

PR #2467, opened 2026-04-15
Update gpu passthrough tests (BugFix)

pedro-avalos:
## Description

- The GPU passthrough script uses environment variables to use local images or modify the VM options, but these are not defined in the `.pxu` jobs
- The tests have also been updated to run mixbench only once
  - Since mixbench scales with the hardware, there is no need to re-run it and time it; if the command works, the test passes


pedro-avalos:
Closing in favor of #2136

PR #2473, opened 2026-04-15
miscellanea/sosreport: requirement now is a tuple of sosreport and sos to work on new and old releases  (bugfix)

### LINKED ISSUE #2472: miscellanea/sosreport still depends on "sosreport" as an executable which does not exist in resolute
### Bug Description

TSIA

miscellanea/sosreport requires "sosreport" to exist but that is deprecated.  There is a transitional package for sosreport in older releases but now the package and executable are called simply sos.

Fix by adding sos as an optional executable

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

Install canonical-certification-server from edge on a resolute system
run any command
sosreport is skipped because "sosreport" no longer exists on Resolute

### Expected Result

sosreport runs using the sos command

### Actual Result

see above

### Environment

Resolute
Checkbox from edge

### Relevant log output

```shell
miscellanea/sosreport	not supported			job cannot be started: resource expression "executable.name == 'sosreport'" evaluates to false
```

### Additional context

_No response_

bladernr:
miscellanea/sosreport: requirement now is a tuple of sosreport and sos to work on new and old releases (Bugfix)

## Description

add 'sos' as an executable to look for in the requires for miscellanea/sosreport to ensure it runs on old and new releases.

## Resolved issues
fixes #2472

## Tests

modified in-situ and ran job on noble and resolute to ensure the sosreport job is executed. It was successful on both releases.

### DIFF
--- a/providers/base/units/miscellanea/jobs.pxu
+++ b/providers/base/units/miscellanea/jobs.pxu
@@ -457,7 +457,7 @@ category_id: com.canonical.plainbox::miscellanea
 estimated_duration: 20.0
 id: miscellanea/sosreport
 user: root
-requires: executable.name == 'sosreport'
+requires: executable.name in ('sosreport', 'sos')
 command: sos report --batch -n lxd --tmp-dir "$PLAINBOX_SESSION_SHARE"
 _summary: Generate baseline sosreport
 _purpose: Generates a baseline sosreport of logs and system data


Hook25 (approved):
why does the requirement ask for sosreport if the command uses `sos`? Shouldn't sos be the executable this needs? either way, lgtm

bladernr:
I just looked and the change happened in 4.0 and is currently existent as far back as Bionic.  INdeed we could just get rid of the sosreport executable requirement, I didn't dig that far back last night.  But that is only if we aren't testing anything older than bionic (anyone still do xenial testing that would be using this?)

PR #2481, opened 2026-04-17
Create RVS snap common directory if it does not exist (BugFix)

pedro-avalos:
## Description

- The RVS snap doesn't seem to create the common directory when it's installed
- The RVS tests also require elevation to root user

## Tests

https://certification.canonical.com/hardware/202606-38900/submission/494795/test-results/pass/


providers/gpgpu/bin/rvs.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, rvs: Path, config: Path) -> None:
         # CHECKBOX-2021: Snap confinement prevents access to /usr/share
```
ajzobro, line 51: Does this still apply if it's run as root now?

pedro-avalos: Yep, snapd will still prevent the snap from accessing that directory

PR #2483, opened 2026-04-20
Migrate idle suspend test cases from manual to semi-auto(DC) and auto(AC) (New)

hanhsuan:
## Description

This PR is going to migrate the manual test cases to semi-auto and auto:
1. `id: power-management/suspend-on-AC-power` to Auto test case
2. `id: power-management/suspend-on-battery-power` to semi-auto test case

The new test groups three test cases
1. setup: for setting dbus configurations of idle to suspend behavior and backup the configurations
2.  test: 
    2.1. Set the rtc wake for full auto AC test only
    2.2. Ensure the system is in the correct power supply mode. (DC/AC)
    2.3. Verifying the suspend is starting at the right timing or not
4. teardown: restore the configurations

At this point, the tolerance is higher for collecting more data to fine tune the gate in the future. 

## Tests

Test
laptop:
    22.04: https://certification.canonical.com/hardware/202001-27667/submission/489824/
    24.04: https://certification.canonical.com/hardware/202001-27667/submission/489823/
desktop: https://certification.canonical.com/hardware/202603-38566/submission/486631/

### DIFF
--- a/providers/base/bin/idle_suspend.py
+++ b/providers/base/bin/idle_suspend.py
@@ -1,0 +1,253 @@
+#!/usr/bin/env python3
+# This file is part of Checkbox.
+#
+# Copyright 2026 Canonical Ltd.
+#
+# Checkbox is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3,
+# as published by the Free Software Foundation.
+#
+# Checkbox is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import glob
+import re
+import subprocess
+import time
+from datetime import datetime, timedelta, timezone
+from typing import List, Optional, Tuple
+
+GSETTINGS_POWER = "org.gnome.settings-daemon.plugins.power"
+GSETTINGS_SESSION = "org.gnome.desktop.session"
+LOGGER_TAG = "idle_suspend_test"
+
+SUSPEND_PATTERNS = [
+    r"PM: suspend entry",
+    r"Suspending system",
+    r"Reached target.*[Ss]leep",
+]
+RESUME_PATTERNS = [
+    r"PM: suspend exit",
+    r"PM: resume",
+    r"Finished.*[Rr]esume",
+    r"ACPI: Waking",
+]
+
+AC_ONLINE_GLOBS = [
+    "/sys/class/power_supply/ADP*/online",
+    "/sys/class/power_supply/AC*/online",
+    "/sys/class/power_supply/ACAD*/online",
+]
+
+
+def run_cmd(cmd: List[str], check: bool = True) -> str:
+    """Run a shell command and return stdout as a stripped string."""
+    result = subprocess.run(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    if check and result.returncode != 0:
+        raise RuntimeError("Command {} failed: {}".format(cmd, result.stderr))
+    return result.stdout.strip()
+
+
+def _find_ac_online_path() -> Optional[str]:
+    """Return the first sysfs AC online path found, or None."""
+    for pattern in AC_ONLINE_GLOBS:
+        paths = glob.glob(pattern)
+        if paths:
+            return paths[0]
+    return None
+
+
+def check_power_mode(mode: str) -> None:
+    """Verify system is on the expected power mode.
+
+    Raises SystemExit if the requirement is not met.
+    """
+    ac_path = _find_ac_online_path()
+    if ac_path is None:
+        raise SystemExit("Cannot determine AC power status.")
+    try:
+        with open(ac_path) as fh:
+            ac_online = fh.read().strip() == "1"
+    except OSError as exc:
+        raise SystemExit("Cannot read power status: {}".format(exc))
+    if mode == "ac" and not ac_online:
+        raise SystemExit("Mode is 'ac' but system is running on battery.")
+    if mode == "battery" and ac_online:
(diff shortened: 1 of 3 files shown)


providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,253 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import glob
+import re
+import subprocess
+import time
+from datetime import datetime, timedelta, timezone
+from typing import List, Optional, Tuple
```
tomli380576: optional: you can quote the types and use newer syntax without breaking 3.5 compatibility, for example:

```py
_find_ac_online_path() -> Optional[str]:
```
can be turned into
```py
_find_ac_online_path() -> "str | None":
```

Example 2:

```py
def run_cmd(cmd: List[str], check: bool = True) -> str:
```

```py
def run_cmd(cmd: "list[str]", check: bool = True) -> str:
```

This way no imports are needed, or we could `import typing as t` and access all the modern types with for example `t.Literal` in quotes

providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,253 @@
...
+    """Parse a short-iso timestamp string to a naive UTC datetime.
+
+    Returns None on parse failure.
+    """
+    normalized = re.sub(r"([+-]\d{2}):(\d{2})$", r"\1\2", ts_str)
+    try:
+        dt = datetime.strptime(normalized, "%Y-%m-%dT%H:%M:%S%z")
+    except ValueError:
+        return None
+    return dt.replace(tzinfo=timezone.utc) - dt.utcoffset()
```
tomli380576: `dt.utcoffset()` apparently could return `None` ([doc](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcoffset)) 

I'm guessing that's rare but let's give it a default value 

```py
return dt.replace(tzinfo=timezone.utc) - (dt.utcoffset() or timedelta())
```

providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,253 @@
...
+        raise SystemExit(
+            "FAIL: Suspend delay {:.1f}s vs expected {}s "
+            "(tolerance {:.1f}s).".format(
+                actual_delta, suspend_seconds, allowed_delta
+            )
+        )
+
+    print("Log at ({})".format(log_start_utc.strftime("%Y-%m-%dT%H:%M:%SZ")))
+    print("Suspend at ({})".format(suspend_utc.strftime("%Y-%m-%dT%H:%M:%SZ")))
+    print("Resume at ({})".format(resume_utc.strftime("%Y-%m-%dT%H:%M:%SZ")))
```
tomli380576: `resume_utc` could be `None` here, I think we need to test for None first before calling strftime

providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,253 @@
...
+def log_timestamp() -> datetime:
+    """Log a start timestamp to the system journal via logger.
+
+    Returns the captured naive UTC datetime.
+    """
+    now = datetime.now(timezone.utc)
+    message = "SUSPEND_TEST_START: {}".format(
+        now.strftime("%Y-%m-%dT%H:%M:%SZ")
+    )
+    run_cmd(["logger", "-t", LOGGER_TAG, message])
```
tomli380576: I think we can directly call `subprocess.check_output` and redirect stderr/suppress `CalledProcessError` if needed.

tomli380576 (changes requested):
please see the inline comments, the main thing are the possible None values

providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,251 @@
...
+    try:
+        dt = datetime.strptime(normalized, "%Y-%m-%dT%H:%M:%S%z")
+    except ValueError:
+        return None
+    return dt.replace(tzinfo=timezone.utc) - (dt.utcoffset() or timedelta())
+
+
+def parse_journal_suspend_times(
+    journal_output: str,
+) -> ("datetime | None", "datetime | None"):
```
tomli380576: optional: python doesn't reject this but the correct syntax is:
```py
def parse_journal_suspend_times(
    journal_output: str,
) -> "tuple[datetime | None, datetime | None]":
```
(basically wrap the entire python3.10 version in strings)

providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,251 @@
...
+        output = exc.output or ""
+    return output.strip()
+
+
+def _parse_ts(ts_str: str) -> "datetime | None":
+    """Parse a short-iso timestamp string to a naive UTC datetime.
+
+    Returns None on parse failure.
+    """
+    normalized = re.sub(r"([+-]\d{2}):(\d{2})$", r"\1\2", ts_str)
```
tomli380576, line 120: could you add a comment for this regex? looks like it removes the last colon from the string

providers/base/bin/idle_suspend.py
```diff
@@ -0,0 +1,251 @@
...
+    """Parse journal output for the latest suspend and resume times.
+
+    Returns a tuple (suspend_utc, resume_utc) of naive UTC datetimes.
+    Either value may be None if not found.
+    """
+    suspend_utc = None
+    resume_utc = None
+    ts_re = re.compile(
+        r"^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}[+-]\d{2}:\d{2})"
+    )
```
tomli380576, line 141: This doesn't match the timestamp in `journalctl --since today --no-pager -o short-iso` for 22.04 and below, because they don't have the final colon in the timezone:
```py
# doesn't match this
'2026-05-07T11:12:53+0800'
# but this will match
'2026-05-07T11:20:15+08:00'
```

I think the easier way to do this is to just take the 1st word of each line in the journal

```py
for line in journal_output.splitlines():
    words = line.strip().split()
    if len(words) < 1:
        continue
    ts = _parse_ts(words[0])
```

tomli380576 (changes requested):
the main thing is the timestamp regex, everything else looks ok

tomli380576 (approved):
lgtm ty!

PR #2486, opened 2026-04-20
Configure UnifiedRunner for ListBootstrapped (bugfix)

Hook25:
## Description

If not configured, the runner won't work in snaps as the user will be None and plz-run will use it and blow up. 

## Resolved issues

Fixes: CHECKBOX-2232

## Documentation

N/A

## Tests

To try this:
```
snap download checkbox24 --edge
unsquashfs checkbox24*.snap
snap try --devmode squashfs-root
checkbox25.checkbox list-bootstrapped sru
# boom
# now patch the subcommands.py
# it should now work as intended
```

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/subcommands.py
+++ b/checkbox-ng/checkbox_ng/launcher/subcommands.py
@@ -1513,6 +1513,10 @@ class ListBootstrapped:
     def sa(self):
         return self.ctx.sa
 
+    @lru_cache(maxsize=1)
+    def get_normal_user(self):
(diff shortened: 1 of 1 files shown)


pieqq (approved):
It works for me!™

It's too bad [the docstring in `start_new_session` is out of date](https://github.com/canonical/checkbox/blob/1acbe00d7ade8cace714c4aef05eeffb7e100cec/checkbox-ng/plainbox/impl/session/assistant.py#L484-L495), cause it's not easy to understand what is a customer runner nor the associated kw args that go with it.

PR #2487, opened 2026-04-21
Fix regex in benchmarks cpu/disk_on_idle (BugFix)

### LINKED ISSUE #2477: Invalid escape char in benchmarks/system/disk_on_idle and cpu_on_idle
### Bug Description

Image

added these two to the cert suite and notice when running on both noble and resolute that there are some warnings about invalid escape sequences.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

install checkbox
run benchmarks/system/disk_on_idle or benchmarks/system/disk_on_idle 
Note the "\S is an invalid escape sequence" warnings in the output

### Expected Result

No warnings during execution

### Actual Result

warnings seen in console output

### Environment

Resolute: $ checkbox-cli --version 7.1.0.dev90
Noble: $ checkbox-cli --version 7.1.0.dev100

### Relevant log output

```shell

```

### Additional context

_No response_

bladernr:
## Description
 * for disk_on_idle and cpu_on_idle, fix regex to resolve warning about invalid escape char in newer python3 versions. (older Python3 just silently passed it)
 * Also added CLAUDE.md to gitignore to avoid accidentally pulling in Claude instructions when making commits. Fixes #2477

## Resolved issues
Resolves #2477
Resolves CHECKBOX-2238

## Documentation

NA 
## Tests

Tested on a fresh Bionic install to verify rawstrings does work back that far:
```
u@bionic-tester:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04 LTS
Release:        18.04
Codename:       bionic
u@bionic-tester:~$ python3 --version
Python 3.6.5
u@bionic-tester:~$ iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall(r"idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
99.787 %
u@bionic-tester:~$ iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall(r"util\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
0.013000000000000001 %
```

For comparison, this is Resolute where the initial warnings appeared for me, showing before and after:
```
bladernr@resolute-test:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu Resolute Raccoon (development branch)
Release:        26.04
Codename:       resolute
bladernr@resolute-test:~$ python3 --version
Python 3.14.4
bladernr@resolute-test:~$ iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall("idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
<string>:1: SyntaxWarning: "\S" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\S"? A raw string is also an option.
98.978 %
bladernr@resolute-test:~$ iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall(r"idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
98.829 %
bladernr@resolute-test:~$ iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall("util\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
<string>:1: SyntaxWarning: "\S" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\S"? A raw string is also an option.
0.0 %
bladernr@resolute-test:~$ iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall(r"util\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
0.0 %
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 node_modules
 
 venv
+
+CLAUDE.md
--- a/providers/base/bin/iostat_benchmark.py
+++ b/providers/base/bin/iostat_benchmark.py
@@ -24,10 +24,7 @@ import sys
 
 
 def parse_iostat_column(output, column):
-    values = [
-        float(n)
-        for n in re.findall(rf"{column}\n.*?(\S+)\n", output)
-    ]
+    values = [float(n) for n in re.findall(rf"{column}\n.*?(\S+)\n", output)]
     if not values:
         print(
             f"ERROR: No '{column}' values found in iostat output",
@@ -48,10 +45,11 @@ def main():
         help="Which metric to report: 'cpu' (idle %%) or 'disk' (util %%)",
     )
     parser.add_argument(
-        "-t","--time",
+        "-t",
+        "--time",
         action="store",
         default=10,
-        help="Time in seconds to run iostat. (default: %(default)s)"
+        help="Time in seconds to run iostat. (default: %(default)s)",
     )
     args = parser.parse_args()
 
--- a/providers/base/bin/iostat_benchmark.py
+++ b/providers/base/bin/iostat_benchmark.py
@@ -60,8 +60,9 @@ def main():
     try:
         result = subprocess.run(
             ["iostat", "-x", "-m", "1", str(args.time)],
-            capture_output=True,
-            text=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
             check=True,
         )
     except subprocess.CalledProcessError as e:
--- a/providers/base/bin/iostat_benchmark.py
+++ b/providers/base/bin/iostat_benchmark.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+# This file is part of Checkbox.
+#
+# Copyright 2026 Canonical Ltd.
+# Written by:
+#   Jeff Lane <jeff@ubuntu.com>
+#
+# Checkbox is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3,
+# as published by the Free Software Foundation.
+#
+# Checkbox is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import re
+import subprocess
+import sys
+
+
+def parse_iostat_column(output, column):
+    values = [
+        float(n)
+        for n in re.findall(rf"{column}\n.*?(\S+)\n", output)
+    ]
+    if not values:
+        print(
+            f"ERROR: No '{column}' values found in iostat output",
+            file=sys.stderr,
+        )
+        return 1
+    print(f"{sum(values) / len(values):.2f}%")
+    return 0
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Measure average CPU or disk utilization from iostat."
+    )
+    parser.add_argument(
+        "metric",
+        choices=["cpu", "disk"],
+        help="Which metric to report: 'cpu' (idle %%) or 'disk' (util %%)",
+    )
+    parser.add_argument(
+        "-t","--time",
+        action="store",
+        default=10,
+        help="Time in seconds to run iostat. (default: %(default)s)"
+    )
+    args = parser.parse_args()
+
+    column = "idle" if args.metric == "cpu" else "util"
+
+    try:
+        result = subprocess.run(
+            ["iostat", "-x", "-m", "1", str(args.time)],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError as e:
+        print(f"ERROR: iostat failed: {e}", file=sys.stderr)
+        return 1
+
+    return parse_iostat_column(result.stdout, column)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/providers/base/bin/iostat_benchmark.py
+++ b/providers/base/bin/iostat_benchmark.py
@@ -24,14 +24,14 @@ import sys
 
 
 def parse_iostat_column(output, column):
-    values = [float(n) for n in re.findall(rf"{column}\n.*?(\S+)\n", output)]
+    values = [float(n) for n in re.findall(column + r"\n.*?(\S+)\n", output)]
     if not values:
         print(
-            f"ERROR: No '{column}' values found in iostat output",
+            "ERROR: No '{}' values found in iostat output".format(column),
             file=sys.stderr,
         )
         return 1
-    print(f"{sum(values) / len(values):.2f}%")
+    print("{:.2f}%".format(sum(values) / len(values)))
     return 0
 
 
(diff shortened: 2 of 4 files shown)


providers/base/units/benchmarks/jobs.pxu
```diff
@@ -88,7 +88,7 @@ category_id: com.canonical.plainbox::benchmarks
 id: benchmarks/system/cpu_on_idle
 estimated_duration: 10.0
 requires: package.name == 'sysstat'
-command: iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall("idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
+command: iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall(r"idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
```
fernando79513: I know this test is 12 years old, but I think we could use the chance to move it to Python.
This test is basically a python script wrapped in bash and I think it could benefit of being written in python and having some unit-tests around it.

fernando79513 (changes requested):

providers/base/units/benchmarks/jobs.pxu
bladernr: ok... it's now a python script with unit tests.

providers/base/bin/iostat_benchmark.py
```diff
@@ -0,0 +1,74 @@
...
+        result = subprocess.run(
+            ["iostat", "-x", "-m", "1", str(args.time)],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError as e:
+        print("ERROR: iostat failed: {}".format(e), file=sys.stderr)
+        return 1
```
fernando79513, line 68: No need to try/except if you are going to raise the error anyway.

.gitignore
```diff
@@ -1,3 +1,5 @@
 node_modules
 
-venv
\ No newline at end of file
+venv
+
+CLAUDE.md
```
fernando79513, line 5: There is already a CLAUDE.md file in checkbox

```suggestion
```

providers/base/bin/iostat_benchmark.py
```diff
@@ -0,0 +1,74 @@
...
+# along with Checkbox.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import re
+import subprocess
+import sys
+
+
+def parse_iostat_column(output, column):
+    values = [float(n) for n in re.findall(column + r"\n.*?(\S+)\n", output)]
```
fernando79513, line 27: I know this regex was here before, but it's only working because the value you are checking is the last one of the table:
```
avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           3.62    0.00    2.46    0.08    0.00   93.84
```

Also, what happens if there is more than one disk?
You are getting only the value of the first disk displayed on the list:
```
Device            r/s     rMB/s   rrqm/s  %rrqm r_await rareq-sz     w/s     wMB/s   wrqm/s  %wrqm w_await wareq-sz     d/s     dMB/s   drqm/s  %drqm d_await dareq-sz     f/s f_await  aqu-sz  %util
sda              0.00      0.00     0.00   0.00    0.00     0.00    0.50      0.00     0.25  33.33    4.00     8.00    0.00      0.00     0.00   0.00    0.00     0.00    0.00    0.00    0.00   0.20
sdb              0.00      0.00     0.00   0.00    0.00     0.00    0.50      0.00     0.25  33.33    4.00     8.00    0.00      0.00     0.00   0.00    0.00     0.00    0.00    0.00    0.00   0.21
```

Is that the intended behavior?

providers/base/bin/iostat_benchmark.py
```diff
@@ -0,0 +1,74 @@
...
+    column = "idle" if args.metric == "cpu" else "util"
+
+    try:
+        result = subprocess.run(
+            ["iostat", "-x", "-m", "1", str(args.time)],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            check=True,
+        )
```
fernando79513, line 65: If you are analyzing the output, it's simpler to use `subprocess.check_output`
```suggestion
        result = subprocess.check_output(
            ["iostat", "-x", "-m", "1", str(args.time)],
            universal_newlines=True,
        )
```

providers/base/bin/iostat_benchmark.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def parse_iostat_column(output, column):
+    values = [float(n) for n in re.findall(column + r"\n.*?(\S+)\n", output)]
+    if not values:
+        print(
+            "ERROR: No '{}' values found in iostat output".format(column),
+            file=sys.stderr,
+        )
+        return 1
```
fernando79513, line 33: Just raise a system exit. Don't print and return 1.

providers/base/bin/iostat_benchmark.py
```diff
@@ -0,0 +1,74 @@
...
+    print("{:.2f}%".format(sum(values) / len(values)))
+    return 0
```
fernando79513, line 35: No need to return 0

fernando79513 (commented):
First of all, thanks a lot for taking the time to translate this test to Python. I think it is much clearer what the test is doing now.

I am still not sure what the purpose of running this test is. AFAIK, it's just informative.

I have a couple of concerns with the previous implementation:
 - The "parse_iostat_column" function only seems to work because we are choosing to parse the last value of the row. If we wanted a different column, we would have to rewrite the functionality. At this point is just a "get_idle_value" or "get_util_value"
 - What happens if there is more than one disk on the device? How do we know which disk to check? I tested it  locally and I got something like this:
```
Device            r/s     rMB/s   rrqm/s  %rrqm r_await rareq-sz     w/s     wMB/s   wrqm/s  %wrqm w_await wareq-sz     d/s     dMB/s   drqm/s  %drqm d_await dareq-sz     f/s f_await  aqu-sz  %util
nvme0n1         42.76      2.47     4.08   8.71    0.12    59.09   19.88      1.41    23.14  53.79    3.09    72.61    0.00      0.00     0.00   0.00    0.00     0.00    0.72    0.35    0.07   0.79
sda              0.00      0.00     0.00   2.16    1.34    23.25    0.00      0.00     0.00  16.67    5.80     4.80    0.00      0.00     0.00   0.00    0.00     0.00    0.00    0.00    0.00   0.00
sdb              0.00      0.00     0.00  51.93    2.70    11.19    0.00      0.00     0.00   0.00    0.00     0.00    0.00      0.00     0.00   0.00    0.00     0.00    0.00    0.00    0.00   0.00
```

fernando79513:
@bladernr I think this issue is still relevant.
Please take a look at the comments so we can merge this PR.

Thanks in advance!

PR #2494, opened 2026-04-22
Fix pre-parsed certification status overrides (bugfix)

Hook25:
## Description

Code didn't correctly handle the yaml certification status overrides as now they are no longer a string blob but a list of overrides. This crashes Checkbox if the field is used in YAML units.

## Resolved issues

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/checkbox-ng/plainbox/impl/xparsers.py
+++ b/checkbox-ng/plainbox/impl/xparsers.py
@@ -456,6 +456,14 @@ class OverrideFieldList(Node):
             entries.append(FieldOverride.parse(line, lineno, col_offset))
         return OverrideFieldList(initial_lineno, col_offset, entries)
 
+    @staticmethod
+    def from_preparsed(override_lst):
+        return OverrideFieldList(
+            0,
+            0,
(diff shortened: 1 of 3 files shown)


checkbox-ng/plainbox/impl/xparsers.py
```diff
@@ -456,6 +456,14 @@ def parse(
             entries.append(FieldOverride.parse(line, lineno, col_offset))
         return OverrideFieldList(initial_lineno, col_offset, entries)
 
+    @staticmethod
+    def from_preparsed(override_lst):
+        return OverrideFieldList(
+            0,
+            0,
```
pieqq, line 463: Are these offsets used anywhere for yaml parsing?

pieqq (approved):
Looks OK, just a quick question.

checkbox-ng/plainbox/impl/xparsers.py
Hook25: no, and as far as I can see, they are unused also for pxu parsing

PR #2495, opened 2026-04-22
Jinja templates spaces break translator (bugfix)

Hook25:
## Description

Currently if a value in a translated field contains a jinja template with a space within the formatters, it is broken into multiple fields, making the translation wrong.

This "resolves" the issue by removing all spaces within the formatters, it is not a perfect solution, as one would probably prefer to threat them semantically, but the translation is best effort and to do the correct solution it is significantly harder (some fields may contain these formatters we didnt anticipate, the current small parser for stringable values doesn't support them etc.)

## Resolved issues

## Documentation

N/A

## Tests

Added tests for the new function. To verify, translate any unit that depends/after/etc. an id with a formmater with a space. It will be broken into multiple lines.

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/test_translator.py
+++ b/checkbox-ng/checkbox_ng/launcher/test_translator.py
@@ -26,6 +26,7 @@ from functools import wraps
 from checkbox_ng.launcher.translator import (
     split_comment,
     split_string_values,
+    no_space_formatters,
     Translator,
     commentable_value,
     CommentedError,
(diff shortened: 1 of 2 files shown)


Hook25:
Example of a broken unit:
```
id: pass-rerun-1
unit: job
summary: Test that passes only at the second try
flags: simple
depends:
  a{{ a }}
after: a{{ b }}
command:
  [ -f $PLAINBOX_SESSION_SHARE/will_pass.txt ] && exit 0
  echo "will pass next time" > $PLAINBOX_SESSION_SHARE/will_pass.txt
  exit 1
```

Wrongly translates to
```
id: pass-rerun-1
unit: job
summary: Test that passes only at the second try
flags:
  - simple
depends:
  - a{{
  - a
  - '}}'
after:
  - a{{
  - b
  - '}}'
command: |
   [ -f $PLAINBOX_SESSION_SHARE/will_pass.txt ] && exit 0
   echo "will pass next time" > $PLAINBOX_SESSION_SHARE/will_pass.txt
   exit 1
```

p-gentili (approved):
I feel you. 

Tested, it makes sense to play it like this, it works as expected.

PR #2496, opened 2026-04-22
Siblings handling wrong in YAML units (bugfix)

Hook25:
## Description

Siblings were wrongly assumed to be a string, therefore passed into the _leaf function that only handles strings. Also, the other assert was wrongly assuming the values passed in would only be single keys, but that is not true for siblings.

## Resolved issues

## Documentation

N/A

## Tests

Added unit tests


pieqq (approved):
Thanks for updating the tests.

PR #2501, opened 2026-04-23
Support json exporting for list bootstrapped (bugfix)

Hook25:
## Description

json exporting is super useful, as it can be piped easily into other tools like `jq` or loaded to do post processing without writing a parser.

## Resolved issues

N/A

## Documentation

Added to option help

## Tests

Test it via
```
checkbox-cli list-bootstrapped --format json sru
```

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/subcommands.py
+++ b/checkbox-ng/checkbox_ng/launcher/subcommands.py
@@ -1543,7 +1543,8 @@ class ListBootstrapped:
             help=_(
                 (
                     "output format, as passed to print function. "
-                    "Use '?' to list possible values"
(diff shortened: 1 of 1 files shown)


fernando79513 (approved):
Thanks for the explanation.

I still think mixing the custom syntax that accepts whatever and json in the same command is a bit weird.

If you had a typo in the json field, you will have a bunch of typos and no error, where usually you may expect a type error:
```
$ checkbox-cli list-bootstrapped --format jsn smoke
jsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsnjsn
```
But if we plan to get rid of this in the future, I think its okay

Issue #2503, opened 2026-04-24
Submission data changed after merge


hanhsuan:
### Bug Description

The `checkbox-cli merged-submissions` are used to merge multiple submissions into one for creating test report. However, it will change the data inside the submissions.

For example:
```
{
      merged                                  |              untouched
      "template_id": null				      |	      "template_id": "com.canonical.certification::firmware/f
      "template_id": null				      |	      "template_id": "com.canonical.certification::info/hdpar
      "template_id": null				      |	      "template_id": "com.canonical.certification::info/hdpar
      "template_id": null				      |	      "template_id": "com.canonical.certification::info/hdpar
      "name": "Attach the installer&amp;#39;s debug log for d |	      "name": "Attach the installer&#39;s debug log for diagn
      "template_id": null				      |	      "template_id": "com.canonical.certification::snapd/log-
      "template_id": null				      |	      "template_id": "com.canonical.certification::snapd/log-
      "template_id": null				      |	      "template_id": "com.canonical.certification::snapd/log-
      "template_id": null				      |	      "template_id": "com.canonical.certification::snapd/log-
							      >	  "description": "CE-QA-PC FC Auto Test",
      "version": "7.2.0.dev13"				      |	      "version": "7.1.0.dev100"
    "version": "7.2.0.dev13"				      |	    "version": "7.1.0.dev100"
      "name": "[DEPRECATED, use &amp;#39;os&amp;#39; instead] |	      "name": "[DEPRECATED, use &#39;os&#39; instead] Collect
      "template_id": null,				      |	      "template_id": "com.canonical.certification::benchmarks
      "template_id": null,				      |	      "template_id": "com.canonical.certification::benchmarks
      "template_id": null,				      |	      "template_id": "com.canonical.certification::benchmarks
      "template_id": null,				      |	      "template_id": "com.canonical.certification::benchmarks
      "template_id": null,				      |	      "template_id": "com.canonical.certification::benchmarks
      "template_id": null,				      |	      "template_id": "com.canonical.certification::benchmarks
      "template_id": null,				      |	      "template_id": "com.canonical.certification::bluetooth4
      "template_id": null,				      |	      "template_id": "com.canonical.certification::disk/apste
      "template_id": null,				      |	      "template_id": "com.canonical.certification::disk/apste
      "template_id": null,				      |	      "template_id": "com.canonical.certification::disk/read_
      "template_id": null,				      |	      "template_id": "com.canonical.certification::disk/read_
[...]

PR #2504, opened 2026-04-24
Fix merge reports submission parsing (bugfix)

pieqq:
## Description

In the JSON report, the template field is called `template_id`, but in a `JobDefinition` it's `template-id`, therefore when the script tries to populate a session, it loses the template information...

I've also modified a bit the MergeReports and MergeSubmissions classes so that they're easier to test, and added some unit tests.

## Resolved issues

- https://warthogs.atlassian.net/browse/CHECKBOX-2246
- #2503 

## Tests

- Unit test added
- Script tested by calling `merge-reports` and `merge-submissions` with a submission and making sure the JSON is the same.


Hook25 (approved):
thanks

PR #2507, opened 2026-04-27
add imx8mm camera test scenario data (New)

stanley31huang:
## Description
add imx8mm camera test scenario data, the ov5640 is the only camera we plan to support on i.MX8M mini platform.


rickwu666666 (approved):
LGTM

PR #2509, opened 2026-04-27
Split skip outcome into 4 possibilites (New)

pieqq:
## Description

Instead of having OUTCOME_SKIP for manually skipped jobs and OUTCOME_NOT_SUPPORTED for any job that is skipped for a specific reason (failed dependencies, failed resources, unmet manifest), 3 new outcomes are created:

- `OUTCOME_SKIPPED_DEPENDENCY`
- `OUTCOME_SKIPPED_RESOURCE`
- `OUTCOME_SKIPPED_MANIFEST`

`OUTCOME_NOT_SUPPORTED` is deprecated but kept for backward compatibility (when using `merge-submissions` with subs made with older versions of Checkbox, for instance).

Finally, `OUTCOME_SKIP` is renamed `OUTCOME_MANUAL_SKIP` to indicate its real nature.

Then the reason why the job was automatically skipped is indicated in the JSON report in the `skip_reason` section.

The HTML report is also updated to take these modifications into account.

**Important:** the "status" is still mapped to `skip` for all of these new outcomes, so that we don't break C3 processing for the moment.

For more information, see the original specification document [CR139](https://docs.google.com/document/d/1kPp2PB_X0ENH1wkBbjUF6Dci4ehDEl8z9g8u67tWeLA/edit?tab=t.0).

## Resolved issues

Fix CHECKBOX-1337

## Tests

- Tested manually by running a few test plans, including the smoke test plan ([submission.zip](https://github.com/user-attachments/files/27141117/submission_2026-04-27T20.27.43.163620.zip) that contains the HTML and the JSON reports)
- Unit tests adjusted
- New unit tests created

### DIFF
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -496,7 +496,9 @@ class RerunWidget(CategoryWidget):
         IJobResult.OUTCOME_SKIP: _("Manually Skipped Jobs"),
         IJobResult.OUTCOME_CRASH: _("Crashed Jobs"),
         IJobResult.OUTCOME_NOT_SUPPORTED: _("Jobs with failed dependencies"),
-        IJobResult.OUTCOME_SKIPPED_DEPENDENCY: _("Jobs with failed dependencies"),
+        IJobResult.OUTCOME_SKIPPED_DEPENDENCY: _(
+            "Jobs with failed dependencies"
+        ),
         IJobResult.OUTCOME_SKIPPED_RESOURCE: _("Jobs with unmet resources"),
         IJobResult.OUTCOME_SKIPPED_MANIFEST: _("Jobs with unmet manifest"),
     }
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -493,9 +493,12 @@ class RerunWidget(CategoryWidget):
 
     section_names = {
         IJobResult.OUTCOME_FAIL: _("Failed Jobs"),
-        IJobResult.OUTCOME_SKIP: _("Skipped Jobs"),
+        IJobResult.OUTCOME_SKIP: _("Manually Skipped Jobs"),
         IJobResult.OUTCOME_CRASH: _("Crashed Jobs"),
         IJobResult.OUTCOME_NOT_SUPPORTED: _("Jobs with failed dependencies"),
+        IJobResult.OUTCOME_SKIPPED_DEPENDENCY: _("Jobs with failed dependencies"),
+        IJobResult.OUTCOME_SKIPPED_RESOURCE: _("Jobs with unmet resources"),
+        IJobResult.OUTCOME_SKIPPED_MANIFEST: _("Jobs with unmet manifest"),
     }
 
     def __init__(self, node):
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -493,7 +493,7 @@ class RerunWidget(CategoryWidget):
 
     section_names = {
         IJobResult.OUTCOME_FAIL: _("Failed Jobs"),
-        IJobResult.OUTCOME_SKIP: _("Manually Skipped Jobs"),
+        IJobResult.OUTCOME_MANUAL_SKIP: _("Manually Skipped Jobs"),
         IJobResult.OUTCOME_CRASH: _("Crashed Jobs"),
         IJobResult.OUTCOME_NOT_SUPPORTED: _("Jobs with failed dependencies"),
         IJobResult.OUTCOME_SKIPPED_DEPENDENCY: _(
--- a/checkbox-ng/plainbox/impl/applogic.py
+++ b/checkbox-ng/plainbox/impl/applogic.py
@@ -75,7 +75,6 @@ def run_job_if_possible(session, runner, config, job, update=True, ui=None):
     return job_state, job_result
 
 
-
 def get_all_exporter_names():
     """
     Get the identifiers (names) of all the supported session state exporters.
--- a/checkbox-ng/plainbox/impl/applogic.py
+++ b/checkbox-ng/plainbox/impl/applogic.py
@@ -32,7 +32,6 @@ from plainbox.abc import IJobResult
 from plainbox.i18n import gettext as _
 from plainbox.impl.result import MemoryJobResult
 from plainbox.impl.result_utils import determine_outcome_and_skip_reason
-from plainbox.impl.secure import config
 from plainbox.impl.secure.qualifiers import select_units
 from plainbox.impl.session import SessionManager
 from plainbox.impl.session.jobs import InhibitionCause
--- a/checkbox-ng/plainbox/impl/applogic.py
+++ b/checkbox-ng/plainbox/impl/applogic.py
@@ -31,6 +31,7 @@ import os
 from plainbox.abc import IJobResult
 from plainbox.i18n import gettext as _
 from plainbox.impl.result import MemoryJobResult
+from plainbox.impl.result_utils import determine_outcome_and_skip_reason
 from plainbox.impl.secure import config
 from plainbox.impl.secure.qualifiers import select_units
 from plainbox.impl.session import SessionManager
@@ -58,27 +59,21 @@ def run_job_if_possible(session, runner, config, job, update=True, ui=None):
     if job_state.can_start():
         job_result = runner.run_job(job, job_state, config, ui)
     else:
-        # Set the outcome of jobs that cannot start to
-        # OUTCOME_NOT_SUPPORTED _except_ if any of the inhibitors point to
-        # a job with an OUTCOME_SKIP outcome, if that is the case mirror
-        # that outcome. This makes 'skip' stronger than 'not-supported'
-        outcome = IJobResult.OUTCOME_NOT_SUPPORTED
-        for inhibitor in job_state.readiness_inhibitor_list:
-            if inhibitor.cause != InhibitionCause.FAILED_DEP:
-                continue
-            related_job_state = session.job_state_map[inhibitor.related_job.id]
-            if related_job_state.result.outcome == IJobResult.OUTCOME_SKIP:
-                outcome = IJobResult.OUTCOME_SKIP
-        job_result = MemoryJobResult(
-            {
-                "outcome": outcome,
-                "comments": job_state.get_readiness_description(),
-            }
+        outcome, skip_reason = determine_outcome_and_skip_reason(
+            job_state, session.job_state_map
         )
+        result_data = {
+            "outcome": outcome,
+            "comments": job_state.get_readiness_description(),
+        }
+        if skip_reason:
+            result_data["skip_reason"] = skip_reason
+        job_result = MemoryJobResult(result_data)
     assert job_result is not None
     if update:
         session.update_job_result(job, job_result)
     return job_state, job_result
+
 
 
 def get_all_exporter_names():
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -150,14 +150,14 @@ class UnifiedRunner(IJobRunner):
                 )
             )
             return JobResultBuilder(
-                outcome=IJobResult.OUTCOME_SKIP,
+                outcome=IJobResult.OUTCOME_MANUAL_SKIP,
                 comments=_("Unsupported plugin type: {}".format(job.plugin)),
             ).get_result()
 
         # resource and attachment jobs are always run (even in dry runs)
         if self._dry_run and job.plugin not in ("resource", "attachment"):
             return JobResultBuilder(
-                outcome=IJobResult.OUTCOME_SKIP,
+                outcome=IJobResult.OUTCOME_MANUAL_SKIP,
                 comments=_("Job skipped in dry-run mode"),
             ).get_result()
 
--- a/checkbox-ng/plainbox/impl/execution.py
+++ b/checkbox-ng/plainbox/impl/execution.py
@@ -150,7 +150,7 @@ class UnifiedRunner(IJobRunner):
                 )
             )
             return JobResultBuilder(
-                outcome=IJobResult.OUTCOME_MANUAL_SKIP,
+                outcome=IJobResult.OUTCOME_CRASH,
                 comments=_("Unsupported plugin type: {}".format(job.plugin)),
             ).get_result()
 
--- a/checkbox-ng/plainbox/impl/exporter/__init__.py
+++ b/checkbox-ng/plainbox/impl/exporter/__init__.py
@@ -273,12 +273,6 @@ class SessionStateExporterBase(ISessionStateExporter):
                     "comments"
                 ] = job_state.result.comments
 
-            # Add skip_reason if present
-            if job_state.result.skip_reason:
-                data["result_map"][job_id][
-                    "skip_reason"
-                ] = job_state.result.skip_reason
-
             # Add Job hash if requested
             if self.OPTION_WITH_JOB_HASH in self._option_list:
                 data["result_map"][job_id]["hash"] = job_state.job.checksum
--- a/checkbox-ng/plainbox/impl/exporter/__init__.py
+++ b/checkbox-ng/plainbox/impl/exporter/__init__.py
@@ -275,7 +275,9 @@ class SessionStateExporterBase(ISessionStateExporter):
 
             # Add skip_reason if present
             if job_state.result.skip_reason:
-                data["result_map"][job_id]["skip_reason"] = job_state.result.skip_reason
+                data["result_map"][job_id][
+                    "skip_reason"
+                ] = job_state.result.skip_reason
 
             # Add Job hash if requested
             if self.OPTION_WITH_JOB_HASH in self._option_list:
--- a/checkbox-ng/plainbox/impl/exporter/__init__.py
+++ b/checkbox-ng/plainbox/impl/exporter/__init__.py
@@ -273,6 +273,10 @@ class SessionStateExporterBase(ISessionStateExporter):
                     "comments"
                 ] = job_state.result.comments
 
+            # Add skip_reason if present
+            if job_state.result.skip_reason:
+                data["result_map"][job_id]["skip_reason"] = job_state.result.skip_reason
+
             # Add Job hash if requested
             if self.OPTION_WITH_JOB_HASH in self._option_list:
                 data["result_map"][job_id]["hash"] = job_state.job.checksum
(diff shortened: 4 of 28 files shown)


checkbox-ng/plainbox/impl/exporter/__init__.py
```diff
@@ -273,6 +273,12 @@ def get_session_data_subset(self, session_manager):
                     "comments"
                 ] = job_state.result.comments
 
+            # Add skip_reason if present
+            if job_state.result.skip_reason:
+                data["result_map"][job_id][
+                    "skip_reason"
+                ] = job_state.result.skip_reason
+
```
pieqq: I've run some tests without these lines (running a test plan and exporting as text, HTML, tar.xz), and it seems to work. I'm not sure if these are needed at all (except maybe with the old XLS exporter?).

checkbox-ng/plainbox/impl/result_utils.py
Hook25, line 1: This is a bit of a weird architectural decision, but I can see why it was done. Probably this being a module on its own avoids bloating the SessionState with more unrelated crap. Also, I can see us moving other result deciding code (like post resume) here.

Commenting to ask your opinion explicitly, you put this here instead of into the session state/session assistant for that reason?

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,134 @@
...
+This module provides utilities for working with job results, particularly
+for determining skip outcomes and extracting skip reason information from
+job readiness inhibitors.
+"""
+
+from plainbox.abc import IJobResult
+from plainbox.impl.session.jobs import InhibitionCause
+
+
+def determine_outcome_and_skip_reason(job_state, job_state_map):
```
Hook25, line 34: It is a bit weird this doesn't handle fail on resource

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        A mapping of job ids to JobState objects (from session)
+    :returns:
+        A tuple of (outcome: str, skip_reason: dict or None)
+        where skip_reason is a dict with optional keys:
+        - related_dependencies: list of job ids that failed
+        - related_resources: list of resource expressions that failed
+        - related_manifests: list of manifest expressions that failed
+    """
+    if not job_state.readiness_inhibitor_list:
+        return IJobResult.OUTCOME_NOT_SUPPORTED, None
```
Hook25: If we get here isn't this a bug? This is specifically for skipped reasons and we have no readiness inhibitor.

Hook25: if it is a bug throw

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,134 @@
...
+                    related_job_state
+                    and related_job_state.result
+                    and related_job_state.result.outcome
+                    == IJobResult.OUTCOME_SKIP
+                ):
+                    outcome = IJobResult.OUTCOME_SKIP
+                else:
+                    skip_reason["related_dependencies"].append(
+                        inhibitor.related_job.id
+                    )
```
Hook25, line 95: Q: How do we get with a failed dep without related job situation? Can you comment on that continue?
```suggestion
            if not inhibitor.related_job:
                continue # explain how do we get here more or less
            related_job_state = job_state_map.get(inhibitor.related_job.id)
            if (
                related_job_state
                and related_job_state.result
                and related_job_state.result.outcome
                == IJobResult.OUTCOME_SKIP
            ):
                outcome = IJobResult.OUTCOME_SKIP 
                # can we break here? or do we need to collect the others as well?
            else:
                # we don't we report the related job id on non skip?
                skip_reason["related_dependencies"].append(
                    inhibitor.related_job.id
                )
```

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,134 @@
...
+                # Check if this is a manifest expression
+                if inhibitor.related_expression.manifest_id_list:
+                    has_failed_manifest = True
+                    skip_reason["related_manifests"].append(
+                        inhibitor.related_expression.text
+                    )
+                else:
+                    skip_reason["related_resources"].append(
+                        inhibitor.related_expression.text
+                    )
```
Hook25, line 108: This is wrong I think. The manifest_id_list is always defined if the expression has a manifest, so this would wrongly classify as manifest skipped jobs that have a true manifest but fail some other resource
```suggestion
            if not inhibitor.related_expression:
                continue # same as above, how do we get here?
            # Check if this is a manifest expression
            if inhibitor.related_expression.manifest_id_list:
                has_failed_manifest = True
                skip_reason["related_manifests"].append(
                    inhibitor.related_expression.text
                )
            else:
                skip_reason["related_resources"].append(
                    inhibitor.related_expression.text
                )
```

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        outcome = IJobResult.OUTCOME_SKIPPED_RESOURCE
+    else:
+        outcome = IJobResult.OUTCOME_NOT_SUPPORTED
+
+    # Return skip_reason only if we have any inhibitors to report
+    if (
+        skip_reason["related_dependencies"]
+        or skip_reason["related_resources"]
+        or skip_reason["related_manifests"]
+    ):
```
Hook25, line 133:
```suggestion
    if bool(skip_reason.values()):
```

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    # Determine outcome based on priority:
+    # FAILED_MANIFEST > FAILED_DEP > FAILED_RESOURCE
+    if has_failed_manifest:
+        outcome = IJobResult.OUTCOME_SKIPPED_MANIFEST
+    elif has_failed_dep:
+        outcome = IJobResult.OUTCOME_SKIPPED_DEPENDENCY
+    elif has_failed_resource:
+        outcome = IJobResult.OUTCOME_SKIPPED_RESOURCE
+    else:
+        outcome = IJobResult.OUTCOME_NOT_SUPPORTED
```
Hook25: How do we get here? In merge reports?

checkbox-ng/plainbox/test-data/html-exporter/without_certification_status.html
Hook25, line 1: not quite sure how this works to be honest, but should we handle the new states here?

checkbox-ng/plainbox/test-data/html-exporter/with_certification_blocker.html
Hook25, line 1: not quite sure how this works to be honest, but should we handle the new states here?

checkbox-ng/plainbox/impl/applogic.py
Hook25, line 51: (not sure if this function is still used)

this doesn't handle fail on resource?

checkbox-ng/plainbox/impl/session/remote_assistant.py
```diff
@@ -557,27 +558,25 @@ def run_job(self, job_id):
         job_state = self._sa.get_job_state(job_id)
 
         if not job_state.can_start():
-            outcome = IJobResult.OUTCOME_NOT_SUPPORTED
+            outcome, skip_reason = determine_outcome_and_skip_reason(
+                job_state, self._sa._context.state.job_state_map
+            )
+            # Override with FAIL for fail-on-resource flag
             for inhibitor in job_state.readiness_inhibitor_list:
```
Hook25: first best: lets handle this in determine_outcome_and_skip_reason
second best: piggy back from outcome,skip_reason? why re-iterate over readiness?

checkbox-ng/plainbox/impl/session/assistant.py
```diff
@@ -1647,28 +1648,24 @@ def run_job(
...
             ui.finished_running(job, job_state, builder.get_result())
         else:
-            # Set the outcome of jobs that cannot start to
-            # OUTCOME_NOT_SUPPORTED _except_ if any of the inhibitors point to
-            # a job with an OUTCOME_SKIP outcome, if that is the case mirror
-            # that outcome. This makes 'skip' stronger than 'not-supported'
-            outcome = IJobResult.OUTCOME_NOT_SUPPORTED
+            # Special handling for fail-on-resource flag
+            outcome = None
             for inhibitor in job_state.readiness_inhibitor_list:
```
Hook25: ditto

checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -493,9 +493,14 @@ class RerunWidget(CategoryWidget):
 
     section_names = {
         IJobResult.OUTCOME_FAIL: _("Failed Jobs"),
-        IJobResult.OUTCOME_SKIP: _("Skipped Jobs"),
+        IJobResult.OUTCOME_SKIP: _("Manually Skipped Jobs"),
```
Hook25: should we rename the outcome to OUTCOME_MANUAL_SKIP?

Hook25 (changes requested):
Not quite sure but I think I've found a couple of issues:
- Over reporting manifest skips
- missing html of new results?

checkbox-ng/plainbox/impl/result_utils.py
pieqq: Yes. At that point, the session state and session assistant modules are thousands of lines long, and I figured it would make sense to split this. I can move this function back into the session state if you think it's better.

checkbox-ng/plainbox/impl/result_utils.py
pieqq: I've done the following test:

1. Create this job:

```yaml
plugin: shell
id: smoke/requirement/manifest_only
requires:
  - manifest.has_audio_playback == 'True'
  - package.name == "unknown-package"
command: 'true'
purpose: blabla
summary: blabla
```

2. Set the manifest to `true`:

```json
# Edit /var/tmp/checkbox-ng/machine-manifest.json and make sure that it contains this:
{
  "com.canonical.certification::has_audio_playback": true
}
```

3. Run this job and check what's processed by the `determine_outcome_and_skip_reason()` function:

```
$ checkbox-cli run "smoke/requirement/manifest_only"

(...)

(Pdb) pp outcome
'skipped-resource'
(Pdb) pp skip_reason
{'related_dependencies': [],
 'related_manifests': [],
 'related_resources': ['package.name == "unknown-package"']}
```

- Manifest has no problem, so it's ignored (since it's `true` as required) **(OK)**
- I don't have a `uknown-package` package on my system, so that resource requirement fails, and this is reflected in the `skip_reason` **(OK)**

4. Modify the machine manifest to set `has_audio_playback` to `false`:

```json
# Edit /var/tmp/checkbox-ng/machine-manifest.json and make sure that it contains this:
{
  "com.canonical.certification::has_audio_playback": false
}
```

5. Re-run the job:

```
$ checkbox-cli run "smoke/requirement/manifest_only"

(...)

(Pdb) pp outcome
'skipped-manifest'
(Pdb) pp skip_reason
{'related_dependencies': [],
 'related_manifests': ["manifest.has_audio_playback == 'True'"],
 'related_resources': []}
```

- This time, the skip reason is because of a failed manifest requirement **(OK)**

So I think your statement

> The manifest_id_list is always defined if the expression has a manifest

is incorrect.

Or maybe I missed something in your reasoning?

checkbox-ng/checkbox_ng/urwid_ui.py
pieqq: Fixed in [4b328da](https://github.com/canonical/checkbox/pull/2509/commits/4b328dabe3e728a0f3b21feb245214c54827c20f)

checkbox-ng/plainbox/impl/session/remote_assistant.py
pieqq: Should be fixed in [f06bbea](https://github.com/canonical/checkbox/pull/2509/commits/f06bbeab9559cf08c80057406973389662bc0936)

checkbox-ng/plainbox/impl/session/assistant.py
pieqq: See [f06bbea](https://github.com/canonical/checkbox/pull/2509/commits/f06bbeab9559cf08c80057406973389662bc0936)

checkbox-ng/plainbox/impl/applogic.py
pieqq: It doesn't look like this function is called at all, except... for its own unit tests.

I guess I should remove it, but I don't know if I should do that in that PR. What do you think?

checkbox-ng/plainbox/impl/result_utils.py
pieqq: Fixed in [f06bbea](https://github.com/canonical/checkbox/pull/2509/commits/f06bbeab9559cf08c80057406973389662bc0936)

checkbox-ng/plainbox/impl/result_utils.py
pieqq: Removed in [039b63b](https://github.com/canonical/checkbox/pull/2509/commits/039b63b8331c373f3edc8672284eb3bcce0aa5c1), it should never be in a state where there are no inhibitor list.

checkbox-ng/plainbox/impl/result_utils.py
pieqq: I removed it in [039b63b](https://github.com/canonical/checkbox/pull/2509/commits/039b63b8331c373f3edc8672284eb3bcce0aa5c1).

checkbox-ng/plainbox/impl/result_utils.py
pieqq: Hm... I don't think this works the way you think it works :)

```python
In [1]:     skip_reason = {
   ...:         "related_dependencies": [],
   ...:         "related_resources": [],
   ...:         "related_manifests": [],
   ...:     }
   ...: 

In [2]: skip_reason
Out[2]: {'related_dependencies': [], 'related_resources': [], 'related_manifests': []}

In [3]: skip_reason.values()
Out[3]: dict_values([[], [], []])

In [4]: bool(skip_reason.values())
Out[4]: True
```

Hook25: sorry, `all`

checkbox-ng/plainbox/impl/applogic.py
Hook25: its the same, you can also leave it there and we will eventually re-discover it

checkbox-ng/plainbox/impl/result_utils.py
Hook25: no its fine

checkbox-ng/plainbox/impl/result_utils.py
Hook25: Interesting, I didnt expect this. What if they are both on the same line? do we get both then? I don't see how it could semantically separate them beside the fact that it is doing lines one by one

checkbox-ng/plainbox/impl/providers/exporters/data/multi-page.html
```diff
@@ -389,7 +389,7 @@ <h3>{{ cat_name }}<span class="ui-li-count {{ OUTCOME_METADATA_MAP[category_outc
     {%- endfor %}
     <p></p>
     {%- set has_any_not_supported = [] %}
-    {%- for job_id, job_state in job_state_map|dictsort if job_state.result.outcome == 'not-supported' and job_state.job.plugin not in ("resource", "attachment") %}
+    {%- for job_id, job_state in job_state_map|dictsort if job_state.result.outcome in ["not-supported", "skipped-dependency", "skipped-resource", "skipped-manifest"] and job_state.job.plugin not in ("resource", "attachment") %}
         {%- if has_any_not_supported.append(1) %}{% endif %}
```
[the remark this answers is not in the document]
pieqq: This is fine, as these outcome are still for jobs that are "not supported".

checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.html
```diff
@@ -344,7 +344,7 @@ <h3>{{ cat_name }}<span class="ui-li-count {{ OUTCOME_METADATA_MAP[category_outc
                     {%- endfor %}
                     <p></p>
                     {%- set has_any_not_supported = [] %}
-                    {%- for job_id, job_state in job_state_map|dictsort if job_state.result.outcome == 'not-supported' and job_state.job.plugin not in ("resource", "attachment") %}
+                    {%- for job_id, job_state in job_state_map|dictsort if job_state.result.outcome in ["not-supported", "skipped-dependency", "skipped-resource", "skipped-manifest"] and job_state.job.plugin not in ("resource", "attachment") %}
                         {%- if has_any_not_supported.append(1) %}{% endif %}
```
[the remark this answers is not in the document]
pieqq: This is fine as this is still for jobs that are not supported, so the explanation shown in the HTML report holds.

checkbox-ng/plainbox/impl/execution.py
```diff
@@ -150,14 +150,14 @@ def run_job(
...
                 comments=_("Unsupported plugin type: {}".format(job.plugin)),
             ).get_result()
 
         # resource and attachment jobs are always run (even in dry runs)
         if self._dry_run and job.plugin not in ("resource", "attachment"):
             return JobResultBuilder(
-                outcome=IJobResult.OUTCOME_SKIP,
+                outcome=IJobResult.OUTCOME_MANUAL_SKIP,
                 comments=_("Job skipped in dry-run mode"),
             ).get_result()
```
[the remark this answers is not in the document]
pieqq: Fixed in https://github.com/canonical/checkbox/pull/2509/commits/6b853e023177575ab90c92323f449dc1b6be8807

(I didn't touch anything for dry run since we don't use it)

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Return skip_reason only if we have any inhibitors to report
+    if (
+        skip_reason["related_dependencies"]
+        or skip_reason["related_resources"]
+        or skip_reason["related_manifests"]
+    ):
+        return outcome, skip_reason
+    else:
+        return outcome, None
```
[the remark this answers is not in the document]
pieqq: Fixed in https://github.com/canonical/checkbox/pull/2509/commits/a138ba46a0e666ab9a66a3e3167030302ee2de7f

checkbox-ng/plainbox/impl/result_utils.py
```diff
@@ -0,0 +1,138 @@
...
+                if inhibitor.related_expression.manifest_id_list:
+                    has_failed_manifest = True
+                    skip_reason["related_manifests"].append(
+                        inhibitor.related_expression.text
+                    )
+                else:
+                    skip_reason["related_resources"].append(
+                        inhibitor.related_expression.text
+                    )
+
```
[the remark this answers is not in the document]
pieqq: Fixed in https://github.com/canonical/checkbox/pull/2509/commits/a138ba46a0e666ab9a66a3e3167030302ee2de7f

Hook25 (dismissed):
Note, this changes outcome assignment of `fail-on-resource` slightly, and I don't think that is a bad idea but maybe we need to document it in a follow up. Before, `fail-on-resource` always marked the job as failed if the resource failed, now only if the resource failed not due to a manifest. I like the change but it may need some docs.

Hook25 (dismissed):

Hook25 (approved):

PR #2533, opened 2026-05-19
fix invalid escape sequence (BugFix)

### LINKED ISSUE #2532: Syntax Warning for invalid escape sequences
### Bug Description

`benchmarks/system/cpu_on_idle` and `benchmarks/system/disk_on_idle` are reporting a syntax error in modern Python version. 
To prevent future issues, we should consider using raw strings or double escaping the sequence.  

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Run any test plan that contains `benchmarks/system/cpu_on_idle` or `benchmarks/system/disk_on_idle`
2. Another option is to run directly the command for those tests. 
3. Syntax error will show

### Expected Result

No warnings on SyntaxErrors

### Actual Result

Syntax Error received on both test cases

### Environment

- OS: Resolute Racoon
- Kernel: 7.0.0-15-generic

### Relevant log output

```shell
<string>:1: SyntaxWarning: "\S" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\S"? A raw string is also an option.
99.959 %`


<string>:1: SyntaxWarning: "\S" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\S"? A raw string is also an option.
0.0 %
```

### Additional context

_No response_

rene-oromtz:
## Description
This is a minor PR to solve an invalid escape sequence on `benchmarks/system/cpu_on_idle` and `benchmarks/system/disk_on_idle` by using raw strings as suggested in https://docs.python.org/3/library/re.html#module-re

## Resolved issues
Resolves #2532 

## Tests
Tested locally on Noble

Before:
```
iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall("idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
<string>:1: SyntaxWarning: invalid escape sequence '\S'
89.652 %
```

After:
```
iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in re.findall(r"idle\n.*?(\S+)\n", lines)]; print(sum(l)/len(l),"%")'
85.881 %
```

### DIFF
--- a/providers/base/units/benchmarks/jobs.pxu
+++ b/providers/base/units/benchmarks/jobs.pxu
@@ -88,7 +88,7 @@ category_id: com.canonical.plainbox::benchmarks
 id: benchmarks/system/cpu_on_idle
 estimated_duration: 10.0
 requires: package.name == 'sysstat'
-command: iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in (re.findall("idle\n.*?(\S+)\n", lines))]; print(sum(l)/len(l),"%")'
+command: iostat -x -m 1 10 | python3 -c 'import sys, re; lines="".join(sys.stdin.readlines()); l=[float(n) for n in re.findall(r"idle\n.*?(\S+)\n", lines)]; print(sum(l)/len(l),"%")'
 _purpose: CPU utilization on an idle system.
 _summary: Measure CPU utilization on an idle system.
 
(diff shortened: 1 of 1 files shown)


fernando79513:
I think this is addressed in this PR:
https://github.com/canonical/checkbox/pull/2487/changes/c28b30cfa9c936b65c8d21f774bdc348f4d2115a

I will review it today.
Let me know if this fixes your issue.

rene-oromtz:
@fernando79513 Yes, that will do it! I swear I searched for a related bug but couldn't find any, maybe I was using a wrong keyword... Anyway, thanks for the heads up, I will close this PR as duplicate

PR #2534, opened 2026-05-20
fix: Update ethernet/multi_iperf3_nic_device job to avoid virtual devices (BugFix)

ajzobro:
## Description

Updated the filter used by the `ethernet/multi_iperf3_nic_device...` job to avoid attempting to do iperf3 testing of virtual ethernet devices, which are typically used for administration of servers.

As seen in https://certification.canonical.com/certificates/request/2605-17514/

## Resolved issues

See above.

## Documentation

N/A

## Tests

Not tested

### DIFF
--- a/providers/base/units/ethernet/jobs.pxu
+++ b/providers/base/units/ethernet/jobs.pxu
@@ -76,7 +76,7 @@ _purpose:
 
 unit: template
 template-resource: device
-template-filter: device.category == 'NETWORK' and device.interface != 'UNKNOWN'
+template-filter: device.category == 'NETWORK' and device.interface != 'UNKNOWN' and device.product_slug != 'Virtual_Ethernet_'
 plugin: shell
 category_id: com.canonical.plainbox::ethernet
 id: ethernet/multi_iperf3_nic_device{__index__}_{interface}


rodwsmith (commented):
This looks reasonable on its face, but I would like some checks that this change won't introduce any false positives on the `device.product_slug != 'Virtual_Ethernet_'` test -- that is, that it won't accidentally exclude something it shouldn't. Maybe write a script to pull down a bunch of submissions (both server and device; say, 100+ submissions of each), check to see if this condition would be met, and then hand-verify that it's all OK?

ajzobro:
Perhaps someone can enlighten me to a case where we do intentionally wish to certify against iperf3 with a virtual ethernet adapter?

rodwsmith:
> Perhaps someone can enlighten me to a case where we do intentionally wish to certify against iperf3 with a virtual ethernet adapter?

I know of no such case; however:

- Manufacturers sometimes throw us a curveball with some configuration that seems weird.
- My expertise is on the server side, which tends to be fairly uniform in configuration. On the devices side, manufacturers do all sorts of crazy things, and I may not be aware of all of them.
- Just because something is marked as a "virtual device" doesn't mean that's what it is. Just look at the DMI data, which is sometimes incorrect, as a precedent.

Of course, no amount of testing now will guarantee that nothing weird crops up in the future; but we can at least perform a due-diligence check to be sure that your proposed exclusion criterion, reasonable as it seems, will not produce large numbers of unintended exclusions today because of something weird in how devices are marked today.

PR #2538, opened 2026-05-22
Add template filter for the Renesas video codec test to prevent it from being run by other projects (Bugfix)

rickwu666666:
## Description
Add template filter for the Renesas video codec test to prevent it from being run by other projects.


stanley31huang (approved):
LGTM

PR #2539, opened 2026-05-22
Fix clocktest false failure on high CPU count systems (bugfix)

mreed8855:
On systems with many cores (e.g., 688), the cumulative time spent switching affinity and yielding exceeds MAX_JITTER (0.2s). This causes the test to misidentify loop overhead as clock skew.

This patch brackets the sampling loop and normalizes each CPU's timestamp against its relative position in the sequence. This isolates actual hardware jitter from the predictable delay of the measurement loop itself.

## Description

This patch brackets the sampling loop and normalizes each CPU's timestamp against its relative position in the sequence. This isolates actual hardware jitter from the predictable delay of the measurement loop itself.

## Resolved issues
https://bugs.launchpad.net/hp/+bug/2143999
This issue was seen on 26.04 and also on 24.04 but only on systems with high core counts.

### DIFF
--- a/providers/base/src/clocktest.c
+++ b/providers/base/src/clocktest.c
@@ -24,7 +24,9 @@
 int test_clock_jitter(){
     cpu_set_t cpumask;
     struct timespec *time;
-    unsigned long nsec;
+    struct timespec t_before, t_after;
+    unsigned long loop_nsec, expected_nsec;
+    long adjusted, min_adjusted, max_adjusted;
     unsigned slow_cpu, fast_cpu;
     double jitter;
     double largest_jitter = 0.0;
@@ -50,6 +52,7 @@ int test_clock_jitter(){
     }
 
     for (iter=0; iter<ITERATIONS; iter++){
+        clock_gettime(CLOCK_REALTIME, &t_before);
         for (cpu=0; cpu < num_cpus; cpu++) {
             CPU_ZERO(&cpumask); CPU_SET(cpu,&cpumask);
             if (setaffinity(cpumask) < 0){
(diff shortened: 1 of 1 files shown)


mreed8855:
HPE tested this.

Testing for clock jitter on 688 cpus
PASSED: largest jitter seen was 0.029482

Testing clock direction for 5 minutes...
PASSED: Iteration 0 delta: 0.000200
PASSED: Iteration 1 delta: 0.000158
PASSED: Iteration 2 delta: 0.000225
PASSED: Iteration 3 delta: 0.000175
PASSED: Iteration 4 delta: 0.000148
clock direction test: sleeptime 60 sec per iteration, failed iterations: 0

providers/base/src/clocktest.c
```diff
@@ -68,15 +71,24 @@ int test_clock_jitter(){
...
+          */
+        loop_nsec = NSEC(t_after) - NSEC(t_before);
         slow_cpu = fast_cpu = 0;
+        min_adjusted = max_adjusted = (long)(NSEC(time[0]) - NSEC(t_before));
         for (cpu=0; cpu < num_cpus; cpu++) {
-            nsec = NSEC(time[cpu]);
-            if (nsec < NSEC(time[slow_cpu])) { slow_cpu = cpu; }
-            if (nsec > NSEC(time[fast_cpu])) { fast_cpu = cpu; }
+            expected_nsec = NSEC(t_before) +
+                (unsigned long)((double)cpu / (num_cpus - 1) * loop_nsec);
```
ajzobro, line 86:
```suggestion
                (unsigned long)((double)cpu / num_cpus * loop_nsec);
```

This way, if there was 1 cpu, `expected_nsec = NSEC(t_before) + 0 / 1 * loop_nsec` = `NSEC(t_before)` and it won't crash due to divide by zero.

If there were 2 cpus:
 - expected_nsec for cpu0 = NSEC(t_before + 0 / 2 * loop_nsec` = `NSEC(t_before)`
 - expected_nsec for cpu1 = NSEC(t_before + 1 / 2 * loop_nsec` = `NSEC(t_before)` + half the loop time

Perhaps we intended instead to do:

```suggestion
                (unsigned long)((double)(cpu +1) / num_cpus * loop_nsec);
```

If there were 2 cpus:
 - expected_nsec for cpu0 = NSEC(t_before + 1 / 2 * loop_nsec` = `NSEC(t_before)` + half the loop time
 - expected_nsec for cpu1 = NSEC(t_before + 2 / 2 * loop_nsec` = `NSEC(t_before)` + full loop time

It seems like we're trying to take the fractional portion of the setup loop time and account for it properly for each cpu.

bladernr: Isn't the single CPU use case already covered here though/? This just exits 0 if there is only a single CPU, and errors if there are NO CPUs detected for some reason.

```
 36     num_cpus = sysconf(_SC_NPROCESSORS_ONLN);
 37     if (num_cpus == 1) {
 38         printf("Single CPU detected. No clock jitter testing necessary.\n");
 39         return 0;
 40     }
 41     else if (num_cpus < 1) {
 42         printf("Get number of cpu failed.\n");
 43         return 1;
 44     }

```

mreed8855: As stated in  the previous comment the num_cpus = 1 is handled in the code  from lines 36 - 40.   The dividing by zero line will never be touched.

mreed8855: The test finds the world clock time for each cpu loop_nsec measures the length of time it took to get all of the times and store them
 t_before is the start of taking the time and t_after is the time after the loop is finished

  loop_nsec = NSEC(t_after) - NSEC(t_before);

For systems with many cpu's the time read at the zero cpu could differ more than .02 than the last cpu like the 688th one
This equation tries to take into account that difference in start time by essentially interpolating the value for the cpu.

 expected_nsec = NSEC(t_before) +
            (unsigned long)((double)cpu / (num_cpus - 1) * loop_nsec);

Take the 10 cpu example where we have 10 cpu's numbered  cpu 0 - 9

cpu / (num_cpus - 1) equals the cpu's position in the line from 0.0 - 1.0

cpu 0 will be at t_before because it is right when the time measurements starts so there is no delay to account for and (unsigned long)((double)cpu / (num_cpus - 1) * loop_nsec) will be zero for cpu 0

cpu's 1 - 9 will be accounted for as we progress starting at cpu = 1 as it the first time measurement that could experience a delay

(unsigned long)((double)cpu / (num_cpus - 1) * loop_nsec)

  cpu 0   t_before
  cpu 1   1/9
  cpu 2   2/9
  cpu 3   3/9
 ...
  cpu 9  9/9 

The tenth cpu which is cpu 9 is essentially the full length of the duration to find and store the time values and provide the full delay needed to take into account the measurement of the last time for the final cpu.

ajzobro: I finally understand:

We are saying that there are 9 times-between-measurements:

```
   0   1   2   3
   |---|---|---|
    \---------/
3 setup times for 4 cpus
```

ajzobro (approved):
LGTM, I finally understand the math, thank you for bearing with me.

PR #2540, opened 2026-05-25
Rename after-suspend-audio/after-suspend-audio-occupied (BugFix)

hanhsuan:
## Description

Rename `after-suspend-audio/after-suspend-audio-occupied` because it is ignored by the Checkbox docs generator. This change ensures the test case appears correctly in the documentation.

## Tests

https://certification.canonical.com/hardware/202001-27667/submission/489829/


Hook25 (approved):
Thanks

PR #2546, opened 2026-05-26
update camera supported format for imx8mm (BugFix)

stanley31huang:
## Description
update camera supported format for imx8m mini EVK, as there's no ISP on the i.MX8M mini EVK board, so the NV12 and RGB color format is not supported.


rickwu666666 (dismissed):
LGTM

rickwu666666 (approved):
LGTM

PR #2551, opened 2026-05-27
Use existing job result if session resumed after all jobs have run (bugfix)

### LINKED ISSUE #2535: If the last test case is user-interact, it will not save status correctly
### Bug Description

If the last test case is user-interact and interrupted the final TUI, it will not save status correctly.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

The steps is in the attached video file.

### Expected Result

The finished test case status should be saved correctly.

### Actual Result

The last one test case status cannot be saved correctly.

### Environment

- OS: 24.04
- Checkbox Type: Debian
- Checkbox Version: 7.2.0.dev22

### Relevant log output

```shell

```

### Additional context

https://github.com/user-attachments/assets/4e33587d-ea92-41ba-8575-885b9e8d8dc1

pieqq:
## Description

If the session is interrupted and resumed when the final job is an interactive one, Checkbox would force the user to set the job outcome again, even if this job has been already run.

To prevent this, a new boolean attribute, `remaining_todo_jobs`, is added to the `SessionMetaData` and is used in:

- the UI to not show the submenu in that case
- the session resume process to not modify the job result and use the existing result

This allows to just continue to the re-run screen directly (or to the finalizing session process if there are no candidates for re-runs).

Since the session object is modified, a new version (version 11) is made available in `suspend.py` and `resume.py` that stores and restores the new `remaining_todo_jobs` attribute.

## Resolved issues

Fix #2535
Fix CHECKBOX-2262

## Tests

Tested successfully both in checkbox local and checkbox remote.

Steps followed:

1. Start a session and select the Smoke test plan
2. When the manual job (which is the last job in this test plan) is displayed, stop the session (`Ctrl-C` on local, and `Ctrl-C` then `Exit and stop the Checkbox agent on the testbed` on remote)
3. Restart Checkbox and resume the session by pressing `R` → when the session is highlighted, you can see "Are there still jobs to run? Yes". Select the session by pressing `Enter` and press `R` to re-run the last job (`smoke/manual`)
4. Set the job to `pass` → the re-run screen is shown
5. Stop the session again (like step 2), then resume it again (like in step 3) → This time, "Are there still jobs to run?" is "No". Select this session by pressing `Enter` → this time, no action is shown, and Checkbox resumes directly in the re-run screen.

***Note:*** at any point, you can check the content of the session metadata by going into the session directory in `/var/tmp/checkbox-ng/sessions/<session id>/` and checking the content of the `session` file which is a gzipped JSON doc: `zcat session | jq ".session.metadata"`

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -514,6 +514,10 @@ class RemoteController(ReportsStage, MainLoopStage):
 
     def _resume_session(self, resume_params):
         metadata = self.sa.prepare_resume_session(resume_params.session_id)
+        # If there are no more jobs to run, resume. The empty dict is sent
+        # so that the last job result is not modified nor re-run.
+        if not metadata.remaining_todo_jobs:
+            return self.resume_by_id(resume_params.session_id, {})
         if "testplanless" not in metadata.flags:
             app_blob = json.loads(metadata.app_blob.decode("UTF-8"))
             test_plan_id = app_blob["testplan_id"]
@@ -626,6 +630,7 @@ class RemoteController(ReportsStage, MainLoopStage):
             (
                 candidate.id,
                 generate_resume_candidate_description(candidate),
+                candidate.metadata.remaining_todo_jobs,
             )
             for candidate in resumable_sessions
         ]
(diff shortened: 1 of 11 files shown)


Hook25 (dismissed):
+1, thanks for the metabox test, works for me

Hook25 (approved):

PR #2552, opened 2026-05-28
Core26 daily builds (infra)

Hook25:
## Description

This adds the core26 runtime snap and the daily builds for it. This will need five followups:
1. To remove the lxd workaround once snapcraft9 is stable
2. To change the build base and to stable
3. To change the promotion workflow for both beta and stable (right now we won't do it)
4. Frontend snaps once we get the channels
5. riscv builds, this is pending actual support from the action

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-2178

## Documentation

Differences from core24 snap:
1. build base/name/python version
2. SNAPCRAFT -> CRAFT + snapcraftctl -> craftctl: names are now deprecated, this is the new name
3. rpi-support-binaries removed: this part installs a library that is long deprecated that is used for camera compatibility on the ancient video stack, it was completely removed in resolute
4. stage package gir1.2* removed, package is long deprecated and now dead, if tests depend on it (which I couldn't figure out) we are going to figure out a replacement
6. stage package: glmark2 + glmark2-es2 removed, packages are x11 specific, the wayland variant remains
7. execstack hack, probably could be removed in 24 as well, execstack is no longer in the repos and the upload is not blocked, so it is not needed anymore

This also lands a change in the core24 recipe: SNAPCRAFT -> CRAFT as the old name was deprecated but not removed and it remained in the yaml due to sloppy `:%s` (forgot a `/g`)

To double check my diff: (surprisingly readable)
```
diff series24/snap/snapcraft.yaml series26/snap/snapcraft.yaml
```

## Tests

The snap is up in the store right now built using this pipeline

### DIFF
--- a/.github/workflows/checkbox-daily-native-builds.yaml
+++ b/.github/workflows/checkbox-daily-native-builds.yaml
@@ -59,6 +59,7 @@ jobs:
           cd checkbox-core-snap/
           sudo apt update && sudo apt install -qq -y python3-setuptools-scm
           ./prepare.sh series${{ matrix.release }}
+          sudo snap refresh lxd --channel=6/beta
 
       - id: snap_build
         uses: Wandalen/wretry.action@71a909ebf09f3ffdc6f42a17bd54ecb43481da49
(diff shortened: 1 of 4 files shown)


pieqq (dismissed):
Looks sensible based on the description provided.

There is not really a way for me to test this locally, but since it managed to push a version in the store, it's good!

pieqq (approved):

PR #2554, opened 2026-05-28
Store resource jobs results in the results section of the JSON report (new)

pieqq:
## Description

Instead of storing the resource jobs results in their own `resource-results` section of the JSON report, they are now stored alongside the other jobs in the `results` section.

The reason is that a failing resource job may have very bad consequences on the test coverage of a test run, but as it stands now, this gets completely ignored because none of our tooling (e.g. C3) actually process the `resource-results` section of the JSON report.

This commit fixes this, and should help identifying problems during enablement.

Note: It's important to note that the `plugin` field of each result allows to check what kind of job is running, so it's very easy to specifically retrieve the resource jobs in the `results` section, since their `plugin` field is set to `resource`.

Note 2: The HTML report is left untouched because it already expose the resource jobs in a dedicated section at the bottom of the page, and this section is clearly marked `failed` if any of the resource jobs is failing:

image

## Resolved issues

C3-1109

## Documentation

JSON schema has been modified accordingly.

## Tests

Tested manually by running a test plan that contains resource jobs. They are now part of the `results` section.


Hook25 (approved):
thanks. Can you please guestimate the blast radius for SRU?

Hook25 (changes requested):
Pending a significant lowering of the failure rate, we can't do this for now.

pieqq:
@Hook25 as discussed this morning, the associated failures in resource jobs have all been fixed (the ones listed in https://warthogs.atlassian.net/browse/C3-814), so we should be able to land this now.

Hook25 (dismissed):

Hook25 (approved):

PR #2577, opened 2026-06-11
Stress-ng tests bugfixes fix user blacklist (new)

### LINKED ISSUE #2529: stress-ng changed behaviour in V0.21.0 and several stressors are now failing
### Bug Description

Since v0.21.0 of stress-ng, the way memory allocation works has been changed: https://github.com/ColinIanKing/stress-ng/commit/3f601ba67210ca1841707e78f77895a3db07b8e1

Instead of allocating a fixed amount of memory, the memory will be dynamically allocated depending on how many stressors are being run.

There are many stressors that we were trying to run using checkbox that never actually ran because they were being skipped. Before V0.21.0 of stress-ng, it would simply print a message that the stressors were being skipped but still return 0 as if it succeeded:
```
$ stress-ng --apparmor 0 --timeout 30 --oom-avoid-bytes 10% --skip-silent --verbose
stress-ng: debug: [39291] invoked with 'stress-ng --apparmor 0 --timeout 30 --oom-avoid-bytes 10% --skip-silent --verbose' by user 1000 'ce-certification-qa'
stress-ng: debug: [39291] stress-ng 0.20.01
stress-ng: debug: [39291] system: Linux ubuntu 5.15.0-1046-nvidia-tegra-igx #46-Ubuntu SMP Thu Apr 2 23:24:52 UTC 2026 aarch64, gcc 11.4.0, glibc 2.35, little endian
stress-ng: debug: [39291] RAM total: 30.0G, RAM free: 23.5G, swap free: 0.0
stress-ng: debug: [39291] temporary file path: '/home/ce-certification-qa', filesystem type: ext4 (47868695 blocks available, /dev/dm-0)
stress-ng: debug: [39291] CPUs have 3 idle states: BUSY, WFI, c7
stress-ng: debug: [39291] 12 processors online, 12 processors configured
stress-ng: info: [39291] setting to a 30 secs run per stressor
stress-ng: debug: [39291] CPU data cache: L1: 64K, L2: 256K, L3: 2048K
stress-ng: debug: [39291] shared cache allocate: shared cache buffer size: 2048K
stress-ng: info: [39291] dispatching hogs:
stress-ng: debug: [39291] starting stressor
stress-ng: debug: [39291] 0 stressors started
stress-ng: debug: [39291] metrics-check: no stressors run
stress-ng: info: [39291] skipped: 12: apparmor (12)
stress-ng: info: [39291] passed: 0
stress-ng: info: [39291] failed: 0
stress-ng: info: [39291] metrics untrustworthy: 0
stress-ng: info: [39291] successful run completed in 0 secs
```

Now however, stress-ng will allocate no memory if we only attempt to be running stressors that are being skipped and therefore the call to `mmap` will be with a size of 0 and the program will exit when trying to check if any memory was actually allocated.

I investigated every stress-ng failure that was reported in our testing and identified why they were failing. Most of the time it was simply because they weren't being run correctly. Most of the failures are because the stressors need superuser permissions to run, others also need to be run with the `--pathological` option because they are known to crash the host systems. Others are also not being run because of a missing kernel configuration or the wrong CPU architecture.

Here is a list of the stressors I investigated and how they should be run on our systems as well as the reason they still don't succeed once they are run with the correct arguments. There might be other stressors that fail on other systems with this recent stress-ng change:

* apparmor: sudo -> causes kernel crash in our kernel as well as generic jammy kernel and potentially newer kernels too (not tested yet)
* bad-ioctl: sudo and --pathological
* bind-mount: sudo and --pathological
* binderfs: sudo -> not supported
* cgroup: sudo
* chroot: sudo
* cpu-online: sudo and --pathological 
* dfp: sudo -> not implemented, kernel needs to be built with decimal _Decimal32, _Decimal64 or _Decimal128 support
* fanotify: sudo
* icmp-flood: sudo
* idle-page: sudo
* ioport: sudo -> not implemented, need x86
* ipsec-mb: sudo -> not x86
* klog: sudo
* loop: sudo
* lsm: sudo → not implemented, kernel needs to be built with linux/lsm.h or lsm_list_modules or lsm_get_self_attr system calls
* memhotplug: sudo
* mlockmany: sudo and --pathological
* module: sudo
* mseal: sudo → mseal system call not supported
* netlink-proc: sudo
* netlink-task: sudo
* oom-pipe: sudo and --pathological
* physmmap: sudo → fails because can’t access /dev/mem, maybe because of CONFIG_STRICT_DEVMEM which is set in kernel config
* physpage: sudo
* plugin: sudo and no plugin so specified
* quota: sudo → skipped because devices are not available, unknown reason
* ramfs: sudo
* rawpkt: sudo
* rawsock: sudo
* rawudp: sudo
* rdrand: sudo → CPU does not support the rdrand instruction
* seccomp: sudo
* secretmem: sudo → memfd_secret system call is not supported
* smi: sudo → kernel built for non-x86 target without sys/io.h or ioperm() or out op-code
* softlockup: sudo
* statmount: sudo → kernel built without Linux statmount or listmount
* swap: sudo
* sysinval: sudo and --pathological
* tsc: sudo → CPU does not support the rdtsc instruction
* tun: sudo
* umount: sudo
* uprobe: sudo
* userfaultfd: sudo 
* usersyscall: sudo -> prctl user dispatch is not working, skipping the stressor
* watchdog: sudo and --pathological
* x86cpuid: sudo → kernel built without x86 cpuid instruction support
* x86syscall: sudo → not x86

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Install checkbox22 and checkbox-ce-oem
sudo snap install checkbox22 --beta
sudo snap install checkbox-ce-oem --channel=uc22/edge --devmode
sudo snap stop checkbox-ce-oem.remote-slave --disable

2. checkbox control <DUTIP> , select "com.canonical.certification::stress-ng-automated" test plan, and run it.

### Expected Result

supported stress-ng jobs should be run correctly and unsupported ones skipped

### Actual Result

some unsupported stress-ng jobs are being run and others are run without the correct arguments which thus leads to failures.

### Environment

- OS: UC22 and Ubuntu Classic 22.04
- Checkbox type: snap
- version: every version that contains stress-ng v0.21.0 or later

### Relevant log output

```shell

```

### Additional context

_No response_

Hook25:
## Description

Given that now stress-ng doesn't silently ignore tests that it can't run anymore, we have discovered that quite a few of the tests that we were supposed to run were not actually running. This is manly due to three things:
1. The test needs super user permission
2. Stress-ng wants a `--pathological` flag to run it (because it may hang the machine)
3. The test is x86-only

Additionally we may need to blacklist some tests because they are not supported on the current platform.

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/2529

## Documentation

N/A

## Tests

This is the list bootstrapped before and after:
- [after_patch.json](https://github.com/user-attachments/files/28842888/after_patch.json)
- [after_ids.txt](https://github.com/user-attachments/files/28842896/after_ids.txt)
- [pre_ids.txt](https://github.com/user-attachments/files/28842903/pre_ids.txt)
- [pre_patch.json](https://github.com/user-attachments/files/28842905/pre_patch.json)


providers/base/bin/capture-stressor.py
```diff
@@ -1,31 +1,119 @@
...
+stressor_attributes["x86cpuid"]["root"] = True
+stressor_attributes["x86syscall"]["root"] = True
+
+stressor_attributes["bad-ioctl"].update(root=True, pathological=True)
+stressor_attributes["bind-mount"].update(root=True, pathological=True)
+stressor_attributes["cpu-online"].update(root=True, pathological=True)
+stressor_attributes["mlockmany"].update(root=True, pathological=True)
+stressor_attributes["oom-pipe"].update(root=True, pathological=True)
+stressor_attributes["sysinval"].update(root=True, pathological=True)
+stressor_attributes["watchdog"].update(root=True, pathological=True)
```
pieqq: This looks like maintenance hell, but I suppose there is no better option?

Maybe we could have a list of stressors to iterate over and set their values to `True`?

providers/base/bin/capture_stressor.py
```diff
@@ -1,31 +1,119 @@
...
+            output = subprocess.run(
+                cmd_get_stressors,
+                capture_output=True,
+                universal_newlines=True,
+                check=True,
+            ).stdout
+        except Exception as err:
+            raise SystemExit(err)
+
+        class_stressor_list = output.split(": ", 1)[1].split()
```
pieqq, line 112:
```suggestion
        # the `stress-ng --class` command returns something like this:
        # class 'cpu' stressors: af-alg atomic bitonicsort ...
        class_stressor_list = output.split(": ", 1)[1].split()

```

providers/base/units/stress/stress-ng.pxu
```diff
@@ -1,40 +1,84 @@
...
-template-resource: stress-ng-stressors
 template-unit: job
+template-resource: stress_ng_stressors
+template-filter: stress_ng_stressors.root == "False"
 id: stress/stress-ng-test-for-stressor-{stressor}
-template-id: stress/stress-ng-test-for-stressor-stressor
+template-id: stress/stress-ng-test-for-stressor-stressor-user
 category_id: com.canonical.plainbox::stress
-_summary: Run the stress-ng for stressor {stressor}
+_summary: Run the stress-ng stressor {stressor}
```
pieqq:
```suggestion
_summary: Run the stress-ng stressor {stressor} (as normal user)
```

providers/base/units/stress/stress-ng.pxu
```diff
@@ -1,40 +1,84 @@
...
+  esac
+
+unit: template
+template-unit: job
+template-resource: stress_ng_stressors
+template-filter: stress_ng_stressors.root == "True"
+id: stress/stress-ng-test-for-stressor-{stressor}
+template-id: stress/stress-ng-test-for-stressor-stressor-root
+category_id: com.canonical.plainbox::stress
+_summary: Run the stress-ng stressor {stressor}
```
pieqq:
```suggestion
_summary: Run the stress-ng stressor {stressor} (as root)
```

providers/base/bin/capture_stressor.py
```diff
@@ -1,31 +1,119 @@
...
+if platform.machine().lower() not in ["x86_64", "amd64"]:
+    # stressors are x86 specific
+    stressor_attributes["ioport"]["unsupported"] = True
+    stressor_attributes["ipsec-mb"]["unsupported"] = True
+    stressor_attributes["x86cpuid"]["unsupported"] = True
+    stressor_attributes["x86syscall"]["unsupported"] = True
+
+
+def main():
+    extra_unsupported = os.getenv("STRESS_NG_EXTRA_UNSUPPORTED", "").split()
```
pieqq, line 88:
```suggestion
def main():
    # STRESS_NG_EXTRA_UNSUPPORTED is a space-separated list of stressors to
    # skip, for instance "ioport x86cpuid x86syscall"
    extra_unsupported = os.getenv("STRESS_NG_EXTRA_UNSUPPORTED", "").split()
```

pieqq (changes requested):

providers/base/bin/capture-stressor.py
Hook25: note that this is not a comprehensive list of stressors, this is a list of overrides. The default state of every stressor is non-root, non-pathological and supported. This list merely overwrites some of these states for some of the available stressors

pieqq: Yes, that I had understood :) I'm just afraid the list will grow longer and/or will fall out of maintenance until we realize, 2 years later, that another stressor required root (but I hope the newer versions of stress-ng will `return 1` and not just silently ignore the issue).

Also, could we have something like that instead?

```python
stressors_requiring_root = ["apparmor", "binderfs", "cgroup", ...]
for stressor in stressors_requiring_root:
    stressor_attributes[stressor]["root"] = True
```

pieqq (approved):

PR #2583, opened 2026-06-17
Fix bug in the environment resource (bugfix)

Hook25:
## Description

The environent resource is bugged and when a envvar contains a `:` it is wrongfully truncated. Similarly if a envvar contains a `\n`, it breaks the resource output.

This fixes the bugs while also translating the script to python and testing it

## Resolved issues

Fixes: CHECKBOX-2280

## Documentation

N/A

## Tests

Adds unit tests. To reproduce the issue run this:
```
BUG="value:with :" checkbox-cli run "environment" | grep BUG
```

Before:
```
BUG: value
```

After:
```
BUG: value:with :
```

### DIFF
--- a/providers/resource/bin/environment_resource.py
+++ b/providers/resource/bin/environment_resource.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python3
+import os
+
+
+def no_newlines(s: str):
+    # newlines break resource output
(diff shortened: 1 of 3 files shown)


p-gentili (approved):
Thanks for taking care of this so quickly.

PR #2587, opened 2026-06-18
Translate executable resource to python (bugfix)

Hook25:
## Description

Similarly to environment, I don't even whant to know what is bugged in the pile of bash. We rely on this resource heavilly, it must be reliable.

## Resolved issues

N/A

## Documentation

N/A

## Tests

Added unittests, compared output, they match

### DIFF
--- a/providers/resource/bin/executable_resource.py
+++ b/providers/resource/bin/executable_resource.py
@@ -7,8 +7,7 @@ from contextlib import suppress
 
 def iter_if_accessible(path: Path):
     with suppress(OSError):
-        return path.iterdir()
-    return iter([])
+        yield from path.iterdir()
 
 
 def main():
--- a/providers/resource/bin/executable_resource.py
+++ b/providers/resource/bin/executable_resource.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python3
+
+import os
+from pathlib import Path
+from contextlib import suppress
+
+
+def iter_if_accessible(path: Path):
+    with suppress(OSError):
+        return path.iterdir()
+    return iter([])
+
+
+def main():
+    paths = {Path(x).resolve() for x in os.get_exec_path()}
+    executables = set()
+    for path in paths:
+        all_files = filter(Path.is_file, iter_if_accessible(path))
+        executables |= {
+            f.name for f in all_files if os.access(str(f), os.X_OK)
+        }
+    for executable in sorted(list(executables), key=str.casefold):
+        print("name:", executable)
+        print()
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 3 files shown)


providers/resource/tests/test_executable_resource.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertEqual(result, [Path("/usr/bin/ls"), Path("/usr/bin/cat")])
+
+    @patch("executable_resource.Path.iterdir")
+    def test_iterdir_oserror(self, iterdir_mock):
+        iterdir_mock.side_effect = OSError("Permission denied")
+        result = list(
+            executable_resource.iter_if_accessible(Path("/no/access"))
+        )
+        self.assertEqual(result, [])
+
```
fernando79513, line 40: This should catch the issue
```suggestion

    @patch("executable_resource.Path.iterdir")
    def test_iterdir_oserror_while_consuming(self, iterdir_mock):
        class FailingIterator:
            def __iter__(self):
                return self

            def __next__(self):
                raise OSError("Permission denied")

        iterdir_mock.return_value = FailingIterator()
        result = list(
            executable_resource.iter_if_accessible(Path("/no/access"))
        )
        self.assertEqual(result, [])

```

providers/resource/bin/executable_resource.py
```diff
@@ -0,0 +1,28 @@
...
+
+import os
+from pathlib import Path
+from contextlib import suppress
+
+
+def iter_if_accessible(path: Path):
+    with suppress(OSError):
+        return path.iterdir()
+    return iter([])
```
fernando79513: I learnt a lot with this PR.
The first thing I learnt is that my PATH was broken since I restored my laptop from the broken PCB.
The second thing I learnt is that `path.iterdir()` is lazy. Since it returns an iterator object, if the creation does not fail immediately, the suppressed error is not caught.
At that later point, execution is already outside the with suppress(OSError) block, so the FileNotFoundError escapes.

```suggestion
def iter_if_accessible(path: Path):
    with suppress(OSError):
        yield from path.iterdir()
```
This turns `iter_if_accessible()` into a generator, and the `suppress(...)` block stays active while the directory iterator is being consumed.

fernando79513 (changes requested):
I think this is a bug with your code.

providers/resource/bin/executable_resource.py
Hook25: good catch thanks, I'm not sure why but on my machine it throws right away. Either way yield from is way better and more legible

fernando79513 (approved):
LGTM +1!

PR #2592, opened 2026-06-18
Update cpuid.py with Clearwater Forest (New)

### LINKED ISSUE #2591: Missing CWF cpuid in cpuid.py
### Enhancement Proposal

cpuid.py is missing the ID for the newly released clearwater forest Xeons.  We need to add this.

bladernr:
## Description
Adds a CWF CPUid to the list of intel cpus in cpuid.py and adjusts the spacing so they all line up pretty like again since "Clearwater Forest" is such a long name.

While I was there, and playing with copilot cli, I learned that unlike Claude it cannot read a user level ~/.github/copilot-instructions.md automatically.  It can only read project level instructions files. BUT apparently referring to the userlevel one in the project level one seems to be a workaround. If the user does not have a global instruction file in ~/.github then copilot will just move on and use the local project instructions only.

## Resolved issues
resolves #2591
resolves CHECKBOX-2284

## Documentation

NA

## Tests

to test, just ran the script on the CWF system I have access to in the Locker:

from a cert test run on the machine before fix:
```
CPU Model: Intel(R) Xeon(R) 6960E+
Unable to determine CPU Family for this CPUID: 0xd06d1
```
modified script run on the machine later:
```
ubuntu@northrup-cwf-716346:~$ sudo ./cpuid.py 
CPU Model: Intel(R) Xeon(R) 6960E+
CPUID: 0xd06d1 which appears to be a Clearwater Forest processor
```

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,6 +1,6 @@
 # Copilot Instructions — Checkbox monorepo
 
-> **Global rules apply** — shell, git, secrets, destructive ops, and PR conventions are in `~/.github/copilot-instructions.md`.
+> Optional: you can add personal Copilot instructions in `~/.github/copilot-instructions.md`; this repository’s rules are defined below.
 
 ## Overview
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,4 +1,6 @@
-# Copilot instructions — Checkbox monorepo
+# Copilot Instructions — Checkbox monorepo
+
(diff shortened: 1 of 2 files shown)


bladernr:
Ahh the checks passed, it's a one line addition to the CPUID list in the script and no other changes, and has been tested on a couple machines to be sure it works as expected.  

The more contentious (perhaps) change is just adding hints  for copilot to look for a global copilot settings file in ~/.github to incorporate in addition to repo local settings files.

Hook25 (approved):

Issue #1808, opened 2025-03-19
Take into consideration iGPU vram in "memory/info" test


omar-selo:
### Bug Description

Some machines allocate VRAM for their iGPU making this test fail. For example [this submission](https://certification.canonical.com/hardware/202111-29710/submission/414906/) shows the following log:
```
Results:
	/proc/meminfo reports:	5.67GiB
	lshw reports:	8GiB

FAIL: Meminfo reports 2503073792 less than lshw, a difference of 29.14%. Only a variance of 15% in reported memory is allowed.
```

Because:
```
dmesg | grep -i vram
[    0.824195] rtc_cmos 00:01: alarms up to one month, 114 bytes nvram, hpet irqs
[    2.870716] amdgpu 0000:03:00.0: amdgpu: VRAM: 2048M 0x000000F400000000 - 0x000000F47FFFFFFF (2048M used)
[    2.870727] [drm] Detected VRAM RAM=2048M, BAR=2048M
[    2.871060] [drm] amdgpu: 2048M of VRAM memory ready
[    3.081871] [drm] vram apper at 0x1AF000000
```

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

Rerun the job http://10.102.156.15:8080/job/cert-oem-sru-focal-desktop-dell-inspiron-15-3525-c29710/

### Expected Result

The test should take into consideration iGPU VRAM and pass.

### Actual Result

The test fails.

### Environment

https://certification.canonical.com/hardware/202111-29710/submission/414906/

### Relevant log output

```shell

```

pieqq:
This issue is a duplicate of #191.

fernando79513:
This issue is still happening. Since this is the one linked in TO, I will reopen this one instead of  #191.
We should address it at some point.

Here I put a list of devices failing this test:
 - https://certification.canonical.com/hardware/202509-37899/submission/492438/
 - https://certification.canonical.com/hardware/202509-37934/submission/492499/
 - https://certification.canonical.com/hardware/202507-37016/submission/492509/
 - https://certification.canonical.com/hardware/202506-36883/submission/492533/
 - https://certification.canonical.com/hardware/202506-36884/submission/492478/
 - https://certification.canonical.com/hardware/202506-36885/submission/492521/
 - https://certification.canonical.com/hardware/202506-36886/submission/492440/

PR #2595, opened 2026-06-22
Memory issue igpu (BugFix)

fernando79513:
## Description

I've added a workaround to try to read the igpu from dmesg and exclude it from the total memory calculation.

 - memory_compare.py now compensates for differences using iGPU VRAM reserved by amdgpu.
 - Unit tests have been added
 - Updated dispatch_lab_job.yaml comments and added a "reboot" step, since it is needed to detect the amdgpu kernel logs, and we do it in our SRU jobs.

## Resolved issues

#1808 

## Documentation

N/A

## Tests

https://github.com/canonical/checkbox/actions/runs/28023950652/job/82990969715


Hook25 (approved):
Thanks, especially for improving the workflow, very useful new reference

One item in the matrix failed but it seems due to that specific machine having an issue, every other looks good

PR #2603, opened 2026-06-23
Retry snap revision info (bugfix)

pieqq:
## Description

Add a retry decorator to the function that uses the snapd `find` command to perform some online queries to retrieve the revisions of a given snap. This will help to prevent errors when there is a network hiccup. The job will still fail if no network is available, as it 100% requires network access to retrieve the revisions in order to run the snap update tests.

## Resolved issues

Fix https://warthogs.atlassian.net/browse/C3-1429

## Tests

Tested locally by running `./snap_update_test.py --resource` without any network connection for a while, then plugging back network. (if no network at all, the test will eventually fail as before)

Added unit tests to test the new `print_logs` param.

### DIFF
--- a/checkbox-support/checkbox_support/helpers/retry.py
+++ b/checkbox-support/checkbox_support/helpers/retry.py
@@ -29,11 +29,13 @@ import time
 from unittest.mock import patch
 
 
-def run_with_retry(f, max_attempts, delay, *args, **kwargs):
+def run_with_retry(f, max_attempts, delay, print_logs=True, *args, **kwargs):
     """
     Run the f function. If it fails, retry for up to max_attempts times, adding
     a backoff and a jitter on top of a delay (in seconds). If none of the runs
-    succeed, raise the encountered exception.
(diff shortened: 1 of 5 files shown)


checkbox-support/checkbox_support/helpers/retry.py
Hook25, line 1: I don't dislike this print logs change but:
- logs should probably go to stderr
- we should probably flush both

Hook25: also consider
```
print = partial(print, flush=True)
```

Hook25 (approved):
lgtm but consider the following

checkbox-support/checkbox_support/helpers/retry.py
pieqq: I just tested this with a resource job that prints stuff to `file=sys.stderr` and indeed, Checkbox ignores this completely and just focuses on stuff printed out on stdout, so I'll close this PR and send a new one with this fix.

PR #2606, opened 2026-06-24
Retry snap revision info stderr (bugfix)

pieqq:
## Description

Add a retry decorator to the function that makes a network call to retrieve revision information for a given snap. In order not to mess up the output of the resource job, the retry function is modified to spit out its informational logs to `stderr` instead of `stdout`.

Note: if network is **really** unavailable for the whole test execution, it will still fail with an exception printed out, but in that case this is expected: we want to investigate this.

## Resolved issues

https://warthogs.atlassian.net/browse/C3-1429

## Tests

- Modified unit tests
- Tested manually by calling `./snap_update_test.py --resource` while offline, waiting until second attempt and connecting back to the network. As you can see, the informational log is now red (`stderr`) while the actual resource output is green (`stdout`):

image

### DIFF
--- a/checkbox-support/checkbox_support/helpers/retry.py
+++ b/checkbox-support/checkbox_support/helpers/retry.py
@@ -25,6 +25,7 @@ delay, backoff and jitter.
 
 import functools
 import random
+import sys
 import time
 from unittest.mock import patch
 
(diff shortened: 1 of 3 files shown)


Hook25 (dismissed):
LGTM, consider flushing both stdout and stderr when you flush stderr

Hook25 (approved):

PR #2612, opened 2026-06-25
Override maximum expected ethernet speed with an environment variable in network.py (New)

### LINKED ISSUE #2568: Allowing user to define the maximum supported speed of network interfaces
### Enhancement Proposal

network.py is relying on the output from `sudo ethtool <INTERFACE>` command to determine the maximum supported speed from a specific network interface.

However for controllers like Nvidia ConnectX8, their advertised maximum speed (400Gbps) can't be achieved without additional setup for GPUDirect, which might not be necessary the default setup out-of-box per information from Nvidia.

Without the setup to enabling GPUDirect, maximum support speed of ConnectX8 becomes 200Gbps, therefore having flexibility for users to define the maximum supported speed of a specific interface in Checkbox would largely improve the user experience when executing iperf3 tests on controllers like Nvidia ConnectX8.

tomli380576:
## Description

This PR adds the relevant logic in network.py to accept a new environment variable `MAX_EXPECTED_SPEED_OVERRIDE`. 

If this environment variable is specified, the logs will show this block:

```
INFO:root:92.85% of expected max 200000 Mb/s
WARNING:root:----------------------------MANUAL OVERRIDE WAS USED----------------------------
WARNING:root:The expected maximum speed of 'enP1p3s0f0np0' was manually overridden to '200000'
WARNING:root:--------------------------------------------------------------------------------
```

indicating that this needs extra attention during review time.

Implements #2568 

## Resolved issues

Some network cards like Nvidia CX8 doesn't support its maximum theoretical speed without special configuration which is the expected behavior. Adding this environment variable will allow these special cards to pass the iperf tests.

There's unfortunately a bit of formatting changes mixed in because the original script won't pass black and flake8 checks.

## Tests

C3:

### DIFF
--- a/providers/base/bin/network.py
+++ b/providers/base/bin/network.py
@@ -369,7 +369,7 @@ class IPerfPerformanceTest(object):
                         "The expected maximum speed of '{}' ".format(
                             self.iface.interface
                         )
-                        + "was manually overridden to '{}'".format(
+                        + "was manually overridden to '{}' Mbps".format(
                             self.expected_max_speed,
                         )
                     )
@@ -722,7 +722,7 @@ def run_test(args, test_target, expected_max_speed: "int | None" = None):
     if expected_max_speed:
         logging.warning(
             "Expected maximum transfer speed of "
-            + "'{}' was overridden to: {}".format(
+            + "'{}' was overridden to: {} Mbps".format(
                 args.interface, expected_max_speed
             )
         )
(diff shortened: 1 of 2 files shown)


tomli380576:
Closing this for now, I will make a PR to implement unit tests and reformat `network.py` first otherwise this is impossible to review down the line.

PR #2626, opened 2026-06-30
Fix the supported_resolutions loop in camera_test.py (Bugfix)

### LINKED ISSUE #2605: camera/display_* failed with USB camera on 26.04
### Bug Description

camera/display_* & camera/led_* would show error "Internal data stream error" with USB camera, bot not for MIPI camera on 26.04

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

1. Run the `com.canonical.certification::camera-base-26-04-manual` test plan via checkbox-cli control 127.0.0.1
2. Choose `camera/display_video0` and `camera/led_video0`

### Expected Result

DUT is able to show camera video preview

### Actual Result

Failed Submission: https://certification.canonical.com/hardware/202606-38926/submission/495946/test-results/fail/
```
Error: Internal data stream error.
Starting video preview
Error: Internal data stream error.
```

### Environment

OS: Ubuntu 26.04
Checkbox Version: 7.3.0.dev42
Checkbox Type: Deb

### Relevant log output

```shell

```

### Additional context

_No response_

tomli380576:
## Description

Inside camera_test.py's `_setup_video_gstreamer` method, the loop gathering the supported resolutions of the viewfinder sink is broken because it does not check if `get_int` has succeeded before collecting the value and doesn't handle int arrays. This PR fixes it and raises an exception if we can't handle the value types.

## Resolved issues

#2605 

## Documentation

Every `<Gst.Structure>.get_*` method like `get_int` returns a tuple of the type `(bool, ValueType)`. For example `get_int` returns `(bool, int)`. If we don't check if [0] is true, then we get junk values in [1]. The exception being `<Gst.Structure>.get_value` which just returns `Any` and expects the caller to test for the type.

There also the easy way out where we just don't set anything for the viewfinder caps and let GStreamer figure it out.

## Tests

C3: 
- Original DUT https://certification.canonical.com/hardware/202606-38926/submission/497757/
- 24.04 https://certification.canonical.com/hardware/202503-36476/submission/497805/
- 22.04 https://certification.canonical.com/hardware/202001-27667/submission/498085/
- 16.04 https://certification.canonical.com/hardware/201812-26713/submission/498046/

(I manually closed the window so there's the "Error: Output window was closed" line)

### DIFF
--- a/providers/base/bin/camera_test.py
+++ b/providers/base/bin/camera_test.py
@@ -43,6 +43,7 @@ import sys
 from glob import glob
 from subprocess import check_call, CalledProcessError, STDOUT, check_output
 from tempfile import NamedTemporaryFile
+from itertools import product
 
 _IOC_NRBITS = 8
 _IOC_TYPEBITS = 8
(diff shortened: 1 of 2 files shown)


hanhsuan (approved):
LGTM

PR #2630, opened 2026-06-30
Setup include requires manifest (new)

Hook25:
## Description

Manifests are very important during setup but up until now there was no way to actually use them properly. The reason is that, although a checkbox-support API offers the capability of reading the machine manifest, editing it or saving it did not work and had to be done manually. 

This introduces:
- requires_manifest: new field to require a manifest (optionally with a specific value) to run a job
- second manifest selection screen before setup include phase that shows all relevant manifests (added via requires_manifest)
- manifest saving also before setup include

#### About the implementation

This adds a new field instead of using resource for two reasons:
1. we've been wanting a separate field for a while for this thing, manifests should be handled separately because it makes it easier to parse them out of metadata/read the test requirements etc.
2. requires (and resources) don't exist/make sense before bootstrap, as bootstrap is the phase where resources are built. To actually allow someone to use requires in setup we would either need a new bootstrap before setup (hard no) or a restricted requires syntax that only allows manifests. In my opinion it is better this way

This adds a new inhibitor to clearly separate the old resource inhibitor from this case, this is just to have code that is a bit easier to read and a better error text in the output. The result is the same as the manifest skip result was added in a previous PR.

This uses the same behaviour that we normally use to decide if the manifest selection screen is shown (that is, do show it if test selection is not forced). This will be changed with a new config in a follow up (`manifest selection`).

This moves the manifest saving code to the session state (because handling the manifest is a state issue, not quite clear to me why it was an assistant thing)

## Resolved issues

Fixes: CHECKBOX-1895

## Documentation

This also adds to the how to and reference all relevant behaviours and fields.

## Tests

Metabox tests + Unit tests (WIP)

### DIFF
--- a/metabox/metabox/metabox-provider/units/setup_include.yaml
+++ b/metabox/metabox/metabox-provider/units/setup_include.yaml
@@ -81,12 +81,26 @@ unit: setup job
 flags: [simple]
 requires_manifest:
   - setup_manifest_true
+  - setup_manifest_false: false
 command: |
-  echo abc
+  echo Setup job
+---
+id: manifest_include_followup
+flags: [simple]
+imports: from com.canonical.plainbox import manifest
+requires:
+  - manifest.setup_manifest_true == "True"
+command: |
+  echo Setup job
 ---
 unit: manifest entry
 id: setup_manifest_true
-name: Setup manifest to be shown before running manifest
+name: Setup manifest to be shown before running manifest that has to be true
+value-type: bool
+---
+unit: manifest entry
+id: setup_manifest_false
+name: Setup manifest to be shown before running manifest that has to be false
 value-type: bool
 ---
 id: manifest_setup_include_tp
@@ -94,4 +108,5 @@ unit: test plan
 name: Test manifest in setup include
 setup_include:
   - manifest_setup_include
-include: []
+include:
+  - manifest_include_followup
--- a/metabox/metabox/metabox-provider/units/setup_include.yaml
+++ b/metabox/metabox/metabox-provider/units/setup_include.yaml
@@ -76,37 +76,46 @@ include:
   - verify_simple_guards
   - verify_resume_guards
 ---
-id: manifest_setup_include
-unit: setup job
-flags: [simple]
-requires_manifest:
-  - setup_manifest_true
-  - setup_manifest_false: false
-command: |
-  echo Setup job
----
-id: manifest_include_followup
-flags: [simple]
-imports: from com.canonical.plainbox import manifest
-requires:
-  - manifest.setup_manifest_true == "True"
-command: |
-  echo Setup job
----
-unit: manifest entry
-id: setup_manifest_true
-name: Setup manifest to be shown before running manifest that has to be true
-value-type: bool
----
 unit: manifest entry
 id: setup_manifest_false
 name: Setup manifest to be shown before running manifest that has to be false
 value-type: bool
 ---
-id: manifest_setup_include_tp
+unit: manifest entry
+id: _setup_manifest_hidden
+name: Setup manifest hidden from the UI
+value-type: bool
+---
+unit: manifest entry
+id: setup_job_manifest
+name: Setup job manifest question shown
+value-type: bool
+---
+unit: setup job
+id: setup_job_requiring_manifest
+flags: [simple]
+requires_manifest:
+  - setup_job_manifest
+  - setup_manifest_false: false
+  - _setup_manifest_hidden
+command: |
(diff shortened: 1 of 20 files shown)


metabox/metabox/metabox-provider/units/setup_include.yaml
```diff
@@ -75,3 +75,48 @@ bootstrap_include:
...
+---
+unit: manifest entry
+id: setup_manifest_false
+name: Setup manifest to be shown before running manifest that has to be false
+value-type: bool
+---
+unit: manifest entry
+id: _setup_manifest_hidden
+name: Setup manifest hidden from the UI
+value-type: bool
```
fernando79513, line 88: We are not validating the metabox providers, but you should add the hidden_reason:

```suggestion
value-type: bool
hidden-reason: This manifest entry is used to test hidden manifests in setup.
```

fernando79513 (changes requested):
First think, I like the `requires_manifest` field. It makes sense to me that manifests are it's separate field and requiring a resource to use them is more convoluted IMO

Second, Very good job here. I tested it locally with the "cer" tool and everything works as expected.

I've added a couple of comments to increase the test coverage, but I didn't found any issue yet. I will continue the review tomorrow, but I don't expect any big issue

metabox/metabox/metabox-provider/units/setup_include.yaml
```diff
@@ -75,3 +75,48 @@ bootstrap_include:
...
+    exit 1
+  fi
+---
+id: manifest_menu_setup_job_tp
+unit: test plan
+name: Test manifest menu shown for setup job
+setup_include:
+  - setup_job_requiring_manifest
+include:
+  - verify_setup_manifest_ran
```
fernando79513, line 123: I ran this test plan with an old version of Checkbox by mistake, and the test was run even without matching manifests

The requires_manifest section was just ignored. It didn't even fail validation to have a `requires_manifest` field that didn't exist on that version of Checkbox; it only ignored it.

I think we should also add a test to check that the job is not run when the manifest is not met.

```
---
id: verify_setup_manifest_not_ran
flags: [simple]
imports:
  - from com.canonical.plainbox import manifest
command: |
  if test -f $PLAINBOX_SESSION_SHARE/setup_manifest_guard; then
    echo "setup job ran, when it should have been skipped"
    exit 1
  else
    echo "setup job did not run, as expected"
  fi
---
id: manifest_menu_setup_job_tp_not_ran
unit: test plan
name: Test manifest menu shown for setup job
setup_include:
  - setup_job_requiring_manifest
include:
  - verify_setup_manifest_not_ran
```

metabox/metabox/scenarios/ui/manifest_menu.py
```diff
@@ -95,3 +95,92 @@ class ManifestSelectionCanSetHiddenLauncher(Scenario):
...
+        2021.com.canonical.certification::setup_manifest_false=False
+        2021.com.canonical.certification::_setup_manifest_hidden=True
+        """)
+
+    steps = [
+        Start(),
+        ExpectNot("System Manifest", timeout=0.5),
+        Expect("2021.com.canonical.certification::verify_setup_manifest_ran"),
+        Expect("job passed"),
+    ]
```
fernando79513, line 186: I think we should also have a test that checks that we are ignoring setup jobs that fail manifest requirements (follow-up):

```
@tag("manual", "interact", "manifest")
class ManifestSelectionUsesLauncherManifestForSetupJobNotRan(Scenario):
    launcher = textwrap.dedent("""
        [launcher]
        launcher_version = 1
        stock_reports = text
        [test plan]
        unit=2021.com.canonical.certification::manifest_menu_setup_job_tp_not_ran
        forced = yes
        [test selection]
        forced = yes
        [manifest]
        2021.com.canonical.certification::setup_job_manifest=False
        """)

    steps = [
        Start(),
        ExpectNot("System Manifest", timeout=0.5),
        Expect("2021.com.canonical.certification::verify_setup_manifest_not_ran"),
        Expect("job passed"),
    ]    
```

fernando79513 (changes requested):
Just a comment to add more metabox tests.

The rest of the implementation looks great. Very good job

metabox/metabox/metabox-provider/units/setup_include.yaml
Hook25: sadly this is how checkbox works, if a field is not in the fields checkbox validates, they are ignored. The new schema is a step in the direction of stopping this (it wouldn't have validated the schema) but it is not used during validate for now. Either way I can add a test to see that the manifest skip is actually working

fernando79513 (approved):
Approved, let's wait until the metabox tests are run to merge it.

PR #2635, opened 2026-07-02
Re-implement max resolution test to handle fractional scaling (New)

### LINKED ISSUE #2616: graphics_max_resolution.py does not account for fractional scaling
### Bug Description

If a DUT is using fractional scaling (now officially supported on 26.04), the test script will round up the fraction to the nearest integer. For example, if the DUT has a 1920x1080 panel and uses 125% scaling, the script will report that it's using 1536x960@200% which is incorrect; the correct answer is 1536x960@125%.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Enable fractional scaling and set the display scaling percentage to something non-integer
2. Run the test script
3. Test script will fail because it's rounding up the scaling percentage

### Expected Result

Test script does not round up the scaling percentage

### Actual Result

Test script rounds up the scaling percentage

### Environment

ubuntu 26.04 desktop

### Relevant log output

```shell

```

### Additional context

_No response_

tomli380576:
## Description

This PR re-implements the `graphics_max_resolution` test to handle fractional scaling that's now officially supported on 26.04.

## Resolved issues

Originally the test case uses the display object returned by `Gdk.Display.get_default()` to get the current application-level resolution and scaling. The scaling value returned by Gdk is [always an integer](https://docs.gtk.org/gdk4/method.Monitor.get_scale_factor.html) which makes the test case unable to handle fractional scaling. The new implementation directly queries GNOME about the physical monitor dimensions instead so we can ignore scaling and wayland/x11 differences altogether. Since we have dropped 16.04, we don't even need to include a unity path.

## Documentation

**This test does not work on 16.04, just like the original version.** The new one doesn't work on 16.04 because the dbus call used by MonitorConfigGnome is not present in unity. The old one doesn't work is because of this line
```py
gi.require_versions({"Gtk": "3.0", "Gdk": "3.0"})
```
the `require_versions` method does not exist in 16.04's python3-gi binding.

## Tests

C3: https://certification.canonical.com/hardware/202503-36476/submission/511274/ pass correctly when both monitors are set to 150% scaling. If I manually lower the resolution it will fail https://certification.canonical.com/hardware/202503-36476/submission/511276/

### DIFF
--- a/checkbox-support/checkbox_support/dbus/gnome_monitor.py
+++ b/checkbox-support/checkbox_support/dbus/gnome_monitor.py
@@ -132,13 +132,13 @@ class PhysicalMonitor(_PhysicalMonitorT):
     def is_builtin(self) -> bool:
         return self.properties.get("is-builtin", False)
 
-    def max_resolution(self) -> "tuple[int, int]":
+    def get_max_resolution(self) -> "tuple[int, int]":
         """Get the maximum physcial resolution of this monitor
 
         :raises ValueError: if there's no supported modes
         :raises RuntimeError: if all modes are 0
         :return: (width, height) pair like (1920, 1080)
-        """        
+        """
         max_w, max_h = 0, 0
         if len(self.modes) == 0:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


hanhsuan (approved):
LGTM

PR #2655, opened 2026-07-07
remove python3.5-3.6 check for ce-oem provider (Bugfix)

stanley31huang:
## Description
remove python3.5 and python3.6 check for ce-oem provider


tomli380576:
finally i can use f strings

seankingyang (approved):
LGTM~~

PR #2657, opened 2026-07-07
Merge (flag) siblings creation with normal sibling creation (bugfix)

Hook25:
## Description

The code was duplicated as both did the same thing and it wans't clear if there were any differences. This merges flag-driven sibling creation (like also after suspend) into the same code path of normal sibling creation (making the sibling fields by hand) removing the duplication.

Minor: this removes the `after` given that all code paths define a `depends` (which is a stronger constraint)

## Resolved issues

Part of: https://warthogs.atlassian.net/browse/CHECKBOX-1928

## Documentation

N/A

## Tests

All unit tests are still passing as before

### DIFF
--- a/checkbox-ng/plainbox/impl/session/state.py
+++ b/checkbox-ng/plainbox/impl/session/state.py
@@ -1232,28 +1232,9 @@ class SessionState:
             if recompute:
                 self._recompute_job_readiness()
 
-    def _add_job_siblings_unit(self, new_job, recompute, via):
-        if new_job.siblings:
-            siblings = new_job.siblings
-            for overrides in siblings:
-                data = {
-                    key: value
-                    for key, value in new_job._data.items()
-                    if not key.endswith("siblings")
-                }
-                data.update(overrides)
-                self._add_job_unit(
-                    JobDefinition(
-                        data,
-                        origin=new_job.origin,
-                        provider=new_job.provider,
-                        controller=new_job.controller,
-                        parameters=new_job.parameters,
-                        field_offset_map=new_job.field_offset_map,
-                    ),
-                    recompute,
-                    via,
-                )
+    def _get_flags_siblings(self, job):
+        siblings = []  # this is exactly what would live in the siblings field
+        # aka a list of overrides
         suspend_siblings = [
             (Suspend.AUTO_FLAG, Suspend.AUTO_JOB_ID, "after-suspend"),
             (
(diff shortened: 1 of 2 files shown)


fernando79513 (commented):
The PR looks fine and it simplifies the logic. So LGTM.

We still have some duplication of the suspend flags logic in:
https://github.com/canonical/checkbox/blob/main/checkbox-ng/plainbox/impl/ctrl.py#L197

Mayve we could move _get_flags_siblings into JobDefinition (in job.py) and expose a single "effective_siblings" accessor, and then use it in `state.py` and `ctlr.py`. This may not be the purpose of this PR, but I don't think its a bad idea either:

Step 1 — move _get_flags_siblings into JobDefinition (in job.py) and expose a single "effective siblings" accessor:
```python
# job.py
from plainbox.suspend_consts import Suspend
from contextlib import suppress

class JobDefinition(...):
    def _get_flag_siblings(self):
        siblings = []
        suspend_siblings = [
            (Suspend.AUTO_FLAG, Suspend.AUTO_JOB_ID, "after-suspend"),
            (Suspend.MANUAL_FLAG, Suspend.MANUAL_JOB_ID, "after-suspend-manual"),
        ]
        for suspend_flag, suspend_job_id, suspend_prefix in suspend_siblings:
            if suspend_flag not in self.get_flag_set():
                continue
            overrides = {}
            # ... (identical body to today's _get_flags_siblings, using self)
            siblings.append(overrides)
        return siblings

    def get_effective_siblings(self):
        """Explicit siblings + siblings implied by flags."""
        return (self.siblings or []) + self._get_flag_siblings()
```

Step 2 — state.py consumes it instead of computing locally:
```python
def _add_job_siblings_unit(self, job, recompute, via):
    for overrides in job.get_effective_siblings():
        data = {k: v for k, v in job._data.items()
                if not k.endswith("siblings")}
        data.update(overrides)
        self._add_job_unit(JobDefinition(data, ...), recompute, via)
```
SessionState._get_flags_siblings disappears.

Step 3 — ctrl.py collapses to one check, dropping the flag branch entirely:
```python
def _is_job_impacting_suspend(self, suspend_job_id, job):
    for sibling in job.get_effective_siblings():
        if suspend_job_id in sibling.get("depends", []):
            return True
    return False
```

Hook25:
@fernando79513 this is a preparation pr to remove all the code in ctrl related to before suspend (as the linked card states) to move it here, with a before field.

fernando79513 (approved):
LGTM +1!

PR #2662, opened 2026-07-08
Expose a normalized chassis in the submission (New)

diohe0311:
Checkbox already normalizes SMBIOS chassis types, but it was only used in resource and never recorded into the submission.

This exposes it as a `chassis` field so C3 get a normalized chassis without having to re-parse anything. The `dmi` resource now send a `chassis` field for CHASSIS devices and the submission template pulls it through.

This change exposes the normalized chassis type through the system information collector. The submission schema now includes this information under the `chassis` field, allowing C3 to consume it without parsing SMBIOS data again.

The chassis list is also updated based on the latest SMBIOS spec, which adds the 7 new types. I've also fix a potential crash where a machine reporting one of the newer types would hit an out-of-range `IndexError` in the chassis lookup. That path now guards its bounds and returns `Unknown` instead of raising error.

Example submission field:

```json
{
  "tool_version": "7.5.1.dev1+g4dcc8d91c.d20260903",
  "success": true,
  "outputs": {
    "payload": {
      "type": "Notebook"
    },
    "stderr": ""
  }
}
```

Tested on a [Qualcomm Addons IQ 9075 EVK](https://certification.canonical.com/hardware/202508-37813/), which does not expose `/sys/class/dmi/id/chassis_type` and returned the expected fallback.
```
Failed to read chassis type. Error: [Errno 2] No such file or directory: '/sys/class/dmi/id/chassis_type'
{
    "type": "Unknown"
}
```
Tested on a [Renesas SMARC EVK](https://certification.canonical.com/hardware/202508-37040/).
```
{
    "type": "Desktop"
}
```

## Description
 - Added ChassisCollector and get_chassis_info to fetch chassis data from DMI
 - Updated the chassis types according to the latest SMBIOS spec

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CHECKBOX-2304

## Tests
Tested locally

### DIFF
--- a/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
+++ b/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
@@ -50,6 +50,11 @@
         {%- endfor %}
     },
 {%- endif %}
+{%- if ns ~ 'dmi' in state.resource_map %}
+{%- for dmi_resource in state.resource_map[ns ~ 'dmi'] if dmi_resource.chassis %}
+    "chassis": "{{ dmi_resource.chassis }}",
+{%- endfor %}
+{%- endif %}
     "results": [
     {%- for job_id, job_state in job_state_map|dictsort if job_state.result.outcome != None and job_state.job.plugin != "attachment" %}
         {
--- a/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
+++ b/checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
@@ -51,9 +51,10 @@
     },
 {%- endif %}
 {%- if ns ~ 'dmi' in state.resource_map %}
-{%- for dmi_resource in state.resource_map[ns ~ 'dmi'] if dmi_resource.chassis %}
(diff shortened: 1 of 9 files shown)


providers/resource/bin/dmi_resource.py
Hook25, line 1: Why?

checkbox-ng/plainbox/impl/providers/exporters/data/checkbox.json
Hook25, line 1: No, we are getting away from that, call the script from a collector, that will add the information as a system_information

submission-schema/schema.json
Hook25, line 1: See above, use the system information and update it!

Hook25 (dismissed):
Please use system information

pieqq (approved):
Tested on my device, it works as expected and exposes a new `chassis` section in the submission's `system_information`:

```json
    "chassis": {
        "tool_version": "7.0.1.dev25+g9a3965953.d20260311",
        "success": true,
        "outputs": {
            "payload": {
                "type": "Desktop"
            },
            "stderr": ""
        }
    },
```

PR #2663, opened 2026-07-08
Use before for jobs that have siblings after suspend (bugfix)

Hook25:
## Description

In Checkbox when a job has a sibling that depends on suspend (including auto generated siblings) it (the job) is implicitly treated as a job that runs before suspend. This relied on a mechanism that pre-dates the `before:` field by a few years introduced here: #1037

In this PR we align this ad-hoc behaviour to how we enforce before ordering now, with the before field, so this effectively reverts (most) of the old change in favour of just using `before`

## Resolved issues

Fixes: CHECKBOX-1928

## Documentation

N/A

## Tests

Added tests to cover all scenarios, namely:
- If a sibling is after suspend (via depends or after) and the job is not after suspend, we set the before
- If a sibling is after suspend and the job is after suspend, we leave it as is
- Same should work for the flag

> Note: For backward compatibility this also supports the manual suspend flag, to a fault, given that if both siblings are present or if both flags are used, both before are defined, but I think that is what a user would expect either way

### DIFF
--- a/checkbox-ng/plainbox/impl/session/state.py
+++ b/checkbox-ng/plainbox/impl/session/state.py
@@ -1278,7 +1278,28 @@ class SessionState:
             siblings.append(overrides)
         return siblings
 
+    def _get_job_data_directly_required_suspend(self, job_data):
+        to_r = set()
+
+        def pxu_compatible_in(con):
+            if isinstance(con, list):
+                return con
+            # this is not perfect but it is good enough for what we need here
+            return con.split()
+
+        for suspend_id in [Suspend.AUTO_JOB_ID, Suspend.MANUAL_JOB_ID]:
+            if suspend_id in pxu_compatible_in(job_data.get("depends", [])):
+                to_r.add(suspend_id)
+            elif suspend_id in pxu_compatible_in(job_data.get("after", [])):
+                to_r.add(suspend_id)
+        return to_r
+
     def _add_job_siblings_unit(self, job, recompute, via):
+        job_requiring_suspend = set(
+            self._get_job_data_directly_required_suspend(job._data)
+        )
+        siblings_requiring_suspend = set()
+
         siblings = job.siblings or []
         siblings += self._get_flags_siblings(job)
         for overrides in siblings:
@@ -1300,6 +1321,24 @@ class SessionState:
                 recompute,
                 via,
             )
+            siblings_requiring_suspend |= (
+                self._get_job_data_directly_required_suspend(data)
+            )
+        sibling_required_not_job = (
+            siblings_requiring_suspend - job_requiring_suspend
+        )
+        # when a job doesn't require to be after suspend but has a sibling that
+        # wants to be the job is always before suspend
+        if not sibling_required_not_job:
+            return
+        # Note: intentionally avoid accessing before directly as it is a
+        #       cached property
+        before = job._data.get("before", [])
+        if isinstance(before, list):
+            before += list(sibling_required_not_job)
+        else:
+            before += " " + " ".join(sibling_required_not_job)
+        job._data["before"] = before
 
     def remove_unit(self, unit, *, recompute=True):
         """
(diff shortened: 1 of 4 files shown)


checkbox-ng/plainbox/impl/session/state.py
```diff
@@ -1278,7 +1278,28 @@ def _get_flags_siblings(self, job):
             siblings.append(overrides)
         return siblings
 
+    def _get_job_data_directly_required_suspend(self, job_data):
```
pieqq, line 1281: the wording of this function is a bit unusual, can you add a docstring to explain what it does?

checkbox-ng/plainbox/impl/session/state.py
```diff
@@ -1278,7 +1278,28 @@ def _get_flags_siblings(self, job):
...
+        to_r = set()
+
+        def pxu_compatible_in(con):
+            if isinstance(con, list):
```
pieqq, line 1294: I assume this is the yaml path? Can you add a comment above?

checkbox-ng/plainbox/impl/session/state.py
```diff
@@ -1300,6 +1321,24 @@ def _add_job_siblings_unit(self, job, recompute, via):
...
+            )
+        sibling_required_not_job = (
+            siblings_requiring_suspend - job_requiring_suspend
+        )
+        # when a job doesn't require to be after suspend but has a sibling that
+        # wants to be the job is always before suspend
+        if not sibling_required_not_job:
+            return
+        # Note: intentionally avoid accessing before directly as it is a
+        #       cached property
```
pieqq, line 1347: I'm not sure I understand this comment. Can you explain?

pieqq (commented):
Just a few questions

checkbox-ng/plainbox/impl/session/state.py
Hook25: properties on jobs are cached so if you access them, you can't change them anymore by modifying `_data` (as it will be read only once). This is done because reading data is (in general) expensive. Issue is that the objective here is to form new _data, so one needs to be careful about not adressing properties he intends to change (like before in this case). 

It would probably not make a difference in this specific situation to do `job.before = xyz`, but doing so leaves _data and the property value misaligned

checkbox-ng/plainbox/impl/session/state.py
```diff
@@ -1278,7 +1278,28 @@ def _get_flags_siblings(self, job):
...
+                return con
+            # this is not perfect but it is good enough for what we need here
+            return con.split()
+
+        for suspend_id in [Suspend.AUTO_JOB_ID, Suspend.MANUAL_JOB_ID]:
+            if suspend_id in pxu_compatible_in(job_data.get("depends", [])):
+                to_r.add(suspend_id)
+            elif suspend_id in pxu_compatible_in(job_data.get("after", [])):
+                to_r.add(suspend_id)
+        return to_r
```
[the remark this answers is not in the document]
Hook25: This is fine and the added complexity is not justified, it is true this doesn't work 100% of the times but it is also true it will work in most situations were we care about it.

checkbox-ng/plainbox/impl/session/state.py
```diff
@@ -1300,6 +1321,24 @@ def _add_job_siblings_unit(self, job, recompute, via):
...
+        if not sibling_required_not_job:
+            return
+        # Note: intentionally avoid accessing before directly as it is a
+        #       cached property
+        before = job._data.get("before", [])
+        if isinstance(before, list):
+            before += list(sibling_required_not_job)
+        else:
+            before += " " + " ".join(sibling_required_not_job)
+        job._data["before"] = before
```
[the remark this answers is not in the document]
Hook25: this is fine, also, the case where there is more than 1 is most likely user error, although we technically support it, so we don't mind the random order

pieqq (dismissed):

pieqq (approved):

Issue #2638, opened 2026-07-03
Checkbox crashes with DependencyDuplicateError during light_sensor_resource when multiple ALS devices are present.


clairlin53:
### Bug Description

The system has two als devices, and the checkbox will crashe during the bootstrap phase.
The checkbox attemps to generate multiple jobs with the exact same ID (com.canonical.certification::power-management/light_sensor_test_als), resulting in a DependencyDuplicateError and halting the Checkbox run.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. checkbox-cli control 127.0.0.1
2. Select `com.canonical.certification::client-cert-desktop-24-04-manual` test plan

### Expected Result

Checkbox should dynamically append a unique identifier (such as the device path, e.g., _iio_device1, _iio_device2) to the generated job ID for each detected sensor. This would allow the test plan to proceed and verify both sensors independently.

### Actual Result

`ubuntu@ubuntu-202601-38304:/sys/bus/iio/devices/iio:device1$ checkbox-cli control 127.0.0.1
============[ Bootstrap com.canonical.certification::device (1/5) ]=============
--------------------------------------------------------------------------------
Outcome: job passed
=========[ Bootstrap com.canonical.certification::graphics_card (2/5) ]=========
--------------------------------------------------------------------------------
Outcome: job passed
=====[ Bootstrap com.canonical.certification::light_sensor_resource (3/5) ]=====
Traceback (most recent call last):
File "/usr/bin/checkbox-cli", line 8, in <module>
sys.exit(main())
^^^^^^
File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/checkbox_cli.py", line 185, in main
return subcmd.invoked(ctx)
^^^^^^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/controller.py", line 209, in invoked
return self.connect_and_run(ctx.args.host, port)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/controller.py", line 339, in connect_and_run
keep_running = self.continue_session()
^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/checkbox_ng/launcher/controller.py", line 513, in continue_session
return self.connection_strategy()[state](self, payload)
[...]

PR #2668, opened 2026-07-09
Fix DependencyDuplicateError during light_sensor_resource when multiple ALS devices are present (Bugfix)

hanhsuan:
## Description

There are some devices that include multiple light sensors with the same name. Therefore, we need to add one more argument, `device`, to ensure the job ID is unique.

## Resolved issues

#2638

## Tests

Single light sensor: https://certification.canonical.com/hardware/202001-27667/submission/499535/
Multiple light sensors:  https://certification.canonical.com/hardware/202601-38307/submission/499500/

### DIFF
--- a/providers/base/bin/light_sensor_test.py
+++ b/providers/base/bin/light_sensor_test.py
@@ -18,6 +18,7 @@ import argparse
 import os
 import time
 from typing import Dict, List, Optional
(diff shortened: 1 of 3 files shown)


tomli380576 (approved):
lgtm

PR #2677, opened 2026-07-13
Fix PTP verify job for ethtool v6.19+ (BugFix)

snalvc:
ethtool v6.19 renamed the PTP clock field from "PTP Hardware Clock: N" to "Hardware timestamp provider index: N", causing PTP_ID to be empty and the test to report "/dev/ptp not found!" even on capable hardware.

Two fixes combined:
- Extend the grep to match both field names with -E alternation, then extract the trailing number with grep -oE '[0-9]+$' instead of awk '{print $NF}'. The awk form also broke Checkbox template expansion, which treated {print $NF} as an unknown resource variable and refused to generate the job unit.
- Replace the `[ "$PTP_ID" = "none" ]` guard with `[ -z "$PTP_ID" ]` since digit-only extraction never yields the string "none".

Signed-off-by: Tony Hsiao <tony.hsiao@canonical.com>

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/ptp/jobs.pxu
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/ptp/jobs.pxu
@@ -48,8 +48,8 @@ estimated_duration: 2s
 imports: from com.canonical.plainbox import manifest
 requires: manifest.has_ptp == 'True'
 command:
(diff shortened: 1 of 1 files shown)


rickwu666666 (approved):
LGTM

tomli380576:
We should probably move to using ioctls to directly talk to the kernel in the future https://gist.github.com/tomli380576/9a3b21ecd50796e51415af416d5d4a3b

PR #2687, opened 2026-07-14
Add test plan pc-sanity-smoke-test-26-04 (New)

xanthein:
## Description

This change add test plan pc-sanity-smoke-test-26-04 which nested pc-sanity-base, client-cert-desktop-26-04-automated and power-management-cpu-cstate-cert-automated


pseudocc (approved):
LGTM

PR #2688, opened 2026-07-15
No longer crash if setup jobs fail in local (bugfix)

Hook25:
## Description

Although this was the intended design, we have decided that this is not the way Checkbox should work. Checkbox should always conclude the test session regardless of what is going on because else we don't get the report.

Note that this internally also makes the setup job results available in the report due to a previous patch, and this is desirable

Note2 this was actually never done for remote, so no need to remove it from there.

## Resolved issues

Fixes: CHECKBOX-2298

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
I tested it locally and it works as expected.
LGTM +1!

PR #2689, opened 2026-07-15
Read WiFi AP info from file if available and group those jobs (Bugfix)

p-gentili:
## Description

The goal is to provide setup/teardown jobs for an external WiFi access point later on.

1. Include WiFi jobs in templated groups
2. Make them read SSID/Password from file, fallback to regular envars

Notes:
The proposed solution to load envar is marked by shellcheck but it's also the most straightforward I've found.
1. The test case doesn't have a clue of the band we're testing
2. ... and it doesn't have a clue of the envars we're using to store SSID and password

Other file parsers just look more complex to add inline.

## Resolved issues

Part of https://warthogs.atlassian.net/browse/ZAP-1502

## Documentation

N/A

## Tests

Current behavior is not impacted:
```
[launcher]
launcher_version = 1
stock_reports = text

[test plan]
unit = com.canonical.certification::wireless-automated
forced = yes

[test selection]
forced = true
match = ".*_ac_.*"

[ui]
type = silent

[environment]
OPEN_AC_SSID = <test-open-ssid>
WPA_AC_SSID = <test-wpa-ssid>
WPA_AC_PSK = <test-pwd>

[manifest]
com.canonical.certification::has_wlan_adapter = true
```

### DIFF
--- a/providers/base/units/wireless/jobs.pxu
+++ b/providers/base/units/wireless/jobs.pxu
@@ -53,6 +53,7 @@ plugin: shell
 user: root
 command:
   net_driver_info.py "$NET_DRIVER_INFO"
+  # shellcheck source=/dev/null
   source "$PLAINBOX_SESSION_SHARE/wifi-bg.env" 2> /dev/null || true
   wifi_nmcli_test.py secured {{ interface }} "$WPA_BG_SSID" "$WPA_BG_PSK"
 category_id: com.canonical.plainbox::wireless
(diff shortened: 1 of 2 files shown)


Hook25 (changes requested):
the solution is fine but please use an if statement to check if the file is present, and if it is, print that you are indeed loading it, given that the content of the file and the credentials in the launcher envvar may differ and it would be a nightmare to debug otherwise

Hook25 (approved):

PR #2693, opened 2026-07-16
Fix Pipewire source/sink change detection (BugFix)

tomli380576:
## Description

The program used to check `["info"]["params"]["Route"]` for each pipewire client but the more accurate way is to check `["info"]["params"]["EnumRoute"]` because sometimes the `Route` list isn't updated upon device removal, but EnumRoute is.

Also added some type annotations and restrict `monitor` and `through` subcommands to only accept `sinks`/`sources` for the `--mode` param.

## Resolved issues

On some desktops, GNOME will correctly report sink/source changes, but `checkbox-support-pipewire-utils` (and notably `wpctl status`) don't.

## Documentation

This would align with how wireplumber detects insertion events.
https://github.com/PipeWire/wireplumber/blob/cb92ef9846e2b2f827e1831515bc3a31e7d2ed1a/src/scripts/device/select-routes.lua#L11

Why the test data was directly modified: EnumRoute and Route seem to share the exact same schema, just updated based on different events. 

## Tests

Unit tests

Regular laptop: https://certification.canonical.com/hardware/202503-36476/submission/500627/

The desktop in question: https://certification.canonical.com/hardware/202607-39073/submission/500626/

### DIFF
--- a/checkbox-support/checkbox_support/scripts/pipewire_utils.py
+++ b/checkbox-support/checkbox_support/scripts/pipewire_utils.py
@@ -249,7 +249,7 @@ class PipewireTest:
         Checks whether the sink is available for the given device.
         This function parse output of pw-dump to find the device type
         equals PipeWire:Interface:Device and media.class equals Audio/Device.
-        For pipewire, the active port will be listed under 
+        For pipewire, the active port will be listed under
             info.params.EnumRoute.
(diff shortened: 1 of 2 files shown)


tomli380576:
~~will upload submissions when C3 comes back~~ **submissions added**

Hook25 (approved):
+1, thanks

PR #2697, opened 2026-07-16
Clean and bugfix session reset (bugfix)

Hook25:
## Description

This adds a new `--clean` option to the `checkbox-cli control` command to force cleanup and reset of an agent. When this option is used:
- No session is resumed
- The agent is brought back to a (semi-)clean state so even if it was stuck in a weird state, reconnection is now possible

This also fixes a very long standing bug we thought was a feature: when a unit is changed and the agent is not restarted before starting a new session with the controller, the unit is not reloaded. We thought this was intended but actually this is entirely incidental (and a bad behaviour) that stems from the fact that the units are loaded at import time (yuck) and "cached" globally in the module. I've decided to add a reset function instead of removing this abomination because I'm scared of what could happen if I do.

## Resolved issues

Fixes: CHECKBOX-2309

## Documentation

Added help command

## Tests

Adjusted tests

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -318,7 +318,13 @@ class RemoteController(ReportsStage, MainLoopStage):
                     # clean is used to recover from the sa being in a weird
                     # unrecoverable state
                     if self._clean:
-                        self._sa.reset_session()
+                        try:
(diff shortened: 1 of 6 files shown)


fernando79513 (approved):
Very good catch. I can't believe this has been there forever and it's actually a bug.
I can't remember the times that I had to reset the agent to include some new changes on a unit...

LGTM +1!

PR #2699, opened 2026-07-17
Add Jetson platform to the CE OEM MIPI camera tests (New)

seankingyang:
## Description

Migrate Jetson camera testing onto the CE OEM MIPI camera framework, adding
`jetson` as a fourth platform alongside `genio`, `imx` and `rz` (per the
"Adding New Platforms" process in `units/camera/README.md`). The change is
purely additive: the legacy `units/Jetson/camera_*.pxu` jobs are untouched,
no test plan is modified, and the new jobs only run on DUTs whose manifest
sets `has_vendor_specific_mipi_camera`.

Ref: https://warthogs.atlassian.net/browse/OEMQA-6806

Data-format changes (scenario JSON and resource output):

- The scenario JSON gains an optional `camera_id` key that overrides
  `v4l2_device_name` (which remains the default identifier) for capture
  methods that don't address the camera by its v4l2 name. On Jetson it holds
  the Argus source index consumed by `nvarguscamerasrc sensor-id=` /
  `nvargus_nvraw --c`. Every item must declare at least one of the two keys.
- The `mipi_camera_resource` output gains `name` (always) and `mode` (only
  when the scenario declares it). All pre-existing fields are unchanged, and
  the rendered job ids are byte-identical for all 12 existing Genio/NXP/RZ
  scenario files (verified by regenerating resources old-vs-new).

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/OEMQA-6806

## Documentation

- `units/camera/README.md`: new job-id/`name` composition, the `camera_id`
  override, and the optional GStreamer plugin-path environment variables.
- `data/Jetson-MIPI-Camera-TestScenario-TestSetup/Test_Scenario_and_Test_Setup.md`:
  Jetson sensor/mode tables, per-connector IMX219 scenario files, and the
  required checkbox environment (deb and snap GStreamer plugin path values).

## Tests


seankingyang:
## Removed `user: root` from the `jetson-core-camera/*` templates

Commit `4fa847281` drops the `user: root` line from the three snap-path
`jetson-core-camera/*` templates in
`contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/Jetson/camera_job.pxu`
(nvargus image, gstreamer image, gstreamer video). The deb-path
`jetson-camera/*` templates never had `user: root` and are untouched.
Net diff: 3 deletions, nothing else.

### Why this is safe — rootless capture proven (OEMQA-6807)

Verified first-hand on the IGX UC22 DUT (2026-07-18), rootless over plain
ssh, uid 1000, no sudo:

- **gst** `nvarguscamerasrc → nvjpegenc` rootless → **346526 B** JPEG.
- **nvargus-nvraw** rootless → **2020549 B**.
- Independently re-verified → **344982 B** JPEG, `GST_ARGUS: Done Success`.

The original blocker was kernel DAC, not AppArmor: the Tegra device nodes
are `root:video 0660` and the runner is not in the `video` group (EACCES
with **no** AppArmor denial). The fix does not add group membership — the
runner already holds an **active seat0** (tty autologin), so logind
`uaccess` ACLs apply once the nodes are tagged. One udev rule file
(`/etc/udev/rules.d/99-tegra-camera-uaccess.rules`):

```
SUBSYSTEM=="video4linux", TAG+="uaccess"
KERNEL=="nvmap", TAG+="uaccess"
KERNEL=="nvhost-*", TAG+="uaccess"
KERNEL=="capture-isp-channel*", TAG+="uaccess"
KERNEL=="capture-vi-channel*", TAG+="uaccess"
KERNEL=="tegra_camera_ctrl", TAG+="uaccess"
SUBSYSTEM=="nvidia-gpu-v2", TAG+="uaccess"
SUBSYSTEM=="nvidia-gpu-v2-power", TAG+="uaccess"
TAG=="uaccess", ENV{MAJOR}!="", RUN{builtin}+="uaccess"
```

The last line is **required**: the stock `uaccess` ACL builtin runs at
`73-seat-late.rules`, before a `99-` file's tag lands, so the tag alone
would be applied too late without the explicit `RUN{builtin}+="uaccess"`
re-trigger.

### Dependency — please weigh before merge

Rootless operation is **not** self-contained in this provider. It requires
both of the following, which are properties of the **image / gadget**, not
of checkbox or the multimedia snap:

1. The runner user holding an **active seat** (autologin console session),
   so logind `uaccess` ACLs apply; and
2. The **uaccess udev rule above shipped in the image**, tagging the Tegra
   camera nodes.

DUTs lacking either of these — e.g. the AGX Orin / Orin Nano used in
earlier validation, unless their images add the rule — **will fail these
jobs rootless**. Reviewers should confirm the rule lands in the target
test images before relying on these jobs without elevation.

Reference: OEMQA-6807 (multimedia snap `home` + `var-tmp` plugs, uaccess
rootless enablement) and OEMQA-6806.

seankingyang:
## Update: Jetson `user: root` restored; rootless scoped to the generic MIPI camera jobs


- **`units/Jetson/camera_job.pxu` — `user: root` restored** on the three
  `jetson-core-camera/*` templates (reverting `4fa847281`). As the previous
  comment's own "Dependency" section flagged, rootless Tegra capture is
  **not self-contained**: it needs the `uaccess` udev rule shipped in the
  **image/gadget**, which is not yet in the target Jetson images. Until that
  rule ships, these templates keep the elevation so the jobs pass on images
  without it. Net effect: this file is now **unchanged vs `main`** (the PR no
  longer touches it).

- **`units/camera/jobs.pxu` — `user: root` removed** from the two generic
  `ce-oem-mipi-camera/*` templates, which run as the invoking user. These
  had `user: root` added earlier in this branch; `main` never had it, so the
  net PR state simply leaves them rootless.

The two generic jobs both run `camera_test.py testing …` against a
vendor-specific MIPI camera (V4L2 / GStreamer): `capture-image-scenario`
captures a `{width}x{height} {format}` still, and `record-video-scenario`
records the same at `{fps}` fps. No device-elevation is part of these jobs
on `main`, so dropping `user: root` returns them to their unelevated state;
any per-platform device-node access is handled at the image level, same as
the Jetson dependency above.

Reference: OEMQA-6807, OEMQA-6806.

baconYao (approved):
LGTM

PR #2702, opened 2026-07-17
Unit tests and minor fixes for network.py (New)

tomli380576:
## Description

This PR increases unit test coverage to 90% for network.py and fixed these issues:
1. removed all uses of unbound variables
2. moved the global `results` variable inside the iperf class and added a thread lock to safely collect results when running iperf tests
3. converted all string paths to `Path` objects
4. replaced magic numbers with named constants
5. added more types to help LSPs

## Resolved issues

Low unit test coverage makes it overly difficult to add new features to network.py because any minor change will cause the CI to complain that there's only 30% patch coverage. 

## Documentation

Tbh the lock is added because I couldn't find a definitive answer as to whether `<list>.append()` is thread safe on all the supported versions. If anyone can prove or disprove it please lmk :smiley_cat: .

## Tests

Single-threaded iperf:

Multi-threaded iperf:

### DIFF
--- a/providers/base/bin/network.py
+++ b/providers/base/bin/network.py
@@ -87,7 +87,8 @@ class IPerfPerformanceTest:
         self.scan_timeout = scan_timeout
         self.iface_timeout = iface_timeout
         self.reverse = reverse
-        self.results = []  # type: list[str]
+
+        self._results = []  # type: list[str]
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
Very good job, thanks!

PR #2703, opened 2026-07-17
Backward compatible resource-results getter (bugfix)

### LINKED ISSUE #2694: Merge-submissions drops resource-results from output submission.json, causing merge-reports to fail
### Bug Description

When using checkbox-cli merge-submissions to combine multiple submission tarballs, the resulting output tarball's submission.json is missing the resource-results field, even when all input submissions contain it.

Fail A: the tarble file cannot safe as new submission archive while using checkbox submission editor

Image

Fail B :  the also causes a subsequent checkbox-cli merge-reports on the merged tarball to fail with:
`Fail to merge. Make sure your input submission provides resource-related information.
'resource-results'
`

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

1. Run
`checkbox-cli merge-submissions A.tar.xz B.tar.xz -o merged.tar.xz`

2. Check the output
`tar -xJOf merged.tar.xz submission.json | python3 -c \
     "import sys,json; d=json.load(sys.stdin); print('resource-results' in d)"`

  → Prints False

3. Run
`checkbox-cli merge-reports merged.tar.xz -o report.html`

→ Fails with KeyError: 'resource-results'

### Expected Result

The merged output submission.json should contain a resource-results field that is the union of all input submissions' resource-results.

### Actual Result

resource-results is silently dropped. merge-reports on the merged tarball then raises a KeyError and aborts.

### Environment

- OS 24.04
- checkbox version 7.4.0.dev79

### Relevant log output

```shell

```

### Additional context

_No response_

Hook25:
## Description

Resource results is no more and this merger expects it to exist. For backward compatibility, this preserves the loader but uses `get` as `[]` is now the actual value that field will always have

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/2694

## Documentation

Added a comment explaining why this line is needed still

## Tests

Added a test without resource-results

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/merge_reports.py
+++ b/checkbox-ng/checkbox_ng/launcher/merge_reports.py
@@ -84,7 +84,8 @@ class MergeReports:
                     self.job_list.append(JobDefinition(result))
                 elif mode == "dict":
                     self.job_dict[result["id"]] = JobDefinition(result)
-            for result in data["resource-results"]:
+            # backward compatibility, now resrouce-result are in results
+            for result in data.get("resource-results", []):
                 result["plugin"] = "resource"
                 result["summary"] = result["name"]
                 # 'id' field in json file only contains partial id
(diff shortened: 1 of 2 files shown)


fernando79513 (approved):
LGTM +1!

Hook25:
I'm merging skipping CI because we have no metabox test for this feature. The failure is due to runners being unreliable today

PR #2704, opened 2026-07-17
Also run setup jobs via Run (bugfix)

Hook25:
## Description

Run was ignoring the setup jobs because of an oversight when implementing them. This fixes it so that run also runs them

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-2297

## Documentation

N/A

## Tests

Adds test to test that the function is indeed called


fernando79513 (changes requested):
Can you remove unused imports?

fernando79513 (approved):
LGTM +1!

PR #2705, opened 2026-07-17
Force refresh snapcraft before uploading (bugfix)

Hook25:
## Description

Uploading works for cross but not for native. The reason is that this old version of snapcraft is no longer able to upload to the store but in cross builds the version of snapcraft is not installed on the runner, but inside the docker environment, so the upload action re-installs it to the latest version.

## Resolved issues

N/A

## Documentation

Added a comment

## Tests

I'm uploading yesterdays build just fine with latest snapcraft on my host


fernando79513 (approved):
LGTM +1!

PR #2719, opened 2026-07-22
Pretty skip reason in UI and comments (New)

Hook25:
## Description

Checkbox used to dump the inhibitors in the console and in the comment of skipped tests but this is not really user friendly as they are hard to read and may not be fully relevant when there is more than one. This PR replaces that mechanism with one that writes the actual skip reason for skipped jobs using the new skip reason system introduced recently. 

## Resolved issues

Fixes: https://warthogs.atlassian.net/browse/CHECKBOX-650

## Documentation

N/A

## Tests

Added metabox tests.
New HTML report: [report_remote.html](https://github.com/user-attachments/files/30261234/report_remote.html)
New submission.json: [submission_remote.json](https://github.com/user-attachments/files/30261244/submission_remote.json)

Try this out with this launcher: [example.txt](https://github.com/user-attachments/files/30261284/example.txt) (txt because GH refuses to upload .conf)
Screenshots:
image

image

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/controller.py
+++ b/checkbox-ng/checkbox_ng/launcher/controller.py
@@ -41,6 +41,7 @@ from plainbox.abc import IJobResult
 from plainbox.impl.color import Colorizer
 from plainbox.impl.config import Configuration
 from plainbox.impl.session.state import SessionMetaData
+from plainbox.impl.result_utils import pretty_skip_reason
 from plainbox.impl.session.resume import (
     IncompatibleJobError,
     CorruptedSessionError,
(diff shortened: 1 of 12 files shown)


checkbox-ng/checkbox_ng/launcher/controller.py
```diff
@@ -933,16 +934,16 @@ def wait_for_job(self, dont_finish=False, suppress_output=False):
...
-            print(_("Job cannot be started because:"))
-            for inhibitor in job_state.readiness_inhibitor_list:
-                SimpleUI.yellow_text(" - {}".format(inhibitor))
+        if result and hasattr(result, "skip_reason"):
+            skip_reason = result.skip_reason
+            try:
+                skip_reason_str = pretty_skip_reason(skip_reason)
+                SimpleUI.yellow_text(skip_reason_str)
+            except ValueError:
+                pass  # unable to pretty print skip reason
```
pieqq: should we log anything here in case?

checkbox-ng/checkbox_ng/launcher/run.py
```diff
@@ -212,6 +213,12 @@ def notify_about_verification(self, job):
             print()
 
     def job_cannot_start(self, job, job_state, result):
+        if hasattr(result, "skip_reason") and result.skip_reason:
+            try:
+                print(self.C.YELLOW(pretty_skip_reason(result.skip_reason)))
+                return
+            except ValueError:
+                pass
```
pieqq:
```suggestion
                pass  # unable to pretty print skip reason
```

pieqq (dismissed):
Nice job, and thanks for taking care of the extra details like the HTML report!

Just a suggestion and a question, but it's good to land as is.

checkbox-ng/checkbox_ng/launcher/controller.py
Hook25: I left it bare intentionally because this is backward compatibility for remote and me being paranoid for local. I don't think there is an actual legitimate situation where this can happen beside with an older agent.

pieqq (dismissed):

pieqq (changes requested):
Can you add a quick unit test to check things don't break when no `skip_reason` are provided?

pieqq (approved):

PR #2721, opened 2026-07-22
Add finish and exit option (new)

### LINKED ISSUE #2720: Unable to exit checkbox
### Bug Description

Exiting checkbox without breaking future usage is difficult. When hitting CTRL-C, we have 4 options. 

Image

- Option 1 (Stop the test case) will not exit
- Option 2 (Pause and disconnect) will exit, but on the next run you will continue the test instead of starting fresh.
- Option 4 seems rather dangerous and only gives a technical explanation -- in what situation would anyone want to do that?
- Option 4 is buggy:
https://github.com/user-attachments/assets/0d5ac763-9169-416f-9025-9cfb8302ce59

There is no "just exit the damn thing" that allows to run checkbox again with a new launcher, for example.

### Cert-blocker Test Case

- [ ] cert-blocker

### To Reproduce

See the attach video. Run a test, try to exit, see what happens

### Expected Result

The abre minimum would be for option 4 to do what it claims - not restart a new test but go to the main menu.

### Actual Result

Option 4 starts a new test.

### Environment

Ubuntu, any

### Relevant log output

```shell

```

### Additional context

_No response_

Hook25:
## Description

When running an automated launcher run (so with forced test selection and test plan selection), the user is unable to exit the test run (finalizing it) back to the cli via the Ctrl+c menu. The reason is that the option doesn't exist, the closest option is to select the `"End this test session preserving its data and launch a new one"` and then rapidly cltr+c twice. The reason is that `"End this test session preserving its data and launch a new one"` finalizes the session and starts a new one, that give that the launcher is provided and automated, immediatly starts testing.

This also looks like a bug because when starting checkbox doesn't signal to the user what it is doing, so there is no way to separate resumes from new sessions from new interactive sessions.

This PR fixes both issues:
1. Introduces a new option to finalize and close the controller (fast key: Q)
2. Introduces a new banner when starting a session that explains how that session was started (handy in general but especially in these situations)

## Resolved issues

Fixes: https://github.com/canonical/checkbox/issues/2720

## Documentation

N/A

## Tests

This was manually tested. The banner looks like this:
```
----------------[ Starting new automated session via launcher ]-----------------
```
This is the launcher used to play around with it
```
#!/usr/bin/env checkbox-cli
#[ui]
#type=silent
[launcher]
launcher_version = 1
stock_reports = text
[test plan]
unit=2021.com.canonical.certification::tired_test_plan
forced = yes
[test selection]
forced = yes
```

Note: a metabox test for this is WIP but it is pretty tricky to develop so lets see if I manage

### DIFF
--- a/checkbox-ng/checkbox_ng/urwid_ui.py
+++ b/checkbox-ng/checkbox_ng/urwid_ui.py
@@ -24,6 +24,7 @@
 
 import os
 import time
+from enum import Enum
 
 from gettext import gettext as _
 
@@ -828,6 +829,15 @@ class TestPlanBrowser:
             return None
 
 
+class InterruptDialogAnswer(Enum):
+    CANCEL = "cancel"
+    KILL_COMMAND = "kill-command"
+    KILL_CONTROLLER = "kill-controller"
+    KILL_AGENT = "kill-agent"
+    FINALIZE = "finalize"
+    FINALIZE_EXIT = "finalize-exit"
+
+
 def interrupt_dialog(host):
     palette = [
         ("body", "light gray", "black", "standout"),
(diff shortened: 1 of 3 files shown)


checkbox-ng/checkbox_ng/urwid_ui.py
```diff
@@ -837,17 +847,24 @@ def interrupt_dialog(host):
...
-        _("End this test session preserving its data and launch a new one"),
-    ]
+        InterruptDialogAnswer.KILL_CONTROLLER: _(
+            "Pause the test session and disconnect from the agent (CTRL+C)"
+        ),
+        InterruptDialogAnswer.KILL_AGENT: _(
+            "Exit and stop the Checkbox agent on the DUT at {}".format(host)
+        ),
+        InterruptDialogAnswer.FINALIZE: _(
+            "End this test session preserving its data and launch a new one"
```
antlassagne: This phrasing is still a bit confusing, since
- if you are using a launcher, it might run the same test again from scratch
- if you're not it will go the the main menu

Who is using this? Why not remove it entirely? (and add it back if everyone screams ofc)

Hook25: I think it is mainly used when someone starts a session with a launcher or an app with test selection not forced but messes up the selection or the manifests, so in order to go back and restart the session without re-calling checkbox, that option is used. I'm not sure, lets see what the others think, as far as I'm concerned we can also remove it. It is also pretty useful for manual sessions, where going back to the main menu is a natural flow

pieqq (dismissed):
Looks good. I agree with Antoine, we should remove that other entry. Please check with QA to know if they are actually using it at all.

pieqq (approved):

PR #2723, opened 2026-07-23
Make shebang rely on env to find python3 instead of hardcoded it (Bugfix)

rickwu666666:
## Description
Make shebang rely on env to find python3 instead of hardcoded it.

## Resolved issues
We found that UC26 does not have Python 3 pre-installed. This will cause Python 3 scripts using the hardcoded shebang `#!/usr/bin/python3` to fail.


baconYao (approved):
LGTM

PR #2634, opened 2026-07-01
Migrate to use libnvml to check nvlink status directly (BugFix)

hanhsuan:
## Description

With nvidia-driver-580, the output of nvidia-smi nvlink --status was empty. However, with nvidia-driver-595, nvidia-smi nvlink --status now returns something, which broke the `prime_offload_test.py` and we don't want to depend on the output from `nvidia-smi nvlink -s` anymore. Therefore, migrating to use `ctype` would be a easier way to skip to add more dependencies into checkbox. 
At the same time, add `__GLX_VENDOR_LIBRARY_NAME=mesa` for all non-nvidia to make sure the default is set to the right one.
 

## Resolved issues

https://github.com/canonical/checkbox/issues/2560

## Tests

26.04 (I+N): https://certification.canonical.com/hardware/202001-27667/submission/498212/
24.04 (I+N): https://certification.canonical.com/hardware/202001-27667/submission/498219/
22.04 run 24.04 test plan (I+N) : https://certification.canonical.com/hardware/202001-27667/submission/498211/
22.04 with NVLINK: https://certification.canonical.com/hardware/202303-31329/submission/498220/
24.04 (A+N): https://certification.canonical.com/hardware/202603-38521/submission/498366/

PR #2727, opened 2026-07-23
Allow different nvidia-smi nvlink -s outputs when it exits 0 (BugFix)

antlassagne:
On some devices, nvidia-smi nvlink -s returns 0, but with the following output: GPU 0: NVIDIA JMJWOA-Generic-GPU (UUID: GPU-483cccbe-9d80-0d7b-5c48-12e6c8696e50) Device does not have or support Nvlink
This does not seem to be a valid reason to system.exit()

### DIFF
--- a/providers/base/bin/prime_offload_tester.py
+++ b/providers/base/bin/prime_offload_tester.py
@@ -253,15 +253,19 @@ class PrimeOffloader:
             ):
                 raise SystemExit("System isn't on-demand mode")
 
-            # prime offload couldn't running on nvlink active or inactive
-            # Therefore, only return empty string is supported environment.
+            # Prime offload cannot run when NVLink is active or inactive.
+            allowed_outputs = [
+                "does not have or support nvlink",
+            ]
             nvlink = subprocess.check_output(
                 ["nvidia-smi", "nvlink", "-s"], universal_newlines=True
             )
             if nvlink:
-                if "error" in nvlink.lower():
-                    raise SystemExit("nvidia driver error")
-                raise SystemExit("NVLINK detected")
+                nvlink_lower = nvlink.lower()
+                if not any(
+                    output in nvlink_lower for output in allowed_outputs
+                ):
+                    raise SystemExit("NVLINK detected")
         except FileNotFoundError:
             self.logger.info(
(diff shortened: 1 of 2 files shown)


Samielakkad (commented):
Retracted.

antlassagne:
> The allowed-output check is too broad for multi-GPU output.
> 
> Right now any occurrence of ?does not have or support nvlink? lets the whole `nvidia-smi nvlink -s` output pass:
> 
> ```python
> if not any(output in nvlink_lower for output in allowed_outputs):
>     raise SystemExit("NVLINK detected")
> ```
> 
> If one GPU reports ?does not have or support Nvlink? but another GPU in the same output reports active/inactive NVLink status, this will accept the machine even though NVLink was detected. The test only covers a single unsupported device.
> 
> I?d parse the output per GPU/section, or at least add a regression test with mixed unsupported + active-link output and require that to still fail.

Hello @Samielakkad, thank you for the review,
Yes I don't consider this ready, I just needed a branch to test it on devices. I can only bring my piece of the puzzle, and your reasoning is valid. I looked for other devices and found one with more GPUs. AFAIU these outputs appeared with GPU driver 595, in case you are also trying.

The newer output I have is 
```
$ nvidia-smi nvlink -s
GPU 0: NVIDIA RTX PRO 2000 Blackwell (UUID: GPU-3727e782-18b9-9991-2bd3-383fd7e4eafc)
Device does not have or support Nvlink
GPU 1: NVIDIA RTX PRO 2000 Blackwell (UUID: GPU-76c47329-a3b9-3716-2763-80779c9b5279)
Device does not have or support Nvlink
GPU 2: NVIDIA GB300 (UUID: GPU-4205f1fd-9c05-ef26-2a33-34578900205c)
NVML: Unable to retrieve Nvlink information as all links are inActive
```
So what are the requirement in multi-gpu setups? If a single gpu is compatible with nvlink then we simply error out and we'll see about a more proper fix when that happens?
To be fair, this test is for portable machines which tend to have only one GPU .But yeah, in theory, it could happen.

antlassagne:
The option we'll take might be [this one](https://github.com/canonical/checkbox/pull/2634)

PR #2728, opened 2026-07-23
More debuggable UsageExpectation (bugfix)

Hook25:
## Description

Currently when we receive a UsageExpectation error in an issue it is really hard to know what went wrong. The reason is that the usage expectations are modified by any function but there is no way to know how we got to a given set. This PR starts tracking the modifications and includes them in the exception. Additionally the error now is clear about it being an error in Checkbox (the old error was weird for our users) and to report it to us.

On the implementation side, the old implementation compared the function code and had to jump through a few hoops to get the actual calling function and its code (given that it could have been decorated). This new implemtation removes all that and just uses the function name. This makes:
- The code way easier to read 
- The code faster (no inspect, no f_code comparison, no frozenset etc. with way less pointless debug logging)
- No difference in my opinion, UA are scoped to the instance, if the function name matches it is the same function (*) 

> (\*) Technically false, it may be a different function if the bound function was changed at runtime, but I don't see a legitimate reason to do that, especially in the SessionAssistant 

## Resolved issues

N/A

## Documentation

N/A

## Tests

To test this remove some UsageExpectation from (for example) `_get_allowed_calls_in_normal_state`. You will see the new error with a history of modifications.

Here is a sample of the new error:
```
Uh, oh...

If you see this message then there is a bug somewhere in Checkbox. We are
sorry for this. Please report this to us.

You are not expected to call SessionAssistant.get_job at this time.
The set of allowed calls, at this time, is:

 - call SessionAssistant.get_job_state() to to access the state of any job.
 - call SessionAssistant.export_to_transport() to to export the results and send them.
 - call SessionAssistant.export_to_file() to to export the results to a file.
 - call SessionAssistant.export_to_stream() to to export the results to a stream.
 - call SessionAssistant.finalize_session() to to mark the session as complete.
 - call SessionAssistant.get_session_id() to to get the id of currently running session.
 - call SessionAssistant.finish_bootstrap() to to finish bootstrapping.

The last 5 modifications were done by (most recent last):

 - SessionAssistant.use_alternate_configuration
 - SessionAssistant.select_test_plan
 - SessionAssistant.start_bootstrap
 - SessionAssistant.finish_bootstrap
 - SessionAssistant.prepare_resume_session
```

### DIFF
--- a/checkbox-ng/plainbox/impl/developer.py
+++ b/checkbox-ng/plainbox/impl/developer.py
@@ -74,6 +74,10 @@ class UnexpectedMethodCall(Exception):
             a part of a string that looks like this: ``' - call {fn_name}() to
             {why}.'``. Developers should use explanations that look natural in
             this context. This text is not meant for internationalization.
+        :param history:
+            A sequence of functions that modified the allowed_pairs and how
+            they did so (fn_name, action), where action is "allow", "disallow",
+            "allow_all_clear", "allow_all_no_clear"
         """
         self.cls = cls
         self.fn_name = fn_name
@@ -91,7 +95,8 @@ class UnexpectedMethodCall(Exception):
                 )
                 for allowed_fn_name, why in self.allowed_pairs
             ),
-            modification_history=" - " + "\n - ".join(self.history),
+            modification_history=" - "
+            + "\n - ".join("{} ({})".format(*x) for x in self.history),
             modification_size=str(MODIFICATION_HISTORY),
         )
 
@@ -155,7 +160,7 @@ class UsageExpectation:
             A string that explains why the function is allowed to be called.
         """
         self._allowed_calls[function.__name__] = reason
-        self._modified(current_function)
+        self._modified(current_function, "allow")
 
     def disallow(self, current_function: Callable, function: Callable):
         """
@@ -168,7 +173,7 @@ class UsageExpectation:
             The function that is now disallowed to be called.
         """
         del self._allowed_calls[function.__name__]
-        self._modified(current_function)
+        self._modified(current_function, "disallow")
 
     def allow_all(
         self,
@@ -195,15 +200,20 @@ class UsageExpectation:
         }
         if clear:
             self._allowed_calls = str_function_reason
+            action = "allow_all, clear"
         else:
             self._allowed_calls.update(str_function_reason)
-        self._modified(current_function)
+            action = "allow_all, don't clear"
+        self._modified(current_function, action)
 
-    def _modified(self, current_function):
+    def _modified(self, current_function, action):
         self.history.append(
-            "{}.{}".format(
-                current_function.__self__.__class__.__name__,
-                current_function.__name__,
+            (
+                "{}.{}".format(
+                    current_function.__self__.__class__.__name__,
+                    current_function.__name__,
+                ),
+                action,
             )
         )
 
--- a/checkbox-ng/plainbox/impl/developer.py
+++ b/checkbox-ng/plainbox/impl/developer.py
@@ -20,6 +20,14 @@
 """Support code for enforcing usage expectations on public API."""
 
 from collections import deque
+from collections.abc import Callable
+
+import sys
+
+if sys.version_info < (3, 9):
+    from typing import Dict
+else:
+    Dict = dict
 
 __all__ = ("UsageExpectation",)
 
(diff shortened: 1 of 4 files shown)


checkbox-ng/plainbox/impl/developer.py
```diff
@@ -157,79 +129,51 @@ def __init__(self, cls):
...
+            self._allowed_calls.update(function_reason)
+        self._modified(current_function)
+
+    def _modified(self, current_function):
+        self.history.append(
+            "{}.{}".format(
+                current_function.__self__.__class__.__name__,
+                current_function.__name__,
+            )
+        )
```
fernando79513: When you get a modification method from the history, you don't know if it was allowed or disallowed. So if the last stack in your call is:
```
        UsageExpectation.of(self).disallow(
            self.start_new_session,
            self.get_test_plan
        )
```
it will fail with something like:

```
The last 5 modifications were done by (most recent last):

 - SessionAssistant.configure_application_restart
 - SessionAssistant.configure_application_restart
 - SessionAssistant.use_alternate_configuration
 - SessionAssistant.start_new_session
 - SessionAssistant.start_new_session
```
but you don't know if the last interacion enabled it or disabled it.
I think it's free to have a more informative logging of the action here:

```python
    def allow(self, current_function, function, reason):
        self._allowed_calls[function.__name__] = reason
        self._modified(current_function, "allowed")
```

```suggestion
    def _modified(self, current_function, action):
        self.history.append(
            "{}.{} -> {}".format(
                current_function.__self__.__class__.__name__,
                current_function.__name__,
                action,
            )
        )
```

checkbox-ng/plainbox/impl/developer.py
```diff
@@ -157,79 +129,51 @@ def __init__(self, cls):
...
 
-    def enforce(self, back=1):
+    def enforce(self, current_function):
         """
         Enforce that usage expectations of the caller are met.
```
fernando79513, line 222: I think a clearer explanation here would help to understand the purpose of all of this UsageExpectation. I hope I'm not making it worse
```suggestion
        Enforce that usage expectations of the caller are met. This method
        checks if the caller is allowed to call the current function by looking
        at the dictionary of allowed calls.
```

checkbox-ng/plainbox/impl/developer.py
```diff
@@ -157,79 +129,51 @@ def __init__(self, cls):
             something goes wrong.
         """
         self.cls = cls
-        self.allowed_calls = {}
+        self._allowed_calls = {}
+        self.history = deque([], maxlen=MODIFICATION_HISTORY)
+
+    def allow(self, current_function, function, reason):
```
fernando79513: If the current function is only here for traceability, I'd add a comment to indicate. Maybe this is too verbose, but It took me a while to get it right. Probably its because of the heat and the late hours, but I think a bit of clarification won't hurt.

```suggestion
    def allow(self, current_function, function, reason):
        """
        Allow the current object to call the given function.

        :param current_function:
            The function that called this method. This is used to record the history
            of modifications.
        :param function:
            The function that is allowed to be called.
        :param reason:
            A string that explains why the function is allowed to be called.
        """
```

fernando79513 (changes requested):
Overall, great improvement, It took me a while to undestand it, but I can get it now, before it was too cryptic, XD

I have added a couple of comments, so we can maybe clarify it even more.

checkbox-ng/plainbox/impl/developer.py
Hook25: I thought about this, but ultimately a function is either going to do one or the other, so I decided to go with the easy thing and if you want the action you go look at the function. Especially because even if you know something was allowed or disallowed, you dont know what was, you just know that function modified that way the expectations.

Hook25: Re-reading your comment I think you are missinterpreting what is being tracked here. This isn't tracking the functions that are allowed or disallowed, that is worthless, this is tracking the series of calls that got you to the current allowed set because when debugging the session assistant, that is the most important information. Usually what happens is that due to a weird combination of calls, you get yourself stuck in a situation where you can't do anything beside Uh Ohing, and when that happens, it is really hard to go back and reproduce it.

checkbox-ng/plainbox/impl/developer.py
Hook25: done

checkbox-ng/plainbox/impl/developer.py
Hook25: done, to all, also added type hints

checkbox-ng/plainbox/impl/developer.py
fernando79513: I get that the most important thing of this modified history dict is to gather which functions called the history list last and modified it, but I think the method that they called is still good to have.
Even if it's just to track that in the code, I think it could be useful.

You may have something like this:
```
        if WEIRD_CONDITION:
            UsageExpectation.of(self).allow= (
                current_func,
                func,
                "weird_condition is met",
            )
        else:
            UsageExpectation.of(self).disallow = {
                current_func,
                func,
            }
```
Does this make sense? Maybe that's not a real use case, because we may never want to explicitly disallow, but we are using something similar here:
https://github.com/canonical/checkbox/blob/9ae657ab0e4b2b3888134ac41375a5cb02654682/checkbox-ng/plainbox/impl/session/assistant.py#L612

Hook25: done

fernando79513 (approved):
LGTM +1!

PR #2729, opened 2026-07-23
Allow setting agent log level (infra)

Hook25:
## Description

While debugging issues (especially with both agent and controller in the same tty), info level logging is a bit too verbose. Give that it is now the default, we have to add a mechanism to tune it down when necessary. This doesn't change the behaviour, just adds the possibility to:
- set debug in the `--log-level` (for uniformity)
- set warning to tune the logging down

## Resolved issues

N/A

## Documentation

Adde Help to explain

## Tests

This prints a lot more (just a bit if you don't start a session):
```
export ALLOW_CHECKBOX_AGENT_NONROOT=1; checkbox-cli run-agent --log-level DEBUG
```
This prints a bit less (a lot less if you start a session):
```
export ALLOW_CHECKBOX_AGENT_NONROOT=1; checkbox-cli run-agent --log-level INFO
```
This prints very little
```
export ALLOW_CHECKBOX_AGENT_NONROOT=1; checkbox-cli run-agent --log-level WARNING
```

### DIFF
--- a/checkbox-ng/checkbox_ng/launcher/agent.py
+++ b/checkbox-ng/checkbox_ng/launcher/agent.py
@@ -20,21 +20,14 @@ functionality.
 """
 
 import gettext
-import json
 import logging
 import os
 import socket
-import sys
 
-from checkbox_ng import app_context
 from checkbox_ng.utils import set_all_loggers_level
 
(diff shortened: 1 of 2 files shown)


p-gentili (approved):
I like it, thanks!

PR #2737, opened 2026-07-30
Update kernel_taint_test bit 2 message (BugFix)

### LINKED ISSUE #2736: taint test shows incorrect info for Bit 2
### Bug Description

The kernel taint test uses old messaging for Bit 2 based on an old kernel spec.  newer kernels changed this messaging and the test needs to change to match to make review easier.

### Cert-blocker Test Case

- [x] cert-blocker

### To Reproduce

Run kernel taint test on any modern CPU (especially recent GNR/SRF/CWF CPUs) and they all throw Bit 2 due to some microcode errors.

### Expected Result

Taint test now still fails on bit 2 but now provides the correct error message (the SMP error message was removed from the kernel taint definitions).

### Actual Result

See above

### Environment

Ubuntu on recent hardware in the server lab

### Relevant log output

```shell

```

### Additional context

```
Kernel taint value is 4101
Taint bit value: 0 (proprietary module was loaded)
*   Proprietary modules found, but they are expected and OK
Taint bit value: 2 (SMP kernel oops on an officially SMP incapable CPU)
Taint bit value: 12 (externally-built ('out-of-tree') module was loaded)
*   Out of Tree modules found, but they are expected and OK
```

bladernr:
## Description
Update the messaging for Bit 2 to use the message in current kernel headers.

## Resolved issues

resolves #2736
resolves CHECKBOX-2312

## Documentation
NA

## Tests
NA

### DIFF
--- a/providers/base/bin/kernel_taint_test.py
+++ b/providers/base/bin/kernel_taint_test.py
@@ -124,7 +124,7 @@ def main():
     taint_meanings = [
         "proprietary module was loaded",
         "module was force loaded",
-        "SMP kernel oops on an officially SMP incapable CPU",
+        "CPU is running outside the specification limits",
         "module was force unloaded",
         "processor reported a Machine Check Exception (MCE)",
         "bad page referenced or some unexpected page flags",


pieqq:
Do you have a pointer to where these messages are defined? I searched the Linux source code but couldn't find either.

Hook25 (approved):
+1 Thanks

PR #2742, opened 2026-08-12
remove manifest in the resource jobs (Bugfix)

stanley31huang:
## Description
remove manifest in the following resource jobs, because of having the manifest in a resource job will cause the manifest won't be applied in current checkbox session.
- iiotg-tsn/tsn-device-ip-match
- ce-oem-monitor/backlight_panel_list


rickwu666666 (approved):
LGTM

PR #2743, opened 2026-08-12
Also export resource results in junit (bugfix)

Hook25:
## Description

This is necessary because we modified the submission but forgot to change it here, so our jenkins, that uses junit, shows different results than C3/TO

## Resolved issues

N/A

## Documentation

N/A

## Tests

Tested locally, it is the same change we did to the checkbox.json template


fernando79513 (approved):
LGTM +1!

PR #2759, opened 2026-08-17
Comment out algif_aead test from test-plan-crypto (Bugfix)

EstebanVg15:
## Description
 
* Currently, when trying to run the `af_alg_aead_gcm_aes_test` Checkbox test case, the following error is displayed.
  ```
    File "/tmp/nest-cvrw56sd.33bd35500f1701f6e719f7f391c8f609fb6fe71c8729b9164c609ead8d8a2f9b/af_alg_test.py", line 15, in create_alg
      sock.bind((crypto_type, name))
  FileNotFoundError: [Errno 2] No such file or directory
  
  During handling of the above exception, another exception occurred:
  
    File "/tmp/nest-cvrw56sd.33bd35500f1701f6e719f7f391c8f609fb6fe71c8729b9164c609ead8d8a2f9b/af_alg_test.py", line 219, in <module>
      main()
  Starting AF_ALG type aead_gcm_aes...
  Error: [Errno 2] No such file or directory
    File "/tmp/nest-cvrw56sd.33bd35500f1701f6e719f7f391c8f609fb6fe71c8729b9164c609ead8d8a2f9b/af_alg_test.py", line 215, in main
      getattr(crypto_test, "test_{}".format(args.type))()
    File "/tmp/nest-cvrw56sd.33bd35500f1701f6e719f7f391c8f609fb6fe71c8729b9164c609ead8d8a2f9b/af_alg_test.py", line 106, in test_aead_gcm_aes
      with self.create_alg("aead", "gcm(aes)") as algo:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/nest-cvrw56sd.33bd35500f1701f6e719f7f391c8f609fb6fe71c8729b9164c609ead8d8a2f9b/af_alg_test.py", line 20, in create_alg
      raise Exception(
  Exception: Error: Not able to use algorithm
  Please check kernel config!
  ```
* This is because `af_alg_aead_gcm_aes_test` runs the `af_alg_test.py` script which creates `AF_ALG` sockets, implicitly loading the `algif_aead` module, and at the moment this fails when trying to test `aead_gcm_aes` since the `/etc/modprobe.d/disable-algif_aead.conf` prevents the `algif_aead` module installation due to: https://ubuntu.com/security/CVE-2026-31431
  ```
  ubuntu@ubuntu:~$ cat /etc/modprobe.d/disable-algif_aead.conf
  # Disable algif_aead module due to CVE-2026-31431 (AKA copy.fail)
  # This will likely be re-enabled in a subsequent update once an updated
  # kernel has been deployed.
  # Blacklisting the module isn't sufficient, we need to do as below:
  install algif_aead /bin/false
  ```
  
* Additionally, as of now, not all kernels have the CVE-2026-3143 kmod fix implemented, so to prevent vulnerabilities, it is better to skip this test.

* This PR proposes commenting out `af_alg_aead_gcm_aes_test` from the crypto test plans, so it is not executed, preventing the `af_alg_test.py script` to try to load the `algif_aead` module. 

## Resolved issues

https://warthogs.atlassian.net/browse/PERI-1483

## Documentation

https://ubuntu.com/security/CVE-2026-31431

## Tests

The `ce-oem-crypto-automated` test plan was executed on a DUT which has the modified `checkbox-provider-ce-oem` side-loaded, and as can be seen the `af_alg_aead_gcm_aes_test` has not been called.

[nvidia-jetson-agx-thor-ce-oem-crypto-automated.txt](https://github.com/user-attachments/files/31322258/nvidia-jetson-agx-thor-ce-oem-crypto-automated.txt)

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/test-plan-crypto.pxu
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/test-plan-crypto.pxu
@@ -23,7 +23,7 @@ include:
     ce-oem-crypto/cryptsetup_benchmark
     ce-oem-crypto/af_alg_hash_crc64_test
     ce-oem-crypto/af_alg_hash_sha256_test
-    ce-oem-crypto/af_alg_aead_gcm_aes_test
+    # ce-oem-crypto/af_alg_aead_gcm_aes_test - CVE-2026-31431
     ce-oem-crypto/af_alg_skcipher_cbc_aes_test
     ce-oem-crypto/af_alg_rng_stdrng_test
 
@@ -52,6 +52,6 @@ include:
     after-suspend-ce-oem-crypto/cryptsetup_benchmark
     after-suspend-ce-oem-crypto/af_alg_hash_crc64_test
     after-suspend-ce-oem-crypto/af_alg_hash_sha256_test
-    after-suspend-ce-oem-crypto/af_alg_aead_gcm_aes_test
+    # after-suspend-ce-oem-crypto/af_alg_aead_gcm_aes_test - CVE-2026-31431
     after-suspend-ce-oem-crypto/af_alg_skcipher_cbc_aes_test
     after-suspend-ce-oem-crypto/af_alg_rng_stdrng_test


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/crypto_generic.pxu
```diff
@@ -54,9 +67,10 @@ category_id: crypto
 flags: also-after-suspend
 estimated_duration: 10
 requires:
-    (cryptoinfo.name == 'gcm(aes)' and cryptoinfo.type == 'aead')
+    (cryptoinfo.name == 'gcm(aes)' and cryptoinfo.type == 'aead' and check_algif_aead.loaded == 'yes')
```
DocSepp: A simple "module.name == algif_aead" could probably already do the trick. I don't think we want to force load a module like this.

Imo we rather just skip it if it's not loaded or at the very least force remove it at the end of the test if it wasn't loaded to begin with.

DocSepp (commented):
Hey Esteban,

thanks for the PR!
I think we can probably simplify this however, refer to the comment I left.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/crypto_generic.pxu
EstebanVg15: Hello @DocSepp ,

CC @pieqq .


Thanks for your suggestion Seb.

I agree that something like below can do the trick in a simpler way (I did some tests with this today).

```
+    module.name == 'algif_aead'
 imports:
     from com.canonical.contrib import cryptoinfo
+    from com.canonical.certification import module
```

However `module.name == 'algif_aead'` will only check if the `algif_aead` module is already loaded (it will not load it), so in most of the cases the test will just be skipped.

Even though this seems to be correct, it is important to say that the `af_alg_test.py` script called by `af_alg_aead_gcm_aes_test` loads the `algif_aead` module automatically on runtime.

This behavior can be seen on the following old Checkbox [submission](https://certification.canonical.com/hardware/202508-37856/submission/483444/) (done before CVE-2026-31431), where [af_alg_aead_gcm_aes_test](https://certification.canonical.com/hardware/202508-37856/submission/483444/test/202983/result/57816891/) succeeded, but if you download the Submission files, you will see that the `algif_aead` module was NOT listed by the `module.name` resource, meaning that `af_alg_test.py` loaded the module on the go.

So, I think that we can proceed with the proposed **modprobe** approach.

Regarding your suggestion, I think that is better not to remove the module on the resource, since my `check_algif_aead`  resource is intended to work with `af_alg_aead_gcm_aes_test` and will not affect it since the test case also loads the module by itself, however if we add `rmmod` on a system where the module was installed by default it could cause runtime issues in my opinion.

stanley31huang: the af_alg_test.py won't loads the module automatically on runtime. So I think we could use the module to do the trick.
https://github.com/canonical/checkbox/blob/main/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/af_alg_test.py

EstebanVg15: Hello @stanley31huang ,

CC @zongminl , @DocSepp ,

Thanks for your response,

Even that the `af_alg_test.py` script does not directly call `insmod` or `modprobe`, it does install the module **implicitly**.

As can be seen on the following example, I bypassed the installation restriction on `/etc/modprobe.d/disable-algif_aead.conf` just for demonstrative purposes, and before running `af_alg_test.py` the `algif_aead` module was not installed, but after running the script the module was automatically loaded.

```
ubuntu@ubuntu:~$ cat /etc/modprobe.d/disable-algif_aead.conf
# Disable algif_aead module due to CVE-2026-31431 (AKA copy.fail)
# This will likely be re-enabled in a subsequent update once an updated
# kernel has been deployed.
# Blacklisting the module isn't sufficient, we need to do as below:
# install algif_aead /bin/false

ubuntu@ubuntu:~$ lsmod | grep algif_aead

ubuntu@ubuntu:~$ /snap/checkbox-ce-oem/7425/providers/checkbox-provider-ce-oem/bin/af_alg_test.py --type aead_gcm_aes
Starting AF_ALG type aead_gcm_aes...
aead: c3b3c41f113a31b73d9a5cd432103069

ubuntu@ubuntu:~$ lsmod | grep algif_aead
algif_aead             12288  0
af_alg                 32768  1 algif_aead
```

That's why I originally proposed the **modprobe** approach.

On the other hand we can proceed with `module.name == 'algif_aead'` but just considering that there will be cases where the test was originally executed, and now it would be skipped if the module is not loaded prior the text execution, I would just like to double check, what do you think?

stanley31huang: Hi @EstebanVg15 ,

Good catch, I didn't noticed that the `af_alg` module will be loaded by execute the `af_alg_test.py` scripts

However, I am thinking to revise the test with following approach, please let me know if that make sense to you.
1. Skip the `af_alg_aead_gcm_aes_test` test due to the test fail by CVE issue you metioned. We could simply apply the way provides by @DocSepp 
2. Adding a new test case to ensure the `af_alg` module is not loaded. If that module been loaded, that means there's a security concerns in that device.

And once the `af_alg` security has been fixed, we could simply remove the this patch.

BTW, one thing I don't understand that why you want to load the module in the resource job?

EstebanVg15: Hello @stanley31huang ,

Thanks for your feedback, considering the inputs we discussed in our internal conversation, I proceeded commenting out the `ce-oem-crypto/af_alg_aead_gcm_aes_test case` from the test plan without modifying the test case itself. Could you please let me know if this looks good to you?

stanley31huang (approved):
LGTM

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/crypto_generic.pxu
stanley31huang: LGTM

PR #2768, opened 2026-08-19
Add vainfo feature-set check against DUT manifest (New)

ahresse:
## Description

The DUT manifest declares the VA-API codec features the device is expected to support, one encoder and decoder entry per VAProfile. This adds a resource job that parses the profiles and entry points advertised by vainfo and checks them in both directions against the manifest.

The resource job emits one record per (profile, direction) pair mapped to its manifest feature. A template generates one pass/fail job per feature so the driver cannot silently advertise an undeclared codec, and a reverse job fails when a declared feature is not exposed by the driver.

## Resolved issues

This is in preparation of: https://github.com/canonical/checkbox/commits/ffmpeg/
Since it relies on vainfo, we want to make sure vainfo is right.

Somehow also related to: https://warthogs.atlassian.net/browse/PEK-2747

## Documentation

Documentation is not affected by these changes.

## Tests

Sanity-check the resource parser on the local machine (should list the profiles the local driver advertises, each mapped to a has_*_decoder/_encoder feature):
cd providers/base/bin && ./va_profile_check.py resource
./va_profile_check.py forward has_h264_high_decoder   # fails (exit 1) with no DUT manifest
./va_profile_check.py reverse                          # passes (exit 0) with no declared feature


providers/base/units/va-api/jobs.pxu
```diff
@@ -10,3 +10,53 @@ plugin: shell
...
+command: va_profile_check.py forward {feature}
+estimated_duration: 0.1
+_summary: Verify that {profile_name} {direction} is declared in the DUT manifest
+_purpose:
+    The VA driver advertises this profile and direction through vainfo. This
+    test checks that the DUT manifest declares the corresponding feature, so
+    that the driver is not exposing capabilities that were not expected on
+    this device.
+
+id: va-api/va-manifest-vs-driver
```
mckees: This job should skip if there are no valid manifest entries rather than the current test pass. For instance, I have a machine manifest from your multimedia branch which includes `"com.canonical.contrib::has_av1_encoder_vaapi": true`. That will obviously not get picked up because the name has changed, but on first run, I thought it had because the job passed simply by having 0 cases

providers/base/units/va-api/test-plan.pxu
```diff
@@ -17,4 +17,7 @@ _name: Hardware video acceleration tests (automated)
 unit: test plan
 include:
     va-api/va-initialize
+    va-api/vainfo
```
mckees: I had to move this to the bootstrap_include  section to get all of the jobs to actually run as intended

providers/base/units/va-api/jobs.pxu
```diff
@@ -10,3 +10,53 @@ plugin: shell
...
+
+unit: template
+template-resource: va-api/vainfo
+template-unit: job
+plugin: shell
+category_id: va-api
+imports: from com.canonical.plainbox import manifest
+id: va-api/va-profile-{profile}-{direction}
+template-id: va-api/va-profile-profile-direction
+requires:
```
mckees: Might be good to add ` requires: manifest.{feature} == 'True'` here?

providers/base/bin/va_profile_check.py
```diff
@@ -0,0 +1,192 @@
...
+    for feature, entry in sorted(features.items()):
+        print("profile: {}".format(entry["profile"]))
+        print("profile_name: {}".format(entry["profile_name"]))
+        print(
+            "entrypoints: {}".format(", ".join(sorted(entry["entrypoints"])))
+        )
+        print("direction: {}".format(entry["direction"]))
+        print("feature: {}".format(feature))
+        print()
+    return 0
```
mckees, line 123: Do we need this to return 0?

providers/base/bin/va_profile_check.py
```diff
@@ -0,0 +1,192 @@
...
+    if missing:
+        print(
+            "ERROR: the following features are declared in the DUT manifest "
+            "but not supported by the VA driver:"
+        )
+        for feature in missing:
+            print("  {}".format(feature))
+        return 1
+    print("OK: all manifest-declared features are supported by the VA driver")
+    return 0
```
mckees, line 168: Same here, is this needed?

providers/base/units/va-api/jobs.pxu
```diff
@@ -10,3 +10,53 @@ plugin: shell
...
+    package.name == 'vainfo'
+    cpuinfo.platform == 'x86_64'
+    manifest.has_va_api == 'True'
+command: va_profile_check.py forward {feature}
+estimated_duration: 0.1
+_summary: Verify that {profile_name} {direction} is declared in the DUT manifest
```
mckees: This ends up running long in some cases. 84 characters here:
Verify that VAProfileH264ConstrainedBaseline decoder is declared in the DUT manifest

Warnings at runtime:
```
warning: /var/tmp/checkbox-providers/base/units/va-api/jobs.pxu:28-47: job 'va-api/va-profile-h264_constrained_baseline-decoder', field 'summary', please stay under 80 characters
warning: /var/tmp/checkbox-providers/base/units/va-api/jobs.pxu:28-47: job 'va-api/va-profile-h264_constrained_baseline-encoder', field 'summary', please stay under 80 characters
```

providers/base/tests/test_va_profile_check.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def test_false_declared_features_are_ignored(self):
+        with patch(
+            "va_profile_check.run_vainfo",
+            return_value=VAINFO_OUTPUT,
+        ):
+            with patch(
+                "va_profile_check.get_manifest",
+                return_value={
+                    NAMESPACE + "::has_h264_main_decoder": False,
```
mckees: This one is present in the vainfo test output, so it is impossible for the test to fail because it will never add a supported feature to the  "missing" list in cmd_reverse().

providers/base/tests/test_va_profile_check.py
```diff
@@ -0,0 +1,293 @@
...
+class TestMain(TestCase):
+    def test_resource(self):
+        with patch(
+            "va_profile_check.run_vainfo",
+            return_value=VAINFO_OUTPUT,
+        ):
+            with redirect_stdout(StringIO()):
+                self.assertEqual(main(["resource"]), 0)
+
+    def test_forward_declared(self):
```
mckees: This looks identical to test_declared_feature_succeed

providers/base/tests/test_va_profile_check.py
```diff
@@ -0,0 +1,293 @@
...
+        with patch(
+            "va_profile_check.get_manifest",
+            return_value={NAMESPACE + "::has_h264_high_decoder": True},
+        ):
+            with redirect_stdout(StringIO()):
+                self.assertEqual(main(["forward", "has_h264_high_decoder"]), 0)
+
+    def test_forward_undeclared(self):
```
mckees: Looks like a duplicate of test_undeclared_feature_fails

mckees (changes requested):
Some feedback here. Once you address these, I'd have copilot take a look in balanced mode. The changes here are complex enough that I don't think the copilot lite version was sufficient

providers/base/units/va-api/jobs.pxu
ahresse: I think, adding `requires: manifest.{feature} == 'True'` to the forward per-profile tests would stop the "vainfo advertises an undeclared codec" case from ever running, defeating the forward test's purpose.
Thus, I would keep forward job running for all advertised profiles so undeclared codecs are caught.

providers/base/bin/va_profile_check.py
ahresse: main() does [return cmd_resource()](https://github.com/canonical/checkbox/pull/2768/changes/BASE..602099213d99e78d6fefd9cc1fe811944ddb3692#diff-511b25d2047f14b89fe9681a0263dd82836cfabd0a2f898a932003a463feca50R183) and [sys.exit(main())](https://github.com/canonical/checkbox/pull/2768/changes/BASE..602099213d99e78d6fefd9cc1fe811944ddb3692#diff-511b25d2047f14b89fe9681a0263dd82836cfabd0a2f898a932003a463feca50R192). 

I think, a python script should exit 0 on success. (+ cmd_resource has no failure path) So, the explicit return 0 is what feeds sys.exit. It is needed for a clean exit code.

providers/base/bin/va_profile_check.py
ahresse: Same as above comment

providers/base/units/va-api/test-plan.pxu
ahresse: resolved in https://github.com/canonical/checkbox/commit/b2c5e7f8c164dec1b67fe771adda4f3f7ebcb17a

providers/base/units/va-api/jobs.pxu
ahresse: resolved in https://github.com/canonical/checkbox/commit/4860e21239094f8bb475624781978bf6b671a8d9

providers/base/tests/test_va_profile_check.py
ahresse: resolved in https://github.com/canonical/checkbox/commit/ad694d2cf4a1a32b7664b1b54d95d1ef788291a6

providers/base/tests/test_va_profile_check.py
ahresse: resolved in https://github.com/canonical/checkbox/commit/ad694d2cf4a1a32b7664b1b54d95d1ef788291a6

providers/base/tests/test_va_profile_check.py
ahresse: resolved in https://github.com/canonical/checkbox/commit/b416a0779bd1ead61b6d32b5947cde9d71a77a7a

providers/base/units/va-api/jobs.pxu
ahresse: resolved in https://github.com/canonical/checkbox/commit/c0ce88a26c01f7def65d494e08c4ae529d0ec9a3

ahresse:
Thank you for the feedback  @mckees : your feedback is mostly resolved: see comments.

I requested a review from copilot balanced.

mckees (dismissed):
LGTM. Aside from the review, I ran the test plan and ensured that anything present in vainfo but not the manifest fails the test. Anything declared "false" explicitly which is actually supported by the driver also fails correctly. I inserted a manual, nonexistent "profile3" for AV1 encode and made sure that it also failed when the driver doesn't advertise that it exists.

mckees:
@ahresse I just noticed that you've used pxu files. Checkbox has started moving to .yaml files instead. Please use `checkbox-cli translate` to convert them

mckees (approved):
Re-ran testing after the PXU to YAML conversions, and all LGTM

ahresse:
@canonical/hardware-certification-solutions-qa : Anyone can take a look at this?

providers/base/bin/va_profile_check.py
```diff
@@ -0,0 +1,202 @@
...
+            profile, entrypoint = line.split(":")
+        except ValueError:
+            continue
+        profiles.append((profile.strip(), entrypoint.strip()))
+    return profiles
+
+
+def profile_to_feature_id(profile):
+    """Turn 'VAProfileHEVCMain10' into 'hevc_main10'."""
+    name = profile[len("VAProfile") :]
```
pieqq:
```suggestion
    name = profile[len("VAProfile"):]
```

I think this is more pep8-aligned.

providers/base/bin/va_profile_check.py
```diff
@@ -0,0 +1,202 @@
...
+    "VAEntrypointIDCT",
+    "VAEntrypointMoComp",
+    "VAEntrypointDeblocking",
+)
+
+ENCODER_ENTRYPOINTS = (
+    "VAEntrypointEncSlice",
+    "VAEntrypointEncSliceLP",
+    "VAEntrypointEncPicture",
+    "VAEntrypointFEI",
```
pieqq: Does these come from some guidelines document? If so, please document that here, because I assume this will evolve with time and it's good to have a reference.

providers/base/bin/va_profile_check.py
```diff
@@ -0,0 +1,202 @@
...
+
+def cmd_forward(feature):
+    if feature_declared(feature):
+        print("OK: {} is declared as supported on this device".format(feature))
+        return 0
+    print(
+        "ERROR: the VA driver advertises {} but this feature is not "
+        "declared in the DUT manifest".format(feature)
+    )
+    return 1
```
pieqq: Regarding this, we have [some guidance](https://github.com/canonical/checkbox/tree/main/providers#python-tests) that mentions

> Don't `print('error message')` and `exit(1)`. `raise SystemExit('error message')` instead.

Please stick to it.

providers/base/bin/va_profile_check.py
```diff
@@ -0,0 +1,202 @@
...
+        help="check that all manifest-declared features are advertised",
+    )
+    args = parser.parse_args(argv)
+    if args.command == "resource":
+        return cmd_resource()
+    if args.command == "forward":
+        return cmd_forward(args.feature)
+    if args.command == "reverse":
+        return cmd_reverse()
+    return 2
```
pieqq: You don't need this, argparse already handles this for you.

providers/base/units/va-api/jobs.yaml
```diff
@@ -0,0 +1,63 @@
+id: va-api/va-initialize
+summary: Detect if the VA API could be loaded
+category_id: va-api
+imports:
+  - from com.canonical.plainbox import manifest
+requires:
+  - package.name == 'vainfo'
```
pieqq: We usually recommend to use `executable.name` instead of `package.name` in these situations, because snap packaging sometimes gets in the way with how packages are determined (and because your `vainfo` executable might come from a snap at some point in the future...).

(This is a general suggestion that applies for all the `requires` fields in the yaml files of this PR.)

pieqq (changes requested):
Sorry for the long wait, since it had been marked as approved we hadn't realized it had not been reviewed by our team yet.

See my inline comments below.

One thing I noticed is that none of the manifest ids noted here are currently defined in the [Feature Set reference in C3](https://github.com/canonical/hexr/blob/main/data/features.yaml). Do you plan to have a PR to update them?

PR #2772, opened 2026-08-20
Fix ptp4l arg interpolation (BugFix)

tomli380576:
## Description

When expanding the ptp4largs variable, we shouldn't quote it because doing so would glue all the arguments into 1 string, which isn't recognized by ptp4l. 

```
+ timeout 30s ptp4l '-i enp2s0 -m -s --network_transport=L2 --tx_timestamp_timeout=5'
+ tee /var/tmp/ptp4l_output.txt
failed to create a clock
```

Also added --transportSpecific=1 because the default server config from ptp4l expects it. See `/usr/share/doc/linuxptp/configs/automotive-master.cfg`

## Tests
Passing result: https://certification.canonical.com/hardware/202601-38351/submission/505713/test/202445/result/62204193/


baconYao (approved):
LGTM. Thanks

PR #2773, opened 2026-08-20
Adopt checkbox-support interface (New)

Hook25:
## Description

This is the last major step to get us to drop dev-mode. After this, a lot of minor things have to be changed but the biggest hurdles are solved. As a side effect this will make apparmor way less angry with us.

Minor: This also lands a small PATH fix to the snap apps.

## Resolved issues

Fixes: CHECKBOX-1846

## Documentation

N/A

## Tests

To test this install the snap (still in devmode) and connect the interface. Run smoke local/remote. Nothing should change. 

Snap built here: 
- https://github.com/canonical/checkbox/actions/runs/32350554365
- https://github.com/canonical/checkbox/actions/runs/32350702746


p-gentili (approved):
I tested the updated PATH by installing `checkbox26`, enabling the `.agent`, and running the `executable` resource job via remote. It listed binaries included in `/snap/bin` only, such as firefox.

The new interface is not testable quite yet.

PR #2774, opened 2026-08-20
Run nxp_ele_list_tv_files when has_nxp_secure_element_ele is set (BugFix)

EstebanVg15:
## Description

 * Not all the hardware devices have a NXP EdgeLock Secure Enclave (ELE) which is intended for i.MX processors, it is better for this test case only to be executed when `has_nxp_secure_element_ele == 'True'`.

## Resolved issues

https://warthogs.atlassian.net/browse/PERI-1566

## Tests

The `nxp_ele_list_tv_files` Resource was called in a Jetson Orin NX DUT (which does not have a NXP Secure Element) having the manifest property enabled and disabled, as can be seen the test is skipped when it is set as "false".

* [Jetson_Orin_NX_has_nxp_secure_element_ele_false.txt](https://github.com/user-attachments/files/31281609/Jetson_Orin_NX_has_nxp_secure_element_ele_false.txt)
* [Jetson_Orin_NX_has_nxp_secure_element_ele_true.txt](https://github.com/user-attachments/files/31281612/Jetson_Orin_NX_has_nxp_secure_element_ele_true.txt)

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/nxp-secure-element.pxu
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/crypto/nxp-secure-element.pxu
@@ -1,4 +1,6 @@
 id: nxp_ele_list_tv_files
+imports: from com.canonical.plainbox import manifest
+requires: manifest.has_nxp_secure_element_ele == 'True'
 plugin: resource
 _summary: Enumerate available NXP ELE test vector files
 _description:


stanley31huang:
Hi @EstebanVg15 ,

Thank you for your help with fixing the nxp_ele_list_tv_files resource. Based on the current Checkbox design, resource jobs should always return an exit code of 0, and they should not depend on the manifest job. Instead, we will use a separate check job to verify that the resource is configured properly. I have submitted a PR to address this issue (https://github.com/canonical/checkbox/pull/2781).

EstebanVg15:
> Hi @EstebanVg15 ,
> 
> Thank you for your help with fixing the nxp_ele_list_tv_files resource. Based on the current Checkbox design, resource jobs should always return an exit code of 0, and they should not depend on the manifest job. Instead, we will use a separate check job to verify that the resource is configured properly. I have submitted a PR to address this issue (#2781).

Hello @stanley31huang ,

Understood, is good to know that you've fixed this, thank you a lot! considering that I will close this PR.

PR #2788, opened 2026-08-25
Use builtin sorted_keys insted of sort by hand (bugfix)

Hook25:
## Description

OrderedDict doesn't work when there are nested collections (like in yaml siblings) so the checksum can mismatch on older python versions

## Resolved issues

Fixes: CHECKBOX-2314

## Documentation

N/A

## Tests

N/A


fernando79513 (approved):
LGTM +1!

PR #2789, opened 2026-08-25
Apply blocker to fwts critical only tests in 26.04 stress tests (New)

tomli380576:
## Description

This is the final change we want to achieve: make fwts critical failures cert blockers, but not the remaining ones (high failures and lower)

## Resolved issues

Prevents a fwts low failure from failing blocker tests.

## Documentation

We will do the migration to yaml after this is merged because this file contains too many other test plans not relevant to the reboot tests.


tomli380576:
wait the base branch is wrong

PR #2794, opened 2026-08-26
Fix wrong include of rs485 auto (Bugfix)

rickwu666666:
## Description
The automated serial test accidentally included the RS485 stress test previously.

### DIFF
--- a/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/serial/test-plan.pxu
+++ b/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/serial/test-plan.pxu
@@ -23,30 +23,6 @@ include:
     ce-oem-serial/serial-transmit-data-tests
     ce-oem-serial/serial-console-tests
 
-id: before-suspend-ce-oem-serial-automated
-unit: test plan
-_name: Before suspend serial automated tests
-_description:
-    Before suspend automated serial tests for devices
-bootstrap_include:
-    ce-oem-serial/serial-list
-    ce-oem-serial/serial-console-list
-include:
-    ce-oem-serial/serial-transmit-data-(?!stress-).*
-    ce-oem-serial/serial-console-(?!list\b).*$
-
-id: after-suspend-ce-oem-serial-automated
-unit: test plan
-_name: After suspend serial automated tests
-_description:
-    After suspend automated serial tests for devices
-bootstrap_include:
-    ce-oem-serial/serial-list
-    ce-oem-serial/serial-console-list
-include:
-    after-suspend-ce-oem-serial/serial-transmit-data-(?!stress-).*
-    after-suspend-ce-oem-serial/serial-console-.*
-
 id: ce-oem-serial-stress
 unit: test plan
 _name: serial stress tests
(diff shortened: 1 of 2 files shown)


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/serial/test-plan.pxu
```diff
@@ -32,7 +32,7 @@ bootstrap_include:
     ce-oem-serial/serial-list
     ce-oem-serial/serial-console-list
 include:
-    ce-oem-serial/serial-transmit-data-.*
+    ce-oem-serial/serial-transmit-data-(?!stress-).*
```
hanhsuan: What is the reason that you can not use [template_id](https://github.com/canonical/checkbox/blob/ba6965ede883ea15709a34b5637881cb8f9c0856/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/serial/jobs.pxu#L124) for this ?

hanhsuan (changes requested):

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/serial/test-plan.pxu
rickwu666666: In previous we want to prevent the test case ordering issue and we split the test plan into before and after suspend. However, you are right we need to make the include part to use template_id instead of using regex to matching job id. But I think that better to do in another PR to go through all of the job ids in ce-oem test plan. What do you think?


https://github.com/canonical/checkbox/blob/main/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/test-plan-ce-oem.pxu#L62
https://github.com/canonical/checkbox/blob/main/contrib/checkbox-ce-oem/checkbox-provider-ce-oem/units/test-plan-ce-oem.pxu#L153

hanhsuan: I thought that it has been done while migrating to the 26.04 . No ?

rickwu666666: Not for ce-oem plan.

hanhsuan: Why not starting this one ?

rickwu666666: Because that require some effort to go through whole test plan. And we can planning that as a task for our engineering focus time. However, the quick fix in this PR can prevent the device from running serial stress test in auto test stage which can save time.

hanhsuan: I could not understand the reason that one line change in this PR will need that much effort.

rickwu666666: I used the template ID as the included job ID and ran a trial run to make sure the after-suspend job actually runs after the auto-suspend job.
```
---------------------------[ Running job 289 / 641 ]----------------------------
---------------------[ Automated test of suspend function ]---------------------       
ID: com.canonical.certification::suspend/suspend_advanced_auto           


---------------------------[ Running job 292 / 641 ]----------------------------                                                  
[ ce-oem-serial/serial-transmit-data-URS485-/dev/ttyLP5-115200 after suspend (S3) ]                                              
ID: com.canonical.contrib::after-suspend-ce-oem-serial/serial-transmit-data-URS485-/dev/ttyLP5-115200
Category: Serial Port        
```

hanhsuan (approved):
LGTM

PR #2797, opened 2026-08-26
Add prebuilt cuda-samples test suite to the base provider (New)

seankingyang:
## Description

This PR adds a new CUDA test suite to the base provider (OEMQA-6843).
It runs every test program from NVIDIA's
[cuda-samples](https://github.com/NVIDIA/cuda-samples) repository as
its own Checkbox job. It works on any device with an NVIDIA GPU,
including Jetson/Tegra boards, and the same jobs work on Ubuntu
classic and Ubuntu Core.

How it works:

- Checkbox does **not** download or build cuda-samples. The tree must
  be cloned and built before the test run (by the test setup on
  classic, or inside a snap on Ubuntu Core). A readiness job checks
  this first and fails early when the tree is missing, not built, or
  at the wrong version.
- The version check: the cuda-samples checkout tag must be the newest
  tag that is not newer than the CUDA version from `nvidia-smi`.
  Example: CUDA 12.6 needs tag `v12.5`, because the next tag is
  `v13.0`.
- The script looks for the tree in this order: `CUDA_SAMPLES_PATH`,
  then `/opt/nvidia/cuda-samples` (NVIDIA's Jetson layout), then
  `/snap/*/current/cuda-samples` (Ubuntu Core snap).
- A resource job lists every sample in the tree. One job is created
  per sample, named `cuda-sample/<samples-category>-<sample>`. Both
  old Makefile trees (<= v12.5) and new cmake trees (>= v13.0) work,
  including the deeper `Samples/8_Platform_Specific/Tegra/` folder.
- Some samples can never run on some devices (for example: they need
  two GPUs, or they need a display). These samples are listed in a
  JSON exclude file, each with a written reason. `exclude_jobs.json`
  is the default list; `exclude_jobs.jetson.json` adds the
  Jetson-only excludes from NVIDIA's own runner, and a project picks
  its list with `CUDA_SAMPLES_EXCLUDE_FILE` in the launcher. A list
  can start from one shipped base list with an `include` key, so no
  entries are copied twice (one include only, no chains, duplicates
  are an error).
- Excluded samples, and samples that were not built, still appear in
  the results as **skipped** jobs with their reason. No sample ever
  disappears silently.
- Each sample runs as the normal user (no root) inside a temporary
  sandbox folder. Because of this, tests also work when the tree is
  owned by root, and a test run never changes the tree itself.
- Every sample has a time limit (default 300 s,
  `CUDA_SAMPLES_TIMEOUT`). Without it, some graphics samples run
  forever on machines without a display.
- 36 unit tests cover the logic. One of them runs the job gate
  expression from `jobs.pxu` through the real plainbox evaluator: the
  parentheses in that expression are required, and the test fails if
  someone removes them (plainbox would then evaluate each part
  separately and the per-sample skip would stop working).

## Resolved issues

Part of OEMQA-6843 (Jetson CUDA test migration). Related: OEMQA-6849
(rootless Tegra device-node setup), OEMQA-6850 (Ubuntu Core
cuda-samples snap).

## Documentation

`providers/base/units/cuda/README.md` documents the prebuilt-only
contract, per-OS setup, environment variables, exclude-list format and
governance, and the sandbox/rootless design.

## Tests

- `manage.py validate` clean (no cuda-related issues in
  `--strict --deprecated`); 36 unit tests green.
- Jetson Orin (riverside, 202406-34151; CUDA 12.6, tree v12.5
  Makefile era): 137 passed / 0 failed / 87 visible skips, 64 min.
  Submission: https://certification.canonical.com/hardware/202406-34151/submission/507019/
- GB300 (202604-38713; CUDA 13.2, tree v13.2 cmake era, in-source):
  136 passed / 0 failed / 67 visible skips, 4 min 35 s.
  Submission:
  https://certification.canonical.com/hardware/202604-38713/submission/507000/
  Full per-job outputs are in the C3 submissions above.


bladernr:
Q: why is this being done in the base provider rather than the gpgpu provider?  IMO we should not really have tests for the same thing in two different places, this just leads down the road to confusion around which provider to put which thing when new GPU tests come up.  The gpgpu provider already has job definitions for a few prebuilt cuda samples that are provided in turn by the cuda-samples snap that @pedro-avalos maintains currently.

To be clear, I like this addition, I just think it needs to be done inside the gpgpu provider to maintain consistency. Plus I would welcome adding these extra samples tests to the gpu test suite for cert as well (this was one of the phase 2 ideas from before anyway)

pedro-avalos:
I would say if this is more reliable, provides wider support, and is easier to maintain than the snap (which I find a pain to manage, especially as it lacks support for some architectures), then I would say, yeah this proposed prebuilt cuda-samples should replace the existing cuda-samples tests (of which there are two in gpgpu provider, I think: cuda-samples snap and I believe there is also one from partner engineering)

providers/base/units/cuda/resource.pxu
```diff
@@ -0,0 +1,16 @@
...
+category_id: cuda
+_summary: Enumerate the samples in the prebuilt cuda-samples tree
+_description:
+ Emits one record per sample source directory in the cuda-samples tree
+ (category/name/path).  Samples listed in the exclude-jobs JSON file
+ are still emitted, with skip set to True and their skip_reason, so
+ the jobs generated for them show up as visible skips instead of
+ silently disappearing from the run.
+estimated_duration: 5s
+imports: from com.canonical.plainbox import manifest
```
rickwu666666, line 13: Can not use manifest in a resource job.

providers/base/bin/cuda_samples.py
```diff
@@ -0,0 +1,603 @@
...
+    silently testing another tree.
+    """
+    env_root = os.environ.get("CUDA_SAMPLES_PATH")
+    if env_root:
+        return [env_root]
+    return [CLASSIC_DEFAULT_ROOT] + sorted(glob.glob(SNAP_ROOT_GLOB))
+
+
+def resolve_root(path=None):
+    """Return the first candidate containing a Samples/ directory."""
```
rickwu666666, line 68: From cuda-samples v13.3, the folder rename from `Samples` to `cpp`.
https://github.com/NVIDIA/cuda-samples/tree/v13.3

rickwu666666:
I've test it on x86 platform with cuda_sapmles v13.3. And seems working.
https://certification.canonical.com/hardware/202607-39081/submission/509302/

providers/base/bin/cuda_samples.py
```diff
@@ -0,0 +1,603 @@
...
+
+def cuda_version():
+    """Return the CUDA version reported by nvidia-smi as (major, minor)."""
+    try:
+        output = subprocess.check_output(
+            ["nvidia-smi"], universal_newlines=True
+        )
+    except (OSError, subprocess.CalledProcessError) as exc:
+        raise SystemExit("Cannot run nvidia-smi: {}".format(exc))
+    match = re.search(r"CUDA Version:\s*(\d+)\.(\d+)", output)
```
rickwu666666, line 296: While I'm using driver 610.57.04, the version pattern changed to `CUDA UMD Version`
```
airport@airport-Nuvo-7000-Series:~$ nvidia-smi 
Fri Sep  4 22:29:43 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 610.57.04              KMD Version: 610.57.04     CUDA UMD Version: 13.3     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA RTX A2000 12GB          On  |   00000000:01:00.0 Off |                  Off |
| 30%   54C    P0             24W /   70W |      12MiB /  12282MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A            2200      G   /usr/bin/gnome-shell                      2MiB |
+-----------------------------------------------------------------------------------------+
```

Issue #1391, opened 2024-08-06
Memory stress test is crashing without any debug logs for the issue.


cvelloth:
### Bug Description

In Desktop Preload Certification Test Suite for 24.04., the stress test memory/memory_stress_ng is failing with the message crashed and terminated. The tar files in the logs, or the html log files do not give any debug data, to understand the cause of failure. How can we debug the cause for the failure ?

In Desktop Preload Certification Test Suite 

### To Reproduce

To Reproduce
Install Checkbox via the following commmands
``` shell
$ sudo snap install checkbox22
$ sudo snap install checkbox --classic
```
Run "memory/memory_stress_ng" job via the following steps
```shell
$ checkbox.checkbox-cli
# Choose "Desktop Preload Certification Tests for 24.04 " test plan
# Select "Stress test of system memory" case in "Stress Tests" set
# Press "T" to perform testing
```

### Environment

- OS: 24.04 Ubuntu Desktop
- Checkbox Type: Snap
- Checkbox version :: 4.1.0.dev25

### Relevant log output

```shell
stress/memory_stress_ng	crashed	blocker		terminated
```

cvelloth:
[submission.json](https://github.com/user-attachments/files/16508504/submission.json)

pieqq:
Hello @cvelloth ! Sorry for the late reply, I missed the issues you opened.

The submission.json you attache shows this:

```json
        {
            "id": "stress/memory_stress_ng",
            "full_id": "com.canonical.certification::stress/memory_stress_ng",
            "name": "Stress test of system memory",
            "certification_status": "blocker",
            "category": "Stress tests",
            "category_id": "com.canonical.plainbox::stress",
            "status": "fail",
            "outcome": "crash",
            "comments": "terminated",
            "io_log": "",
            "type": "test",
            "project": "certification",
            "duration": 4574.1000854969025,
            "plugin": "shell",
            "template_id": null
        }
```

which, as you said, means the test case crashed, and as a result no I/O was logged. (but the test ran for around 1h20, as we see with the `duration` key).

In order to keep investigating this issue, here are a few things you can do:

[...]

PR #2799, opened 2026-08-26
Always set PYTHONUNBUFFERED for jobs and snap apps (bugfix)

Hook25:
## Description

Given the nature of what we are doing in Checkbox, and the fact that any test may brick the machine leaving us with nothing but the log remote was able to salvage, output buffering is antithetical to what we want. This both applies to Checkbox itself but especially our jobs that are mostly written in python. 

Additionally, this is a source of confusion, recurrently pushing people to ask why the job output isn't live printed in remote while in local (given that the session that prints is the session that runs) it was. This partially solves the issue by disabling python output buffering in all jobs and all snaps.

## Resolved issues

Relevant to: https://github.com/canonical/checkbox/issues/1391 (the complaints about missing output)

## Documentation

N/A

## Tests

N/A

### DIFF
--- a/checkbox-core-snap/common_files/config/wrapper_common
+++ b/checkbox-core-snap/common_files/config/wrapper_common
@@ -6,6 +6,8 @@ function append_path() {
   fi
 }
 
+export PYTHONUNBUFFERED=1
+
 if [ -z "$RUNTIME" ]; then
   RUNTIME=$SNAP
 else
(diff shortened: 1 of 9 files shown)


fernando79513 (approved):
LGTM +1!

Hook25:
Merging as all the builds on all arches passed beside a couple who had a runtime fluke (given the nature of the change, there is no difference among them)

PR #2804, opened 2026-08-27
remove manifest from resource job (BugFix)

stanley31huang:
## Description
remove manifest requires field from resource job


rickwu666666 (approved):
LGTM

PR #2806, opened 2026-08-27
Fix salvages skipped (bugfix)

pieqq:
## Description

If a job (job 1) is used to salvage another job (job 2), but job 2 is
being skipped, Checkbox will crash.

This commit prevents this by setting the outcome of job 1 to
`OUTCOME_SKIPPED_DEPENDENCY`.

## Tests

- Add unit test to check `NOT_FAILED_DEP`
- Added a Metabox test


Hook25 (approved):
Thanks

PR #2811, opened 2026-09-01
Add chrony NTP synchronization test (New)

JasonLeonnn:
## Description

The previous `networking/ntp` test depends on `ntpdate` which was deprecated in Ubuntu 15.10 and replaced by systemd-timesyncd. Starting with Ubuntu 25.10, `chrony` has replaced `systemd-timesyncd`.
Because of this, a new test case for NTP synchronization with `chrony`.

## Resolved issues

- Created a new chrony-based script, implement one hour skew, sync using configured NTP sources, and verify.
- Updated networking and dock jobs 

## Documentation

[Jira ticket](https://warthogs.atlassian.net/browse/OEMQA-6861)

## Tests

**Ubuntu 25.10+:** Runs `networking/ntp-chrony`.
**Ubuntu < 25.10:** Falls back to `networking/ntp`.
**Ubuntu Core 26:** Environment check added to automatically skip `networking/ntp-chrony` (UC relies on `systemd-timesyncd` and NTP jobs are excluded from the UC test plan).

**Ubuntu 26.04:** [Result](https://certification.canonical.com/hardware/202511-38064/submission/508835/test-results/)
**Ubuntu 24.04:** [Result](https://certification.canonical.com/hardware/202606-38951/submission/508867/test-results/)
**Ubuntu Core 26:** [Result](https://certification.canonical.com/hardware/202603-38538/submission/509266/test-results/)

### DIFF
--- a/providers/base/bin/network_ntp_chrony_test.py
+++ b/providers/base/bin/network_ntp_chrony_test.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+"""
+Test system clock synchronization using chrony.
+
+The test records the current time, moves the system clock one hour into the
+past using chrony's manual time input, and verifies that chrony can restore
+the correct time from its configured NTP sources. It uses the system's chrony
+configuration instead of contacting a hard-coded NTP server.
+
+The original chrony service state is restored after the test.
+
+Copyright (C) 2026 Canonical Ltd.
+
+Author:
+    Jason Leonhard <jason.leonhard@canonical.com>
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License version 2,
+as published by the Free Software Foundation.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import argparse
+import logging
+import os
+import subprocess
+import sys
+
+DEFAULT_TIMEOUT = 60
+TIME_SKEW_SECONDS = 60 * 60  # 1 hour
+
+
+def run_command(command, timeout=None, check=True):
+    """Run a command and raise an error when it fails."""
+    logging.debug("Running: %s", " ".join(command))
+    result = subprocess.run(
+        command,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+        timeout=timeout,
+    )
+    if check and result.returncode:
+        error = result.stderr.strip() or result.stdout.strip()
+        raise RuntimeError("{}: {}".format(" ".join(command), error))
+    return result
+
+
+def is_chrony_active():
+    """Return whether the chrony service is currently active."""
+    result = run_command(
+        ["systemctl", "is-active", "--quiet", "chrony.service"],
+        check=False,
+    )
+    return result.returncode == 0
+
+
+def skew_time(current_time):
+    """Skew the system time by one hour into the past."""
+    skewed_time = current_time - TIME_SKEW_SECONDS
+    formatted_skewed_time = run_command(
+        [
+            "date",
+            "--date=@{}".format(skewed_time),
+            "+%Y-%m-%d %H:%M:%S",
+        ]
+    ).stdout.strip()
+
+    run_command(["chronyc", "manual", "on"])
+    try:
+        run_command(["chronyc", "settime", formatted_skewed_time])
+        run_command(["chronyc", "-a", "makestep"])
+    finally:
+        run_command(["chronyc", "manual", "reset"])
+        run_command(["chronyc", "manual", "off"])
+
+
+def sync_with_chrony(timeout):
+    """Synchronize the system clock with chrony."""
+    # Re-enable the configured NTP sources.
+    run_command(["chronyc", "online"])
+    # Step the clock if the next update has an offset over 0.1 seconds.
+    run_command(["chronyc", "makestep", "0.1", "1"])
+    # Request one good measurement, with at most four attempts per source.
+    run_command(["chronyc", "burst", "1/4"])
+    # Check once per second, for up to timeout attempts, until synchronized.
+    run_command(
+        ["chronyc", "waitsync", str(timeout), "0.1", "0.0", "1"],
+        timeout=timeout,
+    )
+
+
+def parse_args(argv=None):
+    parser = argparse.ArgumentParser(
+        description="Synchronize the local clock using chrony"
+    )
+    parser.add_argument(
+        "--timeout",
+        default=DEFAULT_TIMEOUT,
+        type=int,
+        help="maximum number of seconds to wait for synchronization",
+    )
+    parser.add_argument(
+        "-d",
+        "--debug",
+        action="store_true",
+        default=False,
+        help="Verbose output for debugging purposes",
+    )
+    return parser.parse_args(argv)
+
+
+def main(argv=None):
+    args = parse_args(argv)
+    logging.basicConfig(
+        format="%(asctime)s %(levelname)s %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+        level=logging.DEBUG if args.debug else logging.INFO,
+    )
+
+    if os.geteuid() != 0:
+        logging.error("You must run this script as root")
+        return 1
+
+    try:
+        # Preserve the service state so the test does not change the system.
+        was_active = is_chrony_active()
+        if not was_active:
+            run_command(["systemctl", "start", "chrony.service"])
+
+        try:
+            # Record the correct time before deliberately changing the clock.
+            current_time = int(run_command(["date", "+%s"]).stdout.strip())
+            current_datetime = run_command(
+                ["date", "+%Y-%m-%d %H:%M:%S"]
+            ).stdout.strip()
+            logging.info("Current system date and time: %s", current_datetime)
+
+            # Move the clock back one hour and apply the change.
+            skew_time(current_time)
+
+            skewed_datetime = run_command(
+                ["date", "+%Y-%m-%d %H:%M:%S"]
+            ).stdout.strip()
+            logging.info("Skewed system date and time: %s", skewed_datetime)
+
+            # Synchronize from NTP and confirm the clock returned to real time.
+            sync_with_chrony(args.timeout)
+
+            synchronized_time = int(
+                run_command(["date", "+%s"]).stdout.strip()
+            )
+            if synchronized_time < current_time:
+                raise RuntimeError("Failed to synchronize the system time")
+            synchronized_datetime = run_command(
+                ["date", "+%Y-%m-%d %H:%M:%S"]
+            ).stdout.strip()
+            logging.info(
+                "Synchronized system date and time: %s",
+                synchronized_datetime,
+            )
+        finally:
+            # Restore the service state found before the test.
+            if not was_active:
+                run_command(["systemctl", "stop", "chrony.service"])
+    except (OSError, RuntimeError, subprocess.TimeoutExpired) as error:
+        logging.error("Time synchronization failed: %s", error)
+        return 1
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/providers/base/bin/network_ntp_chrony_test.py
+++ b/providers/base/bin/network_ntp_chrony_test.py
@@ -7,7 +7,7 @@ past using chrony's manual time input, and verifies that chrony can restore
 the correct time from its configured NTP sources. It uses the system's chrony
 configuration instead of contacting a hard-coded NTP server.
 
-The original chrony service state is restored after the test.
+The chrony service must already be running before the test starts.
 
 Copyright (C) 2026 Canonical Ltd.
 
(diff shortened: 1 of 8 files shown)


providers/base/units/networking/jobs.pxu
```diff
@@ -70,10 +70,9 @@ plugin: shell
 category_id: com.canonical.plainbox::networking
 id: networking/ntp
 flags: also-after-suspend
-requires:
-    (package.name == 'ntpdate' or package.name == 'ntpsec-ntpdate')
+requires: package.name == 'chrony'
 user: root
-command: network_ntp_test.py
+command: network_ntp_chrony_test.py
 _purpose: Test to see if we can sync local clock to an NTP server
```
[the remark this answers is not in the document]
tomli380576: Copilot is right about this, `chrony` doesn't exist on 24.04. Maybe add `lsb.release >= "26.04"` in the job definition or provide a fallback test for 24.04

providers/base/units/networking/packaging.pxu
```diff
@@ -4,13 +4,8 @@ Depends: net-tools, ethtool
...
 unit: packaging meta-data
 os-id: ubuntu
-os-version-id: < 22.04
-Depends: ntpdate
-
-unit: packaging meta-data
-os-id: ubuntu
-os-version-id: >= 22.04
-Depends: ntpsec-ntpdate
+os-version-id: >= 25.10
```
tomli380576: This is probably not necessary, we shouldn't depend on `chrony` to install checkbox. I'm not too sure though, checkbox team please chime in on this.

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+import subprocess
+import sys
+
+DEFAULT_TIMEOUT = 60
+TIME_SKEW_SECONDS = 60 * 60  # 1 hour
+
+
+def run_command(command, timeout=None, check=True):
+    """Run a command and raise an error when it fails."""
+    logging.debug("Running: %s", " ".join(command))
```
tomli380576: Use `logger = logging.getLogger(__name__)` then call `logger`'s methods instead of the module level ones.

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+        command,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+        timeout=timeout,
+    )
+    if check and result.returncode:
+        error = result.stderr.strip() or result.stdout.strip()
+        raise RuntimeError("{}: {}".format(" ".join(command), error))
+    return result
```
tomli380576: [Base provider's style guide](https://github.com/canonical/checkbox/blob/main/providers/README.md#python-tests) explicitly discourages making subprocess wrappers. Directly call `subprocess.run` instead.

tomli380576: if it looks really long in the code, you can `import subprocess as sp` then call `sp.run`

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+    if os.geteuid() != 0:
+        logging.error("You must run this script as root")
+        return 1
+
+    try:
+        # Preserve the service state so the test does not change the system.
+        was_active = is_chrony_active()
+        if not was_active:
+            run_command(["systemctl", "start", "chrony.service"])
```
tomli380576: I think we should immediately exit with 1 here. If chrony isn't even running then there's something wrong with the image.

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+        try:
+            # Record the correct time before deliberately changing the clock.
+            current_time = int(run_command(["date", "+%s"]).stdout.strip())
```
tomli380576: optional: use `time.time()` so you don't have to call all the string methods

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+                ["date", "+%Y-%m-%d %H:%M:%S"]
+            ).stdout.strip()
+            logging.info(
+                "Synchronized system date and time: %s",
+                synchronized_datetime,
+            )
+        finally:
+            # Restore the service state found before the test.
+            if not was_active:
+                run_command(["systemctl", "stop", "chrony.service"])
```
tomli380576: same thing here, we probably shouldn't mutate the service state

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+def is_chrony_active():
+    """Return whether the chrony service is currently active."""
+    result = run_command(
+        ["systemctl", "is-active", "--quiet", "chrony.service"],
+        check=False,
+    )
+    return result.returncode == 0
+
+
+def skew_time(current_time):
```
tomli380576: optional: add type annotations wherever possible. Doing so enables static analysis tools and linters to catch errors without even running the code.

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,181 @@
...
+    run_command(["chronyc", "online"])
+    # Step the clock if the next update has an offset over 0.1 seconds.
+    run_command(["chronyc", "makestep", "0.1", "1"])
+    # Request one good measurement, with at most four attempts per source.
+    run_command(["chronyc", "burst", "1/4"])
+    # Check once per second, for up to timeout attempts, until synchronized.
+    run_command(
+        ["chronyc", "waitsync", str(timeout), "0.1", "0.0", "1"],
+        timeout=timeout,
+    )
```
tomli380576: same as above, directly call `subprocess.run` or `subprocess.check_call` since we are not using the return value.

tomli380576 (changes requested):

clairlin53:
@JasonLeonnn Please provide the submission test result

providers/base/units/networking/jobs.pxu
```diff
@@ -71,12 +71,25 @@ category_id: com.canonical.plainbox::networking
...
 _summary: Test NTP server synchronization capability.
 
+plugin: shell
+category_id: com.canonical.plainbox::networking
+id: networking/ntp-chrony
+flags: also-after-suspend
+requires:
+    lsb.release >= '25.10'
+    package.name == 'chrony'
```
tomli380576: This `package.name` check should probably be removed because it's a dependency of `ubuntu-minimal`. If we don't have `chrony` then something is very wrong. I think you can add a check in the python file if you really want to check this.

tomli380576 (changes requested):
sry just 1 more small comment

providers/base/units/networking/jobs.pxu
baconYao: Agree the perspective of @tomli380576.

There are two things I would like to know
1. Don't we need a manifest for this job?
2. Have we considered the Core Image environment?

baconYao (changes requested):

providers/base/units/networking/jobs.pxu
JasonLeonnn: @tomli380576 You're right, I'll just remove the unnecessary package check since `chrony` is expected by default.

@baconYao Answering your question,
1. Since `chrony` is not optional in this case and missing `chrony` should fail this test, no manifest is needed for this job.
2. I don't think we should consider the Core Image environment since the `networking/ntp` is excluded in the UC26 test plan. So might need to add this new `chrony` test job as excluded job in UC26 test plan.

providers/certification-client/units/client-cert-iot-ubuntucore-26.pxu
```diff
@@ -214,6 +214,8 @@ exclude:
   miscellanea/grub_file_check
   networking/ntp
   after-suspend-networking/ntp
+  networking/ntp-chrony
+  after-suspend-networking/ntp-chrony
```
baconYao: 1. I understand that there may be a reason why `networking/ntp` and `after-suspend-networking/ntp` are excluded from the Ubuntu Core 26 test plan, but I don’t think that alone is sufficient justification for excluding `networking/ntp-chrony` as well.

2. What I’d like to clarify is whether the current Ubuntu Core 26 environment has also deprecated `systemd-timesyncd` and switched to `chrony`, as described in your issue. If so, I think this test case would also be applicable to the Core 26 environment, and the requirements would need to be updated accordingly. If not, we need to understand why UC26 is still using `systemd-timesyncd`.

baconYao (changes requested):

providers/certification-client/units/client-cert-iot-ubuntucore-26.pxu
JasonLeonnn: UC26 has not switched to `chrony` and still uses `systemd-timesyncd` by default. [Ubuntu Core Docs](https://documentation.ubuntu.com/core/how-to-guides/manage-ubuntu-core/set-system-time/).

- `systemd-timesyncd` is integrated directly into `systemd` within immutable base snap and operates strictly as a lightweight SNTP client, avoiding the need for a separate daemon like `chrony`.
- System time process is tightly coupled to the [boots process](https://documentation.ubuntu.com/core/explanation/core-elements/inside-ubuntu-core/#ref-inside-ubuntu-core-inside-ubuntu-core).

Since the Core and Classic are assembled differently and require different approaches, the `chrony` job is appropriate for Classic, but not for Core environment.

If the NTP test coverage is required for Ubuntu Core in the future, I think creating a separate Core-specific job and script using `systemd-timesyncd` would be the better approach.

hanhsuan: We do not prefer to use `excluded` to skip one job, this will make the maintenance getting harder and harder. If the UC26 doesn't align with ubuntu 26.04, the lsb.release >= 25.10 is not a good condition check for your new test cases.

JasonLeonnn: Fixed, remove the test from Core26 exclusion and add environment check to skip the test in Core environment.

baconYao (dismissed):
LGTM

tomli380576 (dismissed):
LGTM

providers/base/units/dock/jobs.pxu
```diff
@@ -653,6 +653,7 @@ category_id: dock-network
 plugin: user-interact-verify
 depends: dock/cold-plug
 requires:
+    lsb.release < '25.10'
```
Hook25:
```suggestion
    os.release < '25.10'
```

providers/base/units/dock/jobs.pxu
```diff
@@ -667,6 +668,26 @@ _steps:
 _verification:
     Was the test able to synchronize system time using the Ethernet connection?
 
+id: dock/networking-ntp-chrony
+category_id: dock-network
+plugin: user-interact-verify
+depends: dock/cold-plug
+requires:
+    lsb.release >= '25.10'
```
Hook25:
```suggestion
    os.release >= '25.10'
```

providers/base/units/dock/jobs.pxu
```diff
@@ -2681,6 +2702,7 @@ plugin: user-interact-verify
 imports: from com.canonical.plainbox import manifest
 requires:
     manifest.has_dock_ethernet_adapter == 'True'
+    lsb.release < '25.10'
```
Hook25:
```suggestion
    os.release < '25.10'
```

providers/base/units/dock/jobs.pxu
```diff
@@ -2701,6 +2723,36 @@ _siblings:
...
+id: dock/all-init-networking-ntp-chrony
+category_id: dock-network
+depends:
+    dock/all-init
+    dock/all-init-ethernet-detect
+plugin: user-interact-verify
+imports: from com.canonical.plainbox import manifest
+requires:
+    manifest.has_dock_ethernet_adapter == 'True'
+    lsb.release >= '25.10'
```
Hook25:
```suggestion
    os.release >= '25.10'
```

providers/base/units/networking/jobs.pxu
```diff
@@ -71,12 +71,25 @@ category_id: com.canonical.plainbox::networking
 id: networking/ntp
 flags: also-after-suspend
 requires:
+    lsb.release < '25.10'
```
Hook25:
```suggestion
    os.release < '25.10'
```

providers/base/units/networking/jobs.pxu
```diff
@@ -71,12 +71,25 @@ category_id: com.canonical.plainbox::networking
...
 command: network_ntp_test.py
 _purpose: Test to see if we can sync local clock to an NTP server
 _summary: Test NTP server synchronization capability.
 
+plugin: shell
+category_id: com.canonical.plainbox::networking
+id: networking/ntp-chrony
+flags: also-after-suspend
+requires:
+    lsb.release >= '25.10'
```
Hook25:
```suggestion
    os.release >= '25.10'
```

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,177 @@
...
+        if synchronized_time < current_time:
+            raise RuntimeError("Failed to synchronize the system time")
+        synchronized_datetime = time.strftime(
+            "%Y-%m-%d %H:%M:%S", time.localtime(synchronized_time)
+        )
+        logger.info(
+            "Synchronized system date and time: %s", synchronized_datetime
+        )
+    except sp.CalledProcessError as error:
+        logger.error(
```
Hook25: Minor: consider `raise SystemExit`

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,177 @@
...
+            "Time synchronization failed: %s",
+            error.output.strip() if error.output else error,
+        )
+        return 1
+    except (OSError, RuntimeError, sp.SubprocessError) as error:
+        logger.error("Time synchronization failed: %s", error)
```
Hook25: Minor: consider `raise SystemExit`

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,177 @@
...
+def main(argv=None) -> int:
+    args = parse_args(argv)
+    logging.basicConfig(
+        format="%(asctime)s %(levelname)s %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+        level=logging.DEBUG if args.debug else logging.INFO,
+    )
+
+    if os.geteuid() != 0:
+        logger.error("You must run this script as root")
```
Hook25: Minor: consider `raise SystemExit`

providers/base/bin/network_ntp_chrony_test.py
```diff
@@ -0,0 +1,177 @@
...
+        return 1
+
+    if not is_chrony_active():
+        logger.error("Chrony service is not active")
```
Hook25: Minor: consider `raise SystemExit`

Hook25 (dismissed):
Consider the following:

Issue #2802, opened 2026-08-27
Test the server cert suite in general using the snaps


mreed8855:
### Enhancement Proposal

A request has been made to ensure the certification-server test suite can be installed  using snaps.

PR #2816, opened 2026-09-01
Enable the test launchers for canonical certification server

mreed8855:
https://github.com/canonical/checkbox/issues/2802

Symlink checkbox-provider-certification-server (and any other provider's) launcher/ scripts, e.g. test-usb, certify-ubuntu-server, into /usr/local/bin from the install and post-refresh hooks of the classic checkbox snap.

Also install a plain wrapper script at /usr/local/bin/checkbox-cli that execs 'snap run checkbox.checkbox-cli', so the launchers' '#!/usr/bin/env checkbox-cli' shebang resolves without manual setup. snapd has no hook API to create a 'snap alias' programmatically, so a wrapper script is used instead.

This removes the need to manually symlink or alias these launchers after installing checkbox26/checkbox24/etc alongside the classic checkbox snap. The remove hook cleans up only what it created.

### DIFF
--- a/checkbox-snap/common_series_classic/snap/hooks/install
+++ b/checkbox-snap/common_series_classic/snap/hooks/install
@@ -14,3 +14,27 @@ ResultAny=yes
 ResultInactive=yes
 ResultActive=yes
 EOF
+
+# Make "checkbox-cli" resolvable on PATH (including under sudo, since
+# Ubuntu's default secure_path already includes /usr/local/bin).
+# Provider launcher scripts (e.g. test-usb, certify-ubuntu-server) use
+# "#!/usr/bin/env checkbox-cli" and rely on this being on PATH. There is
+# no snapd hook API to create a "snap alias" programmatically, so we
+# install a plain wrapper script that runs "snap run" directly instead.
+cat <<'EOF' > /usr/local/bin/checkbox-cli
+#!/bin/sh
+exec snap run checkbox.checkbox-cli "$@"
+EOF
+chmod +x /usr/local/bin/checkbox-cli
+
+# Expose the provider launcher scripts (test-*, certify-*, etc.) shipped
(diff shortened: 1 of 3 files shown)


Hook25:
Hello. Please update your PR so that it maches the PR description template. This not only makes our life easier but also reminds you of the checklist of things you need to do or communicate to us, like, for example, how you tested this change.

With that being said, the deprecation of classic snaps is coming, are you sure:
- you want to use them for this "new" adventure for the server team
- you want to include this hack in them instead of using snap apps properly
- you want to also hack the shebang that way instead of asking for an alias if you really want it (note, you dont need it if you use a snap app)
- you still want to use checkbox local?

Marking this as draft pending update of title and description

bladernr:
Just my comment for why we're looking at this (and trying to help get rid of the Debs)

> With that being said, the deprecation of classic snaps is coming, are you sure:
> 
> you want to use them for this "new" adventure for the server team

The end goal is to try to move to snaps, regardless, this started by exploratory snap installing the current checkbox stuff and trying to run them on Server and finding the gaps between using the snaps to run the various server tests and the current mechanism in Debs.

> you want to include this hack in them instead of using snap apps properly

Happy to use snap apps, we need to look into that instead, this started with a presumption of "Weve not really ever used snaps for server before, lets figure out how, and what is broken"

> you want to also hack the shebang that way instead of asking for an alias if you really want it (note, you dont need it if you use a snap app)

Alias probably better, but as you imply Snap App even better than that.

> you still want to use checkbox local?

We still need to be able to run these locally on a machine. We cannot guarantee that testers at 20+ different OEMs have a means to run things remotely, their infra or laptops are Windows, or MacOS or whatever.  Local execution of some manner (even if it's running control and remote on he same system connecting over localhost, or whatever) is needed, since we cannot control what hardware is being utilized in other companies other than the server being tested and potentially a machine running MAAS.

PR #2827, opened 2026-09-04
add qualcomm ai ml tests (New)

stanley31huang:
## Description
Migrate the Dragonwing AI ML tests case into checkbox-ce-oem.

gstreamer pipeline will access /dev/fastrpc device node, and this node is only available for root permission. But it should be available for normal user. So I side-load the checkbox with `user: root` to ensure the test commands are working.

## Tests
https://certification.canonical.com/hardware/202605-38721/submission/509247/


rickwu666666 (approved):
LGTM

PR #2830, opened 2026-09-04
Avoid crashing on None time duration (bugfix)

Hook25:
## Description

If estimated duration is None, the html exporter crashs. This makes it so the default qualifiers no only removes empty values but also Nones (as undefines)

## Resolved issues

N/A

## Tests

Run the new Basic Manual... test plan. Before this bugfix when exporting the HTML the exporter crashes
After, it doesn't.


pieqq (approved):
Let's land this, but it's weird that Checkbox can spit up `None` durations for jobs.

PR #2831, opened 2026-09-07
include the dragonwing tests for core 24 test plan (New)

stanley31huang:
## Description
include the dragonwing tests for core 24 test plan


rickwu666666 (approved):
LGTM

PR #2864, opened 2026-09-15
Add the Vulkan test via vulkaninfo tool (New)

baconYao:
## Description

To verify whether a device supports Vulkan, the vulkaninfo tool can be used to list all available GPU devices and to validate they are hardware GPU not CPU (sofrware renderer).

It also supports using another external JSON definition file to override or extend the executable binary and LD_LIBRARY_PATH.

For detailed usage instructions, please refer to the [README.md](https://github.com/canonical/checkbox/pull/2864) file.

## Resolved issues

Support the platform check for Vulkan by using vulkaninfo tool (need the vulkan-tools package).

## Documentation

https://warthogs.atlassian.net/browse/OEMQA-6878

## Tests

See https://warthogs.atlassian.net/browse/OEMQA-6878


contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/vulkaninfo_test.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def flush():
+        if current_number is not None:
+            records.append(
+                {
+                    "device_number": current_number,
+                    "device_name": current_fields.get("deviceName", ""),
+                    "device_type": current_fields.get("deviceType", ""),
+                }
+            )
```
tomli380576: this is technically correct, but mutable references inside closures should be avoided or passed in as parameters

maybe we can refactor with pure-ish functions

```py
def _iter_gpu_blocks(output: str) -> "list[tuple[int, list[str]]]":
    blocks: "list[tuple[int, list[str]]]" = []
    device_number: "int | None" = None
    field_lines: "list[str]" = []

    for line in output.splitlines():
        stripped_line = line.strip()
        if stripped_line.startswith("GPU") and stripped_line.endswith(":"):
            candidate = stripped_line[len("GPU") : -1].strip()
            if candidate.isdigit():
                if device_number is not None:
                    blocks.append((device_number, field_lines))
                device_number, field_lines = int(candidate), []
                continue

        if device_number is not None:
            field_lines.append(line)

    if device_number is not None:
        blocks.append((device_number, field_lines))

    return blocks


def _build_record(
    device_number: int, field_lines: "list[str]"
) -> VulkaninfoRecord:
    fields = {}
    for line in field_lines:
        if not line.startswith("\t") or "=" not in line:
            continue
        key, _, value = line.strip().partition("=")
        fields[key.strip()] = value.strip()
    return {
        "device_number": device_number,
        "device_name": fields.get("deviceName", ""),
        "device_type": fields.get("deviceType", ""),
    }


def parse_vulkaninfo_summary(output: str) -> "list[VulkaninfoRecord]":
    return [
        _build_record(device_number, field_lines)
        for device_number, field_lines in _iter_gpu_blocks(output)
    ]
```

tomli380576: device_number should be an int btw since it's guaranteed to be an index

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/vulkaninfo_test.py
```diff
@@ -0,0 +1,361 @@
...
+
+    full_command = " ".join([command, "--summary"])
+    logger.info("Running command: '%s'", full_command)
+    # stdout and stderr are combined because a crashing driver's error
+    # (e.g. 'Segmentation fault (core dumped)') is reported by the
+    # invoking shell on stderr/stdout without a reliable, consistent
+    # stream.
+    result = subprocess.run(
+        full_command,
+        shell=True,
```
tomli380576, line 187: do we have to use `shell=True`?

tomli380576 (changes requested):

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/vulkaninfo_test.py
baconYao: Good catch! We should avoid the `shell=True`. I will revise it.

contrib/checkbox-ce-oem/checkbox-provider-ce-oem/bin/vulkaninfo_test.py
baconYao: Thanks for your suggestion, will fix it.

baconYao: Done

PR #2871, opened 2026-09-16
Add DRP-AI/DRP-AI3 real-device test coverage for RZ V2L/V2H/V2N (NEW)

rickwu666666:
## Description

Adds real-hardware DRP-AI/DRP-AI3 test coverage for Renesas RZ boards
(V2L/V2H/V2N) to `checkbox-provider-ce-oem`, per the spec attached to
KOTO-774 (`KOTO-774-drp-ai-test-spec.md`).

- Split the driver-level device-node smoke test into two
  generation-specific jobs: `driver-nodes-drp-ai` (V2L, single-core
  DRP-AI) and `driver-nodes-drp-ai3` (V2H/V2N, DRP-AI3), instead of one
  OR-gated job, because manifest.has_rz_drp_ai / has_rz_drp_ai3 gate
  different node sets.
  - Follow-up hardware finding during C3 verification: V2N does **not**
    expose `/dev/drp1`, unlike V2H. Confirmed against Renesas' official
    product pages — RZ/V2H has an extra standalone "Dynamically
    reconfigurable processor (DRP)" IP block alongside DRP-AI3 that
    RZ/V2N lacks. This is an expected chip-level difference, not a bug,
    so `/dev/drp1` is now checked only if present and its absence does
    not fail the job.
- Added a single, merged `drp-ai3-object-counter` job (V2H/V2N) built
  on the upstream Q08 Object Counter sample app, using a new
  `DRP_AI_TEST_SNAP` environ so integrators can point at either board's
  differently-named AI-application snap (`rzv-ai-applications` /
  `rzv-ai-applications-v2n`) while sharing one identical command/args
  set.
  - Runs the sample in `TEST` mode against a static, AI-generated dog
    photo (`data/test.png`, no third-party copyright), and asserts success only when the
    application actually logs `[INFO] dog detected.` — not merely that
    the command exits 0.
  - The app must be spawned through a pty (Python's `pty` module)
    because plain stdout redirection never flushed the detection log
    line even after 60s, while a pty (forcing glibc's automatic
    line-buffering) surfaced it within ~2s consistently on real
    hardware.
  - Explicitly unsets a leaked `GST_PLUGIN_PATH` from the snap
    environment before invoking the app, which otherwise broke the
    GStreamer pipeline.
- Both jobs are included in
  `units/rz/ai-accelerator-test-plan.pxu` (`rz-ai-accelerator-automated`).

- Verification note: the two touched jobs (`driver-nodes-drp-ai3`,
  `drp-ai3-object-counter`) were sideloaded and run with a scratch,
  manifest-stripped copy of the provider (manifest gate only removed
  in the scratch copy, never in the committed source) and run scoped
  via `--exact` to just those two jobs, rather than the whole
  `rz-ai-accelerator-automated` test plan -- running the full plan
  against a manifest-stripped sideload also pulls in the unrelated,
  V2L-only `inference` job on V2H/V2N (its manifest gate is stripped
  too), which would not be representative of production behavior.
  `driver-nodes-drp-ai` (V2L) was validated separately on real V2L
  hardware (CID 202507-36926) during spec drafting.

## Resolved issues

Resolves KOTO-774.

## Tests

- **V2H** (CID 202503-36406): https://certification.canonical.com/submissions/status/403366
- **V2N** (CID 202602-38383): https://certification.canonical.com/submissions/status/403365


rickwu666666:
sideload result for server
https://certification.canonical.com/hardware/202602-38383/submission/511898/test-results/pass/

core
https://certification.canonical.com/hardware/202503-36406/submission/511899/test-results/pass/

PR #2039, opened 2025-07-23
Add the test script to trigger and capture cpuidle state (New)

seankingyang:
## Description
- Add the trigger and capture CPU cpuidle state
- For ARM platform also have the CPU idle feature, and this script is similar fwts cstates with more useful info.
- This script also can be run on X86 platform.

## Resolved issues
Miss coverage of the CPU idle.

## Tests

ARM w/o CPUidle driver (this is a true issue):
```
root@ubuntu:/home/ubuntu/CPU# ./cpu_idle_state.py 
[INFO] Starting CPU idle state test
[INFO] This test checks if all processors have the same number of idle states and if idle state transitions happen during alternating load patterns
[INFO] Found 2 CPUs in system
[ERROR] ERROR: No CPUs with cpuidle support found
root@ubuntu:/home/ubuntu/CPU# echo $?
1
```

ARM w/ CPUidle driver (some state can't entry during the trigger idle state testing but the usage of state w/ existing usage so mark as pass )
```
root@mtk-genio:/home/ubuntu/CPU# ./cpu_idle_state.py 
[INFO] Starting CPU idle state test
[INFO] This test checks if all processors have the same number of idle states and if idle state transitions happen during alternating load patterns
[INFO] Found 8 CPUs in system
[INFO] CPU 0: Found 3 idle states
[INFO] CPU 1: Found 3 idle states
[INFO] CPU 2: Found 3 idle states
[INFO] CPU 3: Found 3 idle states
[INFO] CPU 4: Found 3 idle states
[INFO] CPU 5: Found 3 idle states
[INFO] CPU 6: Found 3 idle states
[INFO] CPU 7: Found 3 idle states
[INFO] Testing 8 CPUs with cpuidle support
[INFO] Testing CPU 0 (1/8)
[INFO] CPU 0: State 2 (clusteroff_l) was used
[INFO] CPU 0: State 0 (WFI) was used
[INFO] CPU 0: State 1 (cpuoff_l) was used
[INFO] CPU 0: All idle states reached, stopping early
[INFO] CPU 0: SUCCESS - All states reached: cpuoff_l, clusteroff_l, WFI
[INFO] Testing CPU 1 (2/8)
[INFO] CPU 1: State 2 (clusteroff_l) was used
[INFO] CPU 1: State 0 (WFI) was used
[INFO] CPU 1: NOTE - States with existing usage (not tested): cpuoff_l
[INFO] Testing CPU 2 (3/8)
[...]

PR #2872, opened 2026-09-16
ce-oem: add ARM CPU idle-state test job (New)

rickwu666666:
## Description

Adds `ce-oem-cpu/idle_state_test` job to checkbox-provider-ce-oem, exercising the vendored `cpu_idle_state.py` (from canonical/checkbox#2039) on ARM (`aarch64`/`armv7l`). The script is vendored into this provider's own `bin/`/`tests/` because a base-provider script isn't resolvable on PATH from a ce-oem job's `command:` at runtime (confirmed on a real DUT). Job is gated by a new `has_cpu_idle_states` manifest entry so it's skipped — not falsely failed — on ARM hardware/firmware whose device tree defines no cpuidle states (confirmed on a Renesas RZ/G3S DUT: script executes correctly, and the manifest gate correctly reports `Outcome: job cannot be started (unmet manifest)` instead of failing).

Also verified: Stage 1 static check (`environ_check.py`) passed with 0 findings; vendored `tests/test_cpu_idle_state.py` (97/97 passing); black (`--line-length 79`) / flake8 clean.

## Resolved issues

KOTO-61 (https://warthogs.atlassian.net/browse/KOTO-61)

## Tests

https://certification.canonical.com/submissions/status/403363


rickwu666666:
Since the test can be run on both ARM and x86 platform, we will land this into provider-base instead of provider-ce-oem.

PR #2876, opened 2026-09-17
Add i.MX93 camera support for AP1302 with shared NXP camera handler (New)

stanley31huang:
## Description
This change adds support for the i.MX93 camera path using the AP1302 sensor in the CE OEM camera test framework.

The camera factory previously only accepted imx8m as a supported platform and used an Imx8mBaseCamera name that implied a narrower platform scope than the actual implementation. This patch updates the platform support logic to cover both imx8m and imx9 families, and renames the shared handler to a generic NXP i.MX camera base class to better reflect the common behavior.

It also adds the i.MX93 MIPI camera scenario configuration for the AP1302 sensor, covering both still-image capture and video recording at the supported resolutions and frame rates.

## Tests
https://certification.canonical.com/hardware/202503-36431/submission/511688/test-results/


rickwu666666 (approved):
LGTM

baconYao (approved):
LGTM

